PR #20, opened 2019-08-22
Do not use compiled template outside of production

joelhawksley:
This PR addresses https://github.com/github/actionview-component/issues/19 where template changes were not reflected when reloading a page in development.

It also corrects an error message that referenced the now-removed inline template functionality.

The `modify_file` test helper is shamelessly adapted from `rails/actionview/test/template/compiled_templates_test.rb`.

cc @metade @mikker

### DIFF
--- a/lib/action_view/component.rb
+++ b/lib/action_view/component.rb
@@ -67,8 +67,6 @@ module ActionView
       # We could in theory do this on app boot, at least in production environments.
       # Right now this just compiles the template the first time the component is rendered.
       def compile
-        # Only use compiled method if we're in production. Eventually, we'll probably want to
-        # use something like Rails.application.config for this toggle instead.
         return if @compiled && ActionView::Base.cache_template_loading
 
         class_eval("def call; @output_buffer = ActionView::OutputBuffer.new; #{compiled_template}; end")
--- a/lib/action_view/component.rb
+++ b/lib/action_view/component.rb
@@ -100,7 +100,7 @@ module ActionView
 
         if sibling_files.length == 0
           raise NotImplementedError.new(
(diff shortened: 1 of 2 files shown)


lib/action_view/component.rb
```diff
@@ -67,7 +67,9 @@ def inherited(child)
       # We could in theory do this on app boot, at least in production environments.
       # Right now this just compiles the template the first time the component is rendered.
       def compile
-        return if @compiled
+        # Only use compiled method if we're in production. Eventually, we'll probably want to
+        # use something like Rails.application.config for this toggle instead.
+        return if @compiled && Rails.env.production?
```
jhawthorn: Best to avoid explicitly checking the `Rails.env` value (which is supposed to have a purely user-defined meaning).

I think `ActionView::Base.cache_template_loading` would be a good fit here.

jhawthorn:
In Rails 6 we try pretty hard to keep the cached compiled views even in development mode. It's a [pretty significant speed up](https://github.com/rails/rails/pull/35629#issuecomment-477387164) .

I wonder if this could do something similar? Listen for file updates and then run something like `ActionView::Component.descendants.each(&:clear_class)`? Alternatively it could track the mtime or SHA of the template file (what Rails used to do to avoid recompilation).

tenderlove:
> I wonder if this could do something similar? Listen for file updates and then run something like `ActionView::Component.descendants.each(&:clear_class)`? Alternatively it could track the mtime or SHA of the template file (what Rails used to do to avoid recompilation).

I suggested to @joelhawksley that we do that, but I wasn't sure off the top of my head how.  I *think* this is OK until we figure out the right way (unless you know it?)

Issue #27, opened 2019-08-27
Expose controller context in components


joelhawksley:
Per @metade's [comment](https://github.com/github/actionview-component/issues/15#issuecomment-523433555), let's look at exposing the controller context inside components.

It might be worth considering making this something that can be disabled on a per-component basis (`without_controller_context` macro perhaps?), as coupling components to controllers makes them trickier to test and more difficult to use outside of a request.

@metade if you're still up for this, I'd welcome a PR! If nothing else, perhaps one with some failing test cases for us to look at?

joelhawksley:
I believe this should be addressed in https://github.com/github/actionview-component/releases/tag/v1.3.0.

PR #28, opened 2019-08-27
Expose controller context inside Components

metade:
To start with, just adding a failing test for a component that renders a partial as per @joelhawksley's request in #27 .

For now the partial fails as follows:
```
Error:
ActionView::ComponentTest#test_renders_partial_template:
NoMethodError: undefined method `render' for nil:NilClass
    (eval):1:in `call'
```


metade:
@joelhawksley Sorry, I had to rush off yesterday, just noticed that I got the partial filename wrong.

I can get this test to work by adding the following to `render_in`
```
  @view_renderer ||= view_context.view_renderer
```

Do we have a way to benchmark this to see what kind of effect it has?

metade:
To move this forward I've added the workarounds I shared earlier to add support for partials and button_to/forms in this PR.

I quickly ran the benchmark page on your [demo app](https://github.com/joelhawksley/actionview-component-demo) against these changes and it didn't seem to have a major impact.

I'm afraid I'm not sure how best to handle benchmarking from a Rails gem... I had a quick look at how other popular gems might do it (devise, simple_form, kaminari) and I've not seen anything obvious!

joelhawksley:
@metade this looks great! Mind adding an entry to the changelog on this branch? 

I need to figure out how to run CI on forks, so I've changed the target of this PR to the `contrib` branch. Once we merge this into `contrib`, I'll open a PR to `master` that should run CI.

mkempe:
Hey there,

many thanks for this feature. I think it's very useful (to split "bigger" components). I just tested the branch and works like charm.

Just one thing:

```
= render 'components/hello_component/logo', locals: { slug: slug }
```

works, but

```
= render partial: 'components/hello_component/logo', locals: { slug: slug }
```

throws `undefined method `html_fallback_for_js' for nil:NilClass`.

Thanks again for your work!

mkempe:
Sorry, i think, i was using the `render` method wrong. You don't need to use locales, all assignments are automatically available in the partials. Just use:

```slim
= render 'components/hello_component/logo'
```

metade:
@mkempe Thanks for raising this, I've fixed that particular problem. 

It seems that there were changes in Rails 6 that require extra controller context to be exposed.

We're also finding in our Rails app that we're upgrading to Rails 6.0 that partials aren't setup/compiled correctly with components.

We're getting the following error in the development/cucumber environments:
```
      undefined method `_the_path_to_the_partial_html_haml__1412430037558108033_70350458081300' for #<MyComponent:0x00007ff77bcaaf88> (ActionView::Template::Error)
```

But the individual component tests are working fine.

Unfortunately I've not been able to replicate the error in either this gem or the [demo component app](https://github.com/joelhawksley/actionview-component-demo).

I've been poking around in ActionView to figure out how/where partial templates are defined as methods but I've not had any luck!

Does anyone have any ideas?

metade:
Hmm, this works but it's not great!

```
  delegate_missing_to :view_context

  def view_context
    @view_context ||= controller.view_context
  end
```

mkempe:
@metade Sorry for the delayed response. I am also getting the `undefined method` error (Rails 6.0). I think, it has to do with the code reloading, because i am only getting the error, when i make changes. If you boot the application normally then all is fine. That's why the component tests are working fine too. I will try to find a solution tomorrow.

lib/action_view/component/test_helpers.rb
```diff
@@ -4,7 +4,9 @@ module ActionView
   module Component
     module TestHelpers
       def render_inline(component, **args, &block)
-        Nokogiri::HTML(ApplicationController.new.view_context.render(component, args, &block))
+        controller = ApplicationController.new
```
jcoyne: I would love for this to be extracted to a separate method so that one could stub methods on the controller instance.  e.g.:
```ruby
      def controller
        @controller ||= ApplicationController.new.tap { |c| c.request = ActionDispatch::TestRequest.create }
      end
```

lib/action_view/component/base.rb
```diff
@@ -58,6 +60,9 @@ class Base < ActionView::Base
       #
       def render_in(view_context, *args, &block)
         self.class.compile
+        self.controller = view_context.controller
+        @view_renderer ||= view_context.view_renderer
+        @lookup_context ||= view_context.lookup_context
```
cjbottaro: Can you please add
```
@view_flow ||= view_context.view_flow
```
So that `content_for` works?

metade: I've added that now, together with a test.

metade:
I'm still struggling with rendering partials properly, but I think this only affects components that render partials that are also rendered in another view. The patch I've applied on our app doesn't seem to work here!

I've been able to add an integration test that triggers the error, but I'm skipping the test for Rails 6.0.

@joelhawksley I've updated this PR to the latest master but I had forgotten that you had changed this to point to the `contrib` branch - not sure how best to resolve this!

lib/action_view/component/test_helpers.rb
metade: I've updated this as per your suggestion - is that what you meant?

joelhawksley:
Thanks for your work on this @metade! @tenderlove and I are going to look at why that test is failing on Rails 6 later this week.

I really appreciate all of your work on this ❤️

mkempe:
@metade I just wanna thank you for your constant work on this ❤️. Maybe with the help of @tenderlove we can fix the "partial rendering bug". Really looking forward to use ActionView::Component, it's such an elegant solution to organise your views.

joelhawksley:
@metade thanks, again for this significant contribution. 

@tenderlove and I got things working across all Rails versions in https://github.com/github/actionview-component/pull/47, which includes these changes.

lib/action_view/component/base.rb
```diff
@@ -58,6 +60,11 @@ class Base < ActionView::Base
       #
       def render_in(view_context, *args, &block)
         self.class.compile
+        self.controller = view_context.controller
```
cjbottaro: Can you add
```
self.request = controller.request
```
So that Devise helper methods like `current_user` work? Thanks!

cjbottaro:
Oops, I didn't see it was merged already.

Issue #21, opened 2019-08-22
Any tips for rendering Components from collections?


nogtini:
Before I start playing around with this, I wanted to ask if there is a preferred way, best practice, or even a helper to render components for collections [similar to the rails guide for partials](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-collections)

I think this would be a common use case, and would be willing to submit a PR for a documentation example.

mikker:
An idea for the API,

```ruby
<%= render MyComponent.all(collection, more_options: true) %>
```

Perhaps. But it's a tiny win over

```ruby
<% collection.each do |thing| %>
  <%= render MyComponent.all(thing, more_options: true) %>
<% end %>
```

Then again; you could say that about a lot of Rails 😆

nogtini:
True, though I would imagine if there were to be an API it might look as similar as possible to the partial api.

### Partial
```
<h1>Products</h1>
<%= render partial: "product", collection: @products %>
```

### Component
```
<h1>Products</h1>
<%= render component: Product, collection: @products %>
```

joelhawksley:
@nogtini I don't think I could recommend a _good_ practice here at this point, let along a _best_ practice!

Generally our approach in writing this library has been to stay consistent with Rails APIs, at least as a default. What @nogtini is suggesting is probably the direction I think we'd go in.

joelhawksley:
@nogtini @mikker this is definitely something we'll want to consider.

Right now, we're passing an instance of the component into `render`:

```erb
<%= render MyComponent.new(product: product) %>
```

which means we've already defined what data is going to be used to render the component. 

We might have to change that syntax in order to accommodate the `collection` key. Perhaps we would do something like:

```erb
<%= render MyComponent, collection: products %>
```

Or perhaps we could support both syntaxes?

inopinatus:
Rendering a partial collection is also blazingly fast, because Rails only instantiates each template once, and they're also polymorphic.

AV:C can't be either in it's current architecture: firstly, by using ActiveModel's validations it's obliged to instantiate every component individually, there's no flyweight-pattern option; secondly, there's no `to_partial_path`-equivalent hook to select a component by convention-over-configuration.

I work with Big Heterogeneous Lists Of Stuff and those are both showstoppers for adoption.

nogtini:
[...]

PR #31, opened 2019-08-27
Use render(MyComponent, foo: :bar) syntax

joelhawksley:
In #21, @nogtini pointed out that our existing syntax of calling
`render(MyComponent.new)` makes it difficult to support rendering
collections in the way that Rails allows one to do so with partials.

In order to support rendering collections using components,
and potentially other changes to align with the existing Rails API,
it is necessary to change our syntax used to render components in views.

Now, components are rendered as:
`render(MyComponent, foo: :bar)`, or `render(component: MyComponent, locals: { foo: :bar })`.

The old syntax of:
`render(MyComponent.new(foo: :bar)` is still supported, but will be
deprecated in v2.

### DIFF
--- a/test/action_view/component_test.rb
+++ b/test/action_view/component_test.rb
@@ -77,9 +77,16 @@ class ActionView::ComponentTest < Minitest::Test
     assert_includes result.text, "foo"
     assert_includes result.text, "bar"
   end
-  
+
   def test_renders_erb_template_with_old_syntax
     result = render_inline(ErbComponent.new(message: "bar")) { "foo" }
+
+    assert_includes result.text, "foo"
+    assert_includes result.text, "bar"
+  end
+
+  def test_renders_erb_template_with_hash_syntax
+    result = render_inline(component: ErbComponent, locals: { message: "bar" }) { "foo" }
 
     assert_includes result.text, "foo"
     assert_includes result.text, "bar"


lib/action_view/component/base.rb
```diff
@@ -2,18 +2,29 @@
...
 class ActionView::Base
   module RenderMonkeyPatch
-    def render(component, _ = nil, &block)
-      return super unless component.respond_to?(:render_in)
-
-      component.render_in(self, &block)
+    def render(options = {}, args = {}, &block)
+      if options.respond_to?(:render_in)
+        ActiveSupport::Deprecation.warn(
+          "passing component instances to `render` will be deprecated in actionview-component 2.0. Use `render MyComponent, foo: :bar` instead."
```
rmacklin: I am finding the wording
> will be deprecated in actionview-component 2.0

here (and in the `CHANGELOG` and `README`) a bit confusing.

Since this is already logging a deprecation warning, it sounds like this is deprecated _now_ so is the plan that this syntax will be _removed_ in v2.0?

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+*   The `render_component` test helper has been renamed to `render_inline`.
+
+    *Joel Hawksley*
+
+*   Components are now rendered with `render MyComponent, foo: :bar` syntax. The existing `render MyComponent.new(foo: :bar)` syntax will be deprecated in version 2.0.
+
+    *Joel Hawksley*
+
 *   Components now inherit from ActionView::Component::base
```
rmacklin: I think there's a typo here: `base` -> `Base`

Also, I find it helpful when changelogs contain a heading with the version number that separates each list of changes that were made in that version - would you be open to adding that?

lib/action_view/component/base.rb
rmacklin: Oh, looks like you just noticed this as well and pushed 54bfb4b704ec3c69b2e882e528ef7024df0f0787

However, the confusing wording is still here:
https://github.com/github/actionview-component/blob/54bfb4b704ec3c69b2e882e528ef7024df0f0787/CHANGELOG.md#L5
and here:
https://github.com/github/actionview-component/blob/54bfb4b704ec3c69b2e882e528ef7024df0f0787/README.md#L148

lib/action_view/component/test_helpers.rb
```diff
@@ -3,8 +3,8 @@
 module ActionView
   module Component
     module TestHelpers
-      def render_component(component, &block)
-        Nokogiri::HTML(component.render_in(ApplicationController.new.view_context, &block))
+      def render_inline(component, **args, &block)
```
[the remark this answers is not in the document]
rmacklin: Given that you seem to care about not breaking the 1.0 API until v2.0, I think it's worth noting that renaming `ActionView::Component` to `ActionView::Component::Base` in v1.1.0 was also a breaking change.

I don't think it's particularly difficult for people to fix when they upgrade, but figured I'd point it out.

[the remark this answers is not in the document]
rmacklin: Just some thoughts given the above:

I think you could either treat 1.x as still beta software that can have breaking changes in minor upgrades (the way many packages are in the 0.x stage), or just go ahead and jump to 2.0 now since it's still early in the adoption of this gem and you already have identified a few breaking API changes you want to make. (If you choose the former, documenting it in the README would probably be a good idea.)

lib/action_view/component/base.rb
rmacklin: Resolved by 885bc97b218c1899d5adf16c68e32bd54aa10eed

CHANGELOG.md
rmacklin: Resolved by 8c95681867cd9fef19fc626bb215a1de705a3534 and 06310bb4761a0520494c312834631de695a2774d

PR #39, opened 2019-08-31
Allow component template inheritance.

ushis:
Hi,

this PR changes the way the `template_file_path` is resolved. The idea is to walk up the ancestors chain until an ancestor with a template is found. This enables template inheritance:

```ruby
# link_component.rb
class LinkComponent < ActionView::Component::Base
  def initialize(url:, name:, **options)
    @url = url 
    @name = name 
    @options = options
  end

  private 

  attr_reader :url, :text, :options
end
```
```erb
# link_component.html.erb
<%= link_to name, url, options %>
```
```ruby
# funk_link_component.rb
class FunkyLinkComponent < LinkComponent
  def initialize(name:, url:, **options)
    super(name: name, url: url, options: options.merge(class: 'funky'))
  end
end
```
```ruby
render(LinkComponent, name: "Unfunky", url: unfunky_path)
# => <a href="/unfunky">Unfunky</a>

render(FunkyLinkComponent, name: "Funky", url: funky_path)
# => <a href="/funky" class="funky">Funky</a>
```

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -97,23 +97,24 @@ module ActionView
         end
 
         def template_file_path
-          raise NotImplementedError.new("#{self} must implement #initialize.") unless self.instance_method(:initialize).owner == self
+          ancestors.each do |klass|
+            raise(NotImplementedError, "Could not find a template file for #{self}.") if klass == ActionView::Component::Base
 
-          filename = self.instance_method(:initialize).source_location[0]
-          filename_without_extension = filename[0..-(File.extname(filename).length + 1)]
-          sibling_files = Dir["#{filename_without_extension}.*"] - [filename]
+            next unless klass.is_a?(Class)
+            next unless klass.private_method_defined?(:initialize)
+            next unless klass.instance_method(:initialize).owner == klass
 
-          if sibling_files.length > 1
-            raise StandardError.new("More than one template found for #{self}. There can only be one sidecar template file per component.")
+            filename = klass.instance_method(:initialize).source_location[0]
+            filename_without_extension = Pathname.new(filename).sub_ext("").to_s
+            sibling_files = Dir["#{filename_without_extension}.*"] - [filename]
+            next if sibling_files.empty?
+
+            if sibling_files.length > 1
+              raise StandardError, "More than one template found for #{klass}. There can only be one sidecar template file per component."
+            end
+
+            return sibling_files[0]
           end
-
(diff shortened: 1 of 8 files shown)


joelhawksley:
@ushis interesting! I like the idea of being able to create components that are preset arguments of other components.

I'm wondering if this could instead be accomplished by having `FunkyLinkComponent` render `LinkComponent` in its template, like so:

```erb
<%= render LinkComponent, name: "Funky", url: funky_path, class: "funky" %>
```

If nothing else, this approach ☝️ mimics what I've seen done in React.

What do you think?

ushis:
@joelhawksley You are right. I noticed the given example is a bit too simple to illustrate my use case. I was thinking about components with complex behaviour, i wanted to implement in an abstract component, that is extended incl. its template. While playing around with it, i noticed this approach is much more confusing than building a generic component, which is rendered from inside other components, like you proposed. 

I would be nice, if we could do the rendering inside the component without the need of a template. Something like this, with a simpler API:

```ruby 
class FunkyLinkComponent < ::ActionView::Component
  def render_in(ctx, name:, url:, **options, &block)
    LinkComponent.render_in(ctx, name: name, url: url, options.merge(class: 'funky'), &block)
  end
end
```

A simpler API could be, to override `call`:

```ruby
class FunkyLinkComponent < ::ActionView::Component
  def initialize(name:, url:, **options)
    @name = name 
    @url = url 
    @options = options
  end 

  def call 
    render(LinkComponent, url: url, name: name, **options)
  end

  private 

  attr_reader :url, :name 

  def options 
    @options.merge(class: 'funky')
  end
end
```

joelhawksley:
@ushis I agree! It would be nice to be able to render another component like that.

I think you're on to something here. We used to have inline templates in components that would allow for a similar arrangement, but removed them due to concerns from @soutaro about incompatibility with the typing options coming in Ruby 3. 

Perhaps he could weigh in here?

soutaro:
@joelhawksley The change looks reasonable to me. It is essentially to reuse the `#call` method of a super-class from its sub-classes, which totally makes sense in the context of OO programming.

I'm not very sure if I get the point correctly, so let me know if there is something else to consider.

joelhawksley:
@ushis now that we [support variants](https://github.com/github/actionview-component/pull/92), would you expect us to now overwrite the `call_phone`, `call_tablet` methods, for example?

joelhawksley:
Closing due to inactivity. Feel free to reopen if desired ❤️

PR #54, opened 2019-10-02
Fix implicit partial rendering from components

metade:
It's possible to render a partial by calling render with the partial string, i.e. without using the partial keyword argument.

Updating ActionView::Component::Base#render to handle this.


metade:
The latest release (1.3.1) caused issues with our application as we're rendering partials by just using the implicit method, i.e.
```
<%= render "test_partial" %>
```

PR #57, opened 2019-10-02
Allow sidecar CSS files

### LINKED ISSUE #55: Best practice for CSS?
I have a component structure like so:

```
- /app
  - /components
    - /alert
      - alert.css
      - alert.html.erb
      - alert.rb
```

I'm importing the CSS file in Webpacker's `application.css`.

Unfortunately this setup gives me the error:

```
More than one template found for Alert. There can only be one sidecar template file per component.
```

I'm currently working around this by using `_alert.css` instead of `alert.css`, but it seems odd `actionview-component` is treating this CSS file as a template file?

Is there a best practice here I should be following? Or is this an issue `actionview-component` should fix?

joelhawksley:
This PR tightens our template file lookup logic so that non-template files do not trigger the multiple-sidecar-files error.

[Fixes #55]

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -117,7 +117,7 @@ module ActionView
 
           filename = self.instance_method(:initialize).source_location[0]
           filename_without_extension = filename[0..-(File.extname(filename).length + 1)]
-          sibling_template_files = Dir["#{filename_without_extension}.????.{#{ActionView::Template.template_handler_extensions.join(",")}}"] - [filename]
+          sibling_template_files = Dir["#{filename_without_extension}.????.{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [filename]
 
           if sibling_template_files.length > 1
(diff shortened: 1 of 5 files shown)


lib/action_view/component/base.rb
```diff
@@ -117,19 +117,19 @@ def template_file_path
 
           filename = self.instance_method(:initialize).source_location[0]
           filename_without_extension = filename[0..-(File.extname(filename).length + 1)]
-          sibling_files = Dir["#{filename_without_extension}.*"] - [filename]
+          sibling_template_files = Dir["#{filename_without_extension}.????.{#{ActionView::Template.template_handler_extensions.join(",")}}"] - [filename]
```
kenyonj: Thoughts on using single quotes within the double quotes to make clear that you aren't stopping the outer double quotes?

Also, TIL `????`
```suggestion
          sibling_template_files = Dir["#{filename_without_extension}.????.{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [filename]
```

PR #61, opened 2019-10-05
Provide file and line context for class_eval

jcoyne:
This makes it easier to work with in a debugger


joelhawksley:
Thank you for this very helpful patch ❤️

Issue #65, opened 2019-10-08
[1.3.4] private method `controller' called


afuno:
In version 1.3.3, everything was fine. There was a problem after upgrading to version 1.3.4. I checked - the problem is really in the new version.

One of my components started to throw this error:

```
private method `controller' called for #<My::Component::NameComponent:0x00007f8520434b10>
Did you mean?  controller=
```

The component has input values.

koenhandekyn:
i'm fighting the same issue. 1.3.1 is still running smooth but 1.3.4  is throwing the same. it's throwing it on a rails path helper but i can't drill it down further.

`node_id_merge_path(template_id: node_type_deliverable_type.id, target_id: node.id, action: :node_id) title=t('component.upload.node.open_merged', deliverable_type: "#{deliverable_type_name}" )`

what is 'special' is that these are just 'old fashioned' routes (no rest resources), they are defined dynamically in our case, but this shouldn't cause any trouble - in theory :) 

it might be related maybe ? the explicit 'controller' reference ? seems possibly related to the bizar error message (private method 'controller') ...

but this should definitely just work in my opinion, old fashioned routes still have their purpose :) 

```
    %w(node_id committee_meeting_id instrument_holding_id).each do |action|
      get "/c/(:c)/merge/:template_id/#{action}/:target_id", controller: "merge", as: "#{action}_merge", action: action
      get "/c/(:c)/merge_context/:template_id/#{action}/:target_id", controller: "merge", as: "#{action}_merge_context", action: "#{action}_context"
    end
```

koenhandekyn:
did some more research. at the bottom our full component template.

1. as soon as we reference "node_id_merge_path" helper, we have the above error (private method controller called)
2. BUT ALSO as soon as we have `= form_for Deliverable.new`, even with an empty block, we have the same error.

quite confusing ...


```
tr
  td
    - if deliverables
      a href="#" data-target="##{form_id}_modal" data-toggle="modal"
        i.fa.fa-folder-open
      | &nbsp;
      span.text-left
        .modal.fade id="#{form_id}_modal" aria-hidden="true" aria-labelledby="" role="dialog" tabindex="-1"
          .modal-dialog
            .modal-content
              .modal-header
                button.close aria-hidden="true" data-dismiss="modal" type="button" &times;
                h4.modal-title
                  => deliverable_type_name
                  | archive
              .modal-body
                table.table.table-striped
                  - deliverables.reverse_each do |deliverable_rv|
[...]

PR #68, opened 2019-10-15
Add test for path helpers in nested component

michaelem:
Ilustrating the issue (#65) caused by making controller private in 1.3.4.


joelhawksley:
Thank you for the PR! I've resolved this, using your starting commit, in https://github.com/github/actionview-component/pull/76.

PR #72, opened 2019-10-16
Components should also check view directory for template files

dnwilson:
Whilst working with `ActionView::Components` on one of our projects I noticed that the main components directory was becoming cumbersome with all the different file types in the same directory. So I added a one-liner to allow loading the template files from the `app/views/components` directory. My team really liked having a set structure for our components and not going against already defined rails conventions. So we ended up with the following structure.

```
# assets/stylesheets/components/button.scss
# app/components/button.rb
# app/view/components/button.html.erb
```

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -125,7 +125,8 @@ module ActionView
 
           filename = self.instance_method(:initialize).source_location[0]
           filename_without_extension = filename[0..-(File.extname(filename).length + 1)]
-          sibling_template_files = Dir["#{filename_without_extension}.????.{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [filename]
+          sibling_template_files = file_path_array(filename_without_extension) - [filename]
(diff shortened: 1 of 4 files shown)


joelhawksley:
With that comment in mind, I'm going to respectfully close this PR for now. I'm more than happy to keep this discussion going here, or in private / on a call if you'd prefer. Feel free to reach out to me at: joelhawksley@github.com.

dnwilson:
@joelhawksley understood. Would you also be opposed to a templates/stylesheets/javascripts directory under `app/components`?

joelhawksley:
@dnwilson sorry for the delayed reply, I somehow missed the notification for your comment ❤️ 

In a similar vein to https://github.com/github/actionview-component/issues/67, the concern I have here with including stylesheets/javascripts in the `components` file hierarchy is that, at least to me, it implies that said stylesheets/javascripts would be scoped to / encapsulated within the component. 

I know this is an architecture that folks use in React, but unless we have CSS/JS that we can guarantee is only used in the associated component, I think we should avoid a file/directory structure that implies it is.

PR #84, opened 2019-10-25
Preselect the rendered component in render_inline

elia:
Let Nokogiri point directly at the rendered component instead of requiring the
knowledge about the fact that it's rendered in the context of an HTML document.

A deprecation cycle shouldn't be necessary the change is compatible with the former use (as ensured by the two tests).


joelhawksley:
Thanks for the PR! ❤️

PR #87, opened 2019-10-26
[Fix] Virtual path lookup fixed

juanmanuelramallo:
- Solves https://github.com/github/actionview-component/pull/74#discussion_r338411064
- Whenever the file path contained the pattern /.rb/ the component did not work properly with translations shortcuts
- Updated the method to escape the dot in the regular expression


joelhawksley:
👏 thanks for the clean PR ❤️

Issue #40, opened 2019-09-02
ActionView::Component previews


mikker:
It struck me; AV::Components ought to have previews the same way ActionMailer does.

I realise this might be overkill for a 1.0 release but I'll be happy to throw a first draft together if you folks think it's as good an idea as I do.

Otherwise I might just make it as a gem on its own.

xdmx:
It'd be awesome to have some kind of preview of components, with the ability of grouping the by categories/subcategories. I'm thinking about something like storybook, where you can easily navigate and search the available components.

It'd also be great having a configuration to allowing the access of it from different environments (like mailer previews), that way it could be enabled on a staging environment and having designers accessing it without having to go through the whole rails/project setup.

xdmx:
It would also play very well with the [rails conductor](https://github.com/rails/rails/pull/35489) planned for 6.1, which from my understanding is also the planned version for merging components into rails

joelhawksley:
@mikker this sounds great! I'd happily accept a PR for this that mimics ActionMailer previews as much as possible 😍

tijmenb:
Hi! I'd be very interested in this. [GOV.UK](https://www.gov.uk/) has been working with components for the last couple of years, and one of the key features is the ability to spin up a "component guide" that documents all the components.

The main one is publicly available at: https://components.publishing.service.gov.uk (repo is https://github.com/alphagov/govuk_publishing_components).

For example, this is the [page title component](https://components.publishing.service.gov.uk/component-guide/title):

![Screenshot_2019-10-11 Page title component - Component Guide](https://user-images.githubusercontent.com/233676/66683668-f0fabd00-ec6f-11e9-8d06-73e3fff7261a.png)

A component guide page has:

- A description of how it should be used
- A preview
- A number of usage examples (this is super useful make sure components work visually with all inputs) plus previews
- Code snippet to use it in your template

@mikker have you started on this? I'd be up for spiking something.

mikker:
I haven't yet so feel very free to take over 👍

I love the gov.uk guides and of course you already have something like this going.

juanmanuelramallo:
I'd like to take this one

juanmanuelramallo:
@mikker this issue has been addressed in #96 so I'm closing it

It's just a start btw, we can come back later and make improvements

mikker:
Wow! That's great work! Thanks @juanmanuelramallo 💙💚💛💜❤️

PR #96, opened 2019-11-05
[#40] Previews

juanmanuelramallo:
- Solves #40 
- Copied implementation from Action Mailer previews
- Added a preview class
- Preview classes are associated to a layout in order to properly render the component examples
- Preview classes uses `test/components/previews/layouts/application` as the default layout
- Preview.call class method receives an example method and returns the html of the component
- Preview.call uses [Roadie](https://github.com/Mange/roadie) to inline the styles from the stylesheets in the layout
- Added a railtie class
- The railties define two new routes for the components previewing
- The route `/rails/components/*path` renders the examples of a component
- Added a components controller
- Added some examples in the test app

image


juanmanuelramallo:
Hey @joelhawksley, do you think this can work as a base branch for the previewing feature? 

I basically copied the implementation from Action Mailer previews, but removed some stuff that we might not need, like interceptors and observers.

cover:
Thank you @juanmanuelramallo for working on this :raised_hands: 

One thing that I love about the gov.uk example that @tijmenb shared is all the details you can get inside a component's page, you don't have to browse between pages to see different statuses and how they look or work.

Do you think it'd be possible to achieve such a page?

Maybe by having some methods/conventions that could be used (from the [page title component](https://components.publishing.service.gov.uk/component-guide/title)):

```ruby
class PageTitleComponentPreview < ActionView::Component::Preview
  def self.summary
    'A page title with optional context label'
  end

  def main
    {
      description: "<p>This contains an optional parameter for average title length. The two valid values for this parameter are ‘medium’ or ‘long’. Medium titles are titles where the average is around 30 characters or less. Long titles would have an average length of nearer 50 characters or more.</p><p>On average the titles on government bits of content are 50 characters. The average for bits of general guidance are nearer 27 characters long.</p>",
      props: { title: "My page title" },
      accessibility_acceptance_criteria: "some html code with <uo> lists"
    }
  end

  def with_context
    {
      props: { context: "Publication", title: "My page title" }
    }
  end

  def with_context_language_labelled
    {
      props: { context: "Publication", context_locale: "en", title: "My page title" },
      description: "<p>Sometimes this component appears on a page that has been translated. The title will naturally be supplied in the required language but the context string may fall back to the default. In these instances we need to label the language so the page remains semantic and screenreaders can handle the switch.</p><p>The lang attribute must be set to a valid BCP47 string. A valid code can be the two or three letter language code - for example, English is en or eng, Korean is ko or kor - but if in doubt please check.</p>"
    }
  end
end
```

The conventions would be:
  - a class method `summary` to show the summary/short description in the components list
  - a hash with a `props` (or `params`, `data` or something else) key in each method

This way there would be 1 single page for each component, which would load all public methods and hashes and use their data to render them (by using the data in the `props`).
Any extra key (eg. `description`, `accessibility_acceptance_criteria`, etc) would be rendered on the page as text, keeping the order from the hash.

The result from the above would be something like:

### Page Title
#### Summary
#### Main
##### Description
##### Example
##### Example Code
##### Accessibility acceptance criteria
#### With context
##### Example
##### Example Code
#### With context language labelled
##### Example
##### Example Code
##### Description

juanmanuelramallo:
@cover Hey thanks for the comment! 

1. > One thing that I love about the gov.uk example that @tijmenb shared is all the details you can get inside a component's page, you don't have to browse between pages to see different statuses and how they look or work.
Do you think it'd be possible to achieve such a page?

Definitely, fairly simple change I think. I can update this PR to have that.
Though I would like to double check with @joelhawksley about it. Since current behaviour is just a copy from action mailer previews.

2. >a class method summary to show the summary/short description in the components list

I'd rather use here a class method like the `set_layout` one. So that end users can set the summary by calling `set_summary` in their previews.

3. > accessibility_acceptance_criteria

What's the `accessibility_acceptance_criteria` for? 

4. >description

IMHO writing HTML in a string inside a hash inside these example methods, seems a bit too much for a convention to me. Though I like the idea of optionally providing some explanation about the example. 
But what do you think if instead of having this `description` attribute per example we just use the method name of the example? 

5. >example code

I like it. Maybe we can include it as a subsequent PR. So that we don't make this PR bigger than it already is (?) What do you think?

cover:
2. > I'd rather use here a class method like the set_layout one. So that end users can set the summary by calling set_summary in their previews.

`set_summary` sounds great :raised_hands:

3. > What's the `accessibility_acceptance_criteria` for?

It was just a straight up example from the gov.uk page. On their [page title component page](https://components.publishing.service.gov.uk/component-guide/title) they have this "Accessibility acceptance criteria" section. I'd expect that each company would like to have their own custom list or content to put on their component page (text, links, lists, images, etc).

So the idea, also related to the `description` key, is to just require a `props` key (or with a different name) to be present in the hash, and all other `keys/values` would be treated as section titles/section contents. This way adding a `description` key with some content as value would render a "Description" section with its content on the page, adding a `important_notes` key would add a "Important notes" section, and so on.

The content itself could be a plain string (eg. a simple text description) or a more complex html structure, based on the need of the developers and the information that they want to display.

I'd expect that for many people/teams just the name of the method would be enough to understand what the variant is about. At the same time it'd be nice to provide enough flexibility to add any extra content for those teams that require it :thinking:

Gemfile
```diff
@@ -6,3 +6,4 @@ gemspec
 rails_version = "#{ENV['RAILS_VERSION'] || '6.0.0'}"
 
 gem "rails", rails_version == "master" ? { github: "rails/rails" } : rails_version
+gem "roadie", "~> 4.0", require: false
```
joelhawksley: I can see why we want to use `roadie` for this, but I'd _really_ prefer to see if we can pursue an option that lets us avoid doing so. 

Including this kind of dependency will make it more difficult to merge our code into Rails.

README.md
```diff
@@ -201,6 +201,69 @@ def test_render_component_for_tablet
 end
 ```
 
+### Previewing Components
+Action View Component previews provide a way to see how components look by visiting a special URL that renders them.
```
```suggestion
`ActionView::Component::Preview`s provide a way to see how components look by visiting a special URL that renders them.
```

README.md
```diff
@@ -201,6 +201,69 @@ def test_render_component_for_tablet
...
+In the previous example, the preview class for `TestComponent` should be named `TestComponentPreview` and located in `test/components/previews/test_component_preview.rb`.
```
```suggestion
In the previous example, the preview class for `TestComponent` would be called `TestComponentPreview` and located in `test/components/previews/test_component_preview.rb`.
```

README.md
```diff
@@ -201,6 +201,69 @@ def test_render_component_for_tablet
...
+To see the preview of the component with a given title, implement a method that returns the locals (as a hash) to be passed to the component.
+The method can be named after what you want to preview given the locals you use, and you can define many examples (methods) as you want:
+
+```ruby
+# test/components/previews/test_component_preview.rb
+
+class TestComponentPreview < ActionView::Component::Preview
+  def with_default_title
+    {
```
joelhawksley: I'm curious as to the decision to go with a hash here, instead of something that more closely mimics the way components are rendered in views.

For `ActionMailer::Preview`s, the instance methods contain a direct call to the mailer in question: https://guides.rubyonrails.org/action_mailer_basics.html#previewing-emails. Ideally, we'd do the same here. 

Perhaps we could define our own `render` method on `ActionView::Component::Preview` that uses our `render_inline` test helper under the hood?

README.md
```diff
@@ -201,6 +201,69 @@ def test_render_component_for_tablet
...
+      title: "This is a really long title to see how the component renders this"
+    }
+  end
+end
+```
+
+Then the previews will be available in <http://localhost:3000/rails/components/test_component/with_default_title>
+and <http://localhost:3000/rails/components/test_component/with_long_title>.
+
+To specify the stylesheets and javascript used by your previews you must create a dummy layout file in `test/components/previews/layouts/`.
```
joelhawksley: Hmm. I'm wondering if it might be possible to reuse at least part of the application layout here. I'm worried that having a separate layout file for this would cause the previews to be inaccurate.

README.md
```diff
@@ -201,6 +201,69 @@ def test_render_component_for_tablet
...
+</body>
+</html>
+```
+
+All previews will use `test/components/previews/layouts/application.*` as the default layout. You can set a different layout by calling `set_layout` in your preview class.
+
+```ruby
+# test/components/previews/test_component_preview.rb
+class TestComponentPreview < ActionView::Component::Preview
+  set_layout :admin
```
joelhawksley: Might we call this `layout`, to match https://guides.rubyonrails.org/action_mailer_basics.html#action-mailer-layouts?

lib/action_view/component/base.rb
```diff
@@ -231,6 +234,8 @@ def type
           "text/html"
         end
       end
+
+      ActiveSupport.run_load_hooks(:action_view_component, self)
```
joelhawksley: Do we need to do this and the include above in production?

lib/railties/lib/rails/templates/components/_preview.html.erb
```diff
@@ -0,0 +1,18 @@
+<div class="preview" id="<%= example %>">
```
joelhawksley: Per my read of the Rails source, wouldn't this file go in `railties/lib/rails/templates/rails/components`?

test/test/components/previews/layouts/application.html.erb
```diff
@@ -0,0 +1,22 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Action View Component Test</title>
+  <style>
+    .preview-component {
```
joelhawksley: I'd be interested in having this test layout, if it is necessary, use the `stylesheet_tag` helper to load the main CSS assets in the demo app. I'm wary of having this example define custom styles for the components.

README.md
joelhawksley: At the very least, perhaps we could define a default layout that includes the default application styles and JS? That way developers could override that layout if necessary, but wouldn't have to create one. #conventionoverconfiguration

Gemfile
juanmanuelramallo: Makes sense. 

So the initial approach was just rendering the previews inside an iframe. We can go back that way maybe if it makes it easier for this move forward. 

The main downsides I saw about using iframes is that I couldn't give a proper height size to the iframe itself so that it doesn't have a scroll bar when the component previewed is bigger than the default height.

joelhawksley:
> It was just a straight up example from the gov.uk page. On their [page title component page](https://components.publishing.service.gov.uk/component-guide/title) they have this "Accessibility acceptance criteria" section. I'd expect that each company would like to have their own custom list or content to put on their component page (text, links, lists, images, etc).

@cover I love this idea! However, I want to be cautious with adding functionality that doesn't exist in `ActionMailer::Preview`, at least in this PR. 

As [mentioned in the project README](https://github.com/github/actionview-component#design-philosophy), our goal with this gem is to stick as closely to existing Rails behavior as possible.

That all being said, I think this idea has potential. I'm wondering if it might make sense to start with a PR to Rails that adds this kind of annotation functionality to `ActionMailer:Preview`?

README.md
juanmanuelramallo: Sounds good.

Do you think something like this would work?
```ruby
class TestComponentPreview < ActionView::Component::Preview
  def with_default_title
    render(title: "Default title", subtitle: "Some subtitle")
  end
end
```

And `ActionView::Component::Preview#render` would be an instance method of the preview object, right?

README.md
juanmanuelramallo: Yep, great idea! 

So to sum up:
1. Create a default layout that will be used for rendering the preview examples
2. Developers should be able to override this layout

README.md
juanmanuelramallo: Good call

lib/action_view/component/base.rb
juanmanuelramallo: I'm not sure, I'll double check

lib/railties/lib/rails/templates/components/_preview.html.erb
juanmanuelramallo: yes you're right. I missed the second `rails` folder.

Btw, is it ok that we define these files here in AC:V lib folder? 

I'm aware that these kind of templates actually live in the railties gem, but I wasn't sure where to locate them.

README.md
joelhawksley: I'm thinking we should be even more explicit:

`render(TestComponent, title: "Default title", subtitle: "Some subtitle")`

The benefit of this is that the previews would be examples that could then be copied and pasted directly into views.

test/test/components/previews/layouts/application.html.erb
juanmanuelramallo: Gotcha 👌

Gemfile
joelhawksley: Might we be able to use `render_inline` for this functionality?

README.md
juanmanuelramallo: I see, this is great

I'll update

lib/railties/lib/rails/templates/components/_preview.html.erb
joelhawksley: Hmm. That I don't know. I think as long as we get things _close_, we'll be fine. We're going to have to make other tweaks when merging this upstream, so I'm not worried about getting this perfect 😄

Gemfile
juanmanuelramallo: The `render_inline` does half of the work, because it only returns the html of the rendered component. 

The other half is about including the styles for the component. And since that lives in a stylesheet, and stylesheets linked to a DOM affect the whole DOM, we cannot just link them in the view. 

So that's why I thought about using iframes so that the component is rendered isolated from the main DOM (the one containing all examples from a component) and hence the iframes have the correct stylesheet. 

This can extend to javascript files too I think.

juanmanuelramallo: What do you think?

tijmenb: My initial thought would be using an iframe too. The GOV.UK components sometimes suffer from styles leaking from the "guide" into the components, and vice versa. Another project in UK Government does the same (https://design-system.service.gov.uk/components/details). It solves the sizing issue by giving each example a specific size (not sure how we'd do that here).

joelhawksley: My thinking here is that we should consider what the best default behavior would be. To me, it might be as "simple" as using the existing application layout for these views! What do you think about that?

juanmanuelramallo: Agree, we'll go that way. 

So we'll use the application layout for the examples' views.

test/test/components/previews/preview_component_preview.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class PreviewComponentPreview < ActionView::Component::Preview
+  set_summary "Summary of the preview component"
```
joelhawksley: Is there an equivalent feature in ActionMailer::Preview? If not, I think we should leave this out, for now at least.

juanmanuelramallo: No I don't think there is. I'll remove it

README.md
juanmanuelramallo: Instead of creating a default layout inside the gem, do you think it's fine if we just use the application's layout? https://github.com/github/actionview-component/pull/96/commits/7df33aff785906097ad5fbaa0efa4dac381adac0#diff-67f888315429841b717b9aeef54e71a6R48

Developers will be able to set any other layout they want, but layouts must live inside their app. Makes sense?

lib/action_view/component/base.rb
juanmanuelramallo: 1. The inclusion of ActionView::Component::Previews it's just a module that adds module attribute accessors for `preview_path` and `show_previews`. No more than that. So no harm in there, and we'll actually need them in production, in case the application has been configured to display the previews.

2. If we do not have this `ActiveSupport.run_load_hooks(:action_view_component, self)` we won't get our ActionView::Component configured with the options from the environment config file. Hence routes won't be defined. (This is what I gathered from trial and error testing, and reading the code, so I'm not entirely sure if this answers your question correctly)

juanmanuelramallo:
Hey @joelhawksley, I think all comments have been addressed now. 

Though I have one question, we have something slightly different from how ActionMailer::Preview s work. We display all the examples from a component preview in one template. On the opposite, ActionMailer displays the emails from one mailer one per page. 

Do you think we should also mimic that, and display only one preview of a component in one page?

README.md
juanmanuelramallo: What do you think if we actually use the application's layout? https://github.com/github/actionview-component/pull/96/commits/7df33aff785906097ad5fbaa0efa4dac381adac0#diff-67f888315429841b717b9aeef54e71a6R48 So that way we don't need to create any layout on the gem's side. 

Developers will still be able to override this layout, using any of their application's layouts.

joelhawksley: I think it's worth a shot! If we go that route, do we still need to us iframes?

lib/action_view/component/base.rb
joelhawksley: That makes sense. Perhaps we could put this line behind a conditional that checks to see if previews are enabled for the current environment?

README.md
juanmanuelramallo: Yes, because iframes will contain the component html inside the application's layout. And those iframes will live inside a view that uses the rails' application layout.

Does it make sense to you? Or do you have any other approach in mind?

lib/action_view/component/base.rb
juanmanuelramallo: Yep, good idea

README.md
joelhawksley: I guess I'm curious as to whether we need the iframes at all if we are just using the application layout.

I guess that would only work if you weren't using a different layout. Perhaps we could remove the custom layout functionality for now and see if anyone asks for it?

juanmanuelramallo: Oh maybe we are not talking about the same thing. I forgot to update the readme. 

There are no custom layouts for component previews, developers can only use their application's layouts. By default previews will use the `app/views/layouts/application.html.erb` layout. 

But since the previews are displayed from a rails namespace, they are served [within a layout that comes from rails itself](https://github.com/rails/rails/blob/master/railties/lib/rails/templates/layouts/application.html.erb).

That's why we use iframes, to avoid having that layout when rendering the component. And also to avoid using the application's layout in the `rails/components` and `rails/components/:component_name` routes

I'll update the readme tonight

joelhawksley:
> Do you think we should also mimic that, and display only one preview of a component in one page?

I think that for now it would probably be best to do what ActionMailer::Preview does, which is one per page. We can always deviate from the established convention later 😄

README.md
joelhawksley: Gotcha. Based on my comment below about staying consistent with ActionMailer::Preview, I think this is just fine for now. 😄 👍

juanmanuelramallo:
> > Do you think we should also mimic that, and display only one preview of a component in one page?
> 
> I think that for now it would probably be best to do what ActionMailer::Preview does, which is one per page. We can always deviate from the established convention later 😄

Got it! Will update

lib/action_view/component/base.rb
juanmanuelramallo: Do you have any ideas on how to approach this? 

I couldn't find a way cause it seems like this class is interpreted before rails is initialized.

This is from ActionMailer::Base https://github.com/rails/rails/blob/master/actionmailer/lib/action_mailer/base.rb#L1035

joelhawksley: Oh! Let's just do what ActionMailer does. ☝️  Sorry for the trouble.

juanmanuelramallo: It's ok no prob

README.md
```diff
@@ -203,6 +203,49 @@ def test_render_component_for_tablet
 end
 ```
 
+### Previewing Components
+ActionView::Component::Preview`s provide a way to see how components look by visiting a special URL that renders them.
+In the previous example, the preview class for `TestComponent` would be called `TestComponentPreview` and located in `test/components/previews/test_component_preview.rb`.
+To see the preview of the component with a given title, implement a method that calls the ActionView's render method.
```
```suggestion
To see the preview of the component with a given title, implement a method that renders the component.
```

README.md
```diff
@@ -203,6 +203,49 @@ def test_render_component_for_tablet
...
+The method can be named after what you want to preview given the locals you use, and you can define many examples (methods) as you want:
```
```suggestion
You can define as many examples as you want:
```

README.md
```diff
@@ -203,6 +203,49 @@ def test_render_component_for_tablet
...
+    render(TestComponent, title: "Test component default")
+  end
+
+  def with_long_title
+    render(TestComponent, title: "This is a really long title to see how the component renders this")
+  end
+end
+```
+
+Then the previews will be available in <http://localhost:3000/rails/components/test_component/with_default_title>
```
```suggestion
The previews will be available in <http://localhost:3000/rails/components/test_component/with_default_title>
```

README.md
```diff
@@ -203,6 +203,49 @@ def test_render_component_for_tablet
...
+and <http://localhost:3000/rails/components/test_component/with_long_title>.
+
+Previews will use your application's layout `app/views/layouts/application`. But you can also use any other layout from your app to render the preview.
```
```suggestion
Previews use the application layout by default, but you can also use other layouts from your app:
```

README.md
```diff
@@ -203,6 +203,49 @@ def test_render_component_for_tablet
...
+class TestComponentPreview < ActionView::Component::Preview
+  layout "admin"
+
+  ...
+end
+```
+
+By default, the preview classes live in `test/components/previews`.
+This can be configured using the `preview_path` option.
+For example, if you want to change it to `lib/component_previews`, you can configure it in config/application.rb:
```
```suggestion
For example, if you want to use `lib/component_previews`, set the following in `config/application.rb`:
```

juanmanuelramallo:
> Looking good. All I think that's left is to add some sort of tests. How do you think we might test this?

I didn't add any tests for these since it's an internal tool, the testing would be actually going to the `rails/components` path and checking that the rendering occurs as expected. 

Also I couldn't find any tests for ActionMailer::Preview.

But anyways we could add some tests if you think we need them; I'm not sure how we would do it, but definitely something we can add.

joelhawksley:
@juanmanuelramallo let's just merge this ❤️

Issue #95, opened 2019-11-04
Support `helpers.` syntax


joelhawksley:
It would be wonderful if we could support the `helpers` syntax in components, much like @rafaelfranca added for controllers in https://github.com/rails/rails/pull/24866.

kaspermeyer:
I can take a stab at this, @joelhawksley ✌️

joelhawksley:
@kaspermeyer 😍 thanks! I've assigned this to you so no one else takes it 😄

PR #105, opened 2019-11-11
Add helpers proxy

kaspermeyer:
Adds a `#helpers` method that works as a proxy for components to access helper methods through.

Before this PR you'd have to include helpers on a per-module basis:
```ruby
module IconHelper
  def icon name
    tag.i data: { feather: name.to_s.dasherize }
  end
end

UserComponent < ActionView::Component::Base
  include IconHelper

  def profile_icon
    icon :user   
  end
end
```
Now they can be referenced directly through the `#helpers` proxy:
```ruby
UserComponent < ActionView::Component::Base
  def profile_icon
    helpers.icon :user   
  end
end
```
Or delegated, if you prefer:
```ruby
UserComponent < ActionView::Component::Base
  delegate :icon, to: :helpers
  
  def profile_icon
    icon :user   
  end
end
```
---
The proxy pattern makes all helpers available for "free" and prevents our component classes from bloating with all the additional methods from a given helper module that you might not need.

Inspired by a similar feature in Rails by @rafaelfranca: https://github.com/rails/rails/pull/24866

Resolves: #95

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -93,6 +93,11 @@ module ActionView
         @controller ||= view_context.controller
       end
 
+      # Provides a proxy to access helper methods through
+      def helpers
+        @helpers ||= view_context
+      end
+
       # Looks for the source file path of the initialize method of the instance's class.
       # Removes the first part of the path and the extension.
       def virtual_path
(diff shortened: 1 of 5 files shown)


joelhawksley:
Brilliant, clean PR. This made my day!

kaspermeyer:
Thanks, @joelhawksley 🙏 You just made my day too.

PR #106, opened 2019-11-11
Create component generator

### LINKED ISSUE #97: Add component generator
@vinistock I heard you might have made some progress on this. Would you be interested in sharing what you have so far?

vinistock:
Closes #97. This PR creates a Rails generator for action view components. It should create all that is needed to get going with a new component and the selected attributes.

#### Usage

Running the generator in the following manner will create: a component file, a template file and a unit test.

```bash
bin/rails generate action_view:component Profile name age
      invoke  test_unit
      create    test/components/profile_test.rb
      create  app/components/profile.rb
      create  app/components/profile.html.erb
```

And these will be the content of the files:

```ruby
# app/components/profile.rb

class Profile < ActionView::Component::Base

  def initialize(age:, name:)
    @age = age
    @name = name
  end
end
```

```ruby
# app/components/profile.html.erb

<%= content %>
```

```ruby
# {test|spec}/components/profile_{test|spec}.rb

require "test_helper"

class ProfileTest < ActiveSupport::TestCase
  # test "component does something useful" do
  #   component = Profile.new
  #   assert component
  # end
end
```


lib/generators/action_view/component/component_generator.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      def create_template_file
+        template "component.html.erb", File.join("app/components",  "#{file_name}.html.erb")
+      end
+
+      private
+
+      def initialize_signature
+        if attributes.present?
+          attributes.map { |attr| "#{attr.name}:" }.join(", ")
```
juanmanuelramallo:
```suggestion
          attributes.map(&:name).join(":, ")
```

lib/generators/action_view/component/templates/component.rb.tt
```diff
@@ -0,0 +1,7 @@
+class <%= class_name %> < ActionView::Component::Base
```
juanmanuelramallo: I'm not sure if this should be `ApplicationComponent` instead of `ActionView::Component::Base`

lib/generators/action_view/component/templates/component.rb.tt
```diff
@@ -0,0 +1,7 @@
...
+  validates :content, presence: true
```
juanmanuelramallo: Should we make this configurable too? Because I don't think all components have `content`

lib/generators/action_view/component/component_generator.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module ActionView
```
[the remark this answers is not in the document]
vinistock: My rationale for not adding it under the Rails module is because then invoking the generator would be

```bash
bin/rails g component MyComponent
```

which can potentially create some confusion since it's not generating a "Rails component" but rather an "ActionView component". Having it under the ActionView module makes it explicit that the generator will creator an ActionView component.

We can move it under the Rails module if so desired.

lib/generators/action_view/component/templates/component.rb.tt
vinistock: I would only inherit from `ApplicationComponent` if it starts shipping with Rails (like `ApplicationRecord` or it the generator creates that file as well. Do you think we should generate that file too?

lib/generators/action_view/component/templates/component.rb.tt
vinistock: I'm not sure what the best interface for communicating whether the component has content or not would be. We could use thor to ask for user input indicating if they want their component to have it or not. Something like:

```ruby
def should_add_content?
  ask("Would you like your component to have children? (Y/n)").downcase == "y"
end
```

lib/generators/action_view/component/component_generator.rb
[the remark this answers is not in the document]
joelhawksley: If I had to pick, I would defer to matching the syntax from ActionMailer, `rails g component FooComponent`. 

The less surprise, the better 😄.

lib/generators/action_view/component/templates/component.rb.tt
joelhawksley: This is a tricky one. Could we look for an `ApplicationComponent` and inherit from it if it exists, otherwise inheriting from `ActionView::Component::Base`?

lib/generators/action_view/component/templates/component.rb.tt
joelhawksley: If we can do a check like ☝️ easily, I think it would be a nice touch ❤️ 

Suggested copy: `Would your like MyComponent to require content?`

lib/generators/action_view/component/component_generator.rb
```diff
@@ -0,0 +1,34 @@
...
+
+      argument :attributes, type: :array, default: [], banner: "attribute"
+      hook_for :test_framework
+
+      def create_component_file
+        template "component.rb", File.join("app/components",  "#{file_name}.rb")
+      end
+
+      def create_template_file
+        template "component.html.erb", File.join("app/components",  "#{file_name}.html.erb")
```
joelhawksley: Is it possible/expected to support other view renderers here? What happens in Rails if you have Slim installed and generate a view?

lib/rails/generators/test_unit/templates/component_test.rb.tt
```diff
@@ -0,0 +1,8 @@
+require "test_helper"
+
+class <%= class_name %>Test < ActiveSupport::TestCase
+  # test "component does something useful" do
+  #   component = <%= class_name %>.new
```
joelhawksley: It might be nice to use the `render_inline` helper here ❤️

lib/generators/action_view/component/component_generator.rb
vinistock: I'm not sure how we'd handle that. Is it possible to tell if the application is using Slim or Haml or any other format?

joelhawksley: I'm not sure.

I think it's reasonably safe to assume that ERB will be present in an application, so I'm OK with this.

lib/rails/generators/component/USAGE
```diff
@@ -0,0 +1,13 @@
+Description:
+============
+    Stubs out a new component with its respective template and test.
```
joelhawksley: `stub` is a testing-related word, IMO. How about:

```suggestion
    Creates a new component and test.
```

lib/rails/generators/component/USAGE
```diff
@@ -0,0 +1,13 @@
...
+    Pass the component name, either CamelCased or under_scored, and an optional list of attributes as arguments.
+
+Example:
+========
+    bin/rails generate component Profile name age
+
+    creates a Profile component class, test and template:
```
joelhawksley:
```suggestion
    creates a Profile component and test:
```

vinistock:
@joelhawksley it should be all in there. Look forward to contributing more!

joelhawksley:
@vinistock thanks so much! ❤️❤️ ❤️ 

I look forward to your next PR!

PR #108, opened 2019-11-14
Removing "components" from virtual path

juanmanuelramallo:
Rails default for translation shortcut paths do not include the parent folder where the translations are used.
For instance, when you use `t('.success')` in a controller, the translation key does not include the `controllers` folder.
Therefore, the word `components` get removed from the virtual path of components.
This will allow us to use `t('.title')` in components and expect the translation key not to include the `components` key as the top level key in the translation yml file.

What happens when the developer uses a different folder for components? The virtual path will just be the location of the file, in other words, the output of `Class#source_location`.

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -101,10 +101,9 @@ module ActionView
         @helpers ||= view_context
       end
 
-      # Looks for the source file path of the initialize method of the instance's class.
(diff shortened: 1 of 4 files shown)


juanmanuelramallo:
Does it make sense to you @joelhawksley ?

joelhawksley:
Looks good to me!

Issue #100, opened 2019-11-05
Compile components on application boot


joelhawksley:
In production, it would probably make sense to `compile` all descendents of `ActionView::Component::Base`, so that components render quickly on the first request, instead of compiling lazily as they do now.

kaspermeyer:
I'd be interested in giving this a go, @joelhawksley!

joelhawksley:
Thanks @kaspermeyer! I've assigned it to you. Feel free to unassign yourself if you decide not to take it on.

PR #112, opened 2019-11-19
Pre-compile templates

kaspermeyer:
This PR introduces pre-compilation of component templates during app launch.

I would've liked to write a test for this, but I'm not sure if it's feasible to test something that depends on eager loading inside an initializer, concisely?

Resolves: #100

### DIFF
--- a/lib/action_view/component/railtie.rb
+++ b/lib/action_view/component/railtie.rb
@@ -42,6 +42,12 @@ module ActionView
         end
       end
 
+      initializer "action_view_component.compile_templates" do
+        ActiveSupport.on_load(:after_initialize) do
+          ActionView::Component::Base.descendants.each(&:compile) if config.eager_load
+        end
+      end
+
       config.after_initialize do |app|
         options = app.config.action_view_component
 
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -157,8 +157,8 @@ module ActionView
         end
 
         # Compile templates to instance methods, assuming they haven't been compiled already.
-        # We could in theory do this on app boot, at least in production environments.
-        # Right now this just compiles the first time the component is rendered.
+        #
+        # Templates are pre-compiled during startup when eager loading is enabled in the app.
         def compile
           return if @compiled && ActionView::Base.cache_template_loading
 


lib/action_view/component/railtie.rb
```diff
@@ -42,6 +42,12 @@ class Railtie < Rails::Railtie # :nodoc:
         end
       end
 
+      initializer "action_view_component.compile_templates" do
+        ActiveSupport.on_load(:after_initialize) do
+          ActionView::Component::Base.descendants.each(&:compile) if config.eager_load
```
kaspermeyer: I'm not convinced it's strictly necessary to check for eager loading here since `.descendants` would most likely return an empty array if the classes are lazily loaded. I included it for good measure, as I find it a bit more intention revealing.

kaspermeyer:
Thanks for laying the groundwork for this, @juanmanuelramallo 👏

joelhawksley:
@kaspermeyer this looks good! Can you provide steps to verify this works as intended? Even just how to see it work in local dev.

kaspermeyer:
@joelhawksley absolutely! I created a demo repository to validate the initializer: https://github.com/kaspermeyer/pre-compiled-templates-demo

A small `Railtie` holds the initializer and compiling seems to occur on load like expected:
Screenshot 2019-11-20 at 13 51 16
However, when I tried to require `action_view/component/railtie` (where my initializer along others are located) in `application.rb`, the app crashed with a `NoMethodError`:

```
4: from /railties-6.0.1/lib/rails/initializable.rb:61:in `block in run_initializers'
3: from /railties-6.0.1/lib/rails/initializable.rb:32:in `run'
2: from /railties-6.0.1/lib/rails/initializable.rb:32:in `instance_exec'
1: from /railties-6.0.1/lib/rails/application/finisher.rb:123:in `block in <module:Finisher>'
/railties-6.0.1/lib/rails/application/finisher.rb:123:in `each': undefined method `eager_load!' for ActionView::Component:Module (NoMethodError)
``` 

I suspect it's because we're instructing Rails to eager load our `ActionView::Component` namespace, but we haven't made it autoloadable yet:
https://github.com/github/actionview-component/blob/2bc2bd83a72027b57b26a6c62fa45653246264ba/lib/action_view/component/railtie.rb#L13

[`ActiveSupport::Autoload`](https://api.rubyonrails.org/classes/ActiveSupport/Autoload.html) should be able help us with that. 

I noticed that our `ActionView::Component::Railtie` class isn't being required anywhere under `lib` yet, so the initializers aren't actually loaded at this point. Is this intentional? I can get started working on an autolading strategy, but I'm unsure if the class is still work in progress.

joelhawksley:
@kaspermeyer good detective work! Let's move forward on the autoloading strategy. 

Also, I'd like to see about a way to test this, even if it's in integration. It seems like something that would be easy to break otherwise 😬

kaspermeyer:
@joelhawksley Thanks! I'll get started on it — and I agree that it wouldn't be wise to leave these things untested. I'll do my best to make it happen.

kaspermeyer:
@joelhawksley Sorry for taking so long to get started on this. I haven't been able to set aside as much time to work on this as I would have liked to. I'm still up for it though! I'll get started as soon as I can.

joelhawksley:
@kaspermeyer no worries! I took your work and added some tests to resolve this in https://github.com/github/actionview-component/pull/158. ❤️

PR #120, opened 2019-11-21
Add rudimentary RSpec support to the component generator

dylnclrk:
## Motivation

The component generator does not support RSpec. 

When a project is configured to use RSpec and the component generator is run, no spec is created.

## Solution

@rdavid1099 and I added an RSpec hook to the component generator to create a corresponding spec. We also added some instructions to the README about configuring RSpec to include test helpers like `#render_inline`. 

## Open questions

* Does this functionality belong in actionview-component?
  * This seems useful to include, as establishing patterns for testing components with RSpec will be valuable for a number of users.
  * However, the RSpec generators and configuration should probably be moved in to rspec-rails when actionview-component is provided by Rails, as rails does not include RSpec generators.
* How can we test changes like this to generators? 
  * I'm not sure I know of an example of doing so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,7 +278,7 @@ end
 
 Specs created by the generator should now have access to test helpers like `render_inline`.
 
-To use component previews, you'll want to [configure the preview path](#previewing-components) so that preview
+To use component previews add this into your application.rb: `config.action_view_component.preview_path = "#{Rails.root}/spec/components/previews"`
 classes live somewhere besides the `test/` directory which is not present when RSpec is used.
 
 ## Frequently Asked Questions
--- a/README.md
+++ b/README.md
@@ -278,8 +278,11 @@ end
 
 Specs created by the generator should now have access to test helpers like `render_inline`.
 
-To use component previews add this into your application.rb: `config.action_view_component.preview_path = "#{Rails.root}/spec/components/previews"`
-classes live somewhere besides the `test/` directory which is not present when RSpec is used.
+To use component previews, set the following in `config/application.rb`:
+
+```ruby
+config.action_view_component.preview_path = "#{Rails.root}/spec/components/previews"
+```
 
 ## Frequently Asked Questions
 
--- a/README.md
+++ b/README.md
@@ -260,6 +260,23 @@ For example, if you want to use `lib/component_previews`, set the following in `
 config.action_view_component.preview_path = "#{Rails.root}/lib/component_previews"
 ```
 
+### Setting up RSpec
+
+If you're using RSpec, you can configure component specs to have access to test helpers. Create
+`spec/support/actionview_component.rb` with:
+
+```ruby
+require "action_view/component/test_helpers"
+
+RSpec.configure do |config|
+    # Ensure that the test helpers are available in component specs
+    config.include ActionView::Component::TestHelpers, type: :component
+end
+```
+
+And then ensure it's required in `spec/rails_helper.rb`. Specs created by the generator should now have access to the
+test helpers like `render_inline`.
+
 ## Frequently Asked Questions
 
 ### Can I use other templating languages besides ERB?
--- a/README.md
+++ b/README.md
@@ -278,6 +278,9 @@ end
 
 Specs created by the generator should now have access to test helpers like `render_inline`.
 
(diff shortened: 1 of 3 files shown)


dylnclrk:
@juanmanuelramallo, I agree about adding the config to `rails_helper.rb` instead of a new file. I’ll make that change. 

I’ll look in to getting previews to work. Having a config to override the default seems like a good first step.

Also, re: testing generators. I was looking through the RSpec-Rails docs and they have [some patterns](https://github.com/rspec/rspec-rails/blob/master/spec/generators/rspec/model/model_generator_spec.rb) for testing generators that we may be able to adapt here.

dylnclrk:
@juanmanuelramallo Oh, didn't realize that config was already there. I think that should be sufficient for now, I'll add a note to the RSpec section of the README. Let me know what you think.

What do folks think of using `lib/component_previews/` instead of `test/component_previews/` by default? Has there already been a discussion about why we should put previews in `test/`?

joelhawksley:
> What do folks think of using `lib/component_previews/` instead of `test/component_previews/` by default? Has there already been a discussion about why we should put previews in `test/`?

The [design philosophy of this project](https://github.com/github/actionview-component#design-philosophy) is to match existing/expected Rails behavior as much as possible.

To that end, component previews are designed to match the way ActionMailer previews work: https://guides.rubyonrails.org/action_mailer_basics.html#previewing-emails.

dylnclrk:
@joelhawksley 👍 That totally makes sense. After quickly looking in to it, it seems like [actionmailer's railtie](https://github.com/rails/rails/blob/ed9acb4fcc793ce1ab68a0e5076dc9458cc7f218/actionmailer/lib/action_mailer/railtie.rb#L33) doesn't do anything special for RSpec, either.

dylnclrk:
🤔 gven that the goal is to "match existing/expected Rails behavior as much as possible", do you think any of this RSpec stuff belongs here? I think it will help make it marginally easier for some to adopt and will help exhibit the ease of testing components. But it will need to be moved out when this is merged into Rails.

juanmanuelramallo:
> 🤔 gven that the goal is to "match existing/expected Rails behavior as much as possible", do you think any of this RSpec stuff belongs here? I think it will help make it marginally easier for some to adopt and will help exhibit the ease of testing components. But it will need to be moved out when this is merged into Rails.

I think it's a good start to have them here. Once this gem arrives into rails we can extract the rspec generator to a PR in `rspec-rails`.

README.md
```diff
@@ -278,6 +278,9 @@ end
 
 Specs created by the generator should now have access to test helpers like `render_inline`.
 
+To use component previews, you'll want to [configure the preview path](#previewing-components) so that preview
```
juanmanuelramallo: It'd be great if we could do this by default, as I believe action mailer works out of the box with rspec-rails. 

But it seems we would need to monkey patch `rspec-rails`. Take a look at https://github.com/rspec/rspec-rails/blob/d0f81dfccf92df655caeacfd1ce0ba1808779f99/lib/rspec-rails.rb#L40

Let me know what you think. We can do it in a different PR maybe, just to move this forward.

dylnclrk: Ah, that's interesting! Nice find. @juanmanuelramallo 

I'd vote to discuss that (and any other helpful RSpec changes we could make) in another PR if folks think what we've got here is useful enough on its own.

juanmanuelramallo: Maybe we could just tell them to set the preview path to the same one but using `spec` instead of `test`:

```suggestion
To use component previews add this into your application.rb: `config.action_view_component.preview_path = "#{Rails.root}/spec/components/previews"`
```

juanmanuelramallo:
>I'd vote to discuss that (and any other helpful RSpec changes we could make) in another PR if folks think what we've got here is useful enough on its own.

Yep I think we could add those in a different PR. 

Would be good to have @joelhawksley input regarding this

README.md
dylnclrk: Good suggestion. I like being more direct to get people up-and-running with minimal fuss.

dylnclrk:
Should I rebase and squash all of this in to one commit? Do folks have a preference?

joelhawksley:
Thank you for your contribution @dylnclrk @rdavid1099! ❤️ ❤️ ❤️

PR #123, opened 2019-11-21
Fix edge case issue with extracting variants from less conventional source_locations

rdavid1099:
Problem
=======
I encountered an edge case when trying to use variants. My local development environment has an additional `.` in it (`/example/fake.user/path/to/actionview-component`). The extracting the variant from the `source_location` assumes the path only has `.` in the respective files (`path.split(".").second.split("+")[1]&.to_sym`).

Solution
=======
Using similar patterns to how Rails extracts the variants from views (https://github.com/rails/rails/blob/d2e8b839d2e7236b5f85797f239b01ca2e783024/actionview/lib/action_view/template/resolver.rb#L272), I altered the extraction of the variants in `ActionView::Component::Base#templates`.

_Note_: I did not add any tests for this edge case since, at first, four tests dealing with variants were failing in my local environment. The fix applied has all tests now passing.

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -169,9 +169,13 @@ module ActionView
 
         private
 
+        def matching_views_in_source_location
+          (Dir["#{source_location.sub(/#{File.extname(source_location)}$/, '')}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location])
+        end
+
         def templates
           @templates ||=
-            (Dir["#{source_location.sub(/#{File.extname(source_location)}$/, '')}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location]).each_with_object([]) do |path, memo|
+            matching_views_in_source_location.each_with_object([]) do |path, memo|
               pieces = File.basename(path).split(".")
 
               memo << {
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -172,10 +172,12 @@ module ActionView
         def templates
           @templates ||=
             (Dir["#{source_location.sub(/#{File.extname(source_location)}$/, '')}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location]).each_with_object([]) do |path, memo|
+              pieces = File.basename(path).split(".")
+
               memo << {
                 path: path,
-                variant: path.split(".").second.split("+")[1]&.to_sym,
-                handler: path.split(".").last
+                variant: pieces.second.split("+")[1]&.to_sym,
+                handler: pieces.last
               }
             end
         end
--- a/test/action_view/component/base/unit_test.rb
+++ b/test/action_view/component/base/unit_test.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ActionView::Component::Base::UnitTest < Minitest::Test
+  def test_templates_parses_all_types_of_paths
+    file_path = [
+      "/Users/fake.user/path/to.templates/component/test_component.html+phone.erb",
+      "/_underscore-dash./component/test_component.html+desktop.slim",
+      "/tilda~/component/test_component.html.haml"
+    ]
+    expected = [
+      {variant: :phone, handler: "erb"},
+      {variant: :desktop, handler: "slim"},
+      {variant: nil, handler: "haml"}
+    ]
+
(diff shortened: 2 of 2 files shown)


lib/action_view/component/base.rb
```diff
@@ -172,10 +172,12 @@ def identifier
         def templates
           @templates ||=
             (Dir["#{source_location.sub(/#{File.extname(source_location)}$/, '')}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location]).each_with_object([]) do |path, memo|
+              pieces = File.basename(path).split(".")
+
               memo << {
                 path: path,
-                variant: path.split(".").second.split("+")[1]&.to_sym,
-                handler: path.split(".").last
+                variant: pieces.second.split("+")[1]&.to_sym,
```
juanmanuelramallo: What happens if the path contains several folders with dots in them?

rdavid1099: 💯 Great idea. I get a few tests in there for this edge case.

test/action_view/component/base/unit_test.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ActionView::Component::Base::UnitTest < Minitest::Test
```
rdavid1099: @juanmanuelramallo I added a new unit_test file to isolate the `templates` method and stub out several unique paths. I'm not sure if you want to add unit tests or relocate this test to an existing file. Tell me what you think 😄

juanmanuelramallo: This is good. But I feel it would be better to maintain the current approach of testing. What do you think?

That would be about creating new components in `/test/app/components` and then use those in your test cases in `/test/action_view/component_test.rb` so that we see can assert that the output of the rendering is what we expect.

Those new components can have the path names you used in here:
- fake.user/path/to.templates/component/test_component.html+phone.erb
- /_underscore-dash./component/test_component.html+desktop.slim
- /tilda~/component/test_component.html.haml

lib/action_view/component/base.rb
```diff
@@ -169,13 +169,19 @@ def identifier
 
         private
 
+        def matching_views_in_source_location
```
juanmanuelramallo: If we follow this comment https://github.com/github/actionview-component/pull/123/files#r349892608 we may not need to extract this into a method right? Cause it seems like it was extracted just to be able to stub it in the test no?

test/action_view/component/base/unit_test.rb
rdavid1099: Sounds good. I'll rework this test and follow the existing testing patterns.

rdavid1099: @juanmanuelramallo I'm running into an issue with initializing the component due to Rails conventions.

I have the component and its respective view in the directory path `test/app/components/unconventional.path/_underscore-dash./tilda~/`. When I run the test for this unconventional path:
```ruby
  def test_renders_components_with_less_conventional_paths
    with_variant :phone do
      result = render_inline(UnconventionalPathComponent)

      assert_includes result.text, "Phone"
    end
  end
```
I get a NameError `uninitialized constant ActionView::ComponentTest::UnconventionalPath`. This makes sense since the component is not name spaced using the unconventional directory names. Any ideas on how to get this test to work?

joelhawksley:
@rdavid1099 @juanmanuelramallo any thoughts on how we'd like to proceed with this PR?

test/action_view/component/base/unit_test.rb
juanmanuelramallo: Sorry for the delay, I lost track of time. 

Maybe we have two options:
1. Use modules to namespace according to the folders (following rails convention)
2. Require the file directly in the spec_helper or in the test file

Anyways I now see that the way you tested it is great! So maybe we can make this go forward as it is, and then we can add some tests following the other approach in a follow up PR

lib/action_view/component/base.rb
juanmanuelramallo: One last question, what should happen if the file basename starts with a dot? I don't think we will ever encounter that scenario but better safe than sorry.

juanmanuelramallo:
> @rdavid1099 @juanmanuelramallo any thoughts on how we'd like to proceed with this PR?

I think it's in a good state. 

We can go back to it later and add some tests following the other approach, my only comment was around the testing files.

rdavid1099:
Thanks @joelhawksley @juanmanuelramallo. My schedule's getting packed with "end of year" chaos. Glad I could help out 😄

Issue #122, opened 2019-11-21
[1.5.0] Migration from version 1.4.0


afuno:
Could you please tell about migration from version 1.4.0?

I updated the gem.

I changed this line:

```ruby
require 'action_view/component/base'
```

To this:

```ruby
require 'action_view/component'
```

After that, I can't start the project:

```
% rails s
=> Booting Puma
=> Rails 6.0.1 application starting in development
=> Run `rails server --help` for more startup options
Exiting
Traceback (most recent call last):
	77: from bin/rails:5:in `<main>'
	76: from bin/rails:5:in `load'
	75: from /Users/afuno/Projects/my.app/bin/spring:16:in `<top (required)>'
	74: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	73: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
	72: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/spring-2.1.0/lib/spring/binstub.rb:11:in `<top (required)>'
	71: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/spring-2.1.0/lib/spring/binstub.rb:11:in `load'
	70: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/spring-2.1.0/bin/spring:49:in `<top (required)>'
	69: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/spring-2.1.0/lib/spring/client.rb:30:in `run'
	68: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/spring-2.1.0/lib/spring/client/command.rb:7:in `call'
	67: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/spring-2.1.0/lib/spring/client/rails.rb:28:in `call'
	66: from /Users/afuno/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/spring-2.1.0/lib/spring/client/rails.rb:28:in `load'
	65: from /Users/afuno/Projects/my.app/bin/rails:11:in `<top (required)>'
[...]

PR #124, opened 2019-11-21
[#122] Fix railtie

juanmanuelramallo:
- Thanks to @franciscoj for sharing the fix
- It was failing in Rails 6 due to updating a frozen array
- Updated railties class to work with Rails 6 (source: https://github.com/rails/rails/blob/master/actionmailer/lib/action_mailer/railtie.rb)

### DIFF
--- a/lib/action_view/component/railtie.rb
+++ b/lib/action_view/component/railtie.rb
@@ -27,6 +27,14 @@ module ActionView
         end
       end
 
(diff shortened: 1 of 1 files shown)


joelhawksley:
🙇

joelhawksley:
I'll cut a release with this later today ❤️

Issue #126, opened 2019-11-22
Initializer crashes when eager loading is enabled


kaspermeyer:
When eager loading is enabled, one of the initializers:
https://github.com/github/actionview-component/blob/1aad6f4b9ccdd87c15491d31785dd1c1da5b992e/lib/action_view/component/railtie.rb#L44-L48
crashes with an error:
```
5: from /activesupport-6.0.1/lib/active_support/lazy_load_hooks.rb:67:in `execute_hook'
4: from /activesupport-6.0.1/lib/active_support/lazy_load_hooks.rb:62:in `with_execution_control'
3: from /activesupport-6.0.1/lib/active_support/lazy_load_hooks.rb:74:in `block in execute_hook'
2: from /activesupport-6.0.1/lib/active_support/lazy_load_hooks.rb:74:in `instance_eval'
1: from /actionview-component-1.5.1/lib/action_view/component/railtie.rb:46:in `block (2 levels) in <class:Railtie>'
/actionview-component-1.5.1/lib/action_view/component/railtie.rb:46:in `each': undefined method `action_methods' for EagerComponent:Class (NoMethodError)
```
It makes sense, since `#action_methods` isn't implemented on `ActionView::Component::Base`.

Can you shed some light on this, @juanmanuelramallo? Was the intention to mimic the caching in [`AbstractController`](https://apidock.com/rails/v6.0.0/AbstractController/UrlFor/ClassMethods/action_methods)? Thanks 🙌

PR #127, opened 2019-11-22
Disable initializer `action_view_component.eager_load_actions`

kaspermeyer:
This PR disables an initializer that crashes when eager loading is enabled.

Since attempts to eager load are already disabled for now (https://github.com/github/actionview-component/commit/6dc979793d5eb29478f3fcbfd7408405e767254a) I think it makes sense to disable this initializer as well. When its intent has been sorted out and I get started working on strategies to enable eager/autoloading, it'll probably make sense to enable it again.

Resolves: #126

### DIFF
--- a/lib/action_view/component/railtie.rb
+++ b/lib/action_view/component/railtie.rb
@@ -41,11 +41,12 @@ module ActionView
         end
       end
 
-      initializer "action_view_component.eager_load_actions" do
-        ActiveSupport.on_load(:after_initialize) do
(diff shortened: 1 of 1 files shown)


kaspermeyer:
@joelhawksley It might make sense to bump to version 1.5.2 for this, since every environment with eager loading enabled, like production, will have issues with this.

I've made small demo application to verify the issue: 
https://github.com/kaspermeyer/action-methods-initializer

juanmanuelramallo:
Haha this is a clear example of what a copy/paste can do! 
Sorry for the issues!

PR #129, opened 2019-11-23
Add to_component_class to active record base

### LINKED ISSUE #26: Support rendering models via to_component_class
Per @inopinatus's [comment](https://github.com/github/actionview-component/issues/15#issuecomment-524521196), let's look at supporting `render @model`, where `@model` is an object that implements `to_component_class`.

vinistock:
Closes #26.

This PR aims to implement the `to_component_class` which should map a model to its respective component class. Furthermore, it implements a new branch in the render monkey patch to deal with this particular case.

A couple of concerns for feedback:
1. In order to figure out how to map the record's attributes to the component's attributes, I am listing the parameters from the `initialize` method and slicing the record's attributes. I'm not sure this is the best way forward.
2. Adding active_record to the dummy app forced me to add `database.yml`. Is there a way to get around that?
3. Should we split the ActiveRecord monkey patch into another file?


lib/action_view/component/monkey_patch.rb
```diff
@@ -25,3 +31,11 @@ def render(options = {}, args = {}, &block)
...
   prepend RenderMonkeyPatch
 end
+
+class ActiveRecord::Base
+  def to_component_class
+    class_name = "#{self.class.name}Component"
+
+    class_name.constantize if defined?(class_name)
+  end
+end
```
juanmanuelramallo: Should we move this to its own file?

tomasc:
Thank you for the PR, however please let’s not go this route. Components do not necessarily map to models, in fact their strength is in their conceptual decoupling from models. If this PR is merged a relationship between models and components will be implied, which is a sort antithesis of the whole idea of components.

That being said, if you prefer to use components like this, it should be very easy to implement this in your application, eventually you could make a small gem based on this PR that provides the glue between ActionView::Component and ActiveRecord.

juanmanuelramallo:
Hey @tomasc, this change is solving issue #26. Which was introduced [https://github.com/github/actionview-component/issues/15#issuecomment-524521196](here).

IMHO
```erb
<%= render @book %> 
```

looks a bit better than
```erb
<%= render BookComponent, book: @book %>
```

IMHO this change won't imply that an ActionView::Component **must** be coupled to a model. This is just about making components more railsy. Allowing them to work as common action view partials do. 

Do you still think this change don't belong here?

tomasc:
@juanmanuelramallo hmm, if I understand correctly then I still think this should not be merged.

I have previously used [Draper](https://github.com/drapergem/draper) which, in a way, does what you are proposing: it decorates models with view-related methods (at least that’s how I used it). Over time I found this very misleading: the views, conceptually, should not mirror models. For example you might have a component called `Book`, which has `title`, `author` and `isbn` attributes. These however might (and in many cases should) come from different sources: models, API, value objects, what have you.

The PR you are proposing seems Rails-y, mainly because it looks good and obvious, but it would promote architecture that I believe is already supported through standard Rails views. What I like on the ActionView::Components is that it implements a new, imho more modern, architecture in addition to what is already in Rails today (MVC). And by _not_ supporting direct relation between model and component it would be Rails-y, because opinionated.

But let’s have others chime-in. @joelhawksley what do you think?

lib/action_view/component/monkey_patch.rb
```diff
@@ -25,3 +31,11 @@ def render(options = {}, args = {}, &block)
 
   prepend RenderMonkeyPatch
 end
+
+class ActiveRecord::Base
```
juanmanuelramallo: I checked how active record handles this, but it seems like it does not. It's a responsibility of ActiveModel, so do you think we should use `ActiveModel::Model` here instead?

What would be even better I think is to update this module https://github.com/rails/rails/blob/master/activemodel/lib/active_model/conversion.rb with this new `to_component_class` method.

juanmanuelramallo: @tomasc it seems like this would solve your concern right?

vinistock:
I don't think having multiple ways of doing things in Rails should be a concern. That's pretty common in the framework and most alternative implementations don't stop you from using other types.

Having that said, I understand the concern of needing more information that doesn't all come from a single model. We could also augment this to accept extra arguments in the following way:

```ruby
<%= render @record, some_other_thing: :something %>
```

Also, you could still use the regular rendering with this implementation, but I do believe it needs to be Railsy if we ever intend to upstream this. It will probably have a much lower chance of being accepted if it doesn't fit the Rails way of doing things.

Additionally, let's say we have an active record relation (say `@posts`). Do you a suggestion of how we could avoid having these loops all over our views?

```ruby
<% @posts.each do |post| %>
  <%= render PostComponent, title: post.title %>
<% end %>
```

With the addition of `to_component_class`, the rendered can figure it out on its own and it can simplify the loops to simply doing: 

```ruby
<%= render @posts %>
```

Let me know more of your thoughts.

tomasc:
Hi @vinistock,

I would propose to extend the [standard way of rendering collections](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-collections) in Rails for example like this:

```ruby
<%= render component: BookComponent, collection: […], locals: {…} %>
```

Where the collection argument is is an Array of objects that respond to arguments required by the PostComponent (for example `:title`, `:author` and `:isbn`).

As a consequence, this would, again, promote passing specific arguments to a component, instead of simply handing over whole models. It is more verbose, yes, but I don't think the most concise way of writing render statements is the necessary goal here.

lib/action_view/component/monkey_patch.rb
joelhawksley: Agreed.

lib/action_view/component/monkey_patch.rb
```diff
@@ -15,6 +15,12 @@ def render(options = {}, args = {}, &block)
         options.render_in(self, &block)
       elsif options.is_a?(Class) && options < ActionView::Component::Base
         options.new(args).render_in(self, &block)
+      elsif options.respond_to?(:to_component_class) && !options.to_component_class.nil?
```
joelhawksley: Might we move this case to be last? (Besides the fallback)

vinistock:
I've split the patches and re-ordered the render hierarchy. Let me know if you got other concerns.

Particularly, is there anyway to get rid of `database.yml` if we're only using sqlite3 in memory?

lib/action_view/component/ar_monkey_patch.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ActiveRecord::Base
+  def to_component_class
+    class_name = "#{self.class.name}Component"
+
+    class_name.constantize if defined?(class_name)
```
kaspermeyer: ActiveSupport implements [`#safe_constantize`](https://apidock.com/rails/v6.0.0/ActiveSupport/Inflector/safe_constantize) which returns `nil` if the constant isn't defined, making it possible to refactor this method into a one-liner:
```ruby
def to_component_class
  "#{self.class.name}Component".safe_constantize
end
```
This is only a matter of style though, there's nothing wrong with the implementation as it is 🙂

kaspermeyer:
Great idea @vinistock! I really like the idea of moving closer to the Rails way of rendering and being able to use the shorthand syntax with components:

```ruby
<%= render @post %>
```

My only concern about this, is the fact that we're creating a really tight coupling between the model and the component, when we map all the individual attributes from the model to the component. It's not obvious that you would have to update the parameters on the component initializer every time an attribute is added/removed/renamed on the model itself.

Could we perhaps pass the model instance instead? It would create a much looser coupling and it would mimic the behavior we're used to from partial rendering where `<%= render @post %>` render the `_post.html.erb` partial **and** passes a local named `post` that refers to the model. 

An added bonus of passing the model instead of individual attributes is that any class could implement `#to_component_class` and it would still work as expected.

lib/action_view/component/ar_monkey_patch.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class ActiveRecord::Base
```
juanmanuelramallo: Were you able to take a look at this https://github.com/github/actionview-component/pull/129#discussion_r350170342 ?

vinistock:
@kaspermeyer I agree that the way of passing the attributes to the component might feel a little awkward. However, I don't believe we should pass the instance.

If we pass it, then it allows for querying inside the components, which I understand is not the desired concept.

For example:

```ruby
class MyComponent < ActionView::Component
  def initialize(record)
    @stuff = record.more_records
  end
end
```

We could think of some other way of defining the component's properties that makes extracting those from the model more elegant. I'm open to suggestions.

lib/action_view/component/ar_monkey_patch.rb
vinistock: Will change it to ActiveModel.

joelhawksley:
@vinistock @kaspermeyer IMO we _should_ just pass the instance into the component, as that's how `to_partial_path` handles it. 

I'd prefer we stick to matching existing behavior, at least for now.

vinistock:
@joelhawksley @kaspermeyer okay. Moved the patch to ActiveModel and started accepting the instance in the component.

lib/action_view/component/am_monkey_patch.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
```
joelhawksley: How about we call this file `active_model_conversion_monkey_patch`? I'd prefer to avoid abbreviations/shorthand.

lib/action_view/component/monkey_patch.rb
```diff
@@ -17,6 +17,8 @@ def render(options = {}, args = {}, &block)
         options.new(args).render_in(self, &block)
       elsif options.is_a?(Hash) && options.has_key?(:component)
         options[:component].new(options[:locals]).render_in(self, &block)
+      elsif options.respond_to?(:to_component_class) && !options.to_component_class.nil?
```
joelhawksley: In what case does the `nil?` check help us here?

vinistock: Because we are adding the `to_component_class` to every model, we need to check that the model does indeed have a corresponding component class before trying to render it.

It will return nil if the component class doesn't exist for the model.

vinistock:
@joelhawksley let me know if you like the copy.

README.md
```diff
@@ -163,6 +163,24 @@ Components can be rendered via:
 
 `render(component: TestComponent, locals: { foo: :bar })`
 
+**Rendering components through models**
+
+Passing model instances will cause render to look for its respective component class.
```
```suggestion
Passing model instances will cause `render` to look for its respective component class.
```

README.md
```diff
@@ -163,6 +163,24 @@ Components can be rendered via:
...
+
+The component is instantiated with the rendered model instance.
+
+Example for a `Post` model.
```
```suggestion
Example for a `Post` model:
```

joelhawksley:
Thanks for yet another wonderful contribution @vinistock! ❤️

PR #130, opened 2019-11-25
Line 54

jymok12:
Common Ruby code tools do not handle views which makes it difficult to test for auditing thoroughness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ Currently, Rails encourages testing views via integration or system tests. This 
 
 #### Code Coverage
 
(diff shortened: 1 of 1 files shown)


joelhawksley:
@jymok12 thank you for your contribution! ❤️ 

Might you care to provide a reason for this change? I think it's important in this case to emphasize how the lack of coverage tools for views can lead to testing blindspots.

joelhawksley:
Closing due to inactivity. Please feel free to reopen if desired ❤️

PR #137, opened 2019-11-26
Eliminate sqlite dependency

spdawson:
In #129, `to_component_class` support was added, to allow an Active Model instance to be conveniently rendered using an associated component class.

However, this work initially targeted ActiveRecord, which necessitated a dependency on sqlite for testing. This dependency can be eliminated by using Active model (instead of ActiveRecord) in the tests.


joelhawksley:

juanmanuelramallo:
Loved it! Thanks

PR #139, opened 2019-11-28
Use source location as identifier

vinistock:
I'm working to add components to one of our projects and stumbled upon an incompatibility with one of our gems.

The gem is [better_hml](https://github.com/Shopify/better-html) and the exact line is [here](https://github.com/Shopify/better-html/blob/b5eb06f1f98b7d8d401a61afa6a62d3f2b090f4a/lib/better_html/better_erb.rb#L43).

It uses the template identifier to figure out the extension of the file.

I looked at Rails to try to understand what the template identifier usually is and it seemed to me that it was just the file path. I'm unsure if this would case any problems for the gem or why it was an empty string before, so please do provide feedback.

With this fix, I'm able to render components in our project 😄.

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -164,7 +164,7 @@ module ActionView
         end
 
         def identifier
-          ""
+          source_location
         end
 
         private


lib/action_view/component/base.rb
```diff
@@ -164,7 +164,7 @@ def type
         end
 
         def identifier
-          ""
+          source_location
```
juanmanuelramallo: But this is the source location of the component class, not the html itself no? 

And you would need to it per variant I think 🤔

vinistock: You're 100% right. I overlooked the `source_location` path. We have the template path under `templates.first[:path]`. However, I'm not sure how we'd handle for multiple templates.

Also, how would we take the variant into account?

joelhawksley: It should be possible to return the template for the current variant, right?

joelhawksley:
@vinistock thanks for raising this concern. I'd be cool with you adding the `better-html` gem to the test app inside the project so we can have a test against this case.

lib/action_view/component/base.rb
vinistock: The thing I'm having trouble figuring out is that the better_html gem receives only the component class and not the instance. And identifier is a class method.

I'm trying to understand if the code for better_html would need to be enhanced for components or if we can somehow figure out the variant without having the component's instance (only the class and the source for the template are available at that exact point in the code I linked).

Let me know if you have ideas on how to adequately solve this.

joelhawksley: @vinistock sounds like we might be due for some pairing on this one. Might you be up for some time next week?

vinistock: That makes sense. I'll shoot you an email.

Issue #141, opened 2019-12-03
Should variants fall back to default template?


asgerb:
Thanks a lot for the amazing work! I'm really happy to be using this gem, it's bringing some much needed clarity and testability to my views 🙂

I'm finding the newly implemented variants logic a bit surprising though. Notably the `:variant_exists` validation, which forces me to add templates for every variant. 

`ActionPack::Variants` simply falls back to the default template, in case there is no template defined for the variant and I find this to be a nicer way to go about it as it's a more flexible approach, allowing you to define variant templates only for the components that need it.

What are your thoughts? I'm very open to submitting a PR, but wanted to hear you out first 🙂

joelhawksley:
@asgerb if that's how things work in Rails, a PR would be more than welcome!

joelhawksley:
Closed by https://github.com/github/actionview-component/pull/147 ❤️

PR #142, opened 2019-12-04
Fallback to default template if there is no template defined for a variant

asgerb:
After poking around for a bit, I found a very simple change to implement this feature (#141). I'm not a big fan of the change though – it looks and feels bad – and so I'm not sure this PR should be merged 😉 Perhaps instead it can form the basis for a discussion on the way rendering is handled. 

To me it seems we could rely more on `ActionView` to do the heavy lifting around template lookups and rendering, although it's very possible I'm missing some important nuances, as this is my first time taking a look at the codebase.

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -53,7 +53,13 @@ module ActionView
         @content = view_context.capture(&block) if block_given?
         validate!
 
-        send(self.class.call_method_name(@variant))
+        call_method_name = if self.respond_to?(self.class.call_method_name(@variant))
+          self.class.call_method_name(@variant)
+        else
+          "call"
+        end
+
+        send(call_method_name)
       ensure
         @current_template = old_current_template
       end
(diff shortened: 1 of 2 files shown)


lib/action_view/component/base.rb
```diff
@@ -55,7 +53,13 @@ def render_in(view_context, *args, &block)
         @content = view_context.capture(&block) if block_given?
         validate!
 
-        send(self.class.call_method_name(@variant))
+        call_method_name = if self.respond_to?(self.class.call_method_name(@variant))
```
joelhawksley: This decision tree ends up being pretty similar to `.call_method_name`. Might it be possible to combine the two?

cesariouy:
Thanks @asgerb and @joelhawksley! I've taken this work and used it in #147, incorporating Joel's suggestion.

joelhawksley:
Closed in https://github.com/github/actionview-component/pull/147

PR #143, opened 2019-12-04
Generate test with new naming convention

traels:
Test was created with `render_inline(Thing)` which should be `render_inline(ThingComponent)`.

Also created very basic testcase for generator.


joelhawksley:
❤️ thank you for your contribution!

PR #146, opened 2019-12-09
Avoid dropping elements in the render_inline test helper and adjust tests

### LINKED ISSUE #145: render_inline produces bad output for some templates
The following line in the test helpers produces incorrect results depending on the template being rendered:

https://github.com/github/actionview-component/blob/99fc8dd285cf98f3121fff45f590d6be5da550bc/lib/action_view/component/test_helpers.rb#L7

This isn't a consistent thing though -- it chokes on certain types of input. Consider the following:

```ruby
Nokogiri::HTML('<script>alert("OK")</script>').css('body > *')
```

This returns an empty string. This will similarly produce empty strings depending on the input -- using `noframes`, `frame`, various other elements will all produce weird input. Here's some other fun results:

```ruby
Nokogiri::HTML("<script>alert</script><b>ok</b>").css("body > *").to_s
# => "<b>ok</b>"
Nokogiri::HTML("<script>alert</script>").css("body > *").to_s
# => ""
Nokogiri::HTML("hello world").css("body > *").to_s
# => "<p>hello world</p>"
```

I'd recommend doing the following instead perhaps:

```ruby
Nokogiri::HTML.fragment(controller.view_context.render(component, args, &block))
```

Which produces the expected result of all of the above examples:

```ruby
Nokogiri::HTML.fragment("<script>alert</script><b>ok</b>").to_s
# => "<script>alert</script><b>ok</b>"
Nokogiri::HTML.fragment("<script>alert</script>").to_s
# => "<script>alert</script>"
Nokogiri::HTML.fragment("hello world").to_s
# => "hello world"
```

dark-panda:
The main thrust of this change is to avoid dropping elements by using
`.css("body > *")`, which will strip out some elements in some
situations and produce unexpected test results. At the same time, we'll
clean up some tests to ensure that whitespace stripping is more
consistent and test arguments are given in a more consistent order.

Fixes #145.

### DIFF
--- a/lib/action_view/component/test_helpers.rb
+++ b/lib/action_view/component/test_helpers.rb
@@ -4,7 +4,7 @@ module ActionView
   module Component
     module TestHelpers
       def render_inline(component, **args, &block)
-        Nokogiri::HTML(controller.view_context.render(component, args, &block)).css("body > *")
+        Nokogiri::HTML.fragment(controller.view_context.render(component, args, &block))
       end
 
       def controller
--- a/test/action_view/component_test.rb
+++ b/test/action_view/component_test.rb
@@ -8,15 +8,15 @@ class ActionView::ComponentTest < Minitest::Test
   def test_render_inline
     result = render_inline(MyComponent)
 
-    assert_equal trim_result(result.first.to_html), "<div>hello,world!</div>"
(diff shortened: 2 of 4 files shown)


joelhawksley:
👏 wonderful!

PR #147, opened 2019-12-09
Variant template fallback

cesariouy:
Shout out to @asgerb; this is practically a clone of #142, but with one commit from me that moves the logic to `ActionView::Component::Base.call_method_name` per [Joel's comment](https://github.com/github/actionview-component/pull/142#discussion_r354556290).


joelhawksley:
👏 nice teamwork folks!

PR #148, opened 2019-12-09
feat(previews): pass content block when rendering

cesariouy:
Allow component previews to use content that is passed to the `render` function as a block.

### DIFF
--- a/test/test/components/previews/preview_component_preview.rb
+++ b/test/test/components/previews/preview_component_preview.rb
@@ -8,4 +8,8 @@ class PreviewComponentPreview < ActionView::Component::Preview
   def without_cta
     render(PreviewComponent, title: "More lorem...")
   end
+
+  def with_content
+    render(PreviewComponent, title: "title") { "some content" }
+  end
 end
--- a/lib/action_view/component/preview.rb
+++ b/lib/action_view/component/preview.rb
@@ -30,8 +30,8 @@ module ActionView
       extend ActiveSupport::DescendantsTracker
       include ActionView::Component::TestHelpers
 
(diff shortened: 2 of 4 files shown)


test/test/components/previews/preview_component_preview.rb
```diff
@@ -8,4 +8,8 @@ def default
   def without_cta
     render(PreviewComponent, title: "More lorem...")
   end
+
+  def with_content
+    render(PreviewComponent, title: "title") { "some content" }
```
joelhawksley: What happens if someone wants to pass in content that is ERB? 

We're going to run into the same issue when we add support for rendering components from controllers, as we can't simply allow passing in ERB inside a block.

@juanmanuelramallo any thoughts on how we might enable previews to support ERB blocks? I'm wondering if we'll need to define a template for each preview that then renders the component.

juanmanuelramallo: Sorry for the late response. 

The previews are using the [test helpers' method `render_inline` ](https://github.com/github/actionview-component/blob/master/lib/action_view/component/test_helpers.rb#L7), so as long as we support ERB in there we should be able to support it here too (I'm not sure if it currently does but I think it does not right?)

Anyways we would also need to update [this line from the preview object](https://github.com/github/actionview-component/blob/master/lib/action_view/component/preview.rb#L12) to also receive and pass a block.

joelhawksley: `render_inline` does not support being passed ERB. That being said, I think I'm OK with this change for now. We can always revisit it later 😄

PR #158, opened 2019-12-19
Compile components on application boot when eager_load = true

joelhawksley:
(Based on [existing work](https://github.com/github/actionview-component/pull/112) by @kaspermeyer)

This PR introduces pre-compilation of component templates during app launch.

It was a PAIN to test, but I ended up splitting out the invalid components into their own suite so we could enable eager loading in the main suite.

As an added benefit, this means our suite will more closely mimic production behavior 🙌 

Resolves: #100


joelhawksley:
@juanmanuelramallo I'm not sure. Would you care to investigate?

juanmanuelramallo:
@joelhawksley I will investigate and share my findings in an issue maybe?

Issue #144, opened 2019-12-06
Allow components to have multiple content areas


kylefox:
First of, this library looks _awesome_ and I can't wait to start using it :)

Something that would be really nice is the ability for a component to have more than one output area, or _"outlet"_ (for lack of a better term). Essentially, the idea is very similar to using `yield(:foo)` and `content_for(:foo)` in normal Rails views.

Often, the dynamic content within a component cannot be declared within a single block. One example is a modal window that has a title, body, and row of buttons:

![image](https://user-images.githubusercontent.com/37838/70283669-45f21580-177f-11ea-94be-c8138c9daf62.png)

This would be a really cool way to render this modal with a title (static string), body (block), and footer (block). The block could now be passed an instance of the component (or perhaps something like component builder) which could act to capture explicitly named blocks:

```erb
<%= render(ModalComponent, title: 'Are you sure?') do |modal| %>
  <%= modal.body do %>
    <p>This is some content that would go in the modal body.</p>
  <% end %>

  <%= modal.footer do %>
    <%= link_to 'Cancel', '#', class: 'btn btn-secondary', data: { dismiss: :modal } %>
    <%= button_to 'Confirm', confirmation_path, class: 'btn btn-primary' %>
  <% end %>
<% end %>
```

The component template might look something like this:

```erb
<!-- app/components/modal_component.html.erb -->
<div class="modal-dialog">
  <div class="modal-content">
    <div class="modal-header">
      <h4 class="modal-title"><%= title %></h4>
    </div>

    <div class="modal-body">
      <%= body %>
    </div>

    <div class="modal-footer">
      <%= footer %>
    </div>
  </div>
</div>
```

I think this would make components far more flexible, and allow us to write less repetitive markup when using the components. Without something like this, the developer must remember to include `<div class="modal-body">` and `<div class="modal-footer">` whenever the component is used:

```erb
<%= render(ModalComponent, title: 'Are you sure?') do %>
  <div class="modal-body">
    <p>This is some content that would go in the modal body.</p>
  </div>

  <div class="modal-footer">
    <%= link_to 'Cancel', '#', class: 'btn btn-secondary', data: { dismiss: :modal } %>
    <%= button_to 'Confirm', confirmation_path, class: 'btn btn-primary' %>
  </div>
<% end %>
```

The need to wrap content in these classes feels like an implementation detail that should be the responsibility of the component itself.

[...]

PR #161, opened 2019-12-21
ContentArea support #144

bostanio:
Support ContentAreas which allow rendering of this form:

app/components/modal_component.rb`:
```ruby
class ModalComponent < ActionView::Component::Base
  validates :header, :body, presence: true

  attr_reader :user
  with_content_areas :header, :body

  def initialize(user:)
    @user = user
  end
end
```

`app/components/modal_component.html.erb`:
```erb
<div class="modal">
  <div class="header"><%= header %></div>
  <div class="body"><%= body %>"></div>
</div>
```

We can render it in a view as:

```erb
<%= render(TestComponent, user: {name: 'Jane'}) do |component| %>
  <% component.with(:header) do %>
    Hello <%= user[:name] %>
  <% end %>
  <% component.with(:body) do %>
    <p>Have a great day.</p>
  <% end %>
<% end %>
```

Which returns:

```html
<div class="modal">
  <div class="header">Hello Jane</div>
  <div class="body"><p>Have a great day.</p></div>
</div>
```


lib/action_view/component/content_for.rb
```diff
@@ -0,0 +1,37 @@
...
+      private
+
+      def set_attribute(attr, value)
+        setter_name = "#{attr}=".to_sym
+        instance_variable_name = "@#{attr}".to_sym
+        if respond_to? setter_name
+          send(setter_name, value)
+        else
+          instance_variable_set(instance_variable_name, value)
+        end
```
bostanio: it seems nice to call the setter if it exists other wise assume that its an instance variable that backs the attribute

test/action_view/component_test.rb
```diff
@@ -108,11 +108,20 @@ def test_renders_partial_template
...
   end
 
+  # def test_renders_content_for_template_with_block
+  #   result = render_inline(ContentForComponent, footer: "Bye!") {|comp| "<% content_for(:title) { 'My Title'} %>Have a nice day."}
+  #
+  #   assert_includes result.text, "Hi!"
+  #   assert_includes result.text, "Have a nice day."
+  #   assert_includes result.text, "Bye!"
+  #   assert_html_matches "<div>\nMy Title\nHave a nice day.\nBye!\n</div>", result.to_html
+  # end
```
bostanio: I really want to be able to unit test the various options but I can't figure out the incantations of providing an ERB template. I assume there is a way to do it but I haven't gone spelunking through the Rails helper tests to figure it out yet.

test/action_view/integration_test.rb
```diff
@@ -35,6 +35,15 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     HTML
   end
 
+  test "rendering component with content_for" do
+    get "/content_for"
+    assert_response :success
+
+    assert_includes response.body, "<h1>Hi!</h1>"
+    assert_includes response.body, "<p>Have a great day.</p>"
+    assert_includes response.body, "<h3>Bye!</h3>"
```
bostanio: I'd like to assert all the HTML but there are whitespace and new line differences that are a pain to account for.

test/app/components/content_for_component.html.erb
```diff
@@ -1,2 +1,6 @@
-<% content_for(:foo, "bar") %>
-<div>Hello content for</div>
+<%# content_for(:foo, "bar") %>
```
bostanio: as written uncommenting this content_for breaks because `:foo` isn't an attribute of the component. As mentioned in #144 I think we likely don't want to have a naming collision to avoid confusion like this.

test/app/components/content_for_component.rb
```diff
@@ -1,5 +1,12 @@
 # frozen_string_literal: true
 
 class ContentForComponent < ActionView::Component::Base
-  def initialize(*); end
+  validates :content, :title, :footer, presence: true
+
+  attr_reader :content, :title, :footer
+
+  def initialize(title: nil, footer: nil)
```
bostanio: if you're going to set via content_for then things have to be optional in the initializer.

lib/action_view/component/content_for.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module ActionView
+  module Component # :nodoc:
+    module ContentFor
+      def content_for(attr, content = nil, options = {}, &block)
+        unless respond_to?(attr)
+          raise StandardError.new "Called content_for with unknown attribute '#{attr}'"
+        end
```
bostanio: this seems like a helpful check to prevent mistakes

lib/action_view/component/content_for.rb
```diff
@@ -0,0 +1,37 @@
...
+
+        if block_given?
+          options = content if content
+          content = view_context.capture(&block)
+        end
+
+        existing_content = send(attr)
+        if existing_content && !options[:flush]
+          content = existing_content + content
+        end
```
bostanio: this mimics the options in regular content_for

test/app/views/application/content_for.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render ContentForComponent do |comp| %>
+  <p>Have a great day.</p>
+  <% comp.content_for(:title) do %>
+    <h1>Hi!</h1>
```
joelhawksley: For sanity's sake, might we evaluate some Ruby inside these blocks? Even something like:

```suggestion
    <h1>Hi! <%= 1+ 2 %></h1>
```

test/app/views/application/content_for.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render ContentForComponent do |comp| %>
```
joelhawksley: Nitpick: use full words for intermediate variables:

```suggestion
<%= render ContentForComponent do |component| %>
```

test/config/routes.rb
```diff
@@ -4,6 +4,7 @@
   root to: "application#index"
   get :deprecated, to: "application#deprecated"
   get :component, to: "application#component"
+  get :content_for, to: "application#content_for"
```
joelhawksley:
```suggestion
  get :content_areas, to: "application#content_areas"
```

test/app/components/content_for_component.html.erb
joelhawksley: Let's be sure to uncomment this before moving forward. This should continue to work.

test/app/views/application/content_for.html.erb
```diff
@@ -0,0 +1,9 @@
...
+  <p>Have a great day.</p>
+  <% comp.content_for(:title) do %>
```
joelhawksley: As we discussed on our call, this line could be:

```suggestion
  <% component.with(:title) do %>
```

lib/action_view/component/content_for.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module ActionView
+  module Component # :nodoc:
+    module ContentFor
+      def content_for(attr, content = nil, options = {}, &block)
```
joelhawksley:
```suggestion
      def with(attr, content = nil, options = {}, &block)
```

lib/action_view/component/content_for.rb
```diff
@@ -0,0 +1,38 @@
...
+        unless respond_to?(attr)
```
joelhawksley: Wouldn't we want to be checking for the presence of a setter, not just a reader?

This makes me think we should require these "content_areas" to be registered on the component. If we do that, it might make sense to rename the existing `content` accessor to `body`, so we can have something like:

`content_areas :header, :body, :footer`, and then explicitly check against this list when rendering named blocks.

lib/action_view/component/content_for.rb
```diff
@@ -0,0 +1,38 @@
...
+          raise StandardError.new "Called content_for with unknown attribute '#{attr}'"
+        end
+
+        if block_given?
+          options = content if content
+          content = view_context.capture(&block)
+        end
+
+        existing_content = send(attr)
```
joelhawksley: I wonder if we should support this case. I wouldn't expect components to set content multiple times.

lib/action_view/component/content_for.rb
```diff
@@ -0,0 +1,38 @@
...
+          content = existing_content + content
+        end
+
+        set_attribute(attr, content)
+        nil
+      end
+
+      private
+
+      def set_attribute(attr, value)
```
joelhawksley: If we go with the `content_areas` registration approach above, I think we should consider simplifying this to just use `attr_readers`.

lib/action_view/component/base.rb
```diff
@@ -47,7 +48,8 @@ def render_in(view_context, *args, &block)
         old_current_template = @current_template
         @current_template = self
 
-        @content = view_context.capture(&block) if block_given?
+        @content = view_context.capture(self, &block) if block_given?
```
joelhawksley: What is the reason for this change?

test/app/views/application/content_for.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= render ContentForComponent do |comp| %>
+  <p>Have a great day.</p>
```
joelhawksley: I wonder if we should support this "mixed" syntax, with some named blocks used in combination with an implicit block. It could be confusing to have these bits in a particular order in the template, only to have the component render them in a different order.

test/app/components/content_for_component.rb
```diff
@@ -1,5 +1,12 @@
 # frozen_string_literal: true
 
 class ContentForComponent < ActionView::Component::Base
-  def initialize(*); end
+  validates :content, :title, :footer, presence: true
+
+  attr_reader :content, :title, :footer
```
joelhawksley: Nit: the `content` reader should already be defined in `Base`, and not here.
```suggestion
  attr_reader :title, :footer
```

lib/action_view/component/base.rb
bostanio: this is what makes the component in scope for the block.

lib/action_view/component/content_for.rb
bostanio: My thinking was that attr_reader is enough because we can set the instance variable behind the attribute. I'm ok with registration. That way we can check/require getter and setter mechanics once

lib/action_view/component/content_for.rb
bostanio: yeah I could go either way. It's fair to argue the case for setting content multiple times in the component view must be as a result of logic or loops in the view that might be better encapsulated in the Ruby component file. I wonder if the case of looping would be reasonable. 

Let's decide I'm not fussed. It's probably more docs to write. Easy enough to add later with a real use case behind it.

test/app/views/application/content_for.html.erb
bostanio: interesting. It certainly makes testing a PITA because of the whitespace. 

Would the expectation be that if you supply content areas then we no longer capture the rest of the block as the 'content' (or perhaps renamed to 'body')

test/app/views/application/content_for.html.erb
bostanio: sure

test/config/routes.rb
bostanio: done

test/app/views/application/content_for.html.erb
bostanio: done

test/app/components/content_for_component.html.erb
bostanio: done

lib/action_view/component/content_for.rb
bostanio: killed it

lib/action_view/component/base.rb
```diff
@@ -8,6 +8,7 @@ class Base < ActionView::Base
       include ActiveModel::Validations
       include ActiveSupport::Configurable
       include ActionView::Component::Previewable
+      include ActionView::Component::ContentAreas
```
jcoyne: Does this need to be a module if it's only included on one class?  I think this can make it harder to read the code and understand what interactions are happening.

lib/action_view/component/content_areas.rb
```diff
@@ -0,0 +1,61 @@
...
+    module ContentAreas
+      extend ActiveSupport::Concern
+
+      included do
+        class_attribute :content_areas, default: []
+        self.content_areas = [] # default doesn't work until Rails 5.2
+      end
+
+      class_methods do
+        def set_content_areas(*areas)
```
jcoyne: would. `self.content_areas = [:title, :footer]` be a more idiomatic way?

jcoyne: Can you add documentation on how this method would be used, either inline or in the README?

bostanio: Yup documentation upcoming.

lib/action_view/component/base.rb
bostanio: I think is actually clearer to keep the pieces separate - as we do for Previewable. There is a slightly strange interaction where the Base class interrogates content_areas to decide what to do with the captured content. Maybe we can clean that up.

test/app/components/content_for_component.rb
bostanio: fixed

lib/action_view/component/base.rb
```diff
@@ -93,7 +96,7 @@ def request
         @request ||= controller.request
       end
 
-      attr_reader :content, :view_context
+      attr_reader :body, :view_context
```
bostanio: clearly this is a breaking change. But it feels like the right one.

lib/action_view/component/base.rb
jcoyne: I wonder if ContentAreas should be extracted to a separate class. It seems like a separate set of responsibilities.

lib/action_view/component/content_areas.rb
bostanio: added

lib/action_view/component/base.rb
bostanio: Not convinced that is practical given how closely intertwined in needs to be in terms of having access to the components attributes etc.

README.md
```diff
@@ -199,6 +199,73 @@ An error will be raised:
...
+  <div class="body"><%= body %>"></div>
+</div>
+```
+
+We can render it in a view as:
+
+```erb
+<%= render(TestComponent, header: "Hello Modal") do |component| %>
+  <% component.with(:body) do %>
+    <p>Have a great day %>.</p>
```
kaspermeyer: Isn't there an orphaned `%>` tag here?

lib/action_view/component/content_areas.rb
```diff
@@ -0,0 +1,41 @@
...
+          attr_reader *areas
+          self.content_areas = areas.push(:body)
```
kaspermeyer: Is there a reason that we always need a body to be present? If possible, I think it would be fairer to give users of `set_content_areas` the freedom to choose which types of content is needed for the component, rather than assuming that every component needs a body.

kaspermeyer:
Nice work, @bostanio! 👏 I like the idea of being able to have multiple outlets for content.

The new interface is a little confusing to me though. We're asking folks to define instance variables themselves which we then use to store captured view content in. This means that data and markup are mixed, and it's hard to tell them apart. 

Take the following constructor as an example: 
```ruby
def initialize user: nil, header: nil
  @user = user
  @header = header
end
```

Here `user` is a User object but `header` is a placeholder waiting to be populated with content from a `component.with(:header)` outlet. I don't think that's very clear. 

Maybe we could abstract the content placeholders away somehow?

kaspermeyer:
Also, I think it's a shame that we're breaking backward compatibility by renaming `content` to `body`. Additional content isn't always the body of a component. 

Could we keep `content` when using implicit content, and let folks define their own when using named content with `set_content_areas`? 

This would also mean that we no longer assume that a `body` should be present when using named content areas, which makes the API a bit more flexible:

### Implicit content
```ruby
class ModalComponent < ActionView::Component::Base
end
```

```erb
<div class="modal">
  <%= content %>
</div>
```

### Named content

```ruby
class ModalComponent < ActionView::Component::Base
  validates :header, :body, :footer, presence: true

  set_content_areas :header, :body, :footer
end
```

```erb
<div class="modal">
  <%= header %>
  <%= body %>
  <%= footer %>
</div>
```

bostanio:
@kaspermeyer You make excellent points about content vs body and I agree with you on the approach. We were nervous about the breaking change and as you point out it isn't necessary. 

> The new interface is a little confusing to me though. We're asking folks to define instance variables themselves which we then use to store captured view content in. This means that data and markup are mixed, and it's hard to tell them apart.

A good point. In your example declaring `header` as an argument to the constructor isn't necessary. Calling `set_content_areas` is enough to create the attribute we AV:C then sets the instance variable for you behind the scenes.  If you want to mix data and content you can in that way but its not required. I will update the examples for clarity.

kaspermeyer:
Great! And thank you for clarifying. I didn't catch that an instance variable would be created behind the scenes by `set_content_areas` 👌

lib/action_view/component/content_areas.rb
bostanio: fixed with undoing the content->body rename

bostanio:
I think we should be in better shape now. I reverted the change of content-> body. Feels more natural now thanks @kaspermeyer. The only weird wrinkle now is if you define a content area of content then weird things will happen due to the capture of the entire block into the `content` attribute. We could make `content` a 'reserved' area that you can't choose?

joelhawksley:
@bostanio if it leads to weird behavior, yeah let's try to "reserve" it.

bostanio:
@joelhawksley Added an exception with a test if you try to name :content as a content_area.

README.md
```diff
@@ -199,6 +199,55 @@ An error will be raised:
...
+  <div class="header"><%= header %></div>
+  <div class="body"><%= body %>"></div>
+</div>
+```
+
+We can render it in a view as:
+
+```erb
+<%= render(TestComponent, user: {name: 'Jane'}) do |component| %>
+  <% component.with(:body) do %>
```
dylnclrk: Is this a typo?

```suggestion
  <% component.with(:header) do %>
```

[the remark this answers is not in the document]
bostanio: Nice catch. Fixed

lib/action_view/component/content_areas.rb
```diff
@@ -0,0 +1,44 @@
...
+    module ContentAreas
+      extend ActiveSupport::Concern
+
+      included do
+        class_attribute :content_areas, default: []
+        self.content_areas = [] # default doesn't work until Rails 5.2
+      end
+
+      class_methods do
+        def set_content_areas(*areas)
```
spdawson: This is looking like a really powerful enhancement @bostanio. I wonder whether `define_content_areas` or even `has_content_areas` might be slightly more Rails-y than `set_content_areas`; what do you think?

bostanio: @spdawson Yup this always bugged me as not being Rails-y. Just renamed to `with_content_areas` which mirrors nicely the call to `with` that it enables. That's for the inspiration on a better set of names.

test/action_view/integration_test.rb
```diff
@@ -35,6 +35,24 @@ class IntegrationTest < ActionDispatch::IntegrationTest
...
 
+  test "rendering component with content_for" do
+    get "/content_areas"
+    assert_response :success
+
+    expected_string = %(
+    <div>
+        <h1>Hi!</h1>
+
+        <p>Have a great #{Date.today.strftime("%A")}.</p>
```
dylnclrk: I noticed you changed this line from `<p>Have a great day.</p>` in c0e5f727. What was the motivation for making the assertion a bit more dynamic?

I know it doesn't matter in this case, but comparing the result of calling `Date.today` in a test and in an integration environment is setting off 🚨Time-Zone-mismatch alarms 🚨 in my head.

test/app/components/content_for_component.html.erb
```diff
@@ -1,2 +1,2 @@
 <% content_for(:foo, "bar") %>
-<div>Hello content for</div>
+<div>Hello content for</div>
```
dylnclrk: nit:
```suggestion
<div>Hello content for</div>

```

test/app/views/application/content_areas.html.erb
```diff
@@ -0,0 +1,11 @@
...
+  <% component.with(:body) do %>
+    <p>Have a great <%= Date.today.strftime("%A") %>.</p>
+  <% end %>
+  <% component.with(:title) do %>
+    <h1>Hi!</h1>
+  <% end %>
+  <% component.with(:footer) do %>
+    <h3>Bye!</h3>
+  <% end %>
+<% end %>
```
dylnclrk: Nit: 
```suggestion
<% end %>

```

README.md
```diff
@@ -199,6 +199,55 @@ An error will be raised:
 
 `ActiveModel::ValidationError: Validation failed: Title can't be blank`
 
+#### Content Areas
+
+A component can be configured to have multiple 'content areas' which can be captured and later rendered in the components view. For example:
```
dylnclrk: Nit:

```suggestion
A component can be configured to have multiple 'content areas' which can be captured and later rendered in the component's view. For example:
```

dylnclrk:
Re: @joelhawksley 's comment about whether we should support the "mixed" syntax: https://github.com/github/actionview-component/pull/161#discussion_r362127749

I feel like the implicit style is good for certain things, but once you've used `content_for` you should go all in with it.

What do you think of raising an exception if a component refers to `content` when `with_content_areas ` has been called? I'm not really sure what the error message should be, but something like:

```ruby
      class_methods do
         def with_content_areas(*areas)
           if areas.include?(:content)
             raise StandardError.new ":content is a reserved content_area internal to ActionView:Component. Please use another area name, perhaps ':body'"
           end
           attr_reader *areas
           self.content_areas = areas

           # Prevent users from mixing `content` and `with_content_areas`.
           def content
             raise StandardError.new "Don't mix `content` and `with_content_areas`…"
           end
         end
       end
```

test/action_view/integration_test.rb
bostanio: The idea was to ensure we could render erb escaped syntax inside the `with` blocks. Your right that we could pick a simpler example

bostanio:
@dylnclrk I'm not sure I'm that worried to enforce it. It works, it's not confusing and there are probably places where it's powerful. It feels more Rails-y to permit it. 

For comparison the existing content_for doesn't force you to put everything in a content_for block just because you use it once.

dylnclrk:
@bostanio I see what you're thinking, I could imagine some cases where having both would be powerful. Like if you were to use the content areas as optional enhancements for your component.

Thanks again for adding this functionality 😄 I'm looking forward to using it to clean up a messy component!

README.md
```diff
@@ -124,7 +124,7 @@ An `ActionView::Component` is a Ruby file and corresponding template file (in an
 `app/components/test_component.rb`:
 ```ruby
 class TestComponent < ActionView::Component::Base
-  validates :content, :title, presence: true
+  validates :body, :title, presence: true
```
```suggestion
  validates :content, :title, presence: true
```

README.md
```diff
@@ -199,6 +199,55 @@ An error will be raised:
 
 `ActiveModel::ValidationError: Validation failed: Title can't be blank`
 
+#### Content Areas
+
+A component can be configured to have multiple 'content areas' which can be captured and later rendered in the component's view. For example:
```
```suggestion
A component can declare additional content areas to be rendered in the component. For example:
```

README.md
```diff
@@ -199,6 +199,55 @@ An error will be raised:
...
+
+`app/components/modal_component.rb`:
+```ruby
+class ModalComponent < ActionView::Component::Base
+  validates :user, :header, :body, presence: true
+
+  with_content_areas :header
```
joelhawksley: Do we want to use `body` or the built-in `content` in this example? I'd probably lean towards `body`:

```suggestion
  with_content_areas :header, :body
```

README.md
```diff
@@ -199,6 +199,55 @@ An error will be raised:
...
+  <div class="body"><%= body %>"></div>
+</div>
+```
+
+We can render it in a view as:
+
+```erb
+<%= render(TestComponent, user: {name: 'Jane'}) do |component| %>
+  <% component.with(:header) do %>
+      Hello <%= user[:name] %>
```
joelhawksley: Wouldn't this need to be `@user`?

```suggestion
      Hello <%= @user[:name] %>
```

README.md
```diff
@@ -199,6 +199,55 @@ An error will be raised:
...
+Which returns:
+
+```html
+<div class="modal">
+  <div class="header">Hello Jane</div>
+  <div class="body"><p>Have a great day.</p></div>
+</div>
+```
+
+Content for areas can be either as arguments to the render method or as the content of the block passed to the `with` method.
```
```suggestion
Content for content areas can be passed as arguments to the render method or as named blocks passed to the `with` method.
```

Can you provide an example of this same demo component being rendered via the render args?

More broadly, I wonder if we should support content areas via render args at this point. Perhaps we could stick to just the `with` route for now?

lib/action_view/component/content_areas.rb
```diff
@@ -0,0 +1,44 @@
...
+
+      included do
+        class_attribute :content_areas, default: []
+        self.content_areas = [] # default doesn't work until Rails 5.2
+      end
+
+      class_methods do
+        def with_content_areas(*areas)
+          if areas.include?(:content)
+            raise StandardError.new ":content is a reserved content_area internal to ActionView:Component. Please use another area name, perhaps ':body'"
```
joelhawksley:
```suggestion
            raise ArgumentError.new ":content is a reserved content_area internal to ActionView:Component. Please use another area name, perhaps ':body'"
```

lib/action_view/component/content_areas.rb
```diff
@@ -0,0 +1,44 @@
...
+          end
+          attr_reader *areas
+          self.content_areas = areas
+        end
+      end
+
+      def with(area, content = nil, &block)
+        unless content_areas.include?(area)
+          raise StandardError.new "Unknown content_area '#{area}' - expected one of '#{content_areas}'"
```
joelhawksley:
```suggestion
          raise ArgumentError.new "Unknown content_area '#{area}' - expected one of '#{content_areas}'"
```

lib/action_view/component/content_areas.rb
```diff
@@ -0,0 +1,44 @@
...
+        end
+
+        if block_given?
+          content = view_context.capture(&block)
+        end
+
+        set_area(area, content)
```
joelhawksley:
```suggestion
        instance_variable_set("@#{area}".to_sym, content)
```

test/app/components/content_areas_component.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class ContentAreasComponent < ActionView::Component::Base
+  validates :body, :title, :footer, presence: true
+
+  with_content_areas :title, :body, :footer
+
+  def initialize(title: nil, footer: nil)
+    @title = title
```
joelhawksley: I think it's a little awkward to require this ivar setting / method arguments to make this interface work. 

IMO we should remove this for this first PR and see if we find ourselves needing this interface in the future, and if so, what a simpler implementation might look like at the component usage level.

test/app/components/content_areas_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<div>
+  <%= title %>
```
joelhawksley: It might be nice in this example to wrap each content area in a `div` with a class name like `title`, etc, to make this example feel more realistic.

README.md
bostanio: Sure I can provide the example. I think it's nice to be able to provide both. It 'just works' in that we don't have to do anything. If the Component has constructor args with the same name as content_areas then either version works. It will be clearer with an example.

lib/action_view/component/content_areas.rb
bostanio: 👍

test/app/components/content_areas_component.rb
bostanio: We aren't writing any code to make this work. We'd have to write code to check that content_areas don't have the same name as constructor params. I think we should just leave the option to use both methods. It will be clearer with better examples.

test/app/components/content_areas_component.html.erb
bostanio: sure

README.md
bostanio: No this is valid. All the methods on the components are in scope

test/app/components/content_areas_component.html.erb
bostanio: done

test/app/components/content_areas_component.rb
bostanio: Added some examples to the Readme. I can see places where various combinations are valid/powerful. They might not be the most common cases but IMO none of them are confusing or unexpected. To prevent some or all of those combinations we'd have to write more code. Specifically to prevent content areas having names that are named parameters of the Component constructor. Going out of our way to write more code to restrict flexibility doesn't feel like the right tradeoff to me.

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+<div class="modal">
+  <div class="header"><%= header %></div>
+  <div class="body"><%= body %>"></div>
+</div>
+```
+
+We can render it in a view as:
+
+```erb
+<%= render(TestComponent, user: {name: 'Jane'}) do |component| %>
```
rmacklin:
```suggestion
<%= render(ModalComponent, user: {name: 'Jane'}) do |component| %>
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+  with_content_areas :header, :body
+
+  def initialize(header:)
+    @header = header
+  end
+end
+```
+
+```erb
+<%= render(TestComponent, header: "Hi!") do |component| %>
```
rmacklin:
```suggestion
<%= render(ModalComponent, header: "Hi!") do |component| %>
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+
+  def initialize(header: nil)
+    @header = header
+  end
+end
+```
+
+`app/views/render_arg.html.erb`:
+```erb
+<%= render(TestComponent, header: "Hi!") do |component| %>
```
rmacklin:
```suggestion
<%= render(ModalComponent, header: "Hi!") do |component| %>
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+  <% component.with(:body) do %>
+    <p>Have a great day.</p>
+  <% end %>
+<% end %>
+```
+
+`app/views/with_block.html.erb`:
+```erb
+<%= render(TestComponent) do |component| %>
```
rmacklin:
```suggestion
<%= render(ModalComponent) do |component| %>
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+  <% if header %>
+    <div class="header"><%= header %></div>
+  <% end %>
+  <div class="body"><%= body %>"></div>
+</div>
+```
+
+`app/views/render_arg.html.erb`:
+```erb
+<%= render(TestComponent, header: "Hi!") do |component| %>
```
rmacklin:
```suggestion
<%= render(ModalComponent, header: "Hi!") do |component| %>
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+  <% component.with(:body) do %>
+    <p>Have a great day.</p>
+  <% end %>
+<% end %>
+```
+
+`app/views/with_block.html.erb`:
+```erb
+<%= render(TestComponent) do |component| %>
```
rmacklin:
```suggestion
<%= render(ModalComponent) do |component| %>
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+  <% end %>
+  <% component.with(:body) do %>
+    <p>Have a great day.</p>
+  <% end %>
+<% end %>
+```
+
+`app/views/no_header.html.erb`:
+```erb
+<%= render(TestComponent) do |component| %>
```
rmacklin:
```suggestion
<%= render(ModalComponent) do |component| %>
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
 
 `ActiveModel::ValidationError: Validation failed: Title can't be blank`
 
+#### Content Areas
+
+
```
```suggestion
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+```html
+<div class="modal">
+  <div class="header">Hello Jane</div>
+  <div class="body"><p>Have a great day.</p></div>
+</div>
+```
+
+Content for content areas can be passed as arguments to the render method or as named blocks passed to the `with` method.
+The allows a few different combinations of ways to render the component.
```
```suggestion

The allows a few different combinations of ways to render the component:
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+
+##### Required Render argument optionally overridden or wrapped by a named block
```
```suggestion
##### Required render argument optionally overridden or wrapped by a named block
```

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
...
+  <% component.with(:header) do %>
+    <span class="help_icon">Hello</span> 
+  <% end %>
+  <% component.with(:body) do %>
+    <p>Have a great day.</p>
+  <% end %>
+<% end %>
+```
+
+##### Optional argument passed by render argument or by named block or neither
```
```suggestion
##### Optional argument passed by render argument, by named block, or neither
```

lib/action_view/component/content_areas.rb
```diff
@@ -0,0 +1,37 @@
...
+
+      included do
+        class_attribute :content_areas, default: []
+        self.content_areas = [] # default doesn't work until Rails 5.2
+      end
+
+      class_methods do
+        def with_content_areas(*areas)
+          if areas.include?(:content)
+            raise ArgumentError.new ":content is a reserved content_area internal to ActionView:Component. Please use another area name, perhaps ':body'"
```
joelhawksley:
```suggestion
            raise ArgumentError.new ":content is a reserved content area name. Please use another name, such as ':body'"
```

lib/action_view/component/content_areas.rb
```diff
@@ -0,0 +1,37 @@
...
+  module Component # :nodoc:
+    module ContentAreas
+      extend ActiveSupport::Concern
+
+      included do
+        class_attribute :content_areas, default: []
+        self.content_areas = [] # default doesn't work until Rails 5.2
+      end
+
+      class_methods do
```
joelhawksley: What do you think about putting this in `Base` instead? This feels like a pretty core part of the library.

README.md
```diff
@@ -199,6 +199,174 @@ An error will be raised:
 
 `ActiveModel::ValidationError: Validation failed: Title can't be blank`
 
+#### Content Areas
```
joelhawksley: 👏 thanks for the thorough docs! I'm starting to wonder if we might want to explore other avenues besides the README for documentation. What do you think?

README.md
rmacklin: Should it be
```suggestion
This allows a few different combinations of ways to render the component:
```
?

lib/action_view/component/content_areas.rb
bostanio: Yeah we could certainly do that.

README.md
bostanio: Yeah I think we need some more pages. I'd focus the Readme on simple usage. Detailed examples for ContentAreas and Motivation and design docs in their own pages? Feels like a separate PR.

joelhawksley: Agreed. Just wanted to know what you thought ❤️

joelhawksley:
🙇 incredible work here, @bostanio ❤️

PR #162, opened 2019-12-23
Fixed require path in installation guide [skip ci]

soartec-lab:
## Description

First of all, thank you for creating a nice gem.
I was trying this and ran into one problem.

If you describe require in config / application.rb as per the installation guide, an error will occur.
This can also be checked on the rails console.

```app/components/user_component.rb
class UserComponent < ActionView::Component::Base
  attr_reader :name

  def initialize(name:, style:)
    @name = name
  end
end
```

```
$ rails console
> UserComponent
=> NameError (uninitialized constant ActionView::Component)

> require "'action_view/component/railtie"
=> true

> UserComponent
=> NameError (uninitialized constant ActionView::Component)

> require "action_view/component"
=> true

> UserComponent
=> UserComponent
```

Therefore, I responded by changing the required path.
Guide has been modified accordingly.

## Environments

```
$ rails -v
Rails 6.0.2.1
$ ruby -v
ruby 2.6.4p104 (2019-08-28 revision 67798) [x86_64-darwin17]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ $ bundle
 In `config/application.rb`, add:
 
 ```bash
-require "action_view/component/railtie"
+require "action_view/component/"
 ```
 
 ## Guide


joelhawksley:
@rainerborene this is the same issue as #163, right? We should probably just publish a release 😄 

It might be worth adding a note to the readme that it is for the `master` branch.

afuno:
@joelhawksley Maybe. Just, for example, a completely new user will be very surprised. He puts the gem according to the instructions, and as a result receives an error.

joelhawksley:
My apologies for the confusion here folks. I've published `v1.6.1` which should work with the `rail tie` syntax. Please re-open this if needed ❤️ 🎄

PR #164, opened 2019-12-24
Allow Previews to have no layout 

bostanio:
Fixes the bug that specifying `layout false` in a Preview was ignored because of a truthy check rather than a nil check. 

Also added the ability to override the layout in Preview.call. This allows for other integration options in particular setting the layout to false so we can build controllers that return just the component html without the layout chrome. For example a Rails backend for Storybook #159.

### DIFF
--- a/test/test/components/previews/no_layout_preview.rb
+++ b/test/test/components/previews/no_layout_preview.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class NoLayoutPreview < ActionView::Component::Preview
+  layout false
+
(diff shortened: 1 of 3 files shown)


test/test/components/previews/no_layout_preview.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class NoLayoutPreview < ActionView::Component::Preview
+  layout false
```
bostanio: this didn't work previously

PR #166, opened 2019-12-25
Support dynamic preview args

bostanio:
Allow Previews to be configured with dynamic arguments.

Wire up the previews controller to allow those arguments to be passed via url parameters. 

The user experience here isn't great but it is functional. The big win of this change is that it sets us up to dynamically render Previews with other tools. In particular we can use the dynamic features as the backend for a Storybook and connect [Storybook Knobs](https://github.com/storybookjs/storybook/tree/master/addons/knobs) to the dynamic parameters.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    actionview-component (1.6.1)
+    actionview-component (1.6.2)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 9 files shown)


bostanio:
Closing this in the desire to keep this Gem clean and easy to upstream into Rails. It's likely that the future of this idea is to be its own gem that introduces a StorybookPreview class that mimics Preview but adds the dsl we need to be able to translate AV:Cs in Storybook stories.

PR #171, opened 2020-01-04
Allow component initializers that take no arguments

dylnclrk:
### Motivation

@topherfangio Brought up something that I also found unexpected in [this comment](https://github.com/github/actionview-component/issues/15#issuecomment-531281205):

> 4. It seems like we are forced to pass arguments to a component? Example: I can't get render(MyComponent) to work. I keep getting the error wrong number of arguments (given 1, expected 0) when it calls my initializer. This just seems a bit strange.

While I'm not sure there's a lot of value to components that don't take any props, I do think this error message might be confusing, and could lead a developer that's newer to actionview-component to spend time trying to figure out if they're incorrectly calling `render`. And it's likely to come up for devs when starting a new component with TDD.

### Proposed change

What do people think of changing `Actionview::Base#render` to allow for 0-arity initializers? For example:

```
class ExampleComponent < ActionView::Component::Base
  def initialize
    # Takes no arguments.
  end
end
```

### Other changes considered

Initially, I wanted to avoid changing how render works, and instead change the error message to be more specific than ArgumentError. One way to do this would be to check the airity of the initializer when loading the component. And raise an error message like, "`Component #initialize must accept arguments`". 

However, when I tried to implement this it got a little tricky. I think the place to do the check is in ActionView::Component::Base (in a hook like `::inherited`). But since the component's initializer has not yet been loaded when code from `ActionView::Component::Base::inherited` is run, I'm not if that's possible. It seemed like we'd have to prepend a module and do some alias method chain style trickery with the component's `#initialize`, but that messed with the tactic of using the location of initialize's source location.

### Possible further work

I'm not sure, but I think we may also have to modify [ActionView::Helpers::RenderingHelper](https://github.com/joelhawksley/rails/blob/c221b5b448569771678279216360460e066095a7/actionview/lib/action_view/helpers/rendering_helper.rb#L38-L42) to mirror whatever changes we make here.

### DIFF
--- a/lib/action_view/component/render_monkey_patch.rb
+++ b/lib/action_view/component/render_monkey_patch.rb
@@ -15,10 +15,12 @@ module ActionView
 
           options.render_in(self, &block)
         elsif options.is_a?(Class) && options < ActionView::Component::Base
-          options.new(args).render_in(self, &block)
+          options.safe_new(args).render_in(self, &block)
         elsif options.is_a?(Hash) && options.has_key?(:component)
-          options[:component].new(options[:locals]).render_in(self, &block)
+          options[:component].safe_new(options[:locals]).render_in(self, &block)
         elsif options.respond_to?(:to_component_class) && !options.to_component_class.nil?
+          # In this case, the `options` passed to `new` is the object on which `#to_component_class` is called. So we
+          # want the non-zero arity initializer to be called.
           options.to_component_class.new(options).render_in(self, &block)
         else
           super
--- a/test/action_view/component_test.rb
+++ b/test/action_view/component_test.rb
@@ -17,6 +17,41 @@ class ActionView::ComponentTest < ActionView::Component::TestCase
     assert_html_matches "<div>hello,world!</div>", result.css("div").to_html
   end
 
+  def test_zero_arity_initializers
+      result = render_inline(ZeroArityComponent)
+
+      assert_html_matches "<div>hello,world!</div>", result.to_html
+  end
+
+  def test_zero_arity_initializers_with_hash_syntax
+    result = render_inline(component: ZeroArityComponent, locals: {})
+
+    assert_html_matches "<div>hello,world!</div>", result.to_html
+  end
+
+
+  def test_zero_arity_initializers_with_hash_syntax_and_no_locals
+    result = render_inline(component: ZeroArityComponent)
+
+    assert_html_matches "<div>hello,world!</div>", result.to_html
+  end
+
+  def test_zero_arity_initializer_with_to_component_class
+    exception = assert_raises ArgumentError do
+      render_inline(ZeroArity.new)
+    end
+
+    assert_includes exception.message, "wrong number of arguments (given 1, expected 0)"
+  end
+
+  def test_zero_arity_initializer_arguments_not_dropped
+    exception = assert_raises ArgumentError do
+      render_inline(ZeroArityComponent, unnecessary_attr: "Hello")
+    end
+
+    assert_includes exception.message, "wrong number of arguments (given 1, expected 0)"
+  end
+
   def test_checks_validations
     exception = assert_raises ActiveModel::ValidationError do
       render_inline(WrapperComponent)
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -132,6 +132,21 @@ module ActionView
           super
         end
 
+        def safe_new(options)
+          # Check the arity of `#initialize` to avoid passing an empty
+          # `options` hash when no arguments are expected. Also, check
+          # `options` to avoid dropping arguments passed to `#render`,
+          # which might hide a helpful `ArgumentError`.
+
+          arity = self.instance_method(:initialize).arity
+
+          if (options.blank? && arity.zero?)
+            self.new
+          else
+            self.new(options)
+          end
+        end
+
         def call_method_name(variant)
           if variant.present? && variants.include?(variant)
             "call_#{variant}"
--- a/test/app/components/zero_arity_component.html.erb
+++ b/test/app/components/zero_arity_component.html.erb
@@ -1,0 +1,1 @@
+<div>hello,world!</div>
--- a/test/app/components/zero_arity_component.rb
+++ b/test/app/components/zero_arity_component.rb
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+
+class ZeroArityComponent < ActionView::Component::Base
+  def initialize
+  end
+end
--- a/test/app/models/zero_arity.rb
+++ b/test/app/models/zero_arity.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class ZeroArity
+  # Used to test components that have initializers that take no arguments (arity == 0).
+
+  include ActiveModel::Model
+
+  attr_accessor :title
+end


lib/action_view/component/render_monkey_patch.rb
```diff
@@ -15,7 +15,7 @@ def render(options = {}, args = {}, &block)
 
           options.render_in(self, &block)
         elsif options.is_a?(Class) && options < ActionView::Component::Base
-          options.new(args).render_in(self, &block)
+          options.new(**args).render_in(self, &block)
         elsif options.is_a?(Hash) && options.has_key?(:component)
           options[:component].new(options[:locals]).render_in(self, &block)
         elsif options.respond_to?(:to_component_class) && !options.to_component_class.nil?
```
bostanio: If we want to fix this we need to solve it for the other two cases where we call `new`.

How about we explicitly check the arity of the constructor?

```
def new_component(component_class, options)
  arity = component_class.method(:initialize).arity
  if arity == 0 ? component_class.new : component_class.new(options)
end
```

and then use it in the if-else block `new_component(options, args).render_in(self, &block)` etc. etc.

bostanio: needs test cases for the other branches of the if-else

dylnclrk: Oh yeah, good point! Thanks for taking a look at this. I’ll add the other branches and see how it looks. I have a feeling that the airity check that you suggested will be simpler for the other cases.

 I guess the only thing we’d want to avoid is not dropping arguments (incorrectly) provided by the user to a 0-arity function. As it might make for a slightly confusing debugging experience. But \_(ツ)_/¯ it’s probably ok if we do. 

Also, maybe it’s as simple as something like `if arity ==0 && options.blank? ...` to avoid most cases.

bostanio:
Specs are failing I suspect you need the arity approach to cover all Ruby versions. Also need to handle the `to_component_class` case.

dylnclrk:
@bostanio Yeah, I think I opened this prematurely 😞 I only ran it locally on 2.7 and haven't had a chance to finish the changes you suggested. I'll update here when I've finished; I'm working on them now.

dylnclrk:
@bostanio I addressed one of the other paths, but I'm not sure a test makes sense for the  `render(Post)`/`#to_component_class` style, as in that case the object is always passed right back to new like: `Post.to_component_class.new(Post)`. Let me know what you think! And thanks for all of your help!! 😄

bostanio:
Looks good. I'm in agreement that the `render(Post)/#to_component_class` doesn't need this fix as we're always passing the object back in to the component. Maybe back that one out and this looks good to me.

lib/action_view/component/render_monkey_patch.rb
```diff
@@ -15,15 +15,23 @@ def render(options = {}, args = {}, &block)
...
         end
       end
+
+      private
+
+      def new_component(component_class, options)
+        arity = component_class.instance_method(:initialize).arity
+
+        (options.blank? && arity.zero?) ? component_class.new : component_class.new(options)
+      end
```
dylnclrk: I'm not sure how I feel about my use of `#blank?` and `#zero?` here. If the user is using the `#to_component_class` style of `render` (`render(@post)`) and their object  implements either of those methods, then they might see some confusing results.

That said, I want to inspect the options coming in, as to avoid dropping arguments incorrectly. I think that if the user passes options to a 0-arity initializer, they _should_ see an ArgumentError to help debug.

test/action_view/component_test.rb
```diff
@@ -17,6 +17,33 @@ def test_render_inline_with_old_helper
     assert_html_matches "<div>hello,world!</div>", result.css("div").to_html
   end
 
+  def test_zero_arity_initializers
```
dylnclrk: I wonder if there's a better place for these tests, or just better names…

dylnclrk:
Squashing my changes.

dylnclrk:
Removing the initializer requirement will alleviate this https://github.com/dylnclrk/actionview-component/pull/1, at least for Ruby 2.7+.

lib/action_view/component/render_monkey_patch.rb
bostanio: My style would be to add a code comment around this that details the logic here. As you point out in the PR comment above we're trying to preserve the ArgumentError etc. The tests cover this case as some form of documentation but it might be worth spelling it out here. 

Another though how about we call this method `safe_new_component` to hint at what it's trying to do?

dylnclrk: OK, I made the suggestions you recommended.

dylnclrk:
If we choose to use this change, I sort of feel like there should be a changelog entry.

lib/action_view/component/render_monkey_patch.rb
```diff
@@ -15,15 +15,28 @@ def render(options = {}, args = {}, &block)
...
+
+      private
+
+      def safe_new_component(component_class, options)
+        # Check the arity of `#initialize` to avoid passing an empty
+        # `options` hash when no arguments are expected. Also, check
+        # `options` to avoid dropping arguments passed to `#render`,
+        # which might hide a helpful `ArgumentError`.
+
+
```
dylnclrk: Oops, extra newline.

```suggestion
```

lib/action_view/component/render_monkey_patch.rb
```diff
@@ -15,15 +15,27 @@ def render(options = {}, args = {}, &block)
 
           options.render_in(self, &block)
         elsif options.is_a?(Class) && options < ActionView::Component::Base
-          options.new(args).render_in(self, &block)
+          safe_new_component(options, args).render_in(self, &block)
         elsif options.is_a?(Hash) && options.has_key?(:component)
-          options[:component].new(options[:locals]).render_in(self, &block)
+          safe_new_component(options[:component], options[:locals]).render_in(self, &block)
         elsif options.respond_to?(:to_component_class) && !options.to_component_class.nil?
           options.to_component_class.new(options).render_in(self, &block)
```
joelhawksley: Do we need to update this case as well?

lib/action_view/component/render_monkey_patch.rb
```diff
@@ -15,15 +15,27 @@ def render(options = {}, args = {}, &block)
...
+      private
+
+      def safe_new_component(component_class, options)
+        # Check the arity of `#initialize` to avoid passing an empty
+        # `options` hash when no arguments are expected. Also, check
+        # `options` to avoid dropping arguments passed to `#render`,
+        # which might hide a helpful `ArgumentError`.
+
+        arity = component_class.instance_method(:initialize).arity
+        (options.blank? && arity.zero?) ? component_class.new : component_class.new(options)
```
joelhawksley: It's a nitpick, but I'd prefer we avoid using a ternary here. Mind splitting this into `if else`?

test/action_view/component_test.rb
```diff
@@ -17,6 +17,33 @@ def test_render_inline_with_old_helper
...
+    assert_html_matches "<div>hello,world!</div>", result.to_html
+  end
+
+  def test_zero_arity_initializer_arguments_not_dropped
+    exception = assert_raises ArgumentError do
+      render_inline(ZeroArityComponent, unnecessary_attr: "Hello")
+    end
+
+    assert_includes exception.message, "wrong number of arguments (given 1, expected 0)"
+  end
```
joelhawksley: Can we add a test for rendering a zero arity component via the `to_component_class` interface?

lib/action_view/component/render_monkey_patch.rb
dylnclrk: I went [back and forth on this](https://github.com/github/actionview-component/pull/171#issuecomment-570830559), and I'm leaning towards excluding this case, even though it's sort of weird. But I'd love to hear your thoughts.

If we use `#safe_new_component` for the `#to_component_class` case:
* `options` will be  the object on which `#to_component_class` is called. Which I think means we _always_ want the non-zero arity initializer to be called.
* `#safe_new_component` calls `#blank?` on `options`. And if `options` is a model or whatever responds to `:to_component_class`, I'm not sure if calling `#blank?` is right.
* If we drop the `#blank?` check in `#safe_new_component` all-together and only check the arity of `#initialize` then we run the risk of accidentally dropping arguments a user passed to initialize, and hiding a helpful `ArgumentError`.

If we don't use `#safe_new_component`
* The call to `.new`, is now made in two separate methods (`#render` and `#safe_new_component`). Which feels less ideal.

In general this change I've made feels a little hack-y. But I'm not sure what sort of change would be better.  Do you have any ideas?

It still seems important to me that one can either we support 0-arity initializers or give a more leading error message. Since users aren't calling `#new` on their component, it's not immediately clear why `render(MyComponent)` raises `"Argument Error: wrong number of arguments (1 for 0)"`

lib/action_view/component/render_monkey_patch.rb
dylnclrk: no prob!

Which do you prefer?

```
# option A

if options.blank? && arity.zero?
  component_class.new
else 
  component_class.new(options)
end
```

```
# option B

if options.blank? && arity.zero?
  return component_class.new
end

component_class.new(options)
```

```
# option C
return component_class.new if options.blank? && arity.zero?

component_class.new(options)
```

test/action_view/component_test.rb
dylnclrk: No problem.

I think it should always raise an error (since the object is always passed through to the component)

lib/action_view/component/render_monkey_patch.rb
joelhawksley: 🙇 A, please ❤️

lib/action_view/component/render_monkey_patch.rb
joelhawksley: ❤️ thanks for the explanation. I think you're right to take the approach you did here. 

Mind dropping `options will be the object on which #to_component_class is called. Which I think means we always want the non-zero arity initializer to be called.` in as a code comment above this line?

dylnclrk: ✅ Done. Good thinking! Let me know if you think the comment could be clearer. Right now I have:

> In this case, the `options` passed to `new` is the object on which `#to_component_class` is called. So we want the non-zero arity initializer to be called.

lib/action_view/component/render_monkey_patch.rb
dylnclrk: 🤙

test/action_view/component_test.rb
dylnclrk: Here's the test:

https://github.com/github/actionview-component/pull/171/files#diff-1ead8b7b852d11443af608850a661ed2R39-R45

dylnclrk:
@joelhawksley @bostanio I think I've addressed all of the feedback 😁let me know if there's anything else.

lib/action_view/component/render_monkey_patch.rb
```diff
@@ -15,15 +15,34 @@ def render(options = {}, args = {}, &block)
...
+          # want the non-zero arity initializer to be called.
           options.to_component_class.new(options).render_in(self, &block)
         else
           super
         end
       end
+
+      private
+
+      def safe_new_component(component_class, options)
```
joelhawksley: As this behavior will never change based on the state of the component, why don't we compile this into a method on the component that has the correct arity handling?

dylnclrk: I'm probably wrong about this, but I'm not sure we know the arity of the initialize method when we inherit from `Base`. I think we have to check the arity when the component is initialized. 

What do you think of moving this logic in to a new method,`Base::safe_new`, that the monkey patch calls?

dylnclrk: I'm giving this a shot but sorta getting lost. Maybe we should set up some time to pair/chat synchronously about this. 

Here are some things that are tripping me up: 
* Since `compile!` is called after a component is rendered, it's too late to define the "safe initializer", as the component is already initialized.
* Since we currently inspect the `options` to avoid dropping them, there is _some_ behavior change based on the state of the component. I thought that maybe raising our own error in this situation would work.

Here are the changes to compile that I was considering, but wont work (because compile is called after the first time `initialize` is called).

```rb
        def compile(validate: false)
          return if compiled?

          if template_errors.present?
            raise ActionView::Component::TemplateError.new(template_errors) if validate
            return false
          end

          templates.each do |template|
            class_eval <<-RUBY, __FILE__, __LINE__ + 1
              def #{call_method_name(template[:variant])}
                @output_buffer = ActionView::OutputBuffer.new
                #{compiled_template(template[:path])}
              end
            RUBY
          end

          # To allow component initializers that take no options,
          # check the arity of `#initialize` to avoid passing an
          # empty `options` hash when no arguments are expected.
          if instance_method(:initialize).arity.zero?
            safe_new_method = <<-RUBY
              def safe_new(options)
                unless options.blank? # Check `options` to avoid dropping arguments passed to `#render`
                  raise "Component `#{name}` was rendered with options, but initializer does not accept arguments."
                end
  
                new
              end
            RUBY
          else
            safe_new_method = <<-RUBY
              def safe_new(options)
                new(options)
              end
            RUBY
          end

          class_eval safe_new_method, __FILE__, __LINE__ + 1

          @compiled = true
        end
```

Taking a step back, I was wondering what was behind this suggestion, were you thinking that it was best not to add additional logic to render/initialization if possible?

joelhawksley:
Now that we'll [only be passing component instances](https://github.com/github/actionview-component/issues/206) to `render`, I think we can close this PR.

Thanks for digging into this @dylnclrk ❤️

Issue #163, opened 2019-12-23
[1.6.0 - 1.6.1] uninitialized constant ActionView::Component::Base


afuno:
Rails 6.0.2.1

---

**Gemfile:**

```ruby
gem 'actionview-component'
```

**application.rb:**

```ruby
require 'action_view/component/railtie'
```

**body_component.rb:**

```ruby
class BodyComponent < ActionView::Component::Base
  validates :content, presence: true

  def initialize(data: {})
    @data = data
  end

  private

  attr_reader :data
end
```

**layouts.html.slim:**

```slim
= render(BodyComponent, data: { locale: locale })
```

As a result, I get the error:

> NameError - uninitialized constant ActionView::Component::Base

Gem installed on a new application, following the instructions in the [README](https://github.com/github/actionview-component#installation) file.

afuno:
If replace this line:

```ruby
require 'action_view/component/railtie'
```

On this:

```ruby
require 'action_view/component'
```

That everything works.

afuno:
https://github.com/github/actionview-component/pull/162

rainerborene:
You're looking at the `master` branch. Latest version of `actionview-component` wasn't released yet on RubyGems.

afuno:
@rainerborene Oh, okay.

joelhawksley:
My apologies for the confusion here folks. I've published `v1.6.1` which should work with the `rail tie` syntax. Please re-open this if needed ❤️ 🎄

afuno:
@joelhawksley I updated gem to version 1.6.1. Next, I replaced the code with this:

```ruy
require 'action_view/component/railtie'
```

https://github.com/github/actionview-component/compare/v1.6.0..v1.6.1

The application does not start:

```
/Users/afuno/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/bootsnap-1.4.5/lib/bootsnap/load_path_cache/core_ext/active_support.rb:79:in `block in load_missing_constant': uninitialized constant ActionView::Component::Base (NameError)
```

Everything works successfully with this code:

``ruby
require 'action_view/component'
```

joelhawksley:
@rainerborene I'm seeing this well in my environment. Might you be able to have a look?

rainerborene:
@joelhawksley done. https://github.com/github/actionview-component/pull/167

rainerborene:
I guess the solution is to specify the `require` path on Gemfile.

```ruby
[...]

PR #172, opened 2020-01-04
Fix Uninitialized Constant error #163

bostanio:
This should fix the Uninitialized Constant error. 

Its standard practice to include the entire "gem" in a Railtie and we changed that with 1.6.0

See:
https://github.com/rails/rails/blob/master/actionmailer/lib/action_mailer/railtie.rb#L4
https://github.com/rails/rails/blob/master/actionpack/lib/action_controller/railtie.rb#L4

### DIFF
--- a/lib/action_view/component/railtie.rb
+++ b/lib/action_view/component/railtie.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "rails"
+require "action_view/component"
 
 module ActionView
   module Component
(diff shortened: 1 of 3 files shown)


lib/action_view/component/railtie.rb
```diff
@@ -25,14 +26,22 @@ class Railtie < Rails::Railtie # :nodoc:
...
         end
       end
 
+      initializer "action_view_component.eager_load_actions" do
+        ActiveSupport.on_load(:after_initialize) do
+          ActionView::Component::Base.descendants.each do |descendant|
+            descendant.compile if descendant.has_initializer? && config.eager_load
+          end
+        end
+      end
```
bostanio: this is a somewhat simpler way to handle compiling than polluting base with `eager_load!`. Not necessary for the fix but learned this along the way digging into other railties.

joelhawksley: @bostanio is this change functionally equivalent to what we had before?

Also, do we still need this line, then?

https://github.com/github/actionview-component/pull/158/files#diff-051fdf6a403587a2c819adf69b1ce4f5R28

bostanio: Yes I believe so and it passes the specs around eager load (and manually verified they fail if you turn eager_load off in test). If we upstream this into Rails they have some nice support for testing railties in dev/prod environments. I looked at recreating it but its a LOT of machinery.

No we don't need the line you references. I thought I'd removed it already but maybe missed a push.

PR #174, opened 2020-01-05
Remove initalizer requirement for Ruby 2.7+

dylnclrk:
### Motivation

As discussed in github#93, Ruby 2.7 introduces [`#const_source_location`](https://ruby-doc.org/core-2.7.0/Module.html#method-i-const_source_location) to find the location of constants. This means that we no longer have to find the location of components through the location of their `#initialize` method, and we can remove the initializer requirement for Ruby 2.7+.

### Changes

* Use `#const_source_location` in `ActionView::Component::Base::source_location`, when it's available. 
* Move the initializer-less component test out of invalid, and in to a proper component test. Test it still throws the error for Rubies < 2.7.
* Change eager-loading logic for Ruby 2.7+

### Todo

- [x] Update CI to test Ruby 2.7 once supported by `actions` (actions/setup-ruby#45, actions/virtual-environments#209).
- [x] Rebase to pick up workflow changes in #200 when that's merged.

(The old todo, `Test eager loading change`, is no longer necessary.)

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -142,15 +142,11 @@ module ActionView
 
         def source_location
           @source_location ||=
-              begin
-                # Require #initialize to be defined so that we can use
-                # method#source_location to look up the file name
-                # of the component.
-                #
-                # If we were able to only support Ruby 2.7+,
-                # We could just use Module#const_source_location,
-                # rendering this unnecessary.
-                #
+              if const_source_location_supported?
+                const_source_location(self.name)[0]
+              else
+                # Require `#initialize` to be defined so that we can use `method#source_location`
+                # to look up the filename of the component.
                 initialize_method = instance_method(:initialize)
                 initialize_method.source_location[0] if initialize_method.owner == self
               end
@@ -210,6 +206,10 @@ module ActionView
 
         private
 
+        def const_source_location_supported?
+          respond_to? :const_source_location # introduced in Ruby 2.7
+        end
+
         def matching_views_in_source_location
           return [] unless source_location
           (Dir["#{source_location.chomp(File.extname(source_location))}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location])
@@ -232,7 +232,12 @@ module ActionView
           @template_errors ||=
             begin
               errors = []
-              errors << "#{self} must implement #initialize." if source_location.nil?
+              if source_location.nil? && !const_source_location_supported?
+                # Require `#initialize` to be defined so that we can use `method#source_location`
+                # to look up the filename of the component.
+                errors << "#{self} must implement #initialize."
+              end
+
               errors << "Could not find a template file for #{self}." if templates.empty?
 
               if templates.count { |template| template[:variant].nil? } > 1
--- a/test/action_view/component_test.rb
+++ b/test/action_view/component_test.rb
@@ -372,6 +372,12 @@ class ActionView::ComponentTest < ActionView::Component::TestCase
     assert_html_matches "<span>The Awesome post component!</span>", render_inline(post).to_html
   end
 
+  def test_missing_initializer
+    skip unless const_source_location_supported?
+
+    assert_html_matches "Hello, world!", render_inline(MissingInitializerComponent).text
+  end
+
   private
 
   def modify_file(file, content)
--- a/test/action_view/integration_test.rb
+++ b/test/action_view/integration_test.rb
@@ -186,6 +186,14 @@ class IntegrationTest < ActionDispatch::IntegrationTest
   end
 
   test "does not compile components without initializers" do
+    skip if const_source_location_supported?
+
     assert !MissingInitializerComponent.compiled?
+  end
+
+  test "compiles components without initializers" do
+    skip unless const_source_location_supported?
+
+    assert MissingInitializerComponent.compiled?
   end
 end
--- a/test/action_view/invalid_components_test.rb
+++ b/test/action_view/invalid_components_test.rb
@@ -2,6 +2,8 @@
 
 class ActionView::InvalidComponentTest < ActionView::Component::TestCase
   def test_raises_error_when_initializer_is_not_defined
+    skip if const_source_location_supported?
+
     exception = assert_raises ActionView::Component::TemplateError do
       render_inline(MissingInitializerComponent)
     end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Pending
+
+* Remove initializer requirement for Ruby 2.7+
+
+    *Dylan Clark*
+
 # v1.8.1
 
 * Run validation checks before calling `#render?`.
--- a/README.md
+++ b/README.md
@@ -528,6 +528,11 @@ To use component previews, set the following in `config/application.rb`:
 config.action_view_component.preview_path = "#{Rails.root}/spec/components/previews"
 ```
 
+### Initializer requirement
+
+In Ruby 2.6.x and below, ActionView::Component requires the presence of an `initialize` method in each component. 
+However, `initialize` is no longer required for projects using 2.7.x and above.
+
 ## Frequently Asked Questions
 
 ### Can I use other templating languages besides ERB?
--- a/test/app/components/missing_initializer_component.html.erb
+++ b/test/app/components/missing_initializer_component.html.erb
@@ -1,0 +1,1 @@
+<div>Hello, world!</div>
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -31,3 +31,7 @@ end
 def assert_html_matches(expected, actual)
   assert_equal(trim_result(expected), trim_result(actual))
 end
+
+def const_source_location_supported?
+  Module.respond_to? :const_source_location
+end


lib/action_view/component/base.rb
```diff
@@ -115,21 +115,21 @@ def has_initializer?
...
+            return instance_method(:initialize).source_location[0]
+          end
+
+          Module.const_source_location(self.name)[0]
         end
 
         def eager_load!
           self.descendants.each do |descendant|
-            descendant.compile if descendant.has_initializer?
+            descendant.compile if RUBY_VERSION >= "2.7" || descendant.has_initializer?
```
dylnclrk: Seems like this should be tested. But I'm not sure if I want to test `#eager_load!`

Maybe this should be refactored so the logic can be tested separately. Perhaps by replacing `has_initializer?` with a method like `can_find_template?` that makes sense for both strategies of finding the source. I guess this sort of refactor would be good because it would reduce references to the initializer in the API, too.

lib/action_view/component/base.rb
```diff
@@ -115,21 +115,21 @@ def has_initializer?
...
-          # method#source_location to look up the file name
-          # of the component.
-          #
-          # If we were able to only support Ruby 2.7+,
-          # We could just use Module#const_source_location,
-          # rendering this unnecessary.
-          raise NotImplementedError.new("#{self} must implement #initialize.") unless has_initializer?
-
-          instance_method(:initialize).source_location[0]
+          if RUBY_VERSION < "2.7"
```
bostanio: it would be much cleaner to test if the method we need exits. How about we test `if Module.method_defined(:const_source_location)`?

lib/action_view/component/base.rb
bostanio: I had the same thought. See #172 where I moved this logic into the Railtie. Leaking these detail of the API feels pretty icky. IMO we should just call `compile` here and let it do the no-op if there is no initializer

joelhawksley:
Thanks for taking this on! Let's keep this PR open until we can test these changes in CI.

lib/action_view/component/base.rb
bostanio: I'm going to throw up a quick PR to clean this stuff up that I think will make this easier.

test/action_view/component_test.rb
```diff
@@ -232,6 +232,18 @@ def test_to_component_class
...
+    if RUBY_VERSION < "2.7"
+      exception = assert_raises NotImplementedError do
+        render_inline(NoInitializerComponent)
+      end
+
+      assert_includes exception.message, "must implement #initialize"
+    else
+      assert_html_matches "Hello, world!", render_inline(NoInitializerComponent).text
+    end
+  end
```
bostanio: It's pretty awkward to have  a test that splits what it asserts based on some logic (the name doesn't line up for example). This is a place where mini_test is weak vs rspec or the like. I wonder can we test the ruby version and then define a method or not? That would allow us to keep the invalid test in the invalid file and selectively include a test here for the happy path case.

dylnclrk: Good point. I’m not too familiar with MiniTest, but I just found [this](https://www.rubydoc.info/gems/minitest/2.7.0/MiniTest/Assertions#skip-instance_method). So we could probably split it and have a guard clause on each of the respective tests. Something like `skip unless RUBY_VERSION < “2.7”`.

lib/action_view/component/base.rb
dylnclrk: I had gone back and forth on this, but settled on using RUBY_VERSION, as it might provide a better clue to delete the block when we drop support for rubies < 2.7

But a comment here like `# to support Ruby < 2.7` would also do the trick.

bostanio: Yeah I think comments would fix it. I'd be worried that in Ruby 2.8 they rename `const_source_location` to something else, and we'd break in some harder to understand way.

test/action_view/component_test.rb
bostanio: I wonder if you can do this:

```
if RUBY_VERSION < '2.7'
  define_method :test_no_initializer do
    ...
  end
end
```

lib/action_view/component/base.rb
bostanio: See #175. On the back of that change you'd just need to change `source_location` for Ruby > 2.7 and wouldn't need to worry about it here.

test/app/components/no_initializer_component.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class NoInitializerComponent < ActionView::Component::Base
+end
```
bostanio: With #175 you should be able to use `MissingInitializerComponent` in the tests and drop guy.

lib/action_view/component/base.rb
joelhawksley: What if we checked to see if the component responds_to `const_source_location`?

dylnclrk:
Awesome, @bostanio , When the changes from #175 go in I'll refactor here

test/action_view/component_test.rb
dylnclrk: Having the invalid components in app/component will also make this a little bit nicer.

dylnclrk:
Do folks have a preference of using the Minitest `skip` versus conditionally defining tests with `#define_method` based on the Ruby version?

I see some instances of Rails's test suite making use of `skip`, like [here](https://github.com/rails/rails/blob/f619ac91d33bd431d94726cc97092c3dbeb2afc6/activerecord/test/cases/insert_all_test.rb#L73). So I'm leaning towards that. But here's what changing to conditionally defined tests looks like:

```diff
diff --git a/test/action_view/component_test.rb b/test/action_view/component_test.rb
index 504ad03..e2e359d 100644
--- a/test/action_view/component_test.rb
+++ b/test/action_view/component_test.rb
@@ -353,11 +353,11 @@ class ActionView::ComponentTest < ActionView::Component::TestCase
     assert_html_matches "<span>The Awesome post component!</span>", render_inline(post).to_html
   end
 
-  def test_missing_initializer
-    skip unless RUBY_VERSION > "2.7.0"
-
+  if RUBY_VERSION >= "2.7.0"
+    define_method :test_missing_initializer do
       assert_html_matches "Hello, world!", render_inline(MissingInitializerComponent).text
     end
+  end
 
   private
 
diff --git a/test/action_view/integration_test.rb b/test/action_view/integration_test.rb
index 936ed1d..ad66d87 100644
--- a/test/action_view/integration_test.rb
+++ b/test/action_view/integration_test.rb
@@ -173,9 +173,9 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert UnreferencedComponent.compiled?
   end
 
+  if RUBY_VERSION < "2.7.0"
     test "does not compile components without initializers" do
-    skip if RUBY_VERSION >= "2.7.0"
-
       assert !MissingInitializerComponent.compiled?
     end
   end
+end
diff --git a/test/action_view/invalid_components_test.rb b/test/action_view/invalid_components_test.rb
index be70de0..20d6906 100644
--- a/test/action_view/invalid_components_test.rb
+++ b/test/action_view/invalid_components_test.rb
@@ -1,15 +1,16 @@
 # frozen_string_literal: true
 
 class ActionView::InvalidComponentTest < ActionView::Component::TestCase
-  def test_raises_error_when_initializer_is_not_defined
-    skip if RUBY_VERSION >= "2.7.0"
 
+  if RUBY_VERSION < "2.7.0"
+    define_method :test_raises_error_when_initializer_is_not_defined do
       exception = assert_raises ActionView::Component::TemplateError do
         render_inline(MissingInitializerComponent)
       end
 
       assert_includes exception.message, "must implement #initialize"
     end
+  end
 
   def test_raises_error_when_sidecar_template_is_missing
     exception = assert_raises ActionView::Component::TemplateError do

```

dylnclrk:
@bostanio Refactored to include your changes and it seems like much less of a hack. Thanks!! 🙇

lib/action_view/component/base.rb
```diff
@@ -219,7 +219,13 @@ def template_errors
...
               errors = []
-              errors << "#{self} must implement #initialize." if source_location.nil?
+
+              unless const_source_location_supported?
+                # To support Ruby < 2.7.0, require `#initialize` to be defined so that we can use
+                # `method#source_location`.
+                errors << "#{self} must implement #initialize." if source_location.nil?
+              end
+
               errors << "Could not find a template file for #{self}." if templates.empty?
```
dylnclrk: note that `#templates` should be empty if `#source_location` is `nil` (see above, [base.rb#L201](https://github.com/github/actionview-component/pull/174/files#diff-4ea3d395eff742fa0d8952d23a16d377L201)), so if we're using 2.7+ and `#source_location` is nil for some other reason, this is the error that should be raised. 

I'm not sure what other reasons `source_location` could be nil, so I'm not currently testing this.

test/action_view/component_test.rb
dylnclrk: Do you have thoughts on skip vs. what you recommended? I'm leaning towards `skip` because I see it in the Rails tests. But there's probably instances of the `if … define_method :test_…` way too. 

More context in this comment: https://github.com/github/actionview-component/pull/174#issuecomment-572769270

bostanio:
I like `skip` better than `define_method`. I think I'd make it a named check (like the Ruby example) `skip if supports_.....` We can define that check against a Ruby version or maybe with a similar check for the method we look for on Module.

lib/action_view/component/base.rb
```diff
@@ -129,15 +129,11 @@ def call_method_name(variant)
...
-                # of the component.
-                #
-                # If we were able to only support Ruby 2.7+,
-                # We could just use Module#const_source_location,
-                # rendering this unnecessary.
-                #
+              if const_source_location_supported?
+                const_source_location(self.name)[0]
+              else
+                # For Ruby < 2.7.0, use `method#source_location` on `#initialize`
```
bostanio:
```suggestion
                # Require #initialize to be defined so that we can use	                const_source_location(self.name)[0]
                # method#source_location to look up the file name
                # of the component
```

dylnclrk:
@bostanio, Maybe I'm misunderstanding, but the checks should be named in both the tests and in `base.rb`. 

Perhaps you were looking at an older version of the changes? See a9122edb12afae627af5ff10aba4e772a5288035 and 7f22b0eb3b59c69f89980ecedca03f79fa01922d

test/action_view/integration_test.rb
```diff
@@ -174,6 +174,8 @@ class IntegrationTest < ActionDispatch::IntegrationTest
   end
 
   test "does not compile components without initializers" do
+    skip if const_source_location_supported?
+
```
bostanio: Lets add the inverse of this test. 😄

test/action_view/invalid_components_test.rb
```diff
@@ -2,6 +2,8 @@
 
 class ActionView::InvalidComponentTest < ActionView::Component::TestCase
   def test_raises_error_when_initializer_is_not_defined
+    skip if const_source_location_supported?
+
```
bostanio: similar comment if it is supported lets test that it renders

test/action_view/component_test.rb
bostanio: I like skip

bostanio:
@dylnclrk Yup read and responded your larger comment without seeing the code. Exactly what I was thinking. Nice!

dylnclrk:
@bostanio oh duh, the diff I posted didn't have the predicate methods in it

lib/action_view/component/base.rb
```diff
@@ -197,6 +193,10 @@ def with_content_areas(*areas)
 
         private
 
+        def const_source_location_supported?
```
jcoyne: Perhaps it would be helpful if you added documentation about this method only being in ruby 2.7+

dylnclrk: I added a little inline comment, good thinking!

lib/action_view/component/base.rb
dylnclrk: done.

test/action_view/integration_test.rb
dylnclrk: 👍

dylnclrk:
oof, looks like there's an issue with 2.7 in workflows. I'll dig into it this week.

dylnclrk:
@joelhawksley (+ anyone else that might be watching) I'm Trying to track down an issue with workflows, sorry about all of the CI error notifications you must be getting. 

I'm gonna move this to a new branch on my fork instead. 😬

.github/workflows/ruby_on_rails.yml
```diff
@@ -22,7 +22,7 @@ jobs:
         ruby-version: ${{ matrix.ruby_version }}
     - name: Build and test with Rake
       run: |
-        gem install bundler:1.14.0
+        gem install bundler:1.17.3
```
dylnclrk: @joelhawksley I hope this change is OK, I noticed it was changed in #35 so maybe it was a ruby 2.3 support thing?

But now it matches what's in `Gemfile.lock`

joelhawksley: That's fine. Might you break this into its own PR?

dylnclrk: 👍 no prob!

dylnclrk: I extracted this change in to #200 I'll rebase when that's merged.

dylnclrk: OK, I rebased! :)

dylnclrk:
@joelhawksley OK, I think with the extraction of the CI config, I think I addressed all of your comments. Let me know what ya think!

lib/action_view/component/base.rb
```diff
@@ -232,7 +232,13 @@ def template_errors
           @template_errors ||=
             begin
               errors = []
-              errors << "#{self} must implement #initialize." if source_location.nil?
+
+              unless const_source_location_supported?
```
joelhawksley: Can we combine this conditional with `if source_location.nil?`? It's a little confusing why the two are separate.

dylnclrk: @joelhawksley addressed! But now CI seems to be failing at the bundle step for some reason… 😕 Will try re-running it.

dylnclrk:
Thanks for the reviews everyone! I don't know about you, but I'm stoked to see the initialize requirement go away! 😄

dylnclrk:
Thanks! You got it, I added a little section to the bottom of the readme & a section to the changelog!

joelhawksley:
👏 🙇

PR #175, opened 2020-01-05
Simplify validation of templates and compilation

bostanio:
### Summary

Simplify the way we compile and check validation of the components.
 * Introduced a version of `compile` that doesn't throw and can be called during eager loading
 * Introduced a version of `compile!` that will throw that is called right before rendering
 * Provides a simpler entry point to allow Components without initializers with Ruby 2.7 #174.

With this change we can move the invalid components back into their regular spot. Good proof that we've made the world a little bit better.


lib/action_view/component/base.rb
```diff
@@ -37,7 +37,7 @@ class Base < ActionView::Base
       # <span title="greeting">Hello, world!</span>
       #
       def render_in(view_context, *args, &block)
-        self.class.compile
+        self.class.compile!
```
bostanio: Use the version of compile that throws. This ensures the old behavior of getting exceptions for various mistakes

lib/action_view/component/base.rb
```diff
@@ -110,40 +110,41 @@ def call_method_name(variant)
...
-          # method#source_location to look up the file name
-          # of the component.
-          #
-          # If we were able to only support Ruby 2.7+,
-          # We could just use Module#const_source_location,
-          # rendering this unnecessary.
-          raise NotImplementedError.new("#{self} must implement #initialize.") unless has_initializer?
-
-          instance_method(:initialize).source_location[0]
+          @source_location ||=
```
bostanio: Caching this reflection seems like a great idea. We call this a bunch of times when compiling a template.

lib/action_view/component/base.rb
```diff
@@ -110,40 +110,41 @@ def call_method_name(variant)
...
+                initialize_method = instance_method(:initialize)
+                initialize_method.source_location[0] if initialize_method.owner == self
+              end
         end
 
         def eager_load!
-          self.descendants.each do |descendant|
-            descendant.compile if descendant.has_initializer?
-          end
+          self.descendants.each(&:compile)
```
bostanio: Oh Yeah....

lib/action_view/component/base.rb
```diff
@@ -110,40 +110,41 @@ def call_method_name(variant)
...
 
+        def compile!
+          compile(true)
+        end
+
         # Compile templates to instance methods, assuming they haven't been compiled already.
         # We could in theory do this on app boot, at least in production environments.
         # Right now this just compiles the first time the component is rendered.
-        def compile
+        def compile(raise_if_invalid = false)
```
rainerborene: I suggest following the same protocol of `save!` method and use `validate: false` as argument.

lib/action_view/component/base.rb
```diff
@@ -189,20 +190,32 @@ def templates
             end
         end
 
-        def validate_templates
+        def valid?(raise_if_invalid)
```
rainerborene: Not a common practice to have question mark methods triggering exceptions. I think this should be renamed to something like `validate`.

lib/action_view/component/base.rb
```diff
@@ -189,20 +190,32 @@ def templates
...
+
           if templates.empty?
-            raise NotImplementedError.new("Could not find a template file for #{self}.")
+            raise NotImplementedError.new("Could not find a template file for #{self}.") if raise_if_invalid
+            return false
           end
 
-          if templates.select { |template| template[:variant].nil? }.length > 1
-            raise StandardError.new("More than one template found for #{self}. There can only be one default template file per component.")
+          if templates.count { |template| template[:variant].nil? } > 1
```
dylnclrk: Not really crucial, but won't the invalid variants check below duplicate this check? I wonder if we could merge the two.

test/action_view/integration_test.rb
```diff
@@ -148,4 +148,8 @@ class IntegrationTest < ActionDispatch::IntegrationTest
   test "compiles unreferenced component" do
     assert UnreferencedComponent.compiled?
   end
+
+  test "does not compile components without initializers" do
```
dylnclrk: 👌

test/action_view/invalid_components_test.rb
```diff
@@ -1,11 +1,5 @@
 # frozen_string_literal: true
 
-require "invalid_components_test_helper"
```
dylnclrk: neat! It's nice we can have invalid components in `app/components` now 😄 and that the app wont complain until you try to render the component.

lib/action_view/component/base.rb
dylnclrk: 💅

joelhawksley:
Thanks for digging into this @bostanio!

In general, I'm wary of moving the existing validations out of the eager loading call chain. Doing so would reduce the safety of those checks, as we would then have to ensure we have tests that render every component in an application in order to catch the same set of errors, right?

bostanio:
@joelhawksley I think I agree and it's actually easier that way at least in the sense of the code.  What inspired me to do that was that we we not compiling templates without initializers which are just as broken as components without templates or too many templates. If we are cool with those blowing up eager loading then the code is a bunch simpler.

joelhawksley:
@bostanio what if we added a check for no-initializer components to the eager load step?

bostanio:
We have that already. The intent of this PR was to simplify that so we don't have the deal with initializers as part of eager load (and make the 2.7 support have a smaller blast radius)

Why wouldnt we want want no initializers components to blow up on eager loading? Why are the different from too many or two few template components?

test/action_view/invalid_components_test.rb
joelhawksley: I'm torn as to whether this is a good thing or not. Can you elaborate why you think it's good that the app doesn't raise an error until you try to render the component, vs. when it compiles the component on boot, which only happens when eager loading is turned on?

bostanio: I'm torn too. But in general Rails is lazy about a bunch of things like this and/or as a developer you only figure you broke something when you 'run it' (and lets hope you run it in your test suite). For example
1. I forget to add a view for a controller action
2. The table backing a model is missing/typoed
3. I've typoed the name of some ActiveRecored relation and AR can't find the association by convention
etc. etc.
Its fair to say that some (maybe all) of those cases are hard for Rails to check for at eager load time but that doesn't detract from the fact that those kinds of surprises are not unusual in Rails in general. 

On the flip side I worry a bit that failing at eager load time could be really disruptive. Suppose I'm in the middle of some refactor, I'm moving files around, renaming components, I haven't moved/renamed the views yet because I'm focused on the Component API but now I can't even load the app or maybe run my test suite? That feels pretty ugly.

Lastly, I think we should be consistent and treat all of these 'template error' cases the same way:

1. no initialize
1. no views at all
1. no default view (is this ok if we have variants?)
1. more than one default view
1. to many variant views for the same variant

Note that we currently treat 1 differently than 2, 4 & 5.

joelhawksley: @bostanio 💥 you make a very strong point here.

We've always tried to have this library behave consistently with Rails. With that goal in mind, I think we're right to make this change. 

Thank you for taking the time to explain your perspective ❤️

lib/action_view/component/base.rb
joelhawksley: Agreed. @bostanio mind making this change?

lib/action_view/component/base.rb
joelhawksley: Agreed. We don't really _need_ to have this method have a return value, as it halts the call chain by raising an error if there are any issues.

test/action_view/invalid_components_test.rb
```diff
@@ -36,6 +30,6 @@ def test_raises_error_when_more_than_one_sidecar_template_for_a_variant_is_prese
       render_inline(TooManySidecarFilesForVariantComponent)
     end
 
-    assert_includes error.message, "More than one template found for variant 'test' in TooManySidecarFilesForVariantComponent"
+    assert_includes error.message, "More than one template found for variant(s) 'test, testing' in TooManySidecarFilesForVariantComponent"
```
joelhawksley: What if we refactored the error message to use the `pluralize` method instead of using `(s)`?

joelhawksley: (and to use `to_sentence` for the list of templates)

test/action_view/invalid_components_test.rb
dylnclrk: Sorry for not getting back yesterday.

@bostanio did a great job of making the case for this. Like he said, my thinking was that it's nice to match how Rails doesn't blow up when you forget to write a view for a controller until you call `render`. Which feels less disruptive when you're adding/refactoring a component.

But it can also be helpful to see those errors sooner, so 🤷‍♂I could see an argument for wanting it the other way. But I think the consistency with Rails is valuable, so I'd vote for not blowing up on eager load.

lib/action_view/component/base.rb
bostanio: we do need it to return a value because we need the ability to call compile, not raise an exception but not blow up later because templates are screwy. I'll figure out a solution

lib/action_view/component/base.rb
bostanio: We could do it that way but its a different error so I think its simpler separate.

lib/action_view/component/template_error.rb
```diff
@@ -0,0 +1,11 @@
...
+
+module ActionView
+  module Component
+    class TemplateError < StandardError
+      def initialize(errors)
+        super(errors.join(", "))
+      end
+    end
+  end
+end
```
bostanio: I added a our own Error. We could maybe do better than concating the errors but it probably does for now.

lib/action_view/component/base.rb
```diff
@@ -110,34 +110,40 @@ def call_method_name(variant)
...
         # Right now this just compiles the first time the component is rendered.
-        def compile
+        def compile(validate: false)
           return if compiled?
 
-          validate_templates
+          if template_errors.present?
+            raise ActionView::Component::TemplateError.new(template_errors) if validate
+            return false
+          end
```
bostanio: The logic of compilation is much easier to grok now. Plus we can report on all the mistakes you made with the component in one go.

lib/action_view/component/base.rb
```diff
@@ -177,26 +184,33 @@ def templates
...
+              invalid_variants = templates
+                                   .group_by { |template| template[:variant] }
+                                   .map { |variant, grouped| variant if grouped.length > 1 }
+                                   .compact
 
-            raise StandardError.new("More than one template found for variant '#{variant}' in #{self}. There can only be one template file per variant.")
-          end
+              unless invalid_variants.empty?
+                errors << "More than one template found for #{'variant'.pluralize(invalid_variants.count)} #{invalid_variants.map { |v| "'#{v}'" }.to_sentence} in #{self}. There can only be one template file per variant."
+              end
```
bostanio: @joelhawksley pluralize and to_sentence

lib/action_view/component/railtie.rb
```diff
@@ -35,9 +35,7 @@ class Railtie < Rails::Railtie # :nodoc:
 
       initializer "action_view_component.eager_load_actions" do
         ActiveSupport.on_load(:after_initialize) do
-          ActionView::Component::Base.descendants.each do |descendant|
-            descendant.compile if descendant.has_initializer? && config.eager_load
-          end
+          ActionView::Component::Base.descendants.each(&:compile)
```
joelhawksley: Does this mean we'll always be compiling components even if eager loading is turned off?

bostanio: I don't think so. I think i tested turning eager load off in test.rb and the spec fails.

joelhawksley: Confirmed locally ☝️

PR #180, opened 2020-01-14
Remove the unneeded ComponentExamplesController and simplify preview rendering

jonspalmer:
### Summary

Remove ComponentExamplesController in favor of inline the ComponentsController itself doing the rendering.

One less moving part for this gem to contain AND we don't have to include the TestHelpers in Preview

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -199,7 +199,7 @@ module ActionView
 
         def matching_views_in_source_location
           return [] unless source_location
-          (Dir["#{source_location.sub(/#{File.extname(source_location)}$/, '')}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location])
+          (Dir["#{source_location.chomp(File.extname(source_location))}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location])
         end
 
         def templates
--- a/lib/action_view/component/preview.rb
+++ b/lib/action_view/component/preview.rb
@@ -26,9 +26,7 @@ module ActionView
             layout = @layout.nil? ? "layouts/application" : @layout
           end
 
-          Rails::ComponentExamplesController.render(template: "examples/show",
-                                                    layout: layout,
-                                                    assigns: { example: example_html })
+          ApplicationController.render(inline: "<%= raw('#{example_html}') %>", layout: layout)
         end
 
         # Returns the component object class associated to the preview.
--- a/lib/action_view/component/preview.rb
+++ b/lib/action_view/component/preview.rb
@@ -6,10 +6,9 @@ module ActionView
   module Component # :nodoc:
     class Preview
       extend ActiveSupport::DescendantsTracker
-      include ActionView::Component::TestHelpers
 
-      def render(component, *locals, &block)
-        render_inline(component, *locals, &block)
+      def render(component, **args, &block)
+        { component: component, args: args, block: block }
       end
 
       class << self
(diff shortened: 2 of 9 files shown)


lib/action_view/component/preview.rb
```diff
@@ -26,9 +26,7 @@ def call(example, layout: nil)
             layout = @layout.nil? ? "layouts/application" : @layout
           end
 
-          Rails::ComponentExamplesController.render(template: "examples/show",
-                                                    layout: layout,
-                                                    assigns: { example: example_html })
+          ApplicationController.render(inline: "<%= raw('#{example_html}') %>", layout: layout)
```
jonspalmer: We can just render inline to apply the layout.

lib/action_view/component/base.rb
```diff
@@ -199,7 +199,7 @@ def with_content_areas(*areas)
 
         def matching_views_in_source_location
           return [] unless source_location
-          (Dir["#{source_location.sub(/#{File.extname(source_location)}$/, '')}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location])
+          (Dir["#{source_location.chomp(File.extname(source_location))}.*{#{ActionView::Template.template_handler_extensions.join(',')}}"] - [source_location])
```
jonspalmer: `chomp` is  much more readable (and faster) than regex

lib/action_view/component/preview.rb
```diff
@@ -6,10 +6,9 @@ module ActionView
   module Component # :nodoc:
     class Preview
       extend ActiveSupport::DescendantsTracker
-      include ActionView::Component::TestHelpers
 
-      def render(component, *locals, &block)
-        render_inline(component, *locals, &block)
+      def render(component, **args, &block)
+        { component: component, args: args, block: block }
```
jonspalmer: don't do the render_inline thing from test helpers. Instead just minim the api and return the args. We'll later pass them to a true `render` method.

lib/action_view/component/preview.rb
```diff
@@ -19,21 +18,14 @@ def all
...
-          if layout.nil?
-            layout = @layout.nil? ? "layouts/application" : @layout
-          end
-
-          Rails::ComponentExamplesController.render(template: "examples/show",
-                                                    layout: layout,
-                                                    assigns: { example: example_html })
+        # Returns the arguments for rendering of the component in its layout
+        def render_args(example)
+          new.public_send(example).merge(layout: @layout)
```
jonspalmer: no separate need to render.

lib/railties/lib/rails/components_controller.rb
```diff
@@ -4,6 +4,7 @@
 
 class Rails::ComponentsController < Rails::ApplicationController # :nodoc:
   prepend_view_path File.expand_path("templates/rails", __dir__)
+  prepend_view_path "#{Rails.root}/app/views/" if defined?(Rails.root)
```
jonspalmer: this lets us use the templates in the Rails app that includes the gem

test/action_view/preview_test.rb
```diff
@@ -2,10 +2,12 @@
 
 require "test_helper"
 
-class ActionView::PreviewTest < ActionView::Component::TestCase
+class ActionView::PreviewTest < ActionDispatch::IntegrationTest
```
jonspalmer: Maybe I can just move these into the integration_test but I miss rspec's 'context' blocks to organize tests.

joelhawksley:
@jonspalmer thanks for the refactor / simplification! Keep 'em coming!

PR #181, opened 2020-01-14
Render component by condition

xronos-i-am:
### Summary

This PR adds ability to skip render component if required conditions not meeted. Currently we use the following monkey patch:

```ruby
class PopupComponent < ActionView::Component::Base
  attr_reader :current_user

  def initialize(current_user: nil)
    @current_user = current_user
  end

  def render_in(view_context, *)
     @view_context = view_context
     return '' unless show_popup?
  end

  # show popup for signed users only
  def show_popup?
    current_user.present?
  end
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ class PopupComponent < ActionView::Component::Base
     @current_user = current_user
   end
 
-  # show popup only for signed users
+  # show popup for signed users only
   def show_popup?
     current_user.present?
   end
--- a/README.md
+++ b/README.md
@@ -367,6 +367,28 @@ end
 <% end %>
 ```
 
+### Render by condition
+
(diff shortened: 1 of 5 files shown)


jonspalmer:
Why is it important for the component to decide to no-op vs the parent code that is calling render? This same use case could be more directly solved in the view layer:

```erb
<%= if current_user.present? %>
  <%= render(PopupComponent) %>
<%= end %>
```

kylefox:
By pure coincidence, I happened to find this issue which is exactly what I attempted to solve in https://github.com/github/actionview-component/pull/182

joelhawksley:
Let's go with https://github.com/github/actionview-component/pull/182 for now.

PR #184, opened 2020-01-15
Don't assume ApplicationController

jonspalmer:
### Summary

TestHelpers assume the Rails app has an ApplicationController. Its common practice to have one but not guaranteed. Added simple config to allow it to be set as appropriate.


lib/action_view/component/base.rb
```diff
@@ -118,6 +118,12 @@ def request
 
       attr_reader :content, :view_context
 
+      # The controller used for testing components.
+      # Defaults to ApplicationController. This should be set early
+      # in the initialization process and should be set to a string.
+      mattr_accessor :test_controller
+      @@test_controller = "ApplicationController"
```
joelhawksley: What do you think about checking for the existence of an `ApplicationController`, and if not use `ActionController::Base`?

jonspalmer: @joelhawksley Interesting that you asked. I actually tried that at first and it turns out it doesn't work. If you just use `ActionController:Base` then:
a) the view paths for partials default to `/action/base/`
b) there are no url helpers
c) view helpers aren't included in the controller

I was able to hack around a) and b) but got stumped on c). There is a whole bunch of magic that Rails adds to  controllers included in the app that IMO we should just be happy to use. (We get things like default layout etc. that way too)

BTW I was inspired by the way Devise handles a similar problem:
https://github.com/heartcombo/devise/blob/master/lib/devise.rb#L234

and my mind was then blown that you can do this:
https://github.com/heartcombo/devise/blob/master/app/controllers/devise_controller.rb

joelhawksley:
This patch looks good 👍 

Mind updating the README to describe this functionality?

jonspalmer:
@joelhawksley  README updated.

joelhawksley:
👏

PR #185, opened 2020-01-15
Separate context initialization and render itself

xronos-i-am:
### Summary

This PR separates context initalization from render itself. 

My use cases of this:
1) can override `compile_and_render` to add cache
```ruby
def compile_and_render
  Rails.cache.fetch([cache_key_base, user_type, :footer], expires_in: 30.minutes) do
    super
  end
end
```
Caching at controller side decreases logic complexity of template part. Also it allows do less work to show component when cached, it should affects performance

2) can override `compile_and_render` to add inherited components
```ruby
class PriceComponent < ActionView::Component::Base
  def compile_and_render
    if user_type == 'retail'
      render(Product::RetailPriceComponent)
    else
      render(Product::PartnerPriceComponent)
    end
  end
end
```

Also please note moving `render?` check to `compile_and_render`. It allows to add some additional checks to overrided `render?` method:
```ruby
def render?
  view_context.cookies[:show] == false
end
```

And finally i had to separate `MissingInitializerError`. Several reasons here: 
- `virtual_path` requires initializer check to be done
But also:
- Missing initializer is not "template" error by meaning
- Should raise this error as soon as possible, that's why i move it to beginning of `render_in`

### DIFF
--- a/lib/action_view/component.rb
+++ b/lib/action_view/component.rb
@@ -17,6 +17,7 @@ module ActionView
     autoload :TestCase
     autoload :RenderMonkeyPatch
     autoload :TemplateError
+    autoload :MissingInitializerError
   end
 end
 
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -40,15 +40,20 @@ module ActionView
       # <span title="greeting">Hello, world!</span>
       #
       def render_in(view_context, *args, &block)
-        return "" unless render?
-
-        self.class.compile!
         @view_context = view_context
         @view_renderer ||= view_context.view_renderer
         @lookup_context ||= view_context.lookup_context
         @view_flow ||= view_context.view_flow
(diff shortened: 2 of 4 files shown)


jonspalmer:
There are a few things going on in the PR and I think we'd be smart to separate them:

 1. Do the render check after context is setup. 
 1. Provide a "render" method that can be overridden independently of setting up context.

The first is not contorversial IMO. Let's just do that. I believe it should be as simple as moving the render check and nothing more. (Maybe some code comments about the ordering.

The second is a.bit more nuanced. I don't believe we should allow/expect the the compilation to be overriden. Note in some of you examples you don't call `super`. We shouldn't make it easy for the overrider to forget compilation. I don't believe it's important to be able to cache to avoid the complication code given it runs at most once.

How about two PRs the focus on those issues?

xronos-i-am:
> How about two PRs the focus on those issues?

#186 #187

PR #187, opened 2020-01-16
Separated render method

xronos-i-am:
### Summary

This PR separates context initalization from render itself. 

Use cases:
1) can override `render_template` to add cache
```ruby
def render_template
  Rails.cache.fetch([cache_key_base, user_type, :footer], expires_in: 30.minutes) do
    super
  end
end
```
```

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -48,6 +48,10 @@ module ActionView
         @virtual_path ||= virtual_path
         @variant = @lookup_context.variants.first
 
+        @content = view_context.capture(self, &block) if block_given?
+
+        validate!
+
         return "" unless render?
 
         render_template(&block)
@@ -72,12 +76,6 @@ module ActionView
       def render_template(&block)
         old_current_template = @current_template
         @current_template = self
-
-        @content = view_context.capture(self, &block) if block_given?
-
-        validate!
-
-        return "" unless render?
 
         send(self.class.call_method_name(@variant))
       ensure
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -48,6 +48,28 @@ module ActionView
         @virtual_path ||= virtual_path
         @variant = @lookup_context.variants.first
 
+        return "" unless render?
+
+        render_template(&block)
+      end
+
+      # Method which actual do render. Called by #render_in
+      # For example, it can be used to cache component
+      #
+      # block: optional block to be captured within the view context
+      #
+      # Example subclass:
+      #
+      # app/components/my_component.rb:
+      # class MyComponent < ActionView::Component::Base
+      #   def render_template(&block)
+      #     Rails.cache.fetch(cache_key) do
+      #       super
+      #     end
+      #   end
+      # end
+      #
+      def render_template(&block)
         old_current_template = @current_template
         @current_template = self
 
--- a/README.md
+++ b/README.md
@@ -417,6 +417,21 @@ end
 <%= render(ConfirmEmailComponent, user: current_user) %>
 ```
 
+### Override Render
+
+Components can override a `#render_template` method which renders compiled template. For example, you can wrap rendering with controller-level caching:
+
+```ruby
+# app/components/overrided_render_component.rb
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by joelhawksley** [comment]
@jonspalmer I know you reviewed the original PR including this work, so I'm going to ask for your review here.

@xronos-i-am I'd be interested in mimicking the `around` filter API here: https://guides.rubyonrails.org/action_controller_overview.html#after-filters-and-around-filters

So we'd allow something like `around_render`.

Also, let's add a test and README update for this ❤️

**Comment by jonspalmer** [comment]
I understand the first use case. Adding caching makes sense to me (though I believe we need to make that a first class feature of AV:C eventually).

I struggle a little with the second use case. Overriding `render_template` to completely eliminate the guts of actually rendering the component feels like an anti pattern. (as evidenced by the empty template).

For example IMO opinion it would be MUCH more natural to solve via composition like this:

`price_component.rb`:
```ruby
class PriceComponent < ActionView::Component::Base
  attr_reader :user_type

  def initialize(user_type)
    @user_type = user_type
  end

  def product_component
    user_type == 'retail' ? Product::RetailPriceComponent : Product::PartnerPriceComponent
  end
end
```

`price_component.html.erb`:
```erb
<%= render(product_component) %>
```

It's particularly important for libraries like this that the specs act as documentation of the features we support. As such I think its important we have the spec example be a better example of a good use of overriding `render_template`. Could we actually build some sort of caching example?

jonspalmer:
@joelhawksley I'm all about something like around filter. It doesn't feel like you should be able to completely reimplement or replace the rendering. Instead you can augment it, wrap the content, prevent rendering, return different content (from cache) etc. Its possible we should do the `render?` this way too.

xronos-i-am:
Please note also. Integration test "rendering component with caching" doesn`t really check cache. If cache doesn't work the test not fail

jonspalmer:
To @joelhawksley 's point still needs some Readme additions.

jonspalmer:
Can we add a component_test around this as well as integration?

xronos-i-am:
> Also, let's add a test and README update for this
> To @joelhawksley 's point still needs some Readme additions.
Added readme, but my English is not strong enough, please check

xronos-i-am:
> So we'd allow something like `around_render`.

I started to [add callbacks](https://github.com/xronos-i-am/actionview-component/commit/86ab1604d358d32943c502a96ab894fd814b7747), but not succeeded yet

joelhawksley:
@xronos-i-am thank you for the work on this? What do you think about the proposed `around_render` API? I feel like it would better match existing Rails APIs.

xronos-i-am:
@joelhawksley `around_render` can't replace this PR

this won`t work:
```ruby
around_render :render_cached
def render_cached
    yield
  end
end
```

Around callback asumes that block will yielded. If component is cached the `yield` command will not be called. And in that case `run_callbacks` will return `nil`. That is how `ActiveSupport::Callbacks` works

```ruby
def build_class(callback)
  Class.new {
    include ActiveSupport::Callbacks
    define_callbacks :render
    set_callback :render, :around, callback
    def run; run_callbacks(:render) { 'rendered' }; end
  }
end

klass = build_class(->(instance, block) { block.call })
klass.new.run

klass = build_class(->(instance, block) { "cached" })
klass.new.run
```

```irb
[23] pry(main)> klass = build_class(->(instance, block) { block.call })
=> #<Class:0x00007fa6fd768580>
[24] pry(main)> klass.new.run
=> "rendered"
[25] pry(main)> klass = build_class(->(instance, block) { "cached" })
=> #<Class:0x00007fa6fd5f5478>
[26] pry(main)> klass.new.run
=> nil
```

lib/action_view/component/base.rb
```diff
@@ -50,6 +50,10 @@ def render_in(view_context, *args, &block)
 
         return "" unless render?
 
+        render_template(&block)
+      end
+
+      def render_template(&block)
```
jcoyne: You might want to add some documentation here about the expected use case of a sub-class overriding this method to do caching.

jcoyne: Thanks! This is much better!

PR #189, opened 2020-01-18
Fix generator placing namespaced components in the root directory

asgerb:
### Summary

When generating namespaced components, the generated class names are correct, but the generated files are placed in the root `app/components` directory, and not in the namespace subdirectory.

This means running `rails g component admins/user` will correctly generate `Admins::UserComponent`, but the generated files look like this:
* generated: `app/components/user_component.rb`
should be: `app/components/admins/user_component.rb`
* generated: `app/components/user_component.html.erb` 
should be: `app/components/admins/user_component.html.erb`
* generated: `test/components/user_component_test.rb`
should be: `app/components/admins/user_component_test.rb`

This PR fixes that by adding the `class_path` to the generated file path and adds a test for the generator.

### Other Information

If/when this gets merged I've got another PR ready that separates the template generation into a separate `erb` generator specifically and adds a generator for `slim` (which I use in my projects).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -2,3 +2,7 @@ inherit_gem:
   rubocop-github:
     - config/default.yml
     - config/rails.yml
+
+AllCops:
+  Exclude:
+    - "test/tmp/**/*"
--- a/test/generators/component_generator_test.rb
+++ b/test/generators/component_generator_test.rb
@@ -1,0 +1,52 @@
+require "test_helper"
+require "rails/generators/test_case"
+require "rails/generators/component/component_generator"
+
+Rails.application.load_generators
+
+class ComponentGeneratorTest < Rails::Generators::TestCase
+  tests Rails::Generators::ComponentGenerator
+  destination File.expand_path("../tmp", __dir__)
+  setup :prepare_destination
+
+  arguments %w[user]
+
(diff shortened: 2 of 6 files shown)


.rubocop.yml
```diff
@@ -2,3 +2,7 @@ inherit_gem:
   rubocop-github:
     - config/default.yml
     - config/rails.yml
+
+AllCops:
+  Exclude:
+    - "test/tmp/**/*"
```
jonspalmer: should we .gitignore this too?

test/generators/component_generator_test.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def test_component_with_namespace
+    Thor::LineEditor.stub :readline, "n" do
+      run_generator %w[admins/user]
+    end
+
+    assert_file "app/components/admins/user_component.rb", /class Admins::UserComponent < /
+    assert_file "app/components/admins/user_component.html.erb"
+  end
+end
```
jonspalmer: Nice work 🚀

.rubocop.yml
asgerb: Good point! After the test the directory gets cleaned up, but doesn't hurt to add it to .gitignore too 🙂

PR #190, opened 2020-01-18
Add separate template engine generators (erb & slim)

asgerb:
### Summary

This PR moves the template generation into a separate `erb` generator and adds a generator for the `slim` template engine.

### Other Information

Moving the template generation into a separate generator, prompted me to rethink the current use of `ask?`ing the user if the component should require content.

As I didn't like duplicating the `ask?` in the template generator (it meant the question would be posed twice) my initial approach was to remove it in favor of adding a `--require-content` `class_option`.

This changed the use of the generator though, and as I wasn't sure how welcome of a change it would be, I set about trying to find a way to pass the result of the `ask?` to the template generators. After a bit of digging around the docs of Rails and Thor, I managed to find a solution. This involves having a `--require-content` option in the template generators, invoking them after we've created the component file (and the `ask?` has been called) and then passing the response on to them.

The solution is a bit convoluted and I think it would be cleaner to simply go the `class_option` way, but I'm curious to hear your thoughts!


test/generators/component_generator_test.rb
```diff
@@ -13,42 +13,50 @@ class ComponentGeneratorTest < Rails::Generators::TestCase
...
+      assert_match(/<div>Add User template here<\/div>/, view)
     end
   end
 
-  def test_component_without_required_content
-    Thor::LineEditor.stub :readline, "n" do
-      run_generator
-    end
+  def test_component_with_required_content
+    with_required_content { run_generator }
```
jonspalmer: why reverse the order of these tests. It makes this diff super hard to follow.

lib/rails/generators/slim/component_generator.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "rails/generators/erb/component_generator"
+
+module Slim
+  module Generators
+    class ComponentGenerator < Erb::Generators::ComponentGenerator
```
jonspalmer: Why inherit from the ERB version but effectively override everything? Can we just have a true base class and just parameterize the template? In fact do we need classes at all? Can't we just parameterize a single class?

lib/rails/generators/slim/component_generator.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "rails/generators/erb/component_generator"
+
+module Slim
+  module Generators
```
jonspalmer: these module names don't line up with the file locations. I'm surprised Rails doesn't bark about this.

jonspalmer:
Couple of comments:
a) how do we test the new slim version?
b) can we add Haml while we're here?

lib/rails/generators/slim/component_generator.rb
asgerb: I had the same feeling but I've simply followed the pattern as I saw it in the erb generators from [rails](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/erb/scaffold/scaffold_generator.rb) and [slim-rails](https://github.com/slim-template/slim-rails/blob/master/lib/generators/slim/scaffold/scaffold_generator.rb)

lib/rails/generators/slim/component_generator.rb
asgerb: Good point, I've simply followed the pattern from `slim-rails` (see https://github.com/slim-template/slim-rails/blob/master/lib/generators/slim/scaffold/scaffold_generator.rb) and honestly did not put much thought into refactoring/cleaning it up. Will take another look!

test/generators/component_generator_test.rb
asgerb: Good point and I'm not sure why I did this 😬 Added a new commit, should be easier to follow now.

asgerb:
I'll take a stab at a) and b)

lib/rails/generators/slim/component_generator.rb
asgerb: Come to think of it, my thought was to make it easier in the future to move the generators from `actionview-component` to what I thought would be their final destinations (slim-rails and rails).

jonspalmer: yeah ok. The existing convention is super weird but seems like we should stick with it.

asgerb:
@jonspalmer added the haml generator and separate tests for each template engine. Also added tests in the component generator for the template engines (where I simply pass the template-engine as an option) just to be sure the implementation works.

lib/rails/generators/haml/component_generator.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "rails/generators/erb/component_generator"
+
+module Haml
+  module Generators
+    class ComponentGenerator < Erb::Generators::ComponentGenerator
```
jonspalmer: should we inherit from `Base` here too?

asgerb: It seems to be the norm to inherit from the corresponding Erb generator: https://github.com/haml/haml-rails/blob/master/lib/generators/haml/scaffold/scaffold_generator.rb#L5

joelhawksley:
@jonspalmer mind following up here?

lib/rails/generators/slim/component_generator.rb
jonspalmer: I understand the patterns that Rails uses given they have lots of generators. Personally I think slim-rails copying that probably isn't what I'd do. We're not going to upstream this anymore so the value of matching that pattern is reduced. 

I think I'd be tempted to simplify down to a single class that get parameterized but this is generator code some I'm not as bothered about the clean factoring.  If we find we need to keep changing this and making parallel changes across the implementations maybe we refactor then

joelhawksley:
@asgerb mind updating the README and CHANGELOG for these changes?

asgerb:
@joelhawksley you bet, perhaps I better rebase first?

asgerb:
@joelhawksley I've added a CHANGELOG entry and updated the README. Let me know if there is an issue with the rebasing, my git skills aren't the best!

PR #192, opened 2020-01-20
add (before|around|after)_render hooks

xronos-i-am:
### Summary

Render callbacks #187

### DIFF
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -8,11 +8,14 @@ module ActionView
       include ActiveModel::Validations
       include ActiveSupport::Configurable
       include ActionView::Component::Previewable
+      include ActionView::Component::Callbacks
 
       delegate :form_authenticity_token, :protect_against_forgery?, to: :helpers
 
       class_attribute :content_areas, default: []
       self.content_areas = [] # default doesn't work until Rails 5.2
+
+      attr_accessor :rendered_output
 
       # Entrypoint for rendering components. Called by ActionView::Base#render.
       #
@@ -48,6 +51,7 @@ module ActionView
         @virtual_path ||= virtual_path
         @variant = @lookup_context.variants.first
 
+<<<<<<< HEAD
         old_current_template = @current_template
         @current_template = self
 
@@ -60,10 +64,18 @@ module ActionView
         send(self.class.call_method_name(@variant))
       ensure
         @current_template = old_current_template
+=======
+        run_callbacks :render do
+          render_template(&block)
+        end
+
+        rendered_output
+>>>>>>> add (before|around|after)_render hooks
       end
 
-      def render?
-        true
+      # method used in (before|around)_render hooks to halt the filter chain
+      def render_performed?
+        rendered_output != nil
       end
 
       def initialize(*); end
@@ -112,6 +124,21 @@ module ActionView
       end
 
       private
+
+      # Method which actual do render. Called by #render_in
+      # Sets @rendered_output
+      def render_template(&block)
+        old_current_template = @current_template
+        @current_template = self
+
+        @content = view_context.capture(self, &block) if block_given?
+
+        validate!
+
+        self.rendered_output = send(self.class.call_method_name(@variant))
+      ensure
+        @current_template = old_current_template
+      end
 
       def request
         @request ||= controller.request
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -51,26 +51,15 @@ module ActionView
         @virtual_path ||= virtual_path
         @variant = @lookup_context.variants.first
 
-<<<<<<< HEAD
-        old_current_template = @current_template
-        @current_template = self
-
         @content = view_context.capture(self, &block) if block_given?
 
         validate!
 
-        return "" unless render?
-
-        send(self.class.call_method_name(@variant))
-      ensure
-        @current_template = old_current_template
-=======
         run_callbacks :render do
           render_template(&block)
         end
 
         rendered_output
->>>>>>> add (before|around|after)_render hooks
       end
 
       # method used in (before|around)_render hooks to halt the filter chain
@@ -130,10 +119,6 @@ module ActionView
       def render_template(&block)
         old_current_template = @current_template
         @current_template = self
-
-        @content = view_context.capture(self, &block) if block_given?
-
-        validate!
 
         self.rendered_output = send(self.class.call_method_name(@variant))
       ensure
--- a/lib/action_view/component/callbacks.rb
+++ b/lib/action_view/component/callbacks.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require "active_support/concern"
+require "active_support/callbacks"
+
+module ActionView
+  module Component
+    module Callbacks
+      extend ActiveSupport::Concern
+      include ActiveSupport::Callbacks
+
+      included do
+        define_callbacks :render, terminator: ->(component_instance, result_lambda) do
+          result_lambda.call
+          component_instance.render_performed?
+        end
+      end
+
+      module ClassMethods
+        def before_render(*filters, &block)
+          set_callback(:render, :before, *filters, &block)
+        end
+
+        def after_render(*filters, &block)
+          set_callback(:render, :after, *filters, &block)
+        end
+
+        def around_render(*filters, &block)
+          set_callback(:render, :around, *filters, &block)
+        end
+      end
+    end
+  end
+end
--- a/test/action_view/component_test.rb
+++ b/test/action_view/component_test.rb
@@ -350,6 +350,7 @@ class ActionView::ComponentTest < ActionView::Component::TestCase
     assert_includes render_inline(ConditionalRenderComponent, should_render: true).to_html,
                     "<div>component was rendered</div>"
 
+<<<<<<< HEAD
     assert_equal render_inline(ConditionalRenderComponent, should_render: false).to_html,
                     ""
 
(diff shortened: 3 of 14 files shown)


lib/action_view/component/callbacks.rb
```diff
@@ -0,0 +1,31 @@
...
+      include ActiveSupport::Callbacks
+
+      included do
+        define_callbacks :render
+      end
+
+      module ClassMethods
+        def before_render(*filters, &block)
+          set_callback(:render, :before, *filters, &block)
+        end
```
jonspalmer: What does `filters` mean or get used for?

test/app/components/conditional_render_component.rb
```diff
@@ -1,11 +1,11 @@
 # frozen_string_literal: true
 
 class ConditionalRenderComponent < ActionView::Component::Base
-  def initialize(should_render:)
-    @should_render = should_render
+  before_render do
+    throw(:abort) unless @should_render
```
jonspalmer: Can we use a terminator so that returning `false` halts execution? That feels more natural that having to `throw`

lib/action_view/component/callbacks.rb
xronos-i-am: with filters you can send method name instead of block
```ruby
  before_render :render?
  private def render?; end
```
https://github.com/rails/rails/blob/master/activejob/lib/active_job/callbacks.rb

test/app/components/callbacks_component.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+class CallbacksComponent < ActionView::Component::Base
+  before_render :before
+  around_render :around
+  after_render  :after
+
+  attr_accessor :events
```
jonspalmer: this could be `attr_reader`

test/app/components/callbacks_component.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def initialize(events: [])
+    @events = events
+  end
```
jonspalmer: I know its just test code but why allow events to be passed in 
```suggestion
 def initialize()
    @events = []
  end
```

test/app/components/conditional_render_component.rb
xronos-i-am: ActiveSupport::Callbacks use `throw` terminator by default. We can change to `true`/`false` resolvation, but i`m not sure it`s good idea. It this case you have to control return value of `before_render`

test/action_view/component_test.rb
```diff
@@ -354,6 +354,15 @@ def test_conditional_rendering
                     ""
   end
 
+  def test_render_callbacks
+    events = []
+
+    component_instance = CallbacksComponent.new(events: events)
+
+    assert_includes component_instance.render_in(controller.view_context), "Rendered"
+    assert_equal [:before, :before_around, :after, :after_around], events
```
jonspalmer: passing the events and relying on it being used by reference feels a bit strange. 
```suggestion
    assert_equal [:before, :before_around, :after, :after_around], component_instance.events
```

test/app/components/callbacks_component.rb
xronos-i-am: ok

lib/action_view/component/base.rb
```diff
@@ -40,15 +41,20 @@ class Base < ActionView::Base
...
         @view_flow ||= view_context.view_flow
         @virtual_path ||= virtual_path
         @variant = @lookup_context.variants.first
+
+        run_callbacks :render do
+          render_template(&block)
+        end
+      end
+
+      def render_template(&block)
```
jonspalmer: Can we make `render_template` private?

test/app/components/conditional_render_component.rb
jonspalmer: controlling for the return value would be fine in most cases. I wonder what the alternatives are (throwing exceptions for control flow is an anti pattern most of the time). In Rails Controllers a before filter that renders or redirects results in the filter chain halting. Can we isolate some set of method calls that would have the same effect?

xronos-i-am: This set may include something like `stop_render` and I can't suggest nothing more. I think it would be even worse than `throw` because it is additional DSL user need to know.

xronos-i-am: This is not principal for me actually. The decision is on your side

jcoyne:
Can you add an explanation of why you are making this change?  A link to #187 is not enough for me to get the context.

xronos-i-am:
@jcoyne `before_render` seems to be more "rails" way of preventing render. `around_render` can be used for instrumentation for example

jonspalmer:
I like this approach a great deal. It feels more natural to support halting the rendering this way and allows for instrumentation and a bunch of other extensions. The gap wrt #187 is how to support caching but I think we can solve that. The trick is to stop relying on the return value of the callbacks. How about we add an attr_accessor like `rendered_output` to `Base`. Change the render_template to set it:

```ruby
      def render_template(&block)
        old_current_template = @current_template
        @current_template = self

        validate!

        rendered_output = send(self.class.call_method_name(@variant))
      ensure
      end
```

and then in the render_in method instead or relying on the return value of run_callbacks we do this

```ruby
run_callbacks :render do
  render_template(&block)
end

rendered_output
```

That way any before, after or around filter could manipulate the rendered_output and everything works.

For example the simple caching version would be:

```
around_render :render_cached
def render_cached
  rendered_output = Rails.cache.fetch(cache_key) do
    yield
  end
end
``` 

I suspect we could also 'automate' filter halting by checking for a non nil value of rendered_output (similar to how Rails Controllers halt the chain if you call `render` or `redirect_to`).

jcoyne:
@xronos-i-am I'm not criticizing the substance of your PR, I think this is a big improvement, I'm just asking for a better commit message that explains why this work was worth undertaking.  I think that kind of thing can really help future contributors as they search through git history.

xronos-i-am:
@jcoyne is it better?

xronos-i-am:
@jonspalmer tried your approach

> I suspect we could also 'automate' filter halting by checking for a non nil value of rendered_output (similar to how Rails Controllers halt the chain if you call `render` or `redirect_to`).

```ruby
before_render do
  self.rendered_output = '' unless render?
end
```

Don't sure it is much better than `throw` or `true/false` resolving

> That way any before, after or around filter could manipulate the rendered_output and everything works.

```ruby
def render_cached
  self.rendered_output = Rails.cache.fetch(cache_key) do
    yield
  end
end
```

It may be confusing. We have to assign `@rendered_output` to return of `fetch` command. It's easy to write the following and hard to find why it doesn't work:

```ruby
def render_cached
    self.rendered_output = yield
  end
end
```

xronos-i-am:
#187 should be closed if we use approach with assigning `@rendered_output`

jonspalmer:
Taking a step back a little bit on the caching question. Why is it important to do the caching in the ruby class vs in the template where we have access to all the usual Rails cache helpers?

xronos-i-am:
@jonspalmer i want to have minimum overhead when component is cached. There is no need to run additional ruby code which renders a template. Caching whole the component seems to be the same anti-pattern as conditional `render?` in template, which was moved early. I'm pretty sure templates should by as clean as possible, caching logic should be in ActionView::Component class

rainerborene:
I agree with @jonspalmer. There is something else to consider: ActiveSupport callbacks are slow as hell. By the way, that's **one** of the reasons why `sequel` gem is much more faster than `activerecord`. On my `ApplicationComponent` file I usually redefine the `validate!` to an empty method, as native Ruby keywords works just fine for checking arguments. 

It's worth writing a benchmark code, like rendering a thousand list items, to measure the performance of callbacks before proceeding with this PR (if merged).

jonspalmer:
@rainerborene do we have a sense of why validate and callbacks are so slow? The code they run to check things is relatively straightforward so what's the overhead they introduce?

rainerborene:
I didn't have time to go through the AS callback internals. See the results of the benchmark:

```
     with validation     9.901k i/100ms
  without validation   126.863k i/100ms
     with validation    106.268k (± 1.6%) i/s -    534.654k in   5.032420s
  without validation      1.640M (± 3.4%) i/s -      8.246M in   5.035358s

Comparison:
  without validation:  1639630.3 i/s
     with validation:   106268.1 i/s - 15.43x  slower
```

Here is the code of the benchmark itself.

```ruby
require "active_support/all"
require "action_view"
require "action_view/component"
require "benchmark/ips"

class ActionView::Component::Base
  def self.inherited(_child)
    # no routes available here
  end
end

class Row < ActionView::Component::Base
  validates :name, presence: true

  attr_reader :name

  def initialize(name:)
    @name = name
  end
end

Benchmark.ips do |x|
  x.report("with validation") do
    row = Row.new(name: "blah")
    row.validate!
  end

  x.report("without validation") do
    Row.new(name: "blah")
  end

  x.compare!
end
```

jonspalmer:
Got it. That benchmark is apples to oranges though. My question would be can you do the validation checks manually for equivalent effect but faster than the Rails validations?

rainerborene:
@jonspalmer There are dozens of alternative validation libraries out there. Need to see which one performs better than `activemodel` and `activesupport` together. But I think our premises are completely wrong here. 

If I need to validate input data from the component perspective which is the rendering layer, then  something is completely wrong with my design. And what if someone wants to use `dry-validation` or something else? You'll need to monkey patch or override a bunch of methods? I vote to wipe out validation from `actionview-component`.

Using validations just to check the presence (calling `blank?`) doesn't worth this unreasonable performance cost for me. If someone needs validation, then just include ActiveModel modules into the base component class. 

I don't think it's necessary to enforce a validation library just to use components at all.

xronos-i-am:
> I don't think it's necessary to enforce a validation library to use components at all.

totally agree

This PR doesn't have sense for me anymore. Caching of whole component is not possible with merged #196.

PR #198, opened 2020-01-29
Support applying a layout to a block within a template

nickh:
`ActionView::PartialRenderer` supports applying a layout to a block within a template using syntax like:

```ruby
<%= render(layout: "administrator", locals: { user: chief }) do %>
  Title: <%= chief.title %>
<% end %>
```

This currently raises an exception when used in a component template, eg:

```
ActionView::Template::Error: undefined method `_app_views_integration_examples__test_layout_html_erb__2545028457847958506_70198999886460' for #<LayoutComponent:0x00007fb1057557e8>
    app/views/integration_examples/layout.html.erb:2
    test/action_view/integration_test.rb:75:in `block in <class:IntegrationTest>'

```

This change includes a failing test, and adds support for rendering partials with the `layout:` option.

### DIFF
--- a/test/app/views/integration_examples/_test_layout.html.erb
+++ b/test/app/views/integration_examples/_test_layout.html.erb
@@ -1,1 +1,1 @@
-<div>hello,layout world!</div>
+<div class="layout"><%= yield %></div>
--- a/test/app/views/integration_examples/_test_layout.html.erb
+++ b/test/app/views/integration_examples/_test_layout.html.erb
@@ -1,0 +1,1 @@
+<div>hello,layout world!</div>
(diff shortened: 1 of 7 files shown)


test/app/views/integration_examples/layout.html.erb
```diff
@@ -0,0 +1,2 @@
+layout:<%= render partial: "test_layout" %>
+component:<%= render LayoutComponent %>
```
joelhawksley: Might we pass block content to this component and the partial render call above to ensure the two behaviors are the same?

test/app/views/integration_examples/_test_layout.html.erb
```diff
@@ -0,0 +1 @@
+<div>hello,layout world!</div>
```
joelhawksley: Might we have this template `yield` content passed to it?

joelhawksley:
Closing as stale. I've attempted to find a solution to this issue in this repo and in Rails itself, and have come up short.

PR #200, opened 2020-01-31
Update workflow config to include Ruby 2.7

dylnclrk:
### Summary

PR #174 introduces a feature that requires Ruby 2.7. In order to test this feature, need to run CI against 2.7.  

Let's update our workflow configuration accordingly.

### Other Information

Note that the version of bundler used in the workflow was updated to match our Gemfile.lock. It was pinned to 1.14 in #35, but seems to be OK to update (and the old version was causing issues with 2.7).


joelhawksley:
🙇 thanks @dylnclrk!

PR #202, opened 2020-02-03
Optional validation

xronos-i-am:
### Summary

This PR remove hardcoded inclusion of `ActiveModel::Validations` and allow to add it optionally. 

Component with validation [15 times slower](https://github.com/github/actionview-component/pull/192#issuecomment-580746645). I suppose it usable for 1 of 100 cases. Validation is too sloow to run it for every component

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,67 @@ Which returns:
 <span title="my title">Hello, World!</span>
 ```
 
+##### Validation
+
+Components inherited from `ActionView::Component::Base` extends with `ActionModel::Validations` by default. You can skip this behaviour by config:
+
+`config/application.rb`:
+```ruby
+config.action_view_component.validation_module = 'DryValidation'
+```
+
+You can define your own validation mechanism. `validation_module` should be set to the string that represents module name. Validation module has to provide `validate!` method:
+
+`app/contracts/profile_contract.rb`:
+```ruby
+class ProfileContract < Dry::Validation::Contract
+  params do
+    required(:email).filled(:string)
+  end
+
+  rule(:email) do
+    unless value.include?('@')
+      key.failure('has invalid format')
+    end
+  end
+end
+```
+
+`lib/dry_validation.rb`
+```ruby
+module DryValidation do
+  def validate!
+    result = self.class.contract.call(validation_params)
+
+    unless result.errors.empty?
+      raise RuntimeError.new(result.errors.inspect)
+    end
+  end
+end
+```
+
+`app/components/profile_component.rb`:
+```ruby
+class ProfileComponent < ActionView::Component::Base
+  class << self
+    def contract
+      @contract ||= ProfileContract.new
+    end
+  end
+
+  attr_reader :email
+  private :email
+
+  def initialize(email:)
+    @email = email
+  end
+
+  private def validation_params
+    {email: email}
+  end
+end
+```
+
 ##### Supported `render` syntaxes
 
 Components can be rendered via:
--- a/lib/action_view/component.rb
+++ b/lib/action_view/component.rb
@@ -17,6 +17,7 @@ module ActionView
     autoload :TestCase
     autoload :RenderMonkeyPatch
     autoload :TemplateError
+    autoload :NoOpValidations
   end
 end
 
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -5,7 +5,6 @@ require "active_support/configurable"
 module ActionView
   module Component
     class Base < ActionView::Base
-      include ActiveModel::Validations
       include ActiveSupport::Configurable
       include ActionView::Component::Previewable
 
@@ -13,6 +12,10 @@ module ActionView
 
       class_attribute :content_areas, default: []
       self.content_areas = [] # default doesn't work until Rails 5.2
+
+      # `validation_module` should be set to a string if present
+      class_attribute :validation_module, default: "ActiveModel::Validations"
+      self.validation_module = "ActiveModel::Validations" # default doesn't work until Rails 5.2
 
       # Entrypoint for rendering components. Called by ActionView::Base#render.
       #
@@ -128,6 +131,7 @@ module ActionView
       class << self
         def inherited(child)
           child.include Rails.application.routes.url_helpers unless child < Rails.application.routes.url_helpers
+          child.include self.validation_module ? self.validation_module.safe_constantize : ActionView::Component::NoOpValidations
 
           super
         end
--- a/test/action_view/component_test.rb
+++ b/test/action_view/component_test.rb
@@ -365,6 +365,13 @@ class ActionView::ComponentTest < ActionView::Component::TestCase
     assert_empty render_inline(RenderCheckComponent).text, ""
   end
 
+  def test_custom_validation
+    exception = assert_raises RuntimeError do
+      render_inline(CustomValidationComponent)
+    end
+    assert_equal exception.message, "validate! was called"
+  end
+
   def test_to_component_class
     post = Post.new(title: "Awesome post")
 
--- a/lib/action_view/component/no_op_validations.rb
+++ b/lib/action_view/component/no_op_validations.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+require "active_support/concern"
+
+module ActionView
+  module Component # :nodoc:
+    module NoOpValidations
+      extend ActiveSupport::Concern
+
+      included do
+        def validate!
+        end
+      end
+    end
+  end
+end
--- a/test/app/components/custom_validation_component.html.erb
+++ b/test/app/components/custom_validation_component.html.erb
@@ -1,0 +1,1 @@
+Rendered
--- a/test/app/components/custom_validation_component.rb
+++ b/test/app/components/custom_validation_component.rb
@@ -1,0 +1,15 @@
+# frozen_string_literal: true
+
+module CustomValidations
+  def validate!
+    raise RuntimeError.new("validate! was called")
+  end
+end
+
+ActionView::Component::Base.validation_module = "CustomValidations"
+
+class CustomValidationComponent < ActionView::Component::Base
+  def initialize(*); end
+end
+
+ActionView::Component::Base.validation_module = ActiveModel::Validations.name


jonspalmer:
I've run a variant benchmark myself and found slightly different results. If you simply don't use validation , but still call `validate` it's about 3 times slower. However, running a more realistic benchmark that actually includes rendering a view component and prove there was a difference between the two. I'm remain unconvinced that we need this at all.

```
class ErbComponent < ActionView::Component::Base

  def initialize(message:)
    @message = message
  end
end

class ErbNoValidateComponent < ActionView::Component::Base
  def initialize(message:)
    @message = message
  end
end
```

and the benchmark run in the test suite:

```
  def test_benchmark
    Benchmark.ips do |x|
      x.report("with validation") do
        render_inline(ErbComponent, message: "bar") { "foo" }
      end
    
      x.report("without validation") do
        render_inline(ErbNoValidateComponent, message: "bar") { "foo" }
      end
    
      puts x.compare!
    end
  end
```

```
     with validation   759.000  i/100ms
  without validation   783.000  i/100ms
     with validation      8.021k (± 4.8%) i/s -     40.227k in   5.027458s
  without validation      8.045k (± 5.0%) i/s -     40.716k in   5.074260s

Comparison:
  without validation:     8045.3 i/s
     with validation:     8020.6 i/s - same-ish: difference falls within error
```

this issue is rendering dominates validate in terms of total time spent.

jonspalmer:
If we want to support 'turning' off validation there are MUCH simpler ways to do it. Rather that the approach here just add a class_attribute

```ruby
class_attribute :skip_validation

def self.without_validation
  skip_validation = true
end

def self.with_validation
  skip_validation = false
end

....

validate! unless skip_validation

...
```

Them if you want to turnoff validation entirely just call `without_validation` in your ApplicationComponent and you're done (and could turn it back on in other components if you wanted)

xronos-i-am:
> If we want to support 'turning' off validation there are MUCH simpler ways to do it. Rather that the approach here just add a class_attribute

keep in mind ability to change `ActionModel::Validations` to `dry-validation`

dylnclrk:
For what it's worth, I prefer the on-by-default approach that @jonspalmer described. I wonder how many  users will want to make the performance tradeoff and turn them off. 

It seems better to have fewer hurdles to make use of validations.

joelhawksley:
@xronos-i-am thanks for all of your work on this and the related PRs you've had in flight ❤️ 

In order to avoid introducing a breaking change, I'd prefer that we start by making it possible to disable validations, as well as to swap out the validation library. I think you're pretty close here in this PR.

xronos-i-am:
@joelhawksley done

jonspalmer:
I'm concerned that as coded this is still a breaking change that requires the inclusion of the validation module. I have a few other comments but no time for a little while. Will provide a more complete review later today.

lib/action_view/component/validatable.rb
```diff
@@ -0,0 +1,40 @@
...
+
+module ActionView
+  module Component # :nodoc:
+    module Validatable
+      extend ActiveSupport::Concern
+
+      included do
+        if ActionView::Component.extend_validations
+          include ActiveModel::Validations
+        end
```
jonspalmer: What is the importance of making this configurable? What's the cost of including this always even if someone is overriding `validate!`?

lib/action_view/component/validatable.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        # method `validate!` actually performs validation and can be overrided
+        def validate_input!
+          return if self.class.validation_skiped?
+          validate!
+        end
+      end
```
jonspalmer: Why do we need the `validate_input!` wrapper vs just inlining in base as

```validate! unless skip_validation```

File: lib/action_view/component/validatable.rb [hunk 3]
```diff
@@ -0,0 +1,40 @@
...
+
+      class_methods do
+        attr_accessor :skip_validation
+        attr_accessor :validation_decorator
```
jonspalmer: is `validation_decorator` unused?

test/action_view/component_test.rb
```diff
@@ -365,6 +365,30 @@ def test_render_check
...
+  end
+
+  def test_dry_validation
+    assert_includes render_inline(DryValidationComponent, required: "text").text, "Rendered"
+
+    exception = assert_raises RuntimeError do
+      render_inline(DryValidationComponent)
+    end
+    assert_equal exception.message, "required: must be filled"
+  end
```
jonspalmer: Is it important to test dry-validation here? The real test here is that "if I override `validate!` and throw exceptions those exceptions are not rescued` We could do that with any kind of override, there is nothing unique about the fact that it works with dry-validation (which would give us one less dev dependency)

test/app/components/without_validation_component.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class WithoutValidationComponent < ActionView::Component::Base
+  without_validation!
+
+  include ActiveModel::Validations
```
jonspalmer: do we need this `include`?

test/app/components/without_validation_component.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class WithoutValidationComponent < ActionView::Component::Base
```
jonspalmer: Its feels a little strange to have a `WithoutValidationComponent` that has a `validates` method. This isn't how you'd do it in your application and I worry about test code as misleading documentation. 

Instead how about we override the `validate!` method in this class and have it raise

```ruby
def validate!
  raise "validate! was called"
end
```

lib/action_view/component.rb
```diff
@@ -9,14 +9,23 @@ module ActionView
...
 
+    mattr_accessor :extend_validations
+    @@extend_validations = true
+
+    def self.configure
+      yield self
+    end
+
     autoload :Base
+    autoload :Configuration
```
jonspalmer: Why is this needed as a result of these changes?

jonspalmer: What is this module?

README.md
```diff
@@ -155,6 +155,48 @@ Which returns:
...
+      raise RuntimeError.new(result.errors.to_h)
+    end
+  end
+
+  attr_reader :email
+  private :email
+
+  def initialize(email: nil)
+    @email = email
+  end
```
jonspalmer: its just documentation code but the constructor needs to be public.

lib/action_view/component/validatable.rb
xronos-i-am:
```ruby
include ActionView::Component::Validatable
validate_input!
```

vs 

```ruby
include ActionView::Component::Validatable
validate! unless skip_validation
```

In first case you have to know about only one method `validate_input!` of related `Validatable` module in `Base`. While the second one requires you to know about `skip_validation` also. That increase code coupling

lib/action_view/component/validatable.rb
xronos-i-am: yes, mistake

test/action_view/component_test.rb
xronos-i-am: I think it is better to have real world example here

test/app/components/without_validation_component.rb
xronos-i-am: no, mistake

test/app/components/without_validation_component.rb
xronos-i-am: agree

lib/action_view/component.rb
xronos-i-am: mistake

README.md
xronos-i-am: `initialize` is public. The only private method here is `def email`. Do you want to make it public?

lib/action_view/component/validatable.rb
xronos-i-am: I think it is not correct to load module that not used. It might not to has a big cost probably, but it is additional bytes/miliseconds that is unneccessary. People who prefer dry ecosystem and sequel make this choice because of performance as pointed by @rainerborene [here](https://github.com/github/actionview-component/pull/192#issuecomment-580766349) 

If we make optional validation mechanism, I think the next step could be to make `ActiveModel` optional at all

README.md
jonspalmer: ok I misread the code

jonspalmer:
Thanks for the detailed replies. I think we can simplify this entire setup by making the validation logic entirely swappable. How about just do this:

1. Introduce a new configuration setting, a string: `validation_module`. It defaults to "ActiveModel::Validations"
2. Add a new module ActionView::Component::NoOpValidations which has a single method `validate!` that does nothing.
3. in Base add something like the following code:

```ruby
if self.validation_module
  include self.validation_module.constantize
else
  include ActionView::Component::NoOpValidations
end
```

With that setup we are done and with very little code. Users would be free to do a few different things:

- turn off Validation entirely by setting `validation_module` to nil or false in config. 
- Add their own validation module. The contract would be that module HAS to provide a `validate!` method.

In either case users could later override the `validate!` method in any Component to adjust behavior.

Under this setup there is very much less code, and we don't need the indirection and confusion of wrappers like `validate_input` etc. Plus it's more obvious, easier to configure and more flexible.

Thoughts?

xronos-i-am:
> Thoughts?

The problem is users will not be able to override `validation_module`:

```ruby
module AM
  def validate!
    puts "AM validate"
  end
end

module Noop
  def validate!
    puts "Noop"
  end
end

module Dry
  def validate!
    puts "Dry validate"
  end
end

class Base
  cattr_accessor :validation_module
  @@validation_module = 'AM'

  if self.validation_module
    include self.validation_module.constantize
  else
    include Noop
  end
end

class Child < Base
  self.validation_module = 'Dry'
end

Child.new.validate! # this outs "AM validate"
```

The reason is that value of `self.validation_module` of subclass resolves at the moment when subclass is inherited from `Base`. In that moment subclass itself is not evaluated yet. Related question on [SO](https://stackoverflow.com/questions/790626)

jonspalmer:
Sure but I'm pretty sure this pattern would work:

```ruby
module AM
  def validate!
    puts "AM validate"
  end
end

module Noop
  def validate!
    puts "Noop"
  end
end

module Dry
  def validate!
    puts "Dry validate"
  end
end

class Base
  cattr_accessor :validation_module
  @@validation_module = 'AM'

  def self.inherited(subclass)
    subclass.include self.validation_module ? self.validation_module.constantize : Noop
  end

class Child < Base
 end

Child.new.validate! # this outs "AM validate"

Base.validation_module = 'Dry'

class Child2 < Base
end

Child2.new.validate! # this outs "Dry validate"
```

Provided the validation_module is set via Config before the subclasses are autoloaded this would give us what we need.

xronos-i-am:
@jonspalmer done

lib/action_view/component/base.rb
```diff
@@ -14,6 +13,13 @@ class Base < ActionView::Base
       class_attribute :content_areas, default: []
       self.content_areas = [] # default doesn't work until Rails 5.2
 
+      # `validation_module` should be set to a string if present
+      class_attribute :validation_module, default: "ActiveModel::Validations"
+      self.validation_module = "ActiveModel::Validations" # default doesn't work until Rails 5.2
+
+      # when set to `true` validation does not run
+      class_attribute :skip_validation
```
rainerborene: I think we should deprecate the usage of `skip_validation: false`, meaning that validations will be gone on the next major version of `actionview-component`. Other than that, this PR is 💯

jonspalmer: I think we just eliminate it entirely. I don't think we need to introduce this concept at all.

test/app/components/dry_validation_component.rb
```diff
@@ -0,0 +1,43 @@
...
+  def initialize(required: nil)
+    @required = required
+  end
+
+  private def component_params
+    {required: required}
+  end
+end
+
+ActionView::Component::Base.validation_module = "ActiveModel::Validations"
```
jonspalmer: Can we use `ActiveModel::Validations.name` here?

test/app/components/dry_validation_component.rb
```diff
@@ -0,0 +1,43 @@
...
+module DryValidations
+  def validate!
+    result = self.class.contract.call(component_params)
+
+    unless result.errors.empty?
+      full_messages = result.errors.to_h.map { |f, messages| "#{f}: #{messages.join(', ')}" }
+      raise RuntimeError.new(full_messages.join(", "))
+    end
+  end
+end
```
jonspalmer: I understand the 'real world' example argument but we're trying to test/prove that you can provide an alternative `validate!` method. This feels like a few more moving parts that we need to do that. Can we simplify (maybe not have to need to use dry-validation and force the reader to understand that gem to grok this test). 

I'm all for documenting using dry-validation in the Readme as a real example.

README.md
```diff
@@ -155,6 +155,35 @@ Which returns:
...
+```
+
+You can also skip validation in some components with `skip_validation`:
+
+`app/components/validation_disabled.rb`:
+```ruby
+class ValidationDisabled < ActionView::Component::Base
+  self.skip_validation = true
+end
+```
```
jonspalmer: I don't see the need for both mechanisms and its confusing to provide the choice. Most folks will stick wit h the built in ActiveModel Validations. If you want finer control you supply your own module. Users are free to build features like `skip_validation` into their own validation module but I don't think we need it out of the box.

joelhawksley:
@xronos-i-am thank you so much for your work on this PR, and to @jonspalmer for the thorough review.

After speaking with @rainerborene, @jonspalmer, and several folks here at work, we think it's best to move forward with removing validations from the library in `v2`. 

We'll start by moving the call to `validate!` to a `before_render_check` method, which can be subclassed to be a no-op.

Then, when we introduce the new `ViewComponent::Base` class, we will remove the `validate!` call and the inclusion of `ActiveModel::Validations`.

xronos-i-am:
@joelhawksley ok. When the second version is planned to delivery?

joelhawksley:
@xronos-i-am before RailsConf / 6.1 😄

xronos-i-am:
@joelhawksley ok ))

PR #204, opened 2020-02-12
update README example to use instance variables

ultrawebmarketing:
### Summary

The examples in the README show component templates using local variables instead of instance variables. If you followed the examples, you would raise "undefined local variable or method" error.

### Other Information

none


joelhawksley:
@ultrawebmarketing @jonspalmer do you think it would be better to recommend use of attr_readers or instance variables? I have a distaste for ivars myself, but it's probably baseless in this context.

ultrawebmarketing:
you make the call. i'm impartial to either.

ultrawebmarketing:
So, maybe my previous comment was a little premature. Here is something to consider. When using attr_reader, although you have to add one more line into the component definition to declare what instance variables you want to create a reader accessor for, it makes porting partials over to components a breeze. Essential I can copy and paste my partial template directly into my component template.

README.md
```diff
@@ -219,8 +219,8 @@ end
 `app/components/modal_component.html.erb`:
 ```erb
 <div class="modal">
-  <div class="header"><%= header %></div>
-  <div class="body"><%= body %>"></div>
+  <div class="header"><%= @header %></div>
+  <div class="body"><%= @body %>"></div>
```
rmacklin: `ModalComponent` is declaring `with_content_areas :header, :body` which should work without changing this template to use instance variables.

That's working for this component template in the test suite: https://github.com/github/actionview-component/blob/17069837bcadf2f6e122627f0fc2f90c73622cba/test/app/components/content_areas_component.html.erb

README.md
```diff
@@ -329,10 +329,10 @@ end
 `app/components/modal_component.html.erb`:
 ```erb
 <div class="modal">
-  <% if header %>
-    <div class="header"><%= header %></div>
+  <% if @header %>
+    <div class="header"><%= @header %></div>
   <% end %>
-  <div class="body"><%= body %>"></div>
+  <div class="body"><%= @body %>"></div>
```
rmacklin: Same here - I would expect this to work without changing to instance variables.

jonspalmer:
My sense is we should encourage attr_reader and I've very comfortable that the examples we have document that.

ultrawebmarketing:
@rmacklin while I agree that most of the example will pass CI without altering them, it's the consistency that I'm after. I don't want people getting tripped up like I did. Which ever way has the final say, I will update the doc.

jcoyne:
I think attr_readers do a great job of protecting against typos because you get a NoMethodError, whereas undefined ivars silently render nil.

ultrawebmarketing:
I updated my pull request to use attr_reader and also corrected some errors when running the examples. For some reason the checks are failing with the following, but I think it's unrelated:

```
Bundler could not find compatible versions for gem "bundler":
  In Gemfile:
    bundler (>= 1.14)

    rails (= 5.0.0) was resolved to 5.0.0, which depends on
      bundler (< 2.0, >= 1.3.0)

  Current Bundler version:
    bundler (2.1.4)
This Gemfile requires a different version of Bundler.
Perhaps you need to update Bundler by running `gem install bundler`?
##[error]Process completed with exit code 6.
```

PR #205, opened 2020-02-12
Add deprecation warning for public methods on components.

joelhawksley:
In practice, we've yet to find a legitimate reason to allow
public methods to be defined on components. Furthermore,
allowing them enables them to be tested in place of the HTML
output of the component, which we consider to be an anti-
pattern in our experience with the library so far. This
has been a common trap for developers new to components
to fall into, so I feel it makes sense to have the library
enforce this rule.

This deprecation warning will change to an exception
in release v2.0.0.

### DIFF
--- a/.github/workflows/ruby_on_rails.yml
+++ b/.github/workflows/ruby_on_rails.yml
@@ -7,8 +7,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rails_version: [5.0.0, 5.2.3, 6.0.0, master]
-        ruby_version: [2.4.x, 2.5.x, 2.6.x, 2.7.x]
+        rails_version: [6.0.0, master]
+        ruby_version: [2.6.x, 2.7.x]
         exclude:
           - rails_version: master
             ruby_version: 2.4.x
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -9,6 +9,8 @@ module ActionView
       include ActiveSupport::Configurable
       include ActionView::Component::Previewable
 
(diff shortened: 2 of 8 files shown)


.github/workflows/ruby_on_rails.yml
```diff
@@ -7,8 +7,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rails_version: [5.0.0, 5.2.3, 6.0.0, master]
-        ruby_version: [2.4.x, 2.5.x, 2.6.x, 2.7.x]
```
jonspalmer: Why are we not testing older versions anymore?

lib/action_view/component/base.rb
```diff
@@ -171,6 +173,14 @@ def compile(validate: false)
             return false
           end
 
+          if self == PublicMethodComponent
```
jonspalmer: I don't understand this check.

joelhawksley: This is what I get for not opening this is a draft 🤦‍♂ . This is still a work in progress! I'll close it for now.

.github/workflows/ruby_on_rails.yml
joelhawksley: I was playing around with some CI issues on this branch. I'll end up looking into them elsewhere.

joelhawksley:
@jonspalmer I'm closing this for now, but to respond to your comment:

The ultimate recommendation here is to have most methods on a component be private. As the template is compiled onto the component as an instance method, there is no need for any properties to be public.

As this is an entirely new pattern to most Rails applications, we felt the need to be strong-handed with defining programmatically enforced conventions. Do you think there could be a better way to do this?

PR #208, opened 2020-02-13
Add deprecation warning for public methods on components.

joelhawksley:
In practice, we've yet to find a legitimate reason to allow
public methods to be defined on components. Furthermore,
allowing them enables them to be tested in place of the HTML
output of the component, which we consider to be an anti-
pattern in our experience with the library so far. This
has been a common trap for developers new to components
to fall into, so I feel it makes sense to have the library
enforce this rule.

This deprecation warning will change to an exception
in release v2.0.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-# Pending
+* Deprecate allowing public methods on components.
+
+    *Joel Hawksley*
 
 * Remove initializer requirement for Ruby 2.7+
 
--- a/README.md
+++ b/README.md
@@ -434,7 +434,7 @@ class MyComponentTest < ActionView::Component::TestCase
 end
 ```
 
-In general, we’ve found it makes the most sense to test components based on their rendered HTML.
+In general, we’ve found it makes the most sense to test components based on their rendered HTML. A warning will be raised if any public methods are defined on a component.
 
 #### Action Pack Variants
 
--- a/lib/action_view/component/base.rb
+++ b/lib/action_view/component/base.rb
@@ -9,6 +9,8 @@ module ActionView
       include ActiveSupport::Configurable
       include ActionView::Component::Previewable
 
+      ALLOWED_PUBLIC_INSTANCE_METHODS = [:default_url_options?, :default_url_options, :default_url_options=].freeze
+
       delegate :form_authenticity_token, :protect_against_forgery?, to: :helpers
 
       class_attribute :content_areas, default: []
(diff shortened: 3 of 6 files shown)


jonspalmer:
I'm concerned about the direction of this PR. Please don't read what follows (and it will likely be long) as a rant. Instead very much intended as a discussion of the pros and cons of enforcing best practices vs leaving the flexibility for consumers to make their own decisions.

My concern is along three broad directions:

1. IMO anti pattern here is that private properties work in the views at all
1. Philosophically Ruby/Rails doesn't and shouldn't warn about anti patterns let alone prevent them with exceptions. 
1. There are some patterns where public methods are convenient


**1. The anti pattern is the private methods work in views at all**

 * IMO its super weird that it works. (In fact I didn't know it did).
 * That private methods are accessible in views unexpected and asymmetric compared to working with public/private methods in Ruby objects. That it works is a quirky side effect of the _implementation detail_ of compiling templates into methods on the component. 
 * If we were starting from scratch I'd argue for an API design that would more Ruby/Rails like:
   * We provide a magic `component` "local" to the view. Only public properties would work.
   * View Component have a helper to declare which properties are available to the view. Much like Rails Controllers `helper_method`
   * For syntactic sugar make public methods available to the view.
 * In practice for the Components we've written we explicitly make that distinction/convention of using public/private methods to delineate the component API. Public properties we expect to be used in the view vs private properties that represent implementation details. 
 

**2. Philosophically Ruby/Rails doesn't and shouldn't warn about anti patterns let alone prevent them with exceptions.**

High level - let consumers of this gem make their own decisions rather than restricting the way it can be used. Yes we should "test what we see" an not the implementation details. The documentation and examples should drive that point home. But sometimes its just not that easy. Sometimes testing the public property is "faster" or "simpler". For a complicated property used in the view we might write _one_ test of the html that property results in. Then exhaustive tests directly of the complexity of the property. We don't do it all the time but sometimes it's the 80/20 we choose. We've made a calculated decision, the framework shouldn't get in our way.

**3. There are some patterns where public methods are convenient**

Perhaps its a slightly weird pattern but we fell across this example recently and liked the simplicity. Rendering some Nav dom to link to a list of rendered components:

```ruby
class ChartComponent < ActionView::Component
  attr_reader :name 

  def initialize(name:)
    @name = name
  end

  def id
    SecureRandom.uuid
  end
end
```
`chart_component.html.haml`
```haml
.chart{id: id}
  %h1=name
  .some_chart
    ......
```

with an `Array<ChartComponent>`  `chart_components`:
`some_view.html.haml`
```haml
%nav
  - chart_components.each do |chart|
    %a{href: "##{chart.id}"}=chart.name

.charts
  - chart_components.each do |chart|
    render(chart)
```

You could argue that we should have some other structure for the id-name pairs that we render the nav with and instantiate the ChartComponents with. Maybe we should have a higher level component that composes the above but thats awkward too (there is lots more DOM between the nav and charts in our actual use case). Either way that extra object is just noise when we already have a nice 'view-model' in the form of a ViewComponent that encapsulates that logic. 


_**phew**_


Overall IMO there are compelling reasons to leave as is. If we do add this change my suggestion would be it should be configurable. I'd argue that the default should be for it to be off.

joelhawksley:
@jonspalmer thanks for the thorough response ❤️ 

I was mostly curious when I opened this PR 😄 . It's helpful to hear how public methods are useful for you. Aren't using components as view models in that way, but I can see how it makes sense in that case. 

I don't see any reason to proceed with this change. Closing!

PR #211, opened 2020-02-14
Deprecate non-instance syntaxes

joelhawksley:
Per https://github.com/github/actionview-component/issues/206, this PR deprecates all of the supported `render` syntaxes _except_ for `render(MyComponent.new(foo: :bar)`.

We're making this change to align the gem with the syntax Rails 6.1 will support 🙏


jonspalmer:
I like it!

joelhawksley:
@jonspalmer mind leaving a ✅ then? 😄

jonspalmer:
Will do. On mobile at the moment and can't find the way to do it. If only I knew someone at GitHub to report that too....

Issue #199, opened 2020-01-31
Improve testing experience


joelhawksley:
In talking with @dylnclrk, we thought it might make sense to explore ways to improve the testing experience with components.

Right now, we generally do something like the following:

```ruby
result = render_inline(WrapperComponent) do
  "content"
end

assert_html_matches result.css("span").to_html, "<span>content</span>"
```

It would be lovely if we could provide an API similar/identical to https://guides.rubyonrails.org/testing.html#testing-views:

```ruby
render_inline(WrapperComponent) do
  "content"
end

assert_select "span", "content"
```

joelhawksley:
@dylnclrk I've got a work in progress here. I'll share it soon 😄

joelhawksley:
Closed by https://github.com/github/actionview-component/pull/216

PR #216, opened 2020-02-19
Use Capybara assertions

joelhawksley:
Related issue: #199 

This PR introduces support for the existing Rails system test assertions (from Capybara) one might expect to work out of the box.


Gemfile
```diff
@@ -6,3 +6,5 @@ gemspec
 rails_version = "#{ENV['RAILS_VERSION'] || '6.0.0'}"
 
 gem "rails", rails_version == "master" ? { github: "rails/rails" } : rails_version
+
+gem "capybara", ">= 3.26"
```
rmacklin: I think this dependency should be declared in the gemspec, since one of the gem's library files explicitly `require`s capybara:
https://github.com/github/actionview-component/blob/d64a6a6118573e86b40d3854be7edcf987cf7fdf/lib/action_view/component/test_helpers.rb#L3

rmacklin:
> This PR introduces support for the existing Rails DOM assertions

This seems like a bit of a misnomer. Rails' `assert_select` is different than Capybara's `assert_selector`, and this PR is introducing support for the latter. (I'm not saying using Capybara's assertions is the wrong way to go, though - but it's different than what the PR description claims and what's in https://github.com/github/actionview-component/issues/199#issue-558410314.)

joelhawksley:
@rmacklin good point.

I ended up using Capybara as I found myself wanting `assert_text`. I do think it's worth noting that this is mimicking what Rails uses for system tests: https://github.com/rails/rails/blob/786ece692cb71703241a1059e982add5605b67b3/actionpack/lib/action_dispatch/system_test_case.rb#L3

Do you think it's OK to go with the Capybara assertions?

rmacklin:
I don't feel to strongly either way, and since most rails app's will probably already have a capybara dependency, it seems okay to make it a dependency of this gem 👍

PR #222, opened 2020-02-20
Add gem version to ISSUE_TEMPLATE

biow0lf:
### Summary

I just open issue about ruby 2.7 and realize that issue template miss gem version. So, adding.


joelhawksley:
Thanks!

PR #224, opened 2020-02-21
Add from_options_with_default to Base

joelhawksley:
`from_options_with_default` is a lightweight helper that
enables components to gracefully fall back to a default
value when invalid input is provided.

We're starting to move away from using validations in our
components internally, as we've found that what we really
need is a guarantee that a value will be one that we expect.
What we don't need is for our view code to raise exceptions
when an unexpected value is passed.

In our ~50 components, we've only used two validation methods:
presence and inclusion. This helper is meant to replace
inclusion validations.


lib/action_view/component/base.rb
```diff
@@ -116,6 +116,14 @@ def with(area, content = nil, &block)
         nil
       end
 
+      def from_options_with_default(value, default, allowed_values)
```
rmacklin: I'll be honest: I'm not in love with the name `from_options_with_default`. I'm also a bit surprised that we want to add this to the public API before v2.0. The fact that we're removing validations (https://github.com/github/actionview-component/pull/223#issue-378489241) gave me the impression that we want to strip out the unnecessary parts from the `ViewComponent::Base` class before v2.0 (which I support). Adding this new method to the core seems counter to that, though. And it'll mean we'd need to support this exact method signature until the next major version (assuming we want to follow semver). That is, if we come up with a better name/signature, we'd still have to keep the old one around in v2.* to avoid introducing a breaking change.

With that said, I'm curious if you've considered putting this in your `ApplicationComponent` base class for now (where it'd be easier to make changes to the interface because you can update all the consumers at once) and letting it marinate for a while before bringing it into the gem.

joelhawksley: @rmacklin we have it in our `ApplicationComponent` now, but were prompted to include this in the gem due to wanting to use it in components outside of our application that do exist in the context of an `ApplicationComponent`.

We intended this method to be a lightweight replacement for the most common use case for validations: inclusion checks, easing the transition away from validations.

I have no attachment to the name. Might you suggest some options?

rmacklin: It usually takes me a while of experimenting with usages to come up with a good name (which is why I didn't offer an alternative in the first comment).

I can think about it, but in the meantime, I'm curious if you have some other examples you can share. I could see the `FallbackToDefaultComponent` example being implemented without `from_options_with_default` as something like
```rb
class FallbackToDefaultComponent < ActionView::Component::Base
  COLOR_OPTIONS = { red: :red, green: :green }.tap { |_| _.default = :blue }

  def initialize(color:)
    @color = COLOR_OPTIONS[color]
  end
end
```

If that was repeated a lot, one could define a helper like
```rb
def one_of(*options, default:)
  options.yield_self { |_| _.zip _ }.to_h.tap { |_| _.default = default }
end
```
to abstract the pattern for constructing the `Hash`
```rb
COLOR_OPTIONS = one_of(:red, :green, default: :blue)
```

Anyway, like I said, I don't want to get in your way if this is something you feel is needed in the gem, but seeing a few other usages might be helpful.

joelhawksley: @rmacklin thanks for the example code. 

I'm going to close this for now and let it simmer a bit more internally.

PR #225, opened 2020-02-22
Restructure Railtie into Engine

seanpdoyle:
Summary
===

Integrate with [Zeitwerk loading conventions][zeitwerk]

Prior to this change, adding this gem as a dependency to a `rails@6.0.2`
with `actiontext` in the bundle, then running the test yields the
following deprecation warning:

```
DEPRECATION WARNING: Initialization autoloaded the constants
ActionText::ContentHelper and ActionText::TagHelper.

Being able to do this is deprecated. Autoloading during initialization
is going to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed
during initialization does not run again. So, if you reload
ActionText::ContentHelper, for example, the expected changes won't be
reflected in that stale Module object.

These autoloaded constants have been unloaded.

Please, check the "Autoloading and Reloading Constants" guide for
solutions.
```

To resolve that issue, this commit replaces the [`Railtie`][railtie]
with an [`Engine`][engine], moves the `Rails::ComponentsController`'s
declaration to `app/controllers/rails/components_controller.rb`, then
re-configures that controller's view template loading paths to account
for the move.

[zeitwerk]: https://github.com/fxn/zeitwerk#file-structure
[railtie]: https://api.rubyonrails.org/v6.0.2/classes/Rails/Railtie.html
[engine]: https://api.rubyonrails.org/classes/Rails/Engine.html


joelhawksley:
:👏 thank you! ❤️

PR #232, opened 2020-02-26
Set template filepath and line number for better backtraces

BlakeWilliams:
Currently `__FILE__` and `__LINE__` are the arguments passed to
`class_eval` which means when an exception is raised in the template it
will point back to the component instead of the correct line of the
template.

This updates the filepath and line number arguments passed to
`class_eval` to point to the template instead of the component.

resolves https://github.com/github/actionview-component/issues/85

Here's some screenshots displaying the difference:

Before:

Screen Shot 2020-02-26 at 5 31 11 PM

After:

Screen Shot 2020-02-26 at 5 30 52 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Improve backtraces from exceptions raised in templates
+
+    *Blake Williams*
+
 # v1.12.0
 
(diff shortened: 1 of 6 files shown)


BlakeWilliams:
👍 Can do!

PR #233, opened 2020-02-28
Allow to access params in preview examples

cover:
### Summary

This is helpful when you want to show dynamic values on the component preview.

One use case is integrating [Storybook](https://storybook.js.org/) and the [Knobs addon](https://github.com/storybookjs/storybook/tree/master/addons/knobs), which allows to tweak the values directly from the component's page.

We can take those values and pass them down as params to the Rails preview page, which will then be used to set the values on the component.

### Other Information

I've been briefly playing with the repo from #159 and adding this would allow us to have something like:

Screenshot_20200228_115618

### DIFF
--- a/test/test/components/previews/preview_component_preview.rb
+++ b/test/test/components/previews/preview_component_preview.rb
@@ -16,4 +16,8 @@ class PreviewComponentPreview < ViewComponent::Preview
   def with_tag_helper_in_content
     render(PreviewComponent.new(title: "title")) { content_tag(:span, "some content") }
   end
+
+  def with_params(cta: "Default CTA", title: "Default title")
+    render(PreviewComponent.new(cta: cta, title: title))
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Add the ability to access params from preview examples.
+
+    *Fabio Cantoni*
+
 # v2.4.0
 
 * Add `#render_to_string` support.
--- a/README.md
+++ b/README.md
@@ -521,6 +521,19 @@ Which generates <http://localhost:3000/rails/view_components/test_component/with
 <http://localhost:3000/rails/view_components/test_component/with_long_title>,
 and <http://localhost:3000/rails/view_components/test_component/with_content_block>.
 
+It's also possible to set dynamic values from the params by setting them as arguments:
+
+`test/components/previews/test_component_preview.rb`
+```ruby
+class TestComponentPreview < ViewComponent::Preview
+  def with_dynamic_title(title: "Test component default")
+    render(TestComponent.new(title: title))
+  end
+end
+```
+
+You'll then be able to pass down a value with <http://localhost:3000/rails/components/test_component/with_dynamic_title?title=Custom+title>.
+
 The `ViewComponent::Preview` base class includes
 [`ActionView::Helpers::TagHelper`][tag-helper], which provides the [`tag`][tag]
 and [`content_tag`][content_tag] view helper methods.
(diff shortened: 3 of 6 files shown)


jonspalmer:
@cover I like this and it works. I've been working on a different approach to support Storybook. Specifically I added a "server" framework for Storybook https://github.com/storybookjs/storybook/pull/9250

Essentially Storybook server takes CSF definitions written in json and a preview url.

What's missing (but nearly finished on my laptop) is a RailsGem that connects the dots. It will allow you to define Storybook stories for View Components and generate json from them. I expect it to be an alternative to Previews. 

The big advantage of this approach is a) you write the stories for view components in Ruby b) we can support hot reloading of Storybook Story any time the json changes. One of the missing pieces of the gem is to add 'watching' of the view components to recompile the json when they change.

As a sneak peak I have a DSL like this mostly working for for describing ViewComponent Stories:

```ruby
class KitchenSinkComponentStories < ActionView::Component::Storybook::Stories
  jane_doe KitchenSinkComponent do
    knobs do
      name "Jane Doe"
      birthday Date.new(1950, 3, 26)
      color(:favorite_color, "red")
      like_people true
      number_pets 2
      sports ["football", "baseball"]
      select(:favorite_food, { hot_dog: "Hot Dog", pizza: "Pizza", burgers: "Burgers", ice_cream: "Ice Cream" }, :ice_cream)
      other_things({ hair: "Brown", eyes: "Blue" })
    end
  end
end
```

lib/action_view/component/preview.rb
```diff
@@ -19,8 +27,8 @@ def all
         end
 
         # Returns the arguments for rendering of the component in its layout
-        def render_args(example)
-          new.public_send(example).merge(layout: @layout)
+        def render_args(example, params: {})
+          new(params: params).public_send(example).merge(layout: @layout)
```
jonspalmer: Instead of making params a property of the Preview my not make them arguments of the example?

```ruby
new.public_send(example, **params).merge(layout: @laout)
```

test/test/components/previews/preview_component_preview.rb
```diff
@@ -2,7 +2,7 @@
 
 class PreviewComponentPreview < ActionView::Component::Preview
   def default
```
jonspalmer: With my suggestion above couldn't we make this

```ruby
def default(cta: "Click me!", title: "Lorem Ipsum")
```

that way we don't have this magic params property to know about.

lib/action_view/component/preview.rb
cover: Hey Jon, thank you for your feedback. I really look forward to see your work related to Storybook, that seems great! :grinning: 

I've been thinking a bit about your proposed change.

On one hand I like it because it simplifies the access of params from the example: you just define them as method arguments and it'd set them from the params if they're given.

The implementation would be a little bit different though, as we receive `ActionController::Parameters` from the controller, which would include also other data (`controller`, `action`, etc), so we need to only take the defined keywords in order to avoid passing unsupported ones:

```ruby
# components_controller.rb
@render_args = @preview.render_args(@example_name, params: params.permit!)

# preview.rb
def render_args(example, params: {})
  method_params = instance_method(example).parameters.map(&:last)
  provided_params = params.slice(*method_params).to_h.symbolize_keys
  new.public_send(example, **provided_params).merge(layout: @layout)
end

# component_preview.rb
def default(cta: "Click me!", title: "Lorem Ipsum")
```

On the other hand there are a couple of things that I'm not 100% sure about this approach:

1. We have some extra logic to make it working, which could break. What happens if we define on the method something else than keywords? Eg. `def default(some_config = nil, title: "Lorem Ipsum")`, we'd then may have to take care of that case.

2. I'm trying to think about someone who is adding the example and what they may think if they want to add a value from url parameters. Would they expect to have them passed when they set a keyword argument, or would they try to access the params directly? I mean, if you're on Rails it's very likely that you know that in a controller you can access them by calling `params`, so I'm wondering if there would be a similar expectation or not :thinking: 

What do you think?

jonspalmer: @cover I certainly share you concerns. We likely will also find problems calling the example with zero arity vs positive arity (we already trie to solve this in constructors for Ruby 2.7 that I think we rolled back)

I could go either way, knowing to check params is probably fine as as you point out similar to Rails controller. In all honesty its challenges like this that in part inspired me to go with the DSL I previewed for declaring Stories in Ruby. IMO the explicitness of what the "params" are feels more natural to me.

cover: Cool, I'll update this to pass down the params through the method/example arguments (pretty much the code above) once the current library renaming is merged :+1:

andrewmcodes:
@cover I’d love to know how you got storybook working. Any chance you could share an example repo or explain? 😄

cover:
Hey @andrewmcodes, here's the [repo](https://github.com/cover/avc_storybook) I used. It's mostly based on Jon's work, I've only done a small change to render the component in an iframe. I hope that helps :wink:

andrewmcodes:
> Hey @andrewmcodes, here's the [repo](https://github.com/cover/avc_storybook) I used. It's mostly based on Jon's work, I've only done a small change to render the component in an iframe. I hope that helps 😉

You are awesome - thanks!

jonspalmer:
Just published the first version of a Gem that ties ViewComponent to Storybook using Storybook Server: https://github.com/jonspalmer/view_component_storybook

joelhawksley:
@cover @jonspalmer what's the status of this work? Is it still something we want to pursue?

cover:
@joelhawksley it was working/ready. Now I see that there are some conflicts. Let me know if you're interested in merging it, in that case I can resolve them :wink:

joelhawksley:
@jonspalmer would be interested in finishing the review cycle on this one?

joelhawksley:
@cover @jonspalmer thank you for your work on this 🎊 

I'd love to see more docs / examples added to the README once you have them 🙏

Issue #38, opened 2019-08-31
Allow rendering components in the controller


ushis:
Hi @joelhawksley 

I am happy to see a view model lib, that integrates smoothly with rails. While playing around with `ActionView::Component` i noticed that it is not possible to render components in the controller. I saw, that you are monkey patching `ActionView::Base#render`. So i tried to override the `render` method in my `ApplicationController` in a similar way, but without success.

At the moment users are forced to write templates with a single line: `<%= render(PostShowComponent, post: @post) %>`. It would be great to have an API like this:

```ruby 
class PostsController < ApplicationController
  def show 
    post = Post.find(params[:id])
    render(PostShowComponent, post: post)
  end
end
```
I could provide a PR with a failing feature test, if that helps. 

Thank you , ushi

P.S. Found a workaround. Maybe its a good starting point:

```ruby
class ApplicationController
  private

  def render_component(component, args, **options)
    options = options.reverse_merge(layout: true)
    render(html: view_context.render(component, args), options)
  end 
end
```

joelhawksley:
@ushis I agree. This is definitely something that should work. 

I have no objections to the proposed API.

One thing I'm curious about: Do you think we should support passing content to components via a block argument when rendering them from a controller?

ushis:
@joelhawksley Thanks for the feedback. 

> One thing I'm curious about: Do you think we should support passing content to components via a block argument when rendering them from a controller?

Depends on the complexity. It would enable the implementation of layouts using components:

```ruby
render(AdminLayoutComponent, current_user: user) do 
  render(PostShowComponent, post: post)
end
```

This approach leads to other questions, because `ActionView` depends on a controller in some cases. E.g. to render `csrf_meta_tags` in the layout, we need a controller, that provides `request_forgery_protection_token`. But i think this is addressed in #28 

What i want to say: It would be nice, but not necessary in the first iteration, if too complex.

How do you want to proceed? Opening a PR that monkeypatches ActionController::Base?

joelhawksley:
@ushis a monkey patch is probably a good place to start! I'd definitely welcome a PR.

joelhawksley:
@kaspermeyer I know you had some interest in this one a while ago. Might you be able to have a crack at it?

kaspermeyer:
Absolutely! Thanks for pinging me on this, @joelhawksley ❤️ I can dedicate some time at the end of this week to get started on this along with tests for pre-compiled templates (#112).

[...]

PR #235, opened 2020-02-28
add rudimentary support for rendering components from controllers

joelhawksley:
This change introduces the ability to render components from
controllers, mimicking the behavior introduced in
https://github.com/rails/rails/pull/37919.

[Fixes: #38]

### DIFF
--- a/test/action_view/integration_test.rb
+++ b/test/action_view/integration_test.rb
@@ -10,6 +10,23 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_select("div", "Foo\n  bar")
   end
 
+  test "rendering component in a controller" do
+    get "/controller_inline_baseline"
+
+    assert_select("div", "bar")
+    assert_response :success
+
+    baseline_response = response.body
+
+    get "/controller_inline"
+    assert_select("div", "bar")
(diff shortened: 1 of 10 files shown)


test/action_view/integration_test.rb
```diff
@@ -10,6 +10,23 @@ class IntegrationTest < ActionDispatch::IntegrationTest
...
+    baseline_response = response.body
+
+    get "/controller_inline"
+    assert_select("div", "bar")
+    assert_response :success
+
+    inline_response = response.body
+
+    assert_equal baseline_response, inline_response
+  end
```
jonspalmer: Wouldn't this be a little better as two tests. I'd like to see a test saying "we didn't break regular rendering with our monkey patch" I think that's what you mean by baseline.

joelhawksley:
@jonspalmer right now, this just renders the component without a layout, matching the behavior of rendering partials from controllers. 

Is that the behavior you'd expect? Or would you expect layouts to work?

jonspalmer:
I think without layouts is what I'd expect. Otherwise you'd get double rendered layouts or the like. I'd need to go review the examples in more detail.

PR #239, opened 2020-03-02
rename ActionView::Component to ViewComponent

joelhawksley:
This commit renames the bulk of the library to the ViewComponent
namespace, preserving backward compatibility with
ActionView::Component::Base and the associated base classes for now.
This change will allow users of the library to migrate to the next
major version ahead of the major v2 release.

ViewComponent::Base and its associated classes implement the
following changes:

1) Validations are not enabled by default.
2) Only the render(MyComponent.new) syntax is supported.


rainerborene:
All good. But I'm wondering if this is the best approach to proceed with this kind of migration. Wouldn't be simpler to create a new repository and start from scratch? New repository link, new `gemspec` and new life. How long we will keep these aliases classes/modules on the source code? Both `actionview-component` and  `view-component` gems will refer to the same codebase?

test/app/components/path_container_component.html.erb
```diff
@@ -1 +1 @@
-<%= render PathComponent %>
\ No newline at end of file
+<%= render PathComponent.new %>
```
juanmanuelramallo: Nitpick: I know it's not related to your change, but maybe we would want to add an empty newline here at the end? for the sake of consistency and also for having unix friendly files 🙈 what do you think?

README.md
```diff
@@ -125,11 +139,11 @@ bin/rails generate component Example title content --template-engine slim
...
 #### Implementation
 
-An `ActionView::Component` is a Ruby file and corresponding template file (in any format supported by Rails) with the same base name:
+A `ViewComponent` is a Ruby file and corresponding template file (in any format supported by Rails) with the same base name:
 
 `app/components/test_component.rb`:
 ```ruby
-class TestComponent < ActionView::Component::Base
+class TestComponent < ViewComponent::Base
   validates :content, :title, presence: true
```
juanmanuelramallo: Is this possible to do with this new change? (taking into account that this example is using `ViewComponent::Base` instead of `ActionView::Component::Base`)

README.md
```diff
@@ -1,12 +1,26 @@
-_Note: This gem is in the process of a name / API change, see https://github.com/github/actionview-component/issues/206_
-
-_You are viewing the README for the development version of ActionView::Component. If you are using the current release version you can find the README at https://github.com/github/actionview-component/blob/v1.11.1/README.md_
-
 # ActionView::Component
```
msalzburg: Guess the headline needs to be renamed to ViewComponent too?

lib/action_view/component.rb
```diff
@@ -1,24 +1,4 @@
-# frozen_string_literal: true
+ # frozen_string_literal: true
```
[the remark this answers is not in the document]
jcoyne: It's curious that rubocop didn't catch this.

lib/view_component/base.rb
```diff
@@ -0,0 +1,273 @@
...
+      end
+
+      # Compile templates to instance methods, assuming they haven't been compiled already.
+      # We could in theory do this on app boot, at least in production environments.
+      # Right now this just compiles the first time the component is rendered.
+      def compile(validate: false)
+        return if compiled?
+
+        if template_errors.present?
+          raise ViewComponent::TemplateError.new(template_errors) if validate
```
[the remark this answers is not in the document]
joelhawksley: 👍 good point. I've renamed it to `raise_template_errors`.

PR #240, opened 2020-03-02
Fix rendering of components that utilize `capture` in a parent view context

BlakeWilliams:
In Rails, when you create a form it saves a reference to the current
template object that is rendering the form. When you render a form
builder helper method, like `form.label` inside of a child component the
template object will be the original template object from where the form
builder was instantiated.

This means in cases where we pass a block to a form builder helper the
`@output_buffer` is the parent component's output buffer. So when that
buffer is overridden via `.capture` it's setting the parent component's
`@output_buffer` to a new, temporary buffer but it doesn't change the
child component that is actually rendering's `@output_buffer`. This
means the block is executed in the context of the child components
actual `@output_buffer` and immediately writes to the buffer. Since
`capture` had nothing written to the temporary buffer, it returns the
value of the block call and then inserts that into the parent
component's `@output_buffer`.

This adds a new module, ActionViewCompatibility which can be included in
a component to ensure that any `capture` calls coming from that
component will be called in the correct context. This is helpful when
using helper methods like `form_for`.

### DIFF
--- a/lib/view_component/action_view_compatibility.rb
+++ b/lib/view_component/action_view_compatibility.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module ActionViewCompatibility
+    extend ActiveSupport::Concern
+
+    def form_for(*args, **kwags, &block)
+      with_compatible_capture do
+        super
+      end
+    end
+
+    def form_with(*args, **kwargs, &block)
+      with_compatible_capture do
+        super
+      end
+    end
+
+    included do
+      alias_method :original_capture, :capture
+      alias_method :capture, :capture_with_compatibility
+    end
+
+    def with_compatible_capture(&block)
+      old_compatible_capture = defined?(@compatible_capture) ? @compatible_capture : false
+      @compatible_capture = true
+      yield
+    ensure
+      @compatible_capture = old_compatible_capture
+    end
+
+    def capture_with_compatibility(*args, &block)
+      if defined?(@compatible_capture) && @compatible_capture
+        receiver_aware_capture(*args, &block)
+      else
+        original_capture(*args, &block)
+      end
+    end
+
+    def receiver_aware_capture(*args, &block)
+      receiver = block.binding.receiver
+
+      if receiver != self && receiver.respond_to?(:output_buffer=)
+        if receiver.respond_to?(:original_capture)
+          receiver.original_capture(*args, &block)
+        else
+          receiver.capture(*args, &block)
+        end
+      else
+        original_capture(*args, &block)
+      end
+    end
+  end
+end
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -7,6 +7,7 @@ require "view_component/compile_cache"
 require "view_component/previewable"
 require "view_component/slotable"
 require "view_component/slotable_v2"
+require "view_component/action_view_compatibility"
 
 module ViewComponent
   class Base < ActionView::Base
--- a/test/app/components/form_for_component.html.erb
+++ b/test/app/components/form_for_component.html.erb
@@ -1,0 +1,3 @@
+<%= form_for Post.new, url: "/" do |form| %>
+  <%= render LabelComponent.new(form: form) %>
+<% end %>
--- a/test/app/components/form_for_component.rb
+++ b/test/app/components/form_for_component.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class FormForComponent < ViewComponent::Base
+  include ViewComponent::ActionViewCompatibility
+
+  def initialize
+  end
+end
--- a/test/app/components/form_with_component.html.erb
+++ b/test/app/components/form_with_component.html.erb
@@ -1,0 +1,3 @@
+<%= form_with Post.new, url: "/" do |form| %>
+  <%= render LabelComponent.new(form: form) %>
+<% end %>
--- a/test/app/components/form_with_component.rb
+++ b/test/app/components/form_with_component.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+class FormWithComponent < ViewComponent::Base
+  include ViewComponent::ActionViewCompatibility
+
+  def initialize
+  end
+end
--- a/test/app/components/incompatible_form_component.html.erb
+++ b/test/app/components/incompatible_form_component.html.erb
@@ -1,0 +1,3 @@
+<%= form_for Post.new, url: "/" do |form| %>
+  <%= render LabelComponent.new(form: form) %>
+<% end %>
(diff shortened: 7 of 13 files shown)


lib/action_view/component/render_monkey_patch.rb
```diff
@@ -4,6 +4,16 @@
 module ActionView
   module Component
     module RenderMonkeyPatch # :nodoc:
+      attr_accessor :child_component
```
BlakeWilliams: Maybe a better name would be `child_view_context`?

thomasbrus:
I also ran into this, any chance of getting this reviewed/merged?

joelhawksley:
@thomasbrus I'm not comfortable with this solution, as it depends on a Rails monkey patch. We've already disabled the existing monkey patches when the library is used with Rails 6.1, and don't plan to re-introduce them.

That being said, we do need to address this bug. It just might require making changes to Rails itself.

thelucid:
Any workarounds for this?

joelhawksley:
@thelucid not that I'm aware of. If you find a solution that works for you, let us know!

thelucid:
@joelhawksley I ended up passing the template instance into the constructor and running any renders and helper calls on that instance. Seems to be working quite well.

To clarify, I have a form component that takes a the template as an initialiser param. I store that in an instance variable and renter anything else, including form fields via that instance.

thelucid:
On further investigation, it seems that if [this method](https://github.com/github/view_component/blob/master/lib/view_component/base.rb#L102) was changed to always use the view context, this wouldn't be a problem. To give some context, I'm currently passing a template around to avoid this issue:

```ruby
module ComponentHelper
  # A simple wrapper around `form_with`, for FormComponent.
  # 
  def component_form_with(**options, &block)
    form_with(local: true, **options) do |form|
      FormComponent.new(form, self).then do |component|
        yield component
        render(component, &block)
      end
    end
  end
end

class FormComponent < ViewComponent::Base
  def initialize(builder, template)
    @builder = builder
    @template = template
  end
  
  def fields(**options, &block)
    @template.render(Form::FieldsComponent.new(@builder, **options), &block)
  end
end
```

I'm happy to create a pull request, but am unsure of any knock on effects of always rendering via the view context.

thelucid:
It seems I can get around this by manually rendering via the `view_context`, and therefore remove the need to pass around the template. It still feels like the view context should always be used:

```ruby
module ComponentHelper
  def component_form_with(**options, &block)
    form_with(local: true, **options) do |form|
      render(FormComponent.new(form), &block)
    end
  end

class FormComponent < ViewComponent::Base
  def initialize(builder)
    @builder = builder
  end
  
  def fields(**options, &block)
    view_context.render(Form::FieldsComponent.new(@builder, **options), &block)
  end
end
```

joelhawksley:
@BlakeWilliams what do you think about always rendering in `view_context`?

BlakeWilliams:
> @BlakeWilliams what do you think about always rendering in `view_context`?

I'd need to jump back in and regain context, but if we always use `view_context` are we bypassing `render_in` and always writing directly to the parent renderers `@output_buffer`? I think there may be some trade-offs, but that may work.

I can update this branch real quick to try that and see if the tests still pass.

BlakeWilliams:
I just pushed up a rebased version with the `view_context.render` change, sadly it looks like the issue still exists.

I think it may be due to the builder created by `form_for` and friends captures the current rendering context at call time and `view_context` may be different by the time the `label` with a block arg is called? It's also possible it's the existing `render_in` issue. I'd have to dive in more later, but if anything looks off please let me know.

BlakeWilliams:
Thinking about it, one possible fix might be to just not use `render_in` in production/dev until the `@output_buffer` issue can be solved. I'm not a huge fan of that, but I think that may unblock things until we can come up with an upstream solution for Rails.

23tux:
We have the same problem, are there any updates on this?

BlakeWilliams:
@23tux This is something we'll be revisiting soon, so I'm hoping we'll have an update on the status of this problem in the near future.

BlakeWilliams:
Sorry for the delay on this, it's a bit of a tricky problem!

I just pushed up a new module that should be includable in components that use a slower, but more correct `capture` behavior when inside `form_for` and `form_with`.

I haven't had a chance to test this out in a real application yet, but if anyone wants to give it a try and report their experience, that would be awesome.

test/test_helper.rb
```diff
@@ -4,9 +4,9 @@
 
 if ENV["MEASURE_COVERAGE"]
   SimpleCov.start do
-    command_name "rails#{ENV["RAILS_VERSION"]}-ruby#{ENV["RUBY_VERSION"]}" if ENV["RUBY_VERSION"]
-
-    formatter SimpleCov::Formatter::Console
+    # command_name "rails#{ENV["RAILS_VERSION"]}-ruby#{ENV["RUBY_VERSION"]}" if ENV["RUBY_VERSION"]
+    #
+    # formatter SimpleCov::Formatter::Console
```
manuelpuyol, line 9: is this supposed to be here?

lib/view_component/action_view_compatibility.rb
```diff
@@ -0,0 +1,50 @@
...
+      alias_method :original_capture, :capture
+      alias_method :capture, :capture_with_compatibility
+    end
+
+    def with_compatible_capture(&block)
+      old_compatbile_capture = defined?(@compatbile_capture) ? @compatbile_capture : false
+      @compatbile_capture = true
+      yield
+    ensure
+      @compatbile_capture = old_compatbile_capture
```
manuelpuyol, line 29:
```suggestion
      old_compatible_capture = defined?(@compatible_capture) ? @compatible_capture : false
      @compatible_capture = true
      yield
    ensure
      @compatible_capture = old_compatible_capture
```

BlakeWilliams: Thanks! I typo'd it once and my editor decided that was the way to autocomplete it from then on. 😅

test/test_helper.rb
BlakeWilliams: Nope! I can remove it.

zernie:
Unfortunately, not even including `ActionViewCompatibility` helps. This is a pretty big issue that stops me from rewriting my forms with `view_component`

23tux:
I'm with @zernie, this bug prevents us from refactoring a lot of views into components. Are there any updates? Or is there any way we can help?

BlakeWilliams:
@23tux Fortunately there is, I took a different approach in https://github.com/github/view_component/pull/974 which looks promising.

Benchmark numbers still aren't great, but might be getting close to acceptable. I still need to test that branch out in a real project and see if it changes how partials/components are rendered.

joelhawksley:
Closing per @BlakeWilliams

percyhanna:
I was able to solve a similar issue using some Ruby method re-binding, which I outlined [here](https://github.com/github/view_component/pull/974#issuecomment-932818184). It may help some others with similar situations.

sdhull:
FWIW this has been fixed in some recent version (for me 3.3.0 did not work but 3.8.0 did) but you need to turn on an experimental mode by adding the following line to config/application.rb:
```
config.view_component.capture_compatibility_patch_enabled = true
```

PR #246, opened 2020-03-04
Bootstrap component case study

dpaola2:


joelhawksley:
👏 Looks good! Would you mind adding some basic info to the top? Your name/handle, basic info about the app you're working in (industry, etc).

PR #247, opened 2020-03-05
use new repo name

joelhawksley:
This PR updates some comments to reference our new repository name, `github/view-component` 🎊


rmacklin:
Per https://guides.rubygems.org/name-your-gem, I think the conventional name for this gem would be `view_component`, rather than `view-component`:

> USE UNDERSCORES FOR MULTIPLE WORDS
> If a class or module has multiple words, use underscores to separate them. This matches the file the user will require, making it easier for the user to start using your gem.
> 
> USE DASHES FOR EXTENSIONS
> If you’re adding functionality to another gem, use a dash. This usually corresponds to a / in the require statement (and therefore your gem’s directory structure) and a :: in the name of your main class or module.

The dash was appropriate when it was an extension of `ActionView`, but now that it's standalone, I think an underscore is appropriate by these conventions.

The repository name of course doesn't have to match the gem name, but typically that's desirable to avoid confusion.

joelhawksley:
@rmacklin thanks for the citation! I've changed the name to use an underscore.

Issue #236, opened 2020-02-29
Use ActiveModel::AttributeAssignment or similar


nicolas-brousse:
### Feature request
Does `ActionView::Component::Base` could include `ActiveModel::AttributeAssignment` and use `#assign_attributes(attributes)` in `#initialize`?

Like following:

```rb
class ApplicationComponent < ActionView::Component::Base
  include ActiveModel::AttributeAssignment

  def initialize(attributes = {})
    assign_attributes(attributes) if attributes

    super()
  end
end
```

Or maybe just manage a similar assignment.

### Motivation

Let use the following component as example.

```rb
class ButtonComponent < ActionView::Component::Base
  validates :text, :href, presence: true

  attr_reader :text, :href

  def initialize(text:, href:)
    @text = text
    @href = href
  end
end
```

I notice two things:
1. We need to write twice the attribute name (in `#initialize` and add it to `.attr_reader`)
2. No matter we write `initialize(text:, href:)` or `initialize(text: nil, href: nil)` since we have validation

joelhawksley:
@nicolas-brousse thanks for filing this!

This is definitely something that would sit nicely in an ApplicationComponent though ❤️ 🙏.

We'll be removing validations shortly in `v2.0.0`, so using `attr_reader` will no longer be as common of a pattern in components.

If you have a specific case where you think this would be useful, might you be able to open a PR with a test case?

nicolas-brousse:
@joelhawksley thanks for your answer.

Ah good to know! No I don't think keeping `ActiveModel::Validations` in `ActionView::Component::Base` is necessary. Since we could create an `ApplicationComponent` in our projet, it sounds good enough.

So if I'm correct, we could imagine the following in the future:
```ruby
class ApplicationComponent < ActionView::Component::Base
  include ActiveModel::AttributeAssignment
  include ActiveModel::Validation

  def initialize(attributes = {})
    assign_attributes(attributes) if attributes

    super()
  end
end
```

Or even
```ruby
class ApplicationComponent < ActionView::Component::Base
  include ActiveModel::Model
end
```

joelhawksley:
@nicolas-brousse that's correct. I'd love to see some example components using this behavior shared here: https://github.com/github/actionview-component/issues/227.

In the meantime, I'm going to close this issue ❤️

nicolas-brousse:
@joelhawksley I'm working on demo project to do some test before we implement it on real projects (https://github.com/nicolas-brousse/avc_form_builder).

[...]

PR #249, opened 2020-03-05
[WIP] Use ActionView::ViewPaths to manage views with ViewComponent

nicolas-brousse:
### Summary

This is a work in progress. I only try locally on an existing project for now.

This change attempt to make `#initialize` method optional in components (as discuss in https://github.com/github/view_component/issues/236#issuecomment-594642108) by switching rendering system on `ActionView::ViewPaths`.

### Notes
* Should we have a custom class for `ActionView::MissingTemplate` exception?
* `ViewComponent::ViewPaths` has implemented the scoped view_paths by class on Rails 6.0.0.

### DIFF
--- a/lib/view_component.rb
+++ b/lib/view_component.rb
@@ -12,8 +12,9 @@ module ViewComponent
   autoload :TestHelpers
   autoload :TestCase
   autoload :RenderMonkeyPatch
+  autoload :Rendering
   autoload :RenderingMonkeyPatch
-  autoload :TemplateError
+  autoload :ViewPaths
 end
 
 module ActionView
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -151,7 +151,11 @@ module ViewComponent
             # Require `#initialize` to be defined so that we can use `method#source_location`
             # to look up the filename of the component.
             initialize_method = instance_method(:initialize)
-            initialize_method.source_location[0] if initialize_method.owner == self
+            if initialize_method.owner == self
+              initialize_method.source_location[0]
+            else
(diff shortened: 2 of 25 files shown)


joelhawksley:
@nicolas-brousse is this still a change you'd like to try and land? If so, this PR will need some updates to catch up with `master`. Thank you for taking the time to work on it either way!

nicolas-brousse:
@joelhawksley is this still something interesting for `view_component`?

If yes, then I think I could take some time to resolve conflicts. Or even restart in a new fresh PR. The thing is that I may need some help for some parts, since I'm not an expert of ActionView.

I also have to go deeper about rails incompatibilities, for versions before 6.0. Since Rails 6.0 implemented [some new interesting code](https://github.com/rails/rails/blob/v6.0.0/actionview/lib/action_view/view_paths.rb#L40-L53), that does not exists in [Rails 5.1](https://github.com/rails/rails/blob/v5.2.0/actionview/lib/action_view/view_paths.rb).

Let me know 🙂

joelhawksley:
@nicolas-brousse given that we have removed the initializer requirement another way: https://github.com/github/view_component/pull/277, do you think these changes would benefit the library? If so, how?

nicolas-brousse:
@joelhawksley Oh! I totally missed the #277 changes.

Since the main idea was:
> This change attempt to make `#initialize` method optional in components

And not sure having multiple view paths is a real needs actually. So lets close this PR for now.
If its something that sounds useful in the future, then I may help.

I guess keeping `view_component` light and simple sounds better 🙂

PR #251, opened 2020-03-06
bypass monkey patch on Rails 6.1 builds.

joelhawksley:
This change disables the monkey patch when the Rails version
is at least 6.1.

This change confirms the compatibility of the library with Rails 6.1.

### DIFF
--- a/lib/view_component/render_monkey_patch.rb
+++ b/lib/view_component/render_monkey_patch.rb
@@ -3,7 +3,7 @@
 module ViewComponent
   module RenderMonkeyPatch # :nodoc:
     def render(options = {}, args = {}, &block)
-      if options.respond_to?(:render_in)
+      if options.respond_to?(:render_in) && Rails.version.to_f < 6.1
         options.render_in(self, &block)
       elsif options.is_a?(Class) && options < ActionView::Component::Base
         ActiveSupport::Deprecation.warn(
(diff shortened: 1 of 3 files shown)


jonspalmer:
Do we have existing tests that exercise this?

joelhawksley:
@jonspalmer we do! As we use matrix builds for CI, every test runs against `5.0.0`, `5.2.3`, `6.0.0`, and `master` 😄

lib/view_component/render_monkey_patch.rb
```diff
@@ -3,7 +3,7 @@
 module ViewComponent
   module RenderMonkeyPatch # :nodoc:
     def render(options = {}, args = {}, &block)
-      if options.respond_to?(:render_in)
+      if options.respond_to?(:render_in) && Rails.version.to_f < 6.1
```
jonspalmer: Should we do just a tiny bit of cleanup here? Perhaps a static method like `ViewComponent::apply_render_monkey_patch(options)` that we use is these two locations?

joelhawksley: I agree that we should just wrap the monkey patch entirely once we ship `v2`. As I plan to do that in the next week or so, I'm not going to sweat this temporary duplication 😄

PR #253, opened 2020-03-07
Make `TagHelper` available in Previews

seanpdoyle:
Make `ActionView::Helpers::TagHelper` available in
`ViewComponent::Previews`:

```ruby
def with_html_content
  render(MyComponent.new) do
    tag.div do
      content_tag(:span, "Hello")
    end
  end
end
````

This commit includes [`ActionView::Helpers::TagHelper`][tag-helper] into
`ViewComponent::Preview` base class, which provides the [`tag`][tag] and
[`content_tag`][content_tag] view helper methods.

[tag-helper]: https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html
[tag]: https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-tag
[content_tag]: https://api.rubyonrails.org/classes/ActionView/Helpers/TagHelper.html#method-i-content_tag

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # master
 
+* Make `ActionView::Helpers::TagHelper` available in Previews
+
+    def with_html_content
+      render(MyComponent.new) do
+        tag.div do
+          content_tag(:span, "Hello")
+        end
+      end
+    end
+
+    *Sean Doyle*
+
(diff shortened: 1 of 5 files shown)


joelhawksley:
Seems reasonable enough. Thanks for the PR ❤️

PR #259, opened 2020-03-12
Allow components to be rendered without a template file

rainerborene:
### Summary

This is useful feature when it comes to create little widgets, toggle buttons, badges, and so on. It removes the need of an extra template file with one single line of code.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -159,6 +159,10 @@ module ViewComponent
         @compiled && ActionView::Base.cache_template_loading
       end
 
+      def inlined?
+        instance_methods(false).grep(/^call/).present? && templates.empty?
+      end
+
       def compile!
         compile(raise_template_errors: true)
       end
@@ -167,7 +171,7 @@ module ViewComponent
       # We could in theory do this on app boot, at least in production environments.
       # Right now this just compiles the first time the component is rendered.
       def compile(raise_template_errors: false)
-        return if compiled?
+        return if compiled? || inlined?
 
         if template_errors.present?
           raise ViewComponent::TemplateError.new(template_errors) if raise_template_errors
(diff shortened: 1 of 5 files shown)


test/app/components/inline_component.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class InlineComponent < ViewComponent::Base
+  inline!
```
joelhawksley: What if we just checked for the presence of a `call` method instead of this explicit flag?

rainerborene: That could work but `compile!` method triggers an exception by default when a template file does not exist.

joelhawksley: What if `compile` checked for the presence of a `call` method?

rainerborene: @joelhawksley done, check it out! It also supports the `call_` variants.

joelhawksley:
@rainerborene mind adding a CHANGELOG entry?

rainerborene:
@joelhawksley :+1:

lib/view_component/base.rb
```diff
@@ -159,8 +159,12 @@ def compiled?
         @compiled && ActionView::Base.cache_template_loading
       end
 
+      def inlined?
+        instance_methods(false).grep(/^call/).present? && templates.empty?
+      end
+
       def compile!
-        compile(raise_template_errors: true)
+        compile(raise_template_errors: true) unless inlined?
```
joelhawksley: Would it be possible to move `unless inlined?` to after `return if compiled?` below? That way we won't run this check at runtime if we've compiled the component.

rainerborene: Let me know if that works.

joelhawksley:
👏 thanks for this contribution!

Issue #206, opened 2020-02-12
Renaming to ViewComponent, Syntax Changes


joelhawksley:
👋 Hi there folks! 

After some conversations with the Rails team, we’ve decided to make the following changes to ActionView::Component:

* Rename the gem to ViewComponent
* Revert to the original `render(MyComponent.new)` syntax

We hope that these changes will better communicate the relationship between this project and Rails. To that end, this library will not be upstreamed into Rails 6.1, but will instead have integrated support via the changes introduced in https://github.com/rails/rails/pull/36388.  We’ll be eager to provide this library as a first-class option for Rails developers.

GitHub continues to see value in the project, and I’m happy to share that I will now be working it full-time 🎊. We’re excited for the future of this library and can’t wait to see where it takes us.

## Renaming to ViewComponent
We’ll be renaming the library to ViewComponent.

As it stands, the name ActionView::Component implies that this library is _already_ a part of Rails, when it is not. While we have strived to integrate with Rails, both technically and conceptually, as a non-Rails codebase this name is ultimately improper.

## Syntax changes
Starting with the next release, we’ll be reverting to supporting the original `render(MyComponent.new)` syntax, per discussion on this PR (https://github.com/rails/rails/pull/38362) and in private with the Rails team. We’ll mark the other syntax options as deprecated, and remove them in `v2.0.0`. 

Once these changes are complete, we’ll be able to disable the monkey patches we have in place when using Rails 6.1+. We have no plans to deprecate support for previous versions of Rails.

## Thanks
I want to reiterate how thankful we are for the support this project has received from the community. We’ve been blown away by the contributions from all of you, both in quantity and quality ❤️

jonspalmer:
Great progress!

rainerborene:
Today I had a great chat with @joelhawksley about the future of `actionview-component`, well, now renamed to `view-component`. Before we get into that, let me share why I decided to use this gem in the first place. 

About a year ago I was using a combination of Rails built-in helpers and [`active_decorators`](https://github.com/amatsuda/active_decorator) at the startup I'm working with, and it quickly became a huge mess over time. I couldn't recognize which method belongs to which view anymore (not to mention the lack of tests). 

After migrating to `actionview-component`, something (bad) happened: all the design flaws of my application were exposed. Now I had the necessary information to refactor and write good code this time by isolating dependencies, and make view classes follow the single responsibility principle. I could test all those expensive pieces of crap code without sacrificing clarity, speed and most important: developer happiness.

[...]

PR #266, opened 2020-03-20
add post-install message about name change

joelhawksley:
This PR sets in motion our migration to `view_component` 🎊 . I've already published `v1.16.0` under the new name. 

With this change, the next release will be the last to be made under `actionview-component`.

### DIFF
--- a/actionview-component.gemspec
+++ b/actionview-component.gemspec
@@ -43,4 +43,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "better_html", "~> 1"
   spec.add_development_dependency "rubocop", "= 0.74"
   spec.add_development_dependency "rubocop-github", "~> 0.13.0"
+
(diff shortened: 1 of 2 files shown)


actionview-component.gemspec
```diff
@@ -43,4 +43,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "better_html", "~> 1"
   spec.add_development_dependency "rubocop", "= 0.74"
   spec.add_development_dependency "rubocop-github", "~> 0.13.0"
+
+  spec.post_install_message = "actionview-component has been renamed to view_component, and will no longer be published in this namespace. Please update your Gemfile to use view_component."
```
rmacklin: Do you think it'd be worth including a link to `https://github.com/github/view_component/issues/206` (https://github.com/github/view_component/issues/206) in this message?

PR #269, opened 2020-03-21
Support Ruby 2.4 in CI

### LINKED ISSUE #209: Support Ruby 2.4 in CI
https://github.com/github/actionview-component/pull/207 removed support for Ruby 2.4 as it was breaking CI: https://github.com/github/actionview-component/runs/444076324.

It would be nice to re-introduce this test coverage.

andrewmcodes:
### Summary

Fixes #209 

Update the `ruby_on_rails` action to re-add testing against Ruby 2.4.
To accomplish this, we need to run `gem update --system`, but we don't
want to do this on any version except `2.4.x` to save build time.

This commit uses the context and expression syntax for GitHub actions to
check if the Ruby versions starts with `2.4`, and update the system gems
in the event that we have a match.

### DIFF
--- a/.github/workflows/ruby_on_rails.yml
+++ b/.github/workflows/ruby_on_rails.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         rails_version: [5.0.0, 5.2.3, 6.0.0, master]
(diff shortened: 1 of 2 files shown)


joelhawksley:
Thanks @andrewmcodes! Mind adding a changelog entry and your name to the README?

PR #272, opened 2020-03-23
release v2.0.1

joelhawksley:


rmacklin:
I don't think this warrants a new release. There are no changes in the gem, only in the dummy app. There was no security vulnerability in this gem

joelhawksley:
@rmacklin fair enough. I was just being overly cautious 😅

PR #274, opened 2020-03-24
Render collections

### LINKED ISSUE #21: Any tips for rendering Components from collections?

tclem:
Fixes #21.

Introduces collection rendering for view components with the following API:

``` HTML+ERB
<%= render(ProductComponent.collection(items: @products, as: :product, foo: foo)) %>
```

Worked with @joelhawksley on the initial sketch of this with the goal of having this API feel very similar to the standard Rails [partial rendering of collections](https://guides.rubyonrails.org/layouts_and_rendering.html#rendering-collections). Opening early for feedback, there are still a few open TODOs.

#### TODO

- [x] Sort out validating `:as`
- [x] Update CHANGELOG


lib/view_component/base.rb
```diff
@@ -5,6 +5,41 @@
...
+    def initialize(component, options)
+      @component = component
+      @options = options
+    end
+
+    def render_in(view_context, &block)
+      as = as_variable(@options) || :item
+      args = @options.except(:items, :as)
+
+      @options[:items].map do |item|
```
rainerborene: I don't think that is the optimal way to render large collections. I suggest creating a string buffer and concatenate each rendered component instead of creating an array, then joining it later. Look for some inspiration on  `ActionView::OutputBuffer` class and see how it works under the hood. :)

lib/view_component/base.rb
```diff
@@ -5,6 +5,41 @@
 require "view_component/previewable"
 
 module ViewComponent
+  class CollectionBase
```
rainerborene: One more thing. Better extract this class into a different file.

rainerborene:
I didn't like the `collection` method name but that is subjective. I suggest following the same pattern of `render` view helper and stick with the same argument names like `collection` and `as` (which is already in use). Maybe `render_all` makes more sense.

lib/view_component/base.rb
tclem: Looks like `ActionView::OutputBuffer < ActiveSupport::SafeBuffer < String` (see [`buffers.rb`](https://github.com/rails/rails/blob/e2cf0b1d780b2e09f5270249ca021d94ce4fff9d/actionview/lib/action_view/buffers.rb#L21), and [`output_safety.rb`](https://github.com/rails/rails/blob/e2cf0b1d780b2e09f5270249ca021d94ce4fff9d/activesupport/lib/active_support/core_ext/string/output_safety.rb#L135)) and [`ActionView::RenderedCollection`](https://github.com/rails/rails/blob/e2cf0b1d780b2e09f5270249ca021d94ce4fff9d/actionview/lib/action_view/renderer/abstract_renderer.rb#L125) does `map/join` as well. We can definitely make this use a buffer, but is that optimal? Do we want to optimize for time? If so, a naive benchmark tells me that `join` is going to be faster.

```
  require "benchmark"
  def test_benchmark
    n = 1000000
    Benchmark.bmbm do |x|
      foo = [1] * n

      buffer = ActionView::OutputBuffer.new
      x.report("using ActionView::OutputBuffer") do
        foo.each do
          buffer << "foobar"
        end
        string = buffer.html_safe
      end

      x.report("using map and join") do
        string = foo.map do
          "foobar"
        end.join
      end
    end
  end
```

```
                                     user     system      total        real
using ActionView::OutputBuffer   0.449913   0.004264   0.454177 (  0.455215)
using map and join               0.085798   0.002497   0.088295 (  0.088758)
```

I'm new to rails internals, is there another builder like object besides `OutputBuffer`? I was also a little surprised that `render_in` had to return anything as I was expecting some internal buffer to be getting appended to as things are rendered. Is it possible that some of this is just for inline test rendering?

tclem:
@rainerborene, thank you so much for the review!

> I didn't like the `collection` method name but that is subjective.

:+1: I made another attempt here, curious to hear what you think about the naming. API looks like this now:

``` ruby
# Verbose version
render(ProductComponent.all(collection: @products, as: :product, foo: foo))

# Does the same thing
render(ProductComponent.all(@products, foo: foo))
```

lib/view_component/base.rb
tclem: Done.

lib/view_component/base.rb
rainerborene: @tclem lol. I thought it were using a string buffer instead of concatenating array items. did you benchmark appending to a simple string? We want to optimize for less string allocations and less memory usage.

rainerborene:
Naming is hard. One pattern that I see emerging from Rails codebase is `verb` + `_all` like `upsert_all`, `destroy_all`, `insert_all`, `touch_all`, and so on. Using only `all` is vague and does not reveal any intention IMO. I suggest using the `locals` attribute to pass in arbitrary local variables as well. See more [here](https://guides.rubyonrails.org/layouts_and_rendering.html#local-variables).

lib/view_component/collection.rb
```diff
@@ -0,0 +1,54 @@
...
+      end
+    end
+
+    # Copied from https://github.com/rails/rails/blob/e2cf0b1d780b2e09f5270249ca021d94ce4fff9d/actionview/lib/action_view/renderer/partial_renderer.rb
+    def as_variable(path, options)
+      if as = options[:as]
+        raise_invalid_option_as(as) unless /\A[a-z_]\w*\z/.match?(as.to_s)
+        as.to_sym
+      else
+        File.basename(path).gsub(/_component/, "").to_sym
```
jonspalmer: I don't think you need to use the path to do this. Unlike ActionView partials we have a concrete class that we can leverage to determine the default variable. You could do 

```ruby
class.name.demodulize.underscore.chomp("_component").to_sym
```

lib/view_component/collection.rb
```diff
@@ -0,0 +1,54 @@
...
+    def collection_variable(object, options)
+      if object.respond_to?(:to_ary)
+        object.to_ary
+      elsif options.key?(:collection)
+        collection = options[:collection]
+        collection ? collection.to_a : []
+      else
+        raise ArgumentError.new("Must specify the option `collection` or pass a valid collection object.")
+      end
+    end
```
jonspalmer: Is there a strong reason to have two ways to call the API? It's more code for us to maintain and a confusing API  (what would/should happen if I supply both options?) Given the user is calling a specific class method on the Component can we just force the first argument to be the collection?

Note that the Rails partial case is very different

```
<h1>Products</h1>
<%= render @products %>
```

This is a little bit of magic from Rails that says if you render an array it will treat it like a collection and infer the partial from the type of object in the array. We're not going to have that situation. The caller is going to pick the ViewComponent to call  `.all` on. 

The equivalent in ViewComponent world would be to support this:

```haml
render ViewComponent.all(@products) 
```

Under the covers ViewComponent would figure out the right ViewComponent class by type. In this case: `ViewComponent.all(@products) => ProductComponent.all(@products)`

However, that's a) maybe not necessary sugar for us b) a separate PR :)

lib/view_component/collection.rb
jonspalmer: Do we need this variable to be dynamic?  Unlike ActionView partials we have concrete classes that we can hand this kind of behavior off.  Wouldn't it be simple to define this in the component using a class_attribute? Something like this:


```ruby
class ProductCouponComponent < ViewComponent::Base

  with_collection_parameter :item # if you don't call this would default to :product_coupon

  def initialize(item:)
    @item = item
  end
end
```

test/view_component/view_component_test.rb
```diff
@@ -386,6 +386,84 @@ def test_backtrace_returns_correct_file_and_line_number
     assert_match %r[app/components/exception_in_template_component\.html\.erb:2], error.backtrace[0]
   end
 
+  # Some example usages:
+  #
+  # <%= render(ProductComponent.all(@products)) %>
+  # <%= render(ProductComponent.all(collection: @products, as: :product, foo: foo)) %>
+  # <%= render(ProductComponent.new(product: @product)) %>
```
jonspalmer: It would be terrific for these tests to be stronger a) to provide deeper "documentation" on how this feature works b) there are tests in here that will pass even if the features are broken.

test/view_component/view_component_test.rb
```diff
@@ -386,6 +386,84 @@ def test_backtrace_returns_correct_file_and_line_number
...
+  end
+
+  def test_render_collection_minimal
+    @products = [OpenStruct.new(title: "Hi"), OpenStruct.new(title: "Bye")]
+    render_inline(ProductComponent.all(@products, extra: "extra"))
+
+    assert_selector("h1", count: 2)
+  end
+
+  def test_render_collection_minimal_specify_as
```
jonspalmer: What's the case your guarding against with this test?

test/view_component/view_component_test.rb
```diff
@@ -386,6 +386,84 @@ def test_backtrace_returns_correct_file_and_line_number
...
+  # <%= render(ProductComponent.new(product: @product)) %>
+
+  def test_render_collection
+    @products = [OpenStruct.new(title: "Hi"), OpenStruct.new(title: "Bye")]
+    render_inline(ProductComponent.all(collection: @products, extra: "extra"))
+
+    assert_selector("h1", count: 2)
+  end
+
+  def test_render_collection_minimal
```
jonspalmer: Can we add assertions that prove the extra arguments get passed to the components? At the moment this test would pass if `extra: "extra"` wasn't passed in.

lib/view_component/base.rb
```diff
@@ -14,6 +15,11 @@ class Base < ActionView::Base
     class_attribute :content_areas, default: []
     self.content_areas = [] # default doesn't work until Rails 5.2
 
+    # Render a component collection.
+    def self.all(*args)
```
jonspalmer: Would it be better to  call this  `collection`?

lib/view_component/collection.rb
```diff
@@ -0,0 +1,54 @@
+
+# frozen_string_literal: true
+
+module ViewComponent
+  class Collection
+    def initialize(component, object = nil, options)
```
jonspalmer: I forget all the rules of public/private in Ruby but can we make this constructor private? You should only interact with it via the class method.

lib/view_component/collection.rb
```diff
@@ -0,0 +1,54 @@
...
+      @options = options
+    end
+
+    def render_in(view_context, &block)
+      as = as_variable(@component.virtual_path, @options)
+      collection = collection_variable(@object, @options)
+      args = @options.except(:collection, :as)
+
+      collection.map do |item|
+        args[as] = item
```
jonspalmer: Probably a bit more defensive to create a copy of the args here just on the off chance that the Component manipulates them. Could just do `args.merge(as: item)`

lib/view_component/base.rb
jonspalmer: Your benchmark isn't really apples to apples because you're doing `html_safe` in one and not the other. I think we'd want to do a benchmark of a real case using the rendering of the view component instead of test strings. My gut says that Rails must be doing html_safe etc. inside the loop. 

Imagine that with rendering components it looks more like this - note the new html_safe in the join case:

```ruby
require "benchmark"
  def test_benchmark
    n = 1000000
    Benchmark.bmbm do |x|
      foo = [1] * n

      buffer = ActionView::OutputBuffer.new
      x.report("using ActionView::OutputBuffer") do
        foo.each do
          buffer << "foobar"
        end
        string = buffer.html_safe
      end

      x.report("using map and join") do
        string = foo.map do
          "foobar".html_safe
        end.join
      end
    end
  end
```

then

```
                                     user     system      total        real
using ActionView::OutputBuffer   0.373818   0.002911   0.376729 (  0.377495)
using map and join               0.821182   0.017310   0.838492 (  0.840472)
```

jonspalmer:
> Naming is hard. One pattern that I see emerging from Rails codebase is `verb` + `_all` like `upsert_all`, `destroy_all`, `insert_all`, `touch_all`, and so on. Using only `all` is vague and does not reveal any intention IMO. I suggest using the `locals` attribute to pass in arbitrary local variables as well. See more [here](https://guides.rubyonrails.org/layouts_and_rendering.html#local-variables).

Naming: how about `with_collection` which would be very Rails.
WRT the locals comment I'm not sure I understand. Are you suggesting nesting the args under locals? That feels more awkward to me.

rainerborene:
> Naming: how about with_collection which would be very Rails.
WRT the locals comment I'm not sure I understand. Are you suggesting nesting the args under locals? That feels more awkward to me.

@jonspalmer If we want this library to be merged on Rails in future, we need to stay as close as possible to Rails internals.

jonspalmer:
> @jonspalmer If we want this library to be merged on Rails in future, we need to stay as close as possible to Rails internals.

Sure that's a good guiding principal however we should step back and understand the reasoning for the rails pattern. Passing locals to a partial is something like this

```erb
<h1>New zone</h1>
<%= render partial: "form", locals: {zone: @zone} %>
```

The value of `locals` in this use case is to distinguish between a) the options provided to the `render` method b) the set of parameters that will be passed to the partial

In our case I'm proposing that we make `as` something you define on the Component class and then simplify the signature to just

```ruby
def self.with_collection(collection, **component_args)
```

There is no confusion with that API. 

It's also nice because it makes the calling mostly "symmetric" when using a collection or not. i.e.

```erb
<%= render ProductComponent.new(product: @product, foo: 1, bar: 2) %>

<%= render ProductComponent.with_collection(@products, foo: 1, bar: 2) %>
```

jonspalmer:
Can we add an 'integration' test of this actually running in the test app?

lib/view_component/base.rb
joelhawksley: It was: https://github.com/github/view_component/pull/274#issuecomment-603619978

I'm fine with either, but have a slight preference for `all` as I think:

`.all(collection:` reads better than `.collection(collection:`.

lib/view_component/base.rb
tclem: Yup, you're right, both cases need to call `html_safe` (which is what's happening at the end of `render_in`), but I think they both need to call it inside the loop. The benchmark you pasted has the `OutputBuffer` case doing `html_safe` once at the end and the `map/join` case doing `html_safe` on each iteration. Let's make them the same.

>  I think we'd want to do a benchmark of a real case using the rendering of the view component instead of test strings. 

Love this suggestion! Here's an updated benchmark that actually calls `render_inline`:

``` ruby
require "benchmark"
def test_benchmark
  n = 10000
  @product = OpenStruct.new(title: "Bye")
  Benchmark.bmbm do |x|
    foo = [1] * n

    buffer = ActionView::OutputBuffer.new
    x.report("using ActionView::OutputBuffer") do
      foo.each do
        buffer << render_inline(ProductComponent.new(product: @product, extra: "abc"))
      end
    end

    x.report("using map and join") do
      foo.map do
        render_inline(ProductComponent.new(product: @product, extra: "abc"))
      end.join
    end
  end
end
```
```
                                     user     system      total        real
using ActionView::OutputBuffer   1.427219   0.023809   1.451028 (  1.461946)
using map and join               1.152875   0.026003   1.178878 (  1.204991)
```


I didn't dig too far, but array joins in ruby are [implemented in c](https://github.com/ruby/ruby/blob/5b287481befe03cc3e3dbc4b5571e21dbc523bae/array.c#L2365) and appear to be specialized for strings.

lib/view_component/base.rb
jonspalmer: my argument would be we shouldn't need a `collection:` argument. My proposal is
```ruby
MyComponent.with_collection(@products, .....)
```

tclem: I'm into the idea of `with_collection` and requiring the first argument to be the collection (something that can `to_ary`) and then dropping all the other special kwargs (like `as:` and `collection:`).

tclem: And I think we could take or leave the `with_` prefix:

``` ruby
MyComponent.collection(@products, extra: foo)
```
``` ruby
MyComponent.with_collection(@products, extra: foo)
```

lib/view_component/collection.rb
tclem: I really like this idea, especially because it mean removing `as:` as one of the special values in the options hash. Implemented in the latest on this branch (along with directly using the demodulized class name of the component).

lib/view_component/collection.rb
tclem: 100% onboard with this.

test/view_component/view_component_test.rb
tclem: I reworked these tests with the new api and tried to make them do a better job of documenting the feature, let me know what you think. There shouldn't be any that pass even if the feature is broken, can you comment on the specific test if that's still the case?

test/view_component/view_component_test.rb
tclem: This one went away.

lib/view_component/collection.rb
tclem: Yup! This constructor is now private.

lib/view_component/collection.rb
tclem: Agreed and implemented (slightly different implementation since the key is actually dynamic).

tclem:
Alright, I made another pass at the API and addressed a bunch of specific feedback. Thanks for the review and some great ideas @jonspalmer! I did add a single integration level test too which was a good idea because that demonstrated the need to use `html_safe` (I'll re-run my benchmarks next).

Here's the current state of the API:

``` ruby
ProductComponent.with_collection(@products, foo: foo, …)
```

And instead of having any special keyword args, if a component needs to have a different parameter name you can do that like so (otherwise in this case the name would be `product_coupon:`):

``` ruby
class ProductCouponComponent < ViewComponent::Base

  def initialize(item:)
    @item = item
  end
end
```

This also means that all additional hash/keyword args are passed directly to the component's constructor which I think is quite nice.

lib/view_component/base.rb
tclem: Re-running benchmarks with the knowledge that map/join will require a final `html_safe`. The `OutputBuffer` version doesn't need this because it's already a safe buffer.

map/join still wins by a little.

``` ruby
require "benchmark"
def test_benchmark
  n = 10000
  @product = OpenStruct.new(name: "Soap")
  Benchmark.bmbm do |x|
    foo = [1] * n

    buffer = ActionView::OutputBuffer.new
    x.report("using ActionView::OutputBuffer") do
      foo.each do
        buffer << render_inline(ProductComponent.new(product: @product, notice: "abc"))
      end
    end

    x.report("using map and join") do
      foo.map do
        render_inline(ProductComponent.new(product: @product, notice: "abc"))
      end.join.html_safe
    end
  end
end
```

```
                                     user     system      total        real
using ActionView::OutputBuffer   1.543178   0.028958   1.572136 (  1.573645)
using map and join               1.285421   0.025860   1.311281 (  1.312286)
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # master
 
+# v2.1.0
+
+* Support rendering collection (e.g., `render(MyComponent.all(@items))`).
```
joelhawksley: Just dropping a note here to make sure we update this to reflect whatever API we land on ❤️

test/view_component/view_component_test.rb
```diff
@@ -386,6 +386,61 @@ def test_backtrace_returns_correct_file_and_line_number
     assert_match %r[app/components/exception_in_template_component\.html\.erb:2], error.backtrace[0]
   end
 
+  # Rendering collections
+
+  def test_render_collection
+    @products = [OpenStruct.new(name: "Radio clock"), OpenStruct.new(name: "Mints")]
```
jonspalmer: Its a nit to pik but do we need instance variables here? can't we just use regular variables given they're internal to the test method.

rainerborene:
Great work @tclem. Make sure to rebase before merging this PR. :fist_oncoming:

test/view_component/view_component_test.rb
tclem: Good catch! Fixed.

lib/view_component/version.rb
```diff
@@ -3,7 +3,7 @@
 module ViewComponent
   module VERSION
     MAJOR = 2
-    MINOR = 0
+    MINOR = 1
```
tclem: @joelhawksley would you prefer to do version bump and release in a separate PR? Just read the contributing guidelines...

joelhawksley: I would, if you don't mind removing this change 😄

test/view_component/view_component_test.rb
```diff
@@ -386,6 +386,61 @@ def test_backtrace_returns_correct_file_and_line_number
     assert_match %r[app/components/exception_in_template_component\.html\.erb:2], error.backtrace[0]
   end
 
+  # Rendering collections
```
joelhawksley:
```suggestion
```

README.md
```diff
@@ -446,15 +446,10 @@ Bug reports and pull requests are welcome on GitHub at https://github.com/github
 |@mellowfish|@horacio|@dukex|@dark-panda|@smashwilson|
```
joelhawksley: Might we add documentation for this feature to the README?

tclem: 👍Done. Take a read though and feel free to edit.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # master
 
+# v2.1.0
+
+* Support rendering collections (e.g., `render(MyComponent.with_collection(@items))`).
+
+    *Tim Clem, Joel Hawksley*
```
joelhawksley:
```suggestion
    *Tim Clem*
```

joelhawksley: I can't possible take enough credit to warrant putting my name on this 🙇

tclem: LOL, fair enough. Thanks for getting me ramped up quickly!

PR #278, opened 2020-03-31
Add support for config.action_view.annotate_template_file_names

joelhawksley:
This PR introduces support for `config.action_view.annotate_template_file_names`, which is coming in Rails 6.1: https://github.com/rails/rails/pull/38848.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -283,7 +283,7 @@ module ViewComponent
 
         if handler.method(:call).parameters.length > 1
           handler.call(self, template)
-        else # remove before upstreaming into Rails
+        else
           handler.call(OpenStruct.new(source: template, identifier: identifier, type: type))
         end
       end
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -77,6 +77,10 @@ module ViewComponent
       true
     end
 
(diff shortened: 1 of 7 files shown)


lib/view_component/base.rb
```diff
@@ -180,8 +184,20 @@ def compile(raise_template_errors: false)
           return false
         end
 
+        # If template name annotations are turned on, a line is dynamically
```
BlakeWilliams: 🙏 I really appreciate this comment

lib/view_component/base.rb
```diff
@@ -180,8 +184,20 @@ def compile(raise_template_errors: false)
...
+        # added with a comment. In this case, we want to return a different
+        # starting line number so errors that are raised will point to the
+        # correct line in the component template.
+        line_number =
+          if ActionView::Base.respond_to?(:annotate_template_file_names) &&
+            ActionView::Base.annotate_template_file_names
+            -2
```
BlakeWilliams: Would it be worth adding a regression test for this case since we already have one for `-1`?

joelhawksley: Our test will cover both cases, since we run both Rails 6.1 and previous versions in our matrix builds ❤️

Issue #280, opened 2020-03-31
v2.2.0 assigns source_location incorrectly


joelhawksley:
### Steps to reproduce
Create a component that inherits from another component, such as ApplicationComponent. Attempt to render the component.

### Expected behavior
The component renders.

### Actual behavior
`ViewComponent::TemplateError: Could not find a template file for MyChildComponent.`

### System configuration
**Rails version**:
6.0.2.2

**Ruby version**:
2.6.3p62

**Gem version**:
2.2.0

joelhawksley:
@nashby it looks like 2.2.0 might be broken. I'm looking into this now.

PR #281, opened 2020-03-31
ignore redefinitions of inherited when inspecting caller locations

joelhawksley:
Given an ApplicationComponent that re-defines `inherited`,
we were failing to find the correct template for the child
component, as we were looking at the top frame in the
call stack.

This change ignores any definitions of `inherited` in
the call stack, avoiding the issue.

[Fixes: #280]

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -149,8 +149,8 @@ module ViewComponent
         end
 
         # Derive the source location of the component Ruby file from the call stack.
-        # We need to ignore `inherited` frames here as they indicated that this method
-        # has been re-defined by the consuming application.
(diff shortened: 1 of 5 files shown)


joelhawksley:
@nashby what do you think of this solution?

nashby:
@joelhawksley hey, looks good to me! The only thing I'm worried that we need to get whole stacktrace for that (not only the first frame as it was before). But I can't really think of any other solution atm.

joelhawksley:
@nashby do you think we could limit the size of the stack trace request to something like 10 frames?

nashby:
@joelhawksley I think practically that should work but in theory there might be the case when it's not enough, not sure (e.g when there're a lot of inheritance happening)

PR #284, opened 2020-04-02
Allow using inline render method(s) defined on a parent

simonrand:
### Summary

Allows defining inline render methods on a parent component, and using them in child components.

Also fixes an issue where inline variant calls were not working correctly as the list of variants was not being populated from the defined variant calls.

Example:

```ruby
class BaseComponent < ViewComponent::Base
  def call
    content_tag :div, class: base_classes + additional_classes
  end

  private

  def base_classes
    %w[base-class]
  end

  def additional_classes
    []
  end
end

class ChildComponent < BaseComponent
  private

  def additional_classes
    %w[additional-class]
  end
end

# result

> BaseComponent.new.call
=> "<div class=\"base-class\"></div>"

> ChildComponent.new.call
=> "<div class=\"base-class additional-class\"></div>"
```


simonrand:
(will update tests)

simonrand:
First attempt at this had a silly mistake, updated now!

simonrand:
Just as an extra note on this, I'm not massively comfortable grepping the all instance methods for what we are looking for here, an alternative much be just to grep the relevant classes, something like:

```ruby
def inlined?
  return false unless templates.empty?

  # Only check if an inline render method has been defined on this class
  # or on an ancestor, up to (but excluding) ViewComponent::Base
  ancestor_classes = (ancestors[0...ancestors.index(ViewComponent::Base)] - included_modules)
  return false unless ancestor_classes

  ancestor_classes.any? { |ancestor_class| ancestor_class.instance_methods(false).grep(/^call/).present? }
end
```

..or maybe there's a better way to do this?

joelhawksley:
@simonrand what if we used a `respond_to?(:call)` check?

joelhawksley:
(It would be helpful to have a test case here, if you don't mind ❤️ )

simonrand:
@joelhawksley thanks for the reply! (..and the work on all this - have been following it for a long time) I did go for that approach first, but I hit a couple of problems, mostly around moving this check from a class to instance method (felt like too big a change for what was needed) - the other problem arose from supporting inline calls for variants.. Leave this with me and I'll take another look (and add tests!)

simonrand:
@joelhawksley have updated!

This feature is made a bit difficult mainly by the variants feature - but I think I've got a reasonable solution at this point (without getting into changing the way variants are implemented)

..there is maybe also an argument for not being so defensive on the template rendering - it a template and/or inline call are missing, instead of trying to catch this, maybe just raising would be an easier option? (perhaps by catching any calls this `call*` methods in `ViewComponent::Base`) I'm not convinced this is good either so, so I will still offer this PR as my solution for now..

I should mention this also fixes a bug - inline variants were not working correctly as the available variants were not being gathered properly and thus this check https://github.com/github/view_component/blob/3550b486aeb5865656a7b2a34bdd51b9065a0895/lib/view_component/base.rb#L160 was always falling.

Thanks again!

lib/view_component/base.rb
```diff
@@ -243,6 +245,25 @@ def collection_parameter_name
 
       private
 
+      def compiled_template(file_path)
```
simonrand, line 248: Note: Moving this here to preserve the (maybe previously unintentional, but intentional from me) alphabetical order of the private methods

README.md
```diff
@@ -211,6 +211,22 @@ class InlineComponent < ViewComponent::Base
...
+    link_to "Default", default_path
+  end
+
+  def call_variant
+    link_to "variant", variant_path
+  end
+end
+```
+
+Using a mixture of templates and inline rendering in a component is not supported (templates will be prioritized).
```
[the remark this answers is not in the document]
simonrand: Yeah I was thinking further about this myself since I original wrote this - it's probably possible with a little more work - on reflection I can see it could be useful - I'll take a look!

simonrand:
@joelhawksley so it seems like I was about 90% there with the work I'd already done! dddd353 implements sidecar + inline method support in the same component, with the template taking priority in situations where both exist (which would seem to be the correct behaviour to me)..

There'll obviously be warnings if an inline render method and template exist (as the inline method will be overwritten in the compile method, which is now called if _any_ template exists unless the component is entirely inlined) - but I think this is probably ok, unless we dig into how the template methods are being defined (which I'm not sure is really necessary at this stage)..

README.md
```diff
@@ -211,6 +211,22 @@ class InlineComponent < ViewComponent::Base
...
 
+It is also possible to render variants inline by creating additional `call_` methods.
+
+```ruby
+class InlineVariantComponent < ViewComponent::Base
+  def call
+    link_to "Default", default_path
+  end
+
+  def call_phone
```
joelhawksley, line 222: Love it!

README.md
```diff
@@ -211,6 +211,22 @@ class InlineComponent < ViewComponent::Base
...
+    link_to "Phone", phone_path
+  end
+end
+```
+
+Using a mixture of templates and inline render methods in a component is supported, however if both exist for a component (or specific variant) the template will take priority.
```
joelhawksley, line 228: Reading this description, I wonder if it would be better to just raise an error in this case. I can't think of a case where we'd ever want to allow both to exist.

lib/view_component/base.rb
```diff
@@ -180,7 +180,7 @@ def compile!
       # We could in theory do this on app boot, at least in production environments.
       # Right now this just compiles the first time the component is rendered.
       def compile(raise_template_errors: false)
-        return if compiled? || inlined?
+        return if compiled? || all_inlined?
```
joelhawksley, line 183: What do you think about refactoring this logic to just use `compiled?` This process should only run once per component anyways.

test/view_component/view_component_test.rb
```diff
@@ -26,11 +26,49 @@ def test_renders_content_from_block
   def test_render_without_template
     render_inline(InlineComponent.new)
 
-    assert_predicate InlineComponent, :inlined?
+    assert_predicate InlineComponent, :all_inlined?
     assert_not_predicate InlineComponent, :compiled?
```
joelhawksley, line 30: To me, `compiled?` simply means that the component has been fully initialized / optimized to render, so perhaps it would make more sense for it to be true in this case?

simonrand:
@joelhawksley thanks for the comments - updated! 

One thing of note - there's a relatively interesting case here: https://github.com/github/view_component/pull/284/files#diff-c7432794f4e04debce14f00887227594R104-R110 - in this example we have a child component which inherits from a fully inlined component, but the child provides a template file for a variant call (it could be a default call either). 

In my changes I am allowing this - my reasoning is there's a clear use case in our usage of ViewComponent for this. i.e. we have a Base component which has a `def call` which covers most cases for any child components that inherit from it, however there are cases when a template file may be required, so allowing this would seem to make some sense. Effectively I'm limiting the scope of the template file and inline render method check to only the render methods on that instance.

Let me know what you think!

lib/view_component/base.rb
```diff
@@ -143,6 +143,10 @@ def request
     class << self
       attr_accessor :source_location
 
+      def all_templates_inlined?
+        @all_templates_inlined ||= inline_calls.any? && templates.empty?
```
joelhawksley, line 147: Do we need this predicate for non-testing purposes? If not, can we remove it, or at least take it out of the rendering path?

lib/view_component/base.rb
```diff
@@ -284,20 +310,44 @@ def template_errors
...
+        calls.reject { |call| call == :call }.map do |variant_call|
+          variant_call.to_s.sub("call_", "").to_sym
         end
       end
+
+      def variant_templates
+        @variant_templates ||= templates.map { |template| template[:variant] }
+      end
+
+      def variants
```
joelhawksley, line 341: What do you think about dynamically defining this method during compilation? Once the component is compiled, the return value won't change.

lib/view_component/base.rb
simonrand: You're absolutely right - no need for this at all, will remove!

simonrand:
@joelhawksley thanks for the comments! I have updated - I think this is looking tidier to me now..

One thing - when defining the `variants` method I've used `define_singleton_method`, it feels a bit more appropriate here (compared to `class_eval`) given the method name and content is pretty static - or would you like to keep things consistent here?

lib/view_component/base.rb
```diff
@@ -284,20 +305,36 @@ def template_errors
...
 
-      def compiled_template(file_path)
-        handler = ActionView::Template.handler_for_extension(File.extname(file_path).gsub(".", ""))
-        template = File.read(file_path)
-
-        if handler.method(:call).parameters.length > 1
-          handler.call(self, template)
-        else
-          handler.call(OpenStruct.new(source: template, identifier: identifier, type: type))
+      def variants_from_inline_calls(calls)
```
joelhawksley, line 326: Do we need to pass in `calls` here or could we have this method access `inline_calls_defined_on_self` directly?

lib/view_component/base.rb
```diff
@@ -284,20 +305,36 @@ def template_errors
...
+        calls.reject { |call| call == :call }.map do |variant_call|
+          variant_call.to_s.sub("call_", "").to_sym
         end
       end
+
+      def view_component_ancestors
```
joelhawksley, line 332: Is there a reason for this to be its own method? Could we inline it above instead? (keeping the nice comment you've written here, of course ❤️ )

lib/view_component/base.rb
simonrand: So I'm actually using this twice, once for all inline calls (`inline_calls`,  https://github.com/github/view_component/pull/284/files#diff-08fe1c0f29a6308ad005c21f4adcfd64R187) and then again for the  `inline_calls_defined_on_self` (https://github.com/github/view_component/pull/284/files#diff-08fe1c0f29a6308ad005c21f4adcfd64R314) so think this is maybe ok as-is?

joelhawksley:
@simonrand excellent work! Thank you for your contribution. I look forward to your next PR ❤️

simonrand:
Nice one! Thanks @joelhawksley - ended up being quite a bit different to what I intended when I started, but I think the result ended up being good!

This is an important feature for something we're working on, I'm hoping to be able to push something solid on that soon - I'll probably let you know when this happens (via issues) as I think it might have some possible (if quite specific) benefits in some use cases..

Thanks again for the work on this!

Issue #292, opened 2020-04-08
Capybara as runtime dependency?


n-rodriguez:
Hi there!

Do we really need `capybara` as runtime dependency? See : https://github.com/github/view_component/blob/master/view_component.gemspec#L36

It's used in test helper : https://github.com/github/view_component/blob/master/lib/view_component/test_helpers.rb

Thank you!

joelhawksley:
@n-rodriguez only one way to find out! Would you be up for opening a PR to move it to development only?

n-rodriguez:
@joelhawksley done https://github.com/github/view_component/pull/293

joelhawksley:
@n-rodriguez - per @jcoyne's comment here: https://github.com/github/view_component/pull/293#issuecomment-611273636, we need Capybara for the test helpers to work. 

While most people will have it in their apps already, I don't think it makes sense to require folks to install it themselves to use the test helpers.

Was there an issue you ran into that prompted the desire to make this change?

n-rodriguez:
> Was there an issue you ran into that prompted the desire to make this change?

Because it will be installed in production environment. I do use Capybara locally but not in prod :)

joelhawksley:
@n-rodriguez is there is a way to have a gem provide different dependencies for `development` and `prod`? 

From what I understand, `development_dependency` functions to identify dependencies needed for the development _of_ the gem, not using the gem _in_ development/test.

n-rodriguez:
> From what I understand, `development_dependency` functions to identify dependencies needed for the development _of_ the gem, not using the gem _in_ development/test.

Good point! At first I thought you were actually using Capybara in *your* tests but it seems that it's not the case.

> @n-rodriguez is there is a way to have a gem provide different dependencies for `development` and `prod`?

I would say... documentation.

rmacklin:
I agree that it's not ideal to require `capybara` in a production environment - given that it's not used in production, it is essentially a waste of memory to load it.

So personally, I lean towards removing the dependency from `view_component` for a few reasons:
- Many (most?) rails apps will already declare a `test` dependency on `capybara` because [Rails System Tests](https://guides.rubyonrails.org/testing.html#system-testing) depend on it, which means they'll still be able to use the `capybara` assertions in their component tests.
- It's still possible to test components without the `capybara` assertions, by directly using the return value of the `render_inline` helper.
[...]

PR #293, opened 2020-04-09
Declare Capybara as a development dependency

n-rodriguez:
Fix https://github.com/github/view_component/issues/292

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,6 @@ PATH
   remote: .
   specs:
     view_component (2.2.2)
-      capybara (~> 3)
 
 GEM
   remote: https://rubygems.org/
@@ -189,6 +188,7 @@ PLATFORMS
 DEPENDENCIES
   better_html (~> 1)
   bundler (~> 1.14)
+  capybara (~> 3)
   haml (~> 5)
   minitest (= 5.1.0)
   rails (= 6.0.2.2)
--- a/view_component.gemspec
+++ b/view_component.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.3.0"
 
-  spec.add_runtime_dependency     "capybara", "~> 3"
+  spec.add_development_dependency "capybara", "~> 3"
   spec.add_development_dependency "bundler", "~> 1.14"
   spec.add_development_dependency "rake", "~> 13.0"
   spec.add_development_dependency "minitest", "= 5.1.0"


jcoyne:
I believe the test helper depends on capybara matchers. https://github.com/github/view_component/blob/master/lib/view_component/test_helpers.rb#L3
  It would be good to document that you need to include capybara in your gemfile if you want to also use that helper.

joelhawksley:
@jcoyne I was wondering if that would be an issue. I wouldn't want to break the existing test helper functionality.

jcoyne:
@joelhawksley It's likely that your application already had capybara as a dependency, in which case it's not a problem.  You could test for `defined? Capybara` in TestHelpers and emit a warning about adding it to the "Gemfile" if it isn't

joelhawksley:
@jcoyne if we're going to expect folks to change their Gemfiles, I'm not sure this change is worth it, IMO.

n-rodriguez:
> It would be good to document that you need to include capybara in your gemfile if you want to also use that helper.

sure

n-rodriguez:
> You could test for `defined? Capybara` in TestHelpers and emit a warning about adding it to the "Gemfile" if it isn't

:+1:

joelhawksley:
@n-rodriguez will this require users of the library to change their Gemfiles to use the test helpers? If so, I'm going to recommend that we stick with our current dependency on Capybara.

md5:
Would it be possible to move the test helpers to a separate gem, say `view_component-test_helpers`, and have the runtime dependency for `capybara` live there?

Then in the docs, you could change to this:

> In `Gemfile`, add:
> 
> ```
> gem "view_component"
> gem "view_component-test_helpers", group: :development
> ```

joelhawksley:
@md5 can you provide an example of an existing (preferably popular) project that has a separate gem for test helpers? While that solution makes sense, it feels cumbersome to me.

@n-rodriguez what is the concern with having `capybara` in production? Is it the unnecessary increase in bundle size?

md5:
> @md5 can you provide an example of an existing (preferably popular) project that has a separate gem for test helpers? While that solution makes sense, it feels cumbersome to me.

I think `rack-test` is probably the best example that I'm seeing, although `Rack::MockRequest` and `Rack::MockResponse` are built into `rack` itself and autoloaded from `rack/mock`.

Speaking for myself, I don't find the `capybara` dependency all that objectionable, since it has pretty tame dependencies itself, but it definitely isn't needed in production. I would say the argument is not just bundle size, but having any code above and beyond what's actually needed to run the app. There's obviously a slippery slope here, since there may be parts of `rails` that people aren't using or other cases where they only use a small subset of a monolithic gem's functionality, but `capybara` is a library for testing and nothing else.

rmacklin:
PSA: #298 has been merged so capybara is no longer a dependency.

joelhawksley:
Closing in favor of https://github.com/github/view_component/pull/298. Thank you for your work on this PR @n-rodriguez @md5 @jcoyne !

PR #296, opened 2020-04-11
Making rails application layout the default for preview index views

### LINKED ISSUE #289: Preview index actions use application layout instead of rails application layout
It seems that `/rails/view_components` page is rendered with "application" layout instead of "rails/application" layout. It is a problem because `RailsApplicationController` is missing methods implemented in `ApplicationController` and potentially used in the "application" layout. I checked the difference in implementation with mails preview and I can't really spot the difference.

juanmanuelramallo:
Resolves #289 

### Summary

- Updated view_components_controller to only prepend application's view paths when rendering a given preview
- Preview index views will use rails' default application layout
- Added integration tests for this new behaviour

image

image

image

### DIFF
--- a/app/controllers/rails/view_components_controller.rb
+++ b/app/controllers/rails/view_components_controller.rb
@@ -4,7 +4,6 @@ require "rails/application_controller"
 
 class Rails::ViewComponentsController < Rails::ApplicationController # :nodoc:
   prepend_view_path File.expand_path("../../../lib/railties/lib/rails/templates/rails", __dir__)
-  prepend_view_path "#{Rails.root}/app/views/" if defined?(Rails.root)
 
   around_action :set_locale, only: :previews
   before_action :find_preview, only: :previews
@@ -18,7 +17,7 @@ class Rails::ViewComponentsController < Rails::ApplicationController # :nodoc:
     @previews = ViewComponent::Preview.all
     @page_title = "Component Previews"
     # rubocop:disable GitHub/RailsControllerRenderPathsExist
-    render "components/index"
+    render "components/index", layout: "application"
     # rubocop:enable GitHub/RailsControllerRenderPathsExist
   end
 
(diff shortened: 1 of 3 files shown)


jonspalmer:
I think I understand how this fix works but it's really opaque. Is the fundamental problem that when we add the Rails app/view then there are two layouts called "application" and the Rails app version gets chosen over the Rails default?

juanmanuelramallo:
What do you suggest @jonspalmer ?

tchak:
No matter what is decided about preview controller inheritance I think listing layout/styles should be from "rails/application". So I am :+1: for this fix. I don't know enough about rails internals (and especially view layer) to know if there is a more elegant fix...

jonspalmer:
Yeah I see the tests. My question is do we add a 4th test that makes three `get` requests inside the same test case. Make sure we're not relying on test order execution or some set of side effects we're not catching?

juanmanuelramallo:
Definitely, good call!

test/view_component/integration_test.rb
```diff
@@ -186,6 +185,35 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_select("div", "hello,world!")
   end
 
+  test "test preview renders application's layout by default" do
+    get "/rails/view_components/preview_component/default"
+
+    assert_includes response.body, "<title>ViewComponent - Test</title>"
```
joelhawksley, line 191: Can we use `assert_select` for these tests?

app/controllers/rails/view_components_controller.rb
```diff
@@ -4,7 +4,6 @@
 
 class Rails::ViewComponentsController < Rails::ApplicationController # :nodoc:
   prepend_view_path File.expand_path("../../../lib/railties/lib/rails/templates/rails", __dir__)
-  prepend_view_path "#{Rails.root}/app/views/" if defined?(Rails.root)
```
joelhawksley, line 7: Do we need a CHANGELOG for this?

juanmanuelramallo: I'll add an entry

juanmanuelramallo: Will this work https://github.com/github/view_component/pull/296/commits/c1a2a949f521612611bad1aebf06da32908756db ?

PR #298, opened 2020-04-12
Make Capybara optional

rmacklin:
### Summary

This is a potential solution for the issue brought up in #292.

`ViewComponent::TestHelpers` was the only module that used `capybara`, and that module is for the test environment. Thus, requiring `capybara` in the production environment is unnecessary. Furthermore, since it's still possible to test components using the return value of `render_inline`, applications that don't already have their own `capybara` dependency could still use `view_component` and test their components without the `capybara` assertions. On the other hand, any rails app that already depends on capybara (including any app that uses [Rails System Tests](https://guides.rubyonrails.org/testing.html#system-testing)) will automatically be able to use capybara assertions in their component tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Remove `capybara` dependency.
+
+    *Richard Macklin*
+
 # v2.3.0
 
 * Allow using inline render method(s) defined on a parent.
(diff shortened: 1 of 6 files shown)


joelhawksley:
@rmacklin I'd be 👍 to moving forward with this.

rmacklin:
Sounds good. I've rebased and updated the `README` and `CHANGELOG` to note this change.

joelhawksley:
@rmacklin thank you for your contribution! I really appreciate your work in coming to a solution for this issue.

rmacklin:
Glad I could help!

PR #299, opened 2020-04-13
Only autoload public API

rmacklin:
## Summary

Some of the modules we were `autoload`ing are internal modules. Unlike `Base`, `Preview`, `TestCase`, and `TemplateError`, we don't intend for users to directly use the `Previewable`, `TestHelpers`, `RenderMonkeyPatch`, or `RenderingMonkeyPatch` modules (AFAIK). So I've removed those modules from the gem's `autoload`s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Remove `autoload`s of internal modules (`Previewable`, `RenderMonkeyPatch`, `RenderingMonkeyPatch`).
+
+    *Richard Macklin*
+
 * Remove `capybara` dependency.
 
     *Richard Macklin*
--- a/lib/view_component.rb
+++ b/lib/view_component.rb
@@ -9,7 +9,5 @@ module ViewComponent
   autoload :Preview
   autoload :TestHelpers
   autoload :TestCase
-  autoload :RenderMonkeyPatch
-  autoload :RenderingMonkeyPatch
   autoload :TemplateError
 end
--- a/lib/view_component.rb
+++ b/lib/view_component.rb
@@ -7,7 +7,6 @@ module ViewComponent
 
   autoload :Base
   autoload :Preview
-  autoload :Previewable
   autoload :TestHelpers
   autoload :TestCase
   autoload :RenderMonkeyPatch
--- a/lib/view_component/engine.rb
+++ b/lib/view_component/engine.rb
@@ -43,11 +43,17 @@ module ViewComponent
 
     initializer "view_component.monkey_patch_render" do
       ActiveSupport.on_load(:action_view) do
-        ActionView::Base.prepend ViewComponent::RenderMonkeyPatch if Rails.version.to_f < 6.1
+        if Rails.version.to_f < 6.1
+          require "view_component/render_monkey_patch"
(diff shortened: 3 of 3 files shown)


lib/view_component/test_case.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require "active_support/test_case"
+require "view_component/test_helpers"
```
joelhawksley, line 4: Are there any cases where this line was not previously required, but now will be due to these changes?

If so, can we update the README and add a CHANGELOG entry?

rmacklin: It looks like the README already has `require "view_component/test_helpers"` in the section about testing with rspec: https://github.com/github/view_component/blob/dc0c0c1671dc116a3c2b822a5031b4045c2eb569/README.md#L553-L562
and I don't know of anywhere else where it'd be necessary.

But after seeing that part of the README, I'm wondering if I should remove b0680eb8ba670ed5ef681b65416e6e0883451264. What do you think?

joelhawksley: Can you be more specific about what you're suggesting? I'm not sure I understand.

rmacklin: I was asking if I should drop that commit from this PR because
```
Remove autoload of ViewComponent::TestHelpers

This is an internal module that's only used in ViewComponent::TestCase
```
is a bit contradictory with the part of the README I came across above.

joelhawksley: Ah. I think you're right, unless there is another way to support RSpec. I think the approach we have in the README is the recommended route. 

If you agree, then yeah, let's remove that part of this PR.

jonspalmer: For RSpec the IMO the "best" solution is to separate `TestHelpers` into two pieces. RSpec users get value out of `render_inline` and the supporting code. We're using that today. What RSpec doesn't need is the Capybara Assertions etc. For RSpec you're going to pull in the rspec Capybara helpers explicitly. It would be entirely normal/expected to configure RSpec like this;

```ruby
require 'capybara/rspec'
require "view_component/test_helpers"

RSpec.configure do |config|
end
```

I wonder should we just push Capybara yup to  `ViewComponent::TestCase` ? Anyone using Minitest is going to inherit from that. Meanwhile other frameworks just get what they need.

rmacklin: Sounds good. I've rebased on top of the latest `master` and dropped that commit.

rmacklin: @jonspalmer I was thinking the same thing about moving the Capybara stuff from `TestHelpers` into `TestCase`. Do you want to open a new issue about that? I think this PR, especially with that commit dropped, is independent.

rmacklin: I've opened a stub issue for now that links to the previous comments about it: https://github.com/github/view_component/issues/304

joelhawksley:
Do you think we need a CHANGELOG here?

rmacklin:
> Do you think we need a CHANGELOG here?

I don't think it'd hurt - I've pushed 9bc63fb086b73ce68bbc0e3e6506fe8a9d4d1725

joelhawksley:
❤️ Thanks for another clean PR! Keep 'em coming!

PR #305, opened 2020-04-15
Declare dependency on activesupport

rmacklin:
### Summary

We're requiring activesupport in several places:
https://github.com/github/view_component/blob/060275c1927f4d7a5a5949866838a880ea896d2a/lib/view_component.rb#L3
https://github.com/github/view_component/blob/060275c1927f4d7a5a5949866838a880ea896d2a/lib/view_component/base.rb#L4
https://github.com/github/view_component/blob/060275c1927f4d7a5a5949866838a880ea896d2a/lib/view_component/preview.rb#L3
https://github.com/github/view_component/blob/060275c1927f4d7a5a5949866838a880ea896d2a/lib/view_component/previewable.rb#L3
https://github.com/github/view_component/blob/060275c1927f4d7a5a5949866838a880ea896d2a/lib/view_component/test_case.rb#L3

Thus, we should declare a formal dependency on it. Since we support Rails 5.0.0 through Rails 6.x, we've declared the dependency as `>= 5.0.0, < 7.0`. (Without the 7.0 upper bound, rubygems would print `WARNING:  open-ended dependency on activesupport (>= 5.0.0) is not recommended` when building the gem.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Declare explicit dependency on `activesupport`.
+
+    *Richard Macklin*
+
 * Remove `autoload`s of internal modules (`Previewable`, `RenderMonkeyPatch`, `RenderingMonkeyPatch`).
 
     *Richard Macklin*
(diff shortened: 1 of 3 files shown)


joelhawksley:
@rmacklin ❤️ You're on fire! 🔥

PR #306, opened 2020-04-15
Issue 301: Access collection counters

franks921:
### Summary

`ViewComponent` also makes an index variable available named after the component name or the name supplied to `with_collection_parameter` followed by `_counter`. To access this variable add it to your constructor as an argument.

### Usage example

`app/components/product_component.rb`
``` ruby
class ProductComponent < ViewComponent::Base
  def initialize(product:, product_counter:)
    @product = product
    @counter = product_counter
  end
end
```

`app/components/product_component.html.erb`
``` erb
<li>
  <%= @counter %> <%= @product.name %>
</li>
```


jonspalmer:
Great work here. I like the overall shape and API. One though on design. 

a) The code here is a little dense and obscure with all the magic of setting attr_readers and properties being added to component classes. It took me a really long time to read and understand how it all worked. 
b) We're setting the iterator object on the class after its constructed which means consumers have less flexibility to manipulate it etc.

I wonder instead could/should we change it so we pass the iteration object to the component constructor and allow the component to decide what to do with it? I think that will be more flexible than having it set magically after construction. It would be something like this:

`app/view/products/index.html.erb`
```erb
<%= render(ProductComponent.with_collection(@products)) %>
```
```ruby
class ProductComponent < ViewComponent::Base
  def initialize(product:, product_iterator:)
    @product = product
    @index = product_iterator.index
  end
end
```

I think we could inspect the arguments of the Component to decide if an iterator is being asked for. You could also make it configurable with an optional second argument to `with_collection_parameter`:

```ruby
class ProductComponent < ViewComponent::Base
  with_collection_parameter :item, :item_iterator
  
  def initialize(item:, item_iterator:)
    @item = item
    @index = item_iterator.index
    @last = item_iterator.last?
  end
end
```

Thoughts?

franks921:
Thanks for taking the time to look at this @jonspalmer. 

Funny you should mention it, but my initial stab at this was to pass the iteration class to the component and deal with it in that class. After looking at how Rails does it with partial rendering, I decided to go the same route and keep it in line with Rails setting those helper methods automagically. My motivation on going route was to stay inline with Rails' methodologies as much as possible.

I even went the route of merging it into `@options` on [collection.rb#L11](https://github.com/github/view_component/blob/ae5dc501ec2e9cbef8cd626c340f344923fdb7ed/lib/view_component/collection.rb#L11) but everything I tried felt very bloated and overcomplicated for the desired end-result.

Would it maybe be preferable to set these methods on the base component, similar to `render?`, setting their defaults, and allowing developers to override those methods if the need arises?

I am happy to open a second PR with that as a proposal.

jonspalmer:
@franks921 I understand the motivation for doing it the way that you have. Staying close to Rails conventions is desirable and consistent with the principal of least surprise. However, that parallels are slightly different.

When you render a collection in a Rails view Rails will automagically add both the item from the collection and the iteration object to the set of Locals variables passed to the inner template.

In ViewComponent land we aren't passing things to a partial via locals. We're passing things to the ViewComponent. We could have made the decision to set item from the collection on the component instance via meta programming, the same way you are setting the iterator here. However, we rightly decided not to do that the. This gives the ViewComponent the ability to choose what to do with the passed in item at construction time. IMO we should be consistent and pass the iterator the same way.

I agree that the code to make this work in collection.rb is somewhat non trivial but I'm pretty convinced that overall it would be more grokable and simpler than the meta-programming, `send` code etc. you have today.

franks921:
Yeah that makes sense @jonspalmer, let me go back to the drawing board. I will keep committing to this PR. 

While I have you here, is there any preference you would like in terms of naming? I see you've used `_iterator` in your example code, should I stick with this?

jonspalmer:
For example I think we could make the render_in method this:

```ruby
  iteration = CollectionIteration.new(@collection.size)
  @collection.map do |item|
    @component.new(component_options(item, iterator)).render_in(view_context, &block)
  end.join.html_safe
end

def component_options(item, iterator)
  as = @component.collection_parameter_name
  options = { as => item }
  options.merge! { iterator_key: iterator } if iterator_key
  options
end

def iterator_key
  # check @component.iterator_parameter_name and/or introspect component constructor args
end
```

jonspalmer:
> While I have you here, is there any preference you would like in terms of naming? I see you've used _iterator in your example code, should I stick with this?

I think sticking with the Rails convention of `_iterator` makes sense with the option to override using an optional argument to `with_collection_parameter`

It would be great to support all of these:


```ruby
class ProductComponent < ViewComponent::Base
  # don't blow up because the :product_iterator isn't part of the constructor
  def initialize(product:)
    @product = product
  end
end

class ProductComponent < ViewComponent::Base
  # use the name of the component to automate the item and the iterator
  def initialize(product:, product_iterator:)
    @product = product
    @index = product_iterator.index
  end
end

class ProductComponent < ViewComponent::Base
  with_collection_parameter :item # defaults the iteration_parameter to :item_iterator. Don't blow up if `:item_iterator` isn't a constructor param
  
  def initialize(item:, item_iterator:)
    @item = item
    @index = item_iterator.index
  end
end

class ProductComponent < ViewComponent::Base
  with_collection_parameter :item, :foo_bar

  def initialize(item:, foo_bar:)
    @item = item
    @index  = foo_bar.index
  end
end
```

franks921:
Thanks for all the input @jonspalmer, I will update when I have a free moment again 👍

lib/view_component/collection_iteration.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module ViewComponent # :nodoc:
+  class CollectionIteration
+
+    attr_reader :size
```
tclem, line 6: Might as well bring over the comment for this too.

lib/view_component/collection_iteration.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+module ViewComponent # :nodoc:
```
tclem, line 3: Worth a comment that this is the same as (copied from) [`PartialIteration`](https://github.com/rails/rails/blob/ab18f60841c11427948be444ee5ff0008fd4c3fb/actionview/lib/action_view/renderer/collection_renderer.rb#L6)?

jonspalmer:
Thinking out loud but I wonder if we can do a little better than Rails. While the convenience methods for first and last are nice to have I would imagine most use cases just care about the index. Can we support this form for those cases:

```ruby
class ProductComponent < ViewComponent::Base
  def initialize(product:, product_index:)
    @product = product
    @index = product_index
  end
end
```

franks921:
> The only major thing I'd love to discuss and hear some feedback on is if this should use `ActionView::PartialIteration` directly or have it's own identical class. I guess since `view_component` only depends on `activesupport`, the way it's implemented in this PR right now makes the most sense. @joelhawksley do you have thoughts here?

I decided to change this to be inline with your rename and not depend on `ActionView` :) I will also add some comments to that class.

franks921:
> Thinking out loud but I wonder if we can do a little better than Rails. While the convenience methods for first and last are nice to have I would imagine most use cases just care about the index. Can we support this form for those cases:

@jonspalmer is your thinking that we allow developers to define their own `first?` and `last?` methods if they need? At the very least we will need to return the index of the current iteration and the size of the entire collection to make this possible.

```.rb
class ProductComponent < ViewComponent::Base
  def initialize(product:, product_index:, product_count:)
    @product = product
    @index = product_index
    @count = product_count
  end
end
```

jonspalmer:
@franks921  I perhaps need more context on the actual use cases. I could image situations where you just need the index. For example to include in input names or to be able to generate unique DOM id inside the loop. You wouldn't need the count for that. 

So I was suggesting a simple solution for that use case. If you need the count I'd propose that you should use the iterator form vs us passing index and count parameters.

joelhawksley:
Wow! What a wonderful discussion. I agree with the direction we're heading here. Thanks for hashing this out on a PR!

> The only major thing I'd love to discuss and hear some feedback on is if this should use `ActionView::PartialIteration` directly or have it's own identical class. I guess since `view_component` only depends on `activesupport`, the way it's implemented in this PR right now makes the most sense. @joelhawksley do you have thoughts here?

I'd prefer to just use the class from Rails, as it will force us to default to consistency with Rails APIs. 

> Thinking out loud but I wonder if we can do a little better than Rails. While the convenience methods for first and last are nice to have I would imagine most use cases just care about the index. 

Not being familiar with the use of this functionality, I'm cautious of straying from the Rails APIs too much. If we're going to implement a pared-down solution, I'd want to get the perspective of someone who uses the existing partial iterator in their application to see if that simpler solution would still work for their use cases.

In my opinion, we should design our APIs to match Rails as much as possible, only deviating when we can make strong, evidence-based argument to do so. If we truly feel the Rails implementation isn't ideal, it might be worth looking into making those changes in Rails itself!

I don't feel the evidence is present/strong enough here to warrant deviating, but I am more than happy to be shown otherwise ❤️

franks921:
So I made a quick [commit](https://github.com/github/view_component/pull/306/commits/c288e1c233bb4c6c82ded40476003855e46e9780) that use another automagic method that Rails uses in partial [collection rendering](https://guides.rubyonrails.org/layouts_and_rendering.html#local-variables) `item_counter`. _Documented in the blue block_

To you point @jonspalmer I've only ever used `item_counter` and never used `partial_name_iteration.index`, so maybe this is the happy medium?

A few things to note:

* The entire iterator class is not being sent to the component, just `_counter` (previously `_index`).
* Total count can be passed in if needed (an example of that is in the test)
* Obviously  work in progress with failing specs and stray code 😛

franks921:
Added [another commit](https://github.com/github/view_component/pull/306/commits/70ad6d93d6e9b3723942562d1274e359c5667a27) that tidies things up a bit.

lib/view_component/collection.rb
```diff
@@ -29,10 +29,22 @@ def collection_variable(object)
...
     def component_options(item, iterator)
-      as = @component.collection_parameter_name
+      @options.merge!(collection_parameter_name => item)
+      @options.merge!(collection_counter_parameter_name => iterator.index) if counter_argument_present?
+      @options
+    end
 
-      @options.merge(as => item, "#{as}_counter": iterator.index)
+    def counter_argument_present?
+      !!@component.instance_method(:initialize).parameters.dup.map(&:second).include?(collection_counter_parameter_name)
```
jonspalmer, line 49: Love this approach.

Nits to pick:

* Do we need the "!!" `include?` always returns a boolean not nil right?
* I don't think we need dup. given we map
* should we memoize all these properties?

test/app/views/integration_examples/products.html.erb
```diff
@@ -1,4 +1,4 @@
 <h1>Products for sale</h1>
 <ul>
-  <%= render(ProductComponent.with_collection(@products, notice: "Today only")) %>
+  <%= render(ProductComponent.with_collection(@products, notice: "Today only", count: @products.count)) %>
```
jonspalmer, line 3: I like this. If you want the count you can do it yourself. We have to help to make the counter line up.

lib/view_component/collection.rb
```diff
@@ -26,5 +28,25 @@ def collection_variable(object)
...
+      "#{collection_parameter_name}_counter".to_sym
+    end
+
+    def component_options(item, iterator)
+      @options[collection_parameter_name]         = item
+      @options[collection_counter_parameter_name] = iterator.index if counter_argument_present?
+      # @options.merge!(:"#{collection_parameter_name}" => item)
+      # @options.merge!(:"#{collection_counter_parameter_name}" => iterator.index) if counter_argument_present?
+      @options
+    end
```
jonspalmer, line 46: I don't think we want to change `@options` here. Lets leave it as set.

```suggestion
  def component_options(item, iterator)
    item_options = { collection_parameter_name => item }
    item_options[collection_counter_parameter_name] = iterator.index if counter_argument_present?
    @options.merge(item_options)
  end
```

lib/view_component/collection.rb
franks921: Yeah, I think it won't hurt to memoize and you are correct about `include?` always returning a boolean.

franks921:
I pushed another commit.

I just want to point out that since we're using a counter now and not an index, the index is getting `1` added to it  when merged into `@options`,

```
item_options[collection_counter_parameter_name] = iterator.index + 1 if counter_argument_present?
```

lib/view_component/collection.rb
```diff
@@ -1,22 +1,24 @@
...
     end
 
     private
 
     def initialize(component, object, **options)
-      @component = component
+      @component  = component
       @collection = collection_variable(object || [])
-      @options = options
+      @options    = options
```
jonspalmer, line 21: Can we revert whitespace changes on lines you didn't touch. It just confused the git history and adds to confusion resolving merge conflicts etc.

jonspalmer, line 21: Can we revert whitespace changes on lines you didn't touch. It just confused the git history and adds to problems resolving merge conflicts etc.

lib/view_component/collection.rb
```diff
@@ -26,5 +28,24 @@ def collection_variable(object)
...
+
+    def component_options(item, iterator)
+      item_options = { collection_parameter_name => item }
+      item_options[collection_counter_parameter_name] = iterator.index + 1 if counter_argument_present?
+
+      @options.merge(item_options)
+    end
+
+    def counter_argument_present?
+      @_counter_argument_present ||= @component.instance_method(:initialize).parameters.map(&:second).include?(collection_counter_parameter_name)
```
jonspalmer, line 48:
```suggestion
      @counter_argument_present ||= @component.instance_method(:initialize).parameters.map(&:second).include?(collection_counter_parameter_name)
```

lib/view_component/collection.rb
franks921: My bad... OCD force of habit 😊 Sorting a PR out now.

franks921:
Done @jonspalmer. 

I had to fix a few style issues in subsequent commits, but here is the commit where I added new tests 6906d8e39bfa229de3eb0338db18556a458bdf85

lib/view_component/collection.rb
```diff
@@ -26,5 +28,24 @@ def collection_variable(object)
...
+
+    def component_options(item, iterator)
+      item_options = { collection_parameter_name => item }
+      item_options[collection_counter_parameter_name] = iterator.index + 1 if counter_argument_present?
+
+      @options.merge(item_options)
+    end
+
+    def counter_argument_present?
+      @counter_argument_present ||= @component.instance_method(:initialize).parameters.map(&:second).include?(collection_counter_parameter_name)
```
joelhawksley, line 48: Why does it matter if the argument is `second`? What happens if we have many arguments to the component? 

Generally, Ruby kwargs should be not be order-dependent. I'd like to avoid introducing such a dependency here if possible.

lib/view_component/collection.rb
```diff
@@ -26,5 +28,24 @@ def collection_variable(object)
...
+    end
+
+    def component_options(item, iterator)
+      item_options = { collection_parameter_name => item }
+      item_options[collection_counter_parameter_name] = iterator.index + 1 if counter_argument_present?
+
+      @options.merge(item_options)
+    end
+
+    def counter_argument_present?
```
joelhawksley, line 47: Can we set this value at compile time instead of at run time?

test/app/components/photo_component.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class PhotoComponent < ViewComponent::Base
```
joelhawksley, line 3: Might we name this to correspond with the feature it is meant to exercise? The actual function / content of the component is excellent 👌 

```suggestion
class CollectionCounterComponent < ViewComponent::Base
```

test/app/components/product_component.rb
```diff
@@ -1,8 +1,10 @@
 # frozen_string_literal: true
 
 class ProductComponent < ViewComponent::Base
-  def initialize(product:, notice:)
+  def initialize(product:, notice:, count: nil, product_counter: nil)
     @product = product
-    @notice = notice
+    @notice  = notice
+    @count   = count
```
joelhawksley, line 7: What is `@count` used for in this example? It doesn't appear to be used in the template.

test/app/views/integration_examples/products.html.erb
joelhawksley: Is this API present in Rails?

lib/view_component/collection.rb
franks921: Hi @joelhawksley 

This is not with the order fo the arguments, it's because `instance_method(:initialize).parameters` returns a nested array and the second array item is the one that holds the argument name. 

Here is an example using my own code
```
$ Atoms::AvatarComponent.instance_method(:initialize).parameters

[[:keyreq, :user], [:key, :size], [:key, :glyph], [:key, :color], [:key, :html]]
```

test/app/components/product_component.rb
franks921: Sorry, that's some stray code. Removing it

lib/view_component/collection.rb
franks921: I am not certain what you mean with this comment @joelhawksley :/

lib/view_component/collection.rb
joelhawksley: 🤦 I I misread the code! Thank you for clarifying ❤️

lib/view_component/collection.rb
joelhawksley: Since the method signature of a component is static, we could set the result of this method as part of `compile` (in `Base`) so that we don't have to do this work at runtime.

franks921: I hope this is what you mean @joelhawksley 1c168aa737bfbdcd357777e7bce9ebefbb2ef3c8

franks921: I think I need to move it into the `compile` method now that I'm reading the code a bit more?

franks921: Here is attempt 2 7c0676b87e063fe214a2605cb41bc708ea7778ea

lib/view_component/base.rb
```diff
@@ -241,6 +241,14 @@ def collection_parameter_name
         (@with_collection_parameter || name.demodulize.underscore.chomp("_component")).to_sym
       end
 
+      def collection_counter_parameter_name
```
joelhawksley: I _believe_ this should be something like `define_singleton_method`, like above:

https://github.com/github/view_component/pull/306/files/302defcc84a73ac0b55f9a6b66f8b6384f6703c7..1c168aa737bfbdcd357777e7bce9ebefbb2ef3c8#diff-08fe1c0f29a6308ad005c21f4adcfd64R186

(My apologies for not sharing this example case in my earlier comment ❤️ )

franks921: No need to apologise 😄 

I think I committed my second attempt as you were writing this comment. I already pushed a fix after I reread everything again 👍 Here is the commit 7c0676b87e063fe214a2605cb41bc708ea7778ea

README.md
```diff
@@ -322,6 +322,25 @@ end
 </li>
 ```
 
+`ViewComponent` also makes a counter variable available named after the component name or the name supplied to `with_collection_parameter` followed by `_counter`. To access this variable add it to your constructor as an argument.
```
joelhawksley, line 325: What if we defined this in terms of `with_collection_parameter`

```suggestion
`ViewComponent` defines a counter variable matching the parameter name above, followed by `_counter`. To access this variable, add it to the constructor as an argument:
```

lib/view_component/collection.rb
```diff
@@ -1,13 +1,15 @@
-
 # frozen_string_literal: true
 
 module ViewComponent
   class Collection
+
```
joelhawksley, line 5:
```suggestion
```

test/app/views/integration_examples/products.html.erb
joelhawksley, line 3: I think my comment here might have been lost: do we need `count` here?

test/app/components/product_component.rb
```diff
@@ -1,8 +1,9 @@
 # frozen_string_literal: true
 
 class ProductComponent < ViewComponent::Base
-  def initialize(product:, notice:)
+  def initialize(product:, notice:, count: nil, product_counter: nil)
```
joelhawksley, line 4: Can we remove the `count` argument? It's unused.

franks921:
Added all the suggestions above @joelhawksley

README.md
```diff
@@ -322,6 +322,25 @@ end
 </li>
 ```
 
+`ViewComponent` defines a counter variable matching the parameter name above, followed by `_counter`. To access this variable, add it to the constructor as an argument:
```
joelhawksley, line 325:
```suggestion
`ViewComponent` defines a counter variable matching the parameter name above, followed by `_counter`. To access this variable, add it to`initialize` as an argument:
```

franks921:
You're welcome! And thanks @joelhawksley, @jonspalmer and @tclem for your patience 🙌

Issue #178, opened 2020-01-11
Overriding default previews layout


AlecRust:
I have a hand-rolled styleguide of `actionview-component` components that I'm looking to migrate to  the built-in component previews introduced in #96.

I want to utilise [this gem's `ComponentsController`](https://github.com/github/actionview-component/blob/master/lib/railties/lib/rails/components_controller.rb) rather than maintain my own, but I'd like to override the layout for the index page at `localhost:3000/rails/components`. This is the layout in question:

https://github.com/github/actionview-component/blob/master/lib/railties/lib/rails/templates/rails/components/index.html.erb

Based on description in #96 I expected to be able to add this layout at `test/components/previews/layouts/application.html.erb` but this doesn't seem to get picked up.

Am I missing something? Where can I add this layout file, or do I need to set up a separate route if I want to achieve this?

joelhawksley:
@juanmanuelramallo mind having a look at this?

juanmanuelramallo:
Hey @AlecRust, the template is actually [being used manually in the controller](https://github.com/github/actionview-component/blob/master/lib/railties/lib/rails/components_controller.rb#L17). So if you want to override it I would suggest overriding the index controller action.

```ruby
# lib/overrides/overrides_components_controller.rb

class OverridesComponentsController < Rails::ComponentsController
  def index
    # These first two lines are just being copied over
    @previews = ActionView::Component::Preview.all
    @page_title = "Component Previews"
    render template: "layouts/application" # or any other template you want to use here
  end
end
```

And in order to use this controller you would need to auto load it in development:
```ruby
# config/application.rb
config.autoload_paths << Rails.root.join('lib/overrides')

# or if you want to use this on production
# config.eager_load_paths << Rails.root.join('lib/overrides')
```

And to make the router use this controller (as it's [done here](https://github.com/github/actionview-component/blob/master/lib/action_view/component/railtie.rb#L59))
```ruby
# config/routes.rb
get "/rails/components"       => "overrides_components#index", :internal => true
```

I haven't tested this. So let me know if this works please.

AlecRust:
I'm afraid that doesn't work for me, I get an empty `<body>` at `/rails/components` URL. The logs say this file was rendered:

```
/Users/alec/.rbenv/versions/2.6.5/lib/ruby/gems/2.6.0/gems/actionview-component-1.7.0/lib/railties/lib/rails/templates/rails/components/index.html
```

[...]

PR #311, opened 2020-04-18
Add default layout option in order to use app/views/layouts

jaredcwhite:
### Summary

I was digging a ton into component previews on a new project today, and I found myself wanting to use a custom layout (in my case `component_preview`) for all of the index pages as well as component preview pages. I started by monkey patching Rails::ApplicationController, but then realized I should simply adjust what's going on in ViewComponentsController, which gave me the idea for a configuration option and hence this PR.

### Other Information

This makes no behavioral change at all out-of-the-box. Everything will work as before. The only difference is if that `default_preview_layout` config option is set, the preview indices and components will pull that in rather than use the Rails internal app layout. Of course individual previews can still override the layout, including setting it to nil. (I also ended up fixing an existing test that would always pass even if a layout _was_ rendered.)

Let me know what you think! Thanks.

### DIFF
--- a/app/controllers/view_components_controller.rb
+++ b/app/controllers/view_components_controller.rb
@@ -28,11 +28,11 @@ class ViewComponentsController < Rails::ApplicationController # :nodoc:
       prepend_preview_examples_view_path
       @example_name = File.basename(params[:path])
       @render_args = @preview.render_args(@example_name, params: params.permit!)
-      layout = determine_layout(@render_args[:layout])[:layout]
+      layout = determine_layout(@render_args[:layout], prepend_views: false)[:layout]
       template = @render_args[:template]
       locals = @render_args[:locals]
       opts = {}
-      opts[:layout] = layout if layout.present?
+      opts[:layout] = layout if layout.present? || layout == false
       opts[:locals] = locals if locals.present?
       render template, opts # rubocop:disable GitHub/RailsControllerRenderLiteral
     end
@@ -67,7 +67,7 @@ class ViewComponentsController < Rails::ApplicationController # :nodoc:
   end
 
   # Returns either {} or {layout: value} depending on configuration
-  def determine_layout(layout_override = nil)
+  def determine_layout(layout_override = nil, prepend_views: true)
     return {} unless defined?(Rails.root)
 
     layout_declaration = {}
@@ -78,6 +78,8 @@ class ViewComponentsController < Rails::ApplicationController # :nodoc:
     elsif default_preview_layout.present?
       layout_declaration[:layout] = default_preview_layout
     end
+
+    prepend_application_view_paths if layout_declaration[:layout].present? && prepend_views
 
     layout_declaration
   end
--- a/app/controllers/view_components_controller.rb
+++ b/app/controllers/view_components_controller.rb
@@ -16,18 +16,19 @@ class ViewComponentsController < Rails::ApplicationController # :nodoc:
   def index
     @previews = ViewComponent::Preview.all
     @page_title = "Component Previews"
+    render "view_components/index", **determine_layout
   end
 
   def previews
     if params[:path] == @preview.preview_name
       @page_title = "Component Previews for #{@preview.preview_name}"
-      render "view_components/previews"
+      render "view_components/previews", **determine_layout
     else
       prepend_application_view_paths
       prepend_preview_examples_view_path
       @example_name = File.basename(params[:path])
       @render_args = @preview.render_args(@example_name, params: params.permit!)
-      layout = @render_args[:layout]
+      layout = determine_layout(@render_args[:layout])[:layout]
       template = @render_args[:template]
       locals = @render_args[:locals]
       opts = {}
@@ -38,6 +39,10 @@ class ViewComponentsController < Rails::ApplicationController # :nodoc:
   end
 
   private
+
+  def default_preview_layout # :doc:
+    ViewComponent::Base.default_preview_layout
+  end
 
   def show_previews? # :doc:
     ViewComponent::Base.show_previews
@@ -59,6 +64,22 @@ class ViewComponentsController < Rails::ApplicationController # :nodoc:
     I18n.with_locale(params[:locale] || I18n.default_locale) do
       yield
     end
+  end
+
+  # Returns either {} or {layout: value} depending on configuration
+  def determine_layout(layout_override = nil)
+    return {} unless defined?(Rails.root)
+
+    layout_declaration = {}
+
+    if !layout_override.nil?
+      # Allow component-level override, even if false (thus no layout rendered)
+      layout_declaration[:layout] = layout_override
+    elsif default_preview_layout.present?
+      layout_declaration[:layout] = default_preview_layout
+    end
+
+    layout_declaration
   end
 
   def prepend_application_view_paths
--- a/lib/view_component/previewable.rb
+++ b/lib/view_component/previewable.rb
@@ -7,6 +7,14 @@ module ViewComponent # :nodoc:
     extend ActiveSupport::Concern
 
     included do
+      # Set a custom default preview layout through app configuration:
+      #
+      #     config.view_component.default_preview_layout = "component_preview"
+      #
+      # This affects preview index pages as well as individual component previews
+      #
+      mattr_accessor :default_preview_layout, instance_writer: false
+
       # Set the location of component previews through app configuration:
       #
       #     config.view_component.preview_paths << "#{Rails.root}/lib/component_previews"
--- a/test/view_component/integration_test.rb
+++ b/test/view_component/integration_test.rb
@@ -281,6 +281,7 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     get "/rails/view_components/no_layout/default"
 
     assert_select("div", "hello,world!")
+    refute_includes response.body, "ViewComponent - Test"
   end
 
   test "test preview renders application's layout by default" do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Add `default_preview_layout` configuration option to load custom app/views/layouts.
+
+    *Jared White, Juan Manuel Ramallo*
+
 # 2.17.0
 
 * Slots return stripped HTML, removing leading and trailing whitespace.
--- a/README.md
+++ b/README.md
@@ -674,6 +674,14 @@ class TestComponentPreview < ViewComponent::Preview
 end
 ```
 
+You can also set a custom layout to be used by default for previews as well as the preview index pages via the `default_preview_layout` configuration option:
+
+`config/application.rb`
+```ruby
+# Set the default layout to app/views/layouts/component_preview.html.erb
+config.default_preview_layout = "component_preview"
+```
+
 Preview classes live in `test/components/previews`, which can be configured using the `preview_paths` option:
 
 `config/application.rb`
--- a/coverage/coverage.txt
+++ b/coverage/coverage.txt
@@ -2,10 +2,10 @@
 Incomplete test coverage
 --------------------------------------------------------------------------------
 
-/app/controllers/view_components_controller.rb: 97.67% (missed: 54)
+/app/controllers/view_components_controller.rb: 98.18% (missed: 59)
 /lib/view_component/base.rb: 97.86% (missed: 178,265,293,356)
 /lib/view_component/engine.rb: 94.64% (missed: 22,25,38)
 /lib/view_component/preview.rb: 94.0% (missed: 47,57,107)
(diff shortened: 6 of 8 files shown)


app/controllers/rails/view_components_controller.rb
```diff
@@ -16,23 +16,31 @@ class Rails::ViewComponentsController < Rails::ApplicationController # :nodoc:
   def index
     @previews = ViewComponent::Preview.all
     @page_title = "Component Previews"
+    opts = default_preview_layout.nil? ? {} : { layout: default_preview_layout }
+    prepend_application_view_paths if opts.key?(:layout)
```
jonspalmer, line 20: Can we set the layout explicitly rather than the indirection of `prepend_application_view_paths`.
```suggestion
    opts = default_preview_layout.nil? ? {} : { layout: Rails.root.join("app/views/layouts",  default_preview_layout) }
```

app/controllers/rails/view_components_controller.rb
```diff
@@ -16,23 +16,31 @@ class Rails::ViewComponentsController < Rails::ApplicationController # :nodoc:
...
       prepend_application_view_paths
       @example_name = File.basename(params[:path])
       @render_args = @preview.render_args(@example_name)
       layout = @render_args[:layout]
-      opts = layout.nil? ? {} : { layout: layout }
+      opts = if layout.nil?
+               default_preview_layout.nil? ? {} : { layout: default_preview_layout }
+             else
+               { layout: layout }
+             end
```
jonspalmer, line 43: Functionally this is great but stylistically nested if conditions like this feel a little hard to read. It would be a little clearer as 
```suggestion
      layout_override = layout.presence? || default_preview_layout.presence?
      opts = layout_override.nil? ? {} : { layout: layout_override }
```

Do we have a test case where layout = false and default_preview_layout is set? Important that we're getting that right by checking nil? presence? etc

lib/view_component/previewable.rb
```diff
@@ -7,6 +7,12 @@ module Previewable
     extend ActiveSupport::Concern
 
     included do
+      # Set a custom default preview layout through app configuration:
```
jonspalmer, line 10: Can we add details that we're using this preview layout for the preview index pages too?

test/view_component/integration_test.rb
```diff
@@ -214,12 +215,28 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_select "title", "Component Previews"
   end
 
+  test "test preview index renders custom application layout if configured" do
+    ViewComponent::Base.default_preview_layout = "admin"
+    get "/rails/view_components"
+    ViewComponent::Base.default_preview_layout = nil
```
jonspalmer, line 221: For clarity on the code under test it feels like we're getting to a place where this would be considerably clearer as its own test class with setup and teardown methods that manipulate `default_preview_layout`. As it's written I did a double take a) we repeat this pattern in many tests b) is it important to nil out `default_preview_layout` before asserting or could doing that have side effects on the asserted conditions?


@joelhawksley what are you thoughts on using RSepc? The ability to use nested describe blocks would be perfect for cases like this where you would use before/after blocks to do common setup and teardown tasks.

test/view_component/integration_test.rb
```diff
@@ -214,12 +215,28 @@ class IntegrationTest < ActionDispatch::IntegrationTest
...
+
+    assert_select "title", "ViewComponent - Admin - Test"
```
jonspalmer, line 223: Is it time for us to write a helper method for this pattern that makes it clearer we're asserting the right layout? That we're proxying the laying being applied by asserting the page title is sort of a weird detail. 
```suggestion
    assert_admin_layout
```

?

app/controllers/rails/view_components_controller.rb
jonspalmer: Similar comment as above about explicitly setting the layout rather than using `prepend_application_view_paths`

test/view_component/integration_test.rb
```diff
@@ -200,6 +200,7 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     get "/rails/view_components/no_layout/default"
 
     assert_select("div", "hello,world!")
+    refute_includes response.body, "ViewComponent - Test"
```
jonspalmer, line 203: Let's add a new test case for no_layout where we set `default_preview_layout` and make sure that there is still no layout.

app/controllers/rails/view_components_controller.rb
jaredcwhite: @jonspalmer Well prepending the app's view path allows the layout to use partials from the app (shared/head, etc.), so that needs to be preserved.

app/controllers/rails/view_components_controller.rb
jaredcwhite: > Functionally this is great but stylistically nested if conditions like this feel a little hard to read. It would be a little clearer as

I'm fine with cleaning that up a bit, but that syntax change didn't pass the a couple tests.

> Do we have a test case where layout = false and default_preview_layout is set? Important that we're getting that right by checking nil? presence? etc

Good call, adding a test to check false and switching logic from .nil? to .blank? did the trick.

test/view_component/integration_test.rb
jaredcwhite: No strong opinions here about the test suite, so I'll happily defer to the rest of y'all.

app/controllers/rails/view_components_controller.rb
jonspalmer: Got it. And I'm guessing that sadly resolving those partial's isn't local to the layout that is picked. It relies of view paths?

jaredcwhite: Let me see if I can refactor the layout hash stuff into a separate method so the individual actions in this controller are simpler.

jaredcwhite:
All right, I refactored the layout logic into its own method and merged the prepend views functionality into that as well. Still slightly non-obvious why the conditional is set up that way, but hopefully added comments make it clearer what to expect. Also added a test for the `layout false` scenario in a component.

test/view_component/integration_test.rb
juanmanuelramallo: IMHO RSpec would be a great addition to the development environment of the gem @jonspalmer

juanmanuelramallo:
This PR will close https://github.com/github/view_component/issues/178

test/view_component/integration_test.rb
joelhawksley: How about we start by moving these tests to their own file?

I’d like to _try_ to stick with our current testing framework as it is consistent with Rails 😄

joelhawksley:
@jonspalmer might you be able to revisit and review this PR?

test/view_component/integration_test.rb
jaredcwhite: @jonspalmer @joelhawksley If you need me to go rejigger the tests, just let me know

juanmanuelramallo: In order not to disrupt the previous config I would suggest to set the value back instead of nil, we can go even a step further and create a helper method for this in the `test/test_helper.rb` file

```ruby
# test/test_helper.rb

...

def with_default_preview_layout(layout)
  old_value = ViewComponent::Base.default_preview_layout
  ViewComponent::Base.default_preview_layout = layout
  yield
  ViewComponent::Base.default_preview_layout = old_value
end
```
Then in our tests we can just use it like:
```ruby
test "test preview index renders custom application layout if configured" do
  with_default_preview_layout("admin") do
    get "/rails/view_components"
    assert_select "title", "ViewComponent - Admin - Test"
  end
end
```

juanmanuelramallo:
I think we need to rebase this PR with master given that the view_components_controller was updated to use the templates from `app/views`. @jaredcwhite let me know if you need a hand with that

jaredcwhite:
@juanmanuelramallo Thanks for the suggestion regarding that test helper. I incorporated it into the tests and also moved them over to a separate file, after rebasing from master.

app/controllers/view_components_controller.rb
```diff
@@ -16,24 +16,26 @@ class ViewComponentsController < Rails::ApplicationController # :nodoc:
   def index
     @previews = ViewComponent::Preview.all
     @page_title = "Component Previews"
+    render "view_components/index", **determine_layout
```
juanmanuelramallo, line 19: Will this work?
```suggestion
    render **determine_layout
```

app/controllers/view_components_controller.rb
```diff
@@ -56,7 +58,26 @@ def set_locale
...
+    layout_declaration = {}
+
+    if !layout_override.nil?
+      # Allow component-level override, even if false (thus no layout rendered)
+      layout_declaration[:layout] = layout_override
+    elsif default_preview_layout.present?
+      layout_declaration[:layout] = default_preview_layout
+    end
+
+    if layout_declaration.key?(:layout) || !!always_use_app_views
```
juanmanuelramallo, line 77: IMHO no need to double negate, it's just a boolean.

```suggestion
    if layout_declaration.key?(:layout) || always_use_app_views
```

test/view_component/custom_layouts_integration_test.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  test "test preview renders without layout even if default layout is configured" do
+    with_default_preview_layout("admin") do
+      get "/rails/view_components/no_layout/default"
+      assert_select("div", "hello,world!")
+      refute_includes response.body, "ViewComponent - Admin - Test"
+    end
+  end
+end
```
juanmanuelramallo, line 44: Maybe we can name this file the same as the config attribute?

`test/view_component/default_preview_layout_integration_test.rb`

test/view_component/custom_layouts_integration_test.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class CustomLayoutsIntegrationTest < ActionDispatch::IntegrationTest
+  test "test preview index renders custom application layout if configured" do
```
juanmanuelramallo, line 6: Maybe we can avoid rewriting "test" in each test case?

```suggestion
  test "preview index renders custom application layout if configured" do
```

jaredcwhite:
Good suggestions @juanmanuelramallo — I just pushed up those changes. Something weird happened with the CI though, nothing ran

juanmanuelramallo:
Weird, can you try rebasing and pushing so that the CI triggers again? I don't think it's related to your changes

jaredcwhite:
@juanmanuelramallo Working better now, but it looks like that one `render` change on the `index` action is running into a Rubocop rule:

`GitHub/RailsControllerRenderLiteral: render must be used with a string literal`

Should we disable it, or put back in the view name string?

juanmanuelramallo:
Oh man, yeah maybe put back the template name no? I'd rather do that than disabling rubocop for that line

jaredcwhite:
Works! ✅

jaredcwhite:
Hi all, just wanted to pop my head back in to see what the status might be for this feature. I haven't followed other PRs too closely lately so I'm not sure if this has been superseded by some other approach or if it's still relevant. Thanks =)

juanmanuelramallo:
Seems relevant to me, but same as you, haven't been following up other PRs closely. Better call @joelhawksley

test/view_component/integration_test.rb
joelhawksley: @jaredcwhite thoughts on @jonspalmer's comment here?

jaredcwhite: @joelhawksley I think I did that in a subsequent commit, but I'll double-check

jaredcwhite:
Sounds good. 👍 I hope to get to it this week.

jaredcwhite:
OK, at long last I was able to rebase and add some info to the README, so I think this is good to go now. Incidentally, there had been a regression on master regarding the context of `layout` set to `false` for a component which wasn't actually getting caught by a test (aka the test was passing even though a layout was in fact rendered), so this PR fixes that and makes sure the test catches any future regression.

app/controllers/view_components_controller.rb
```diff
@@ -28,11 +28,11 @@ def previews
       prepend_preview_examples_view_path
       @example_name = File.basename(params[:path])
       @render_args = @preview.render_args(@example_name, params: params.permit!)
-      layout = determine_layout(@render_args[:layout])[:layout]
+      layout = determine_layout(@render_args[:layout], prepend_views: false)[:layout]
       template = @render_args[:template]
       locals = @render_args[:locals]
       opts = {}
-      opts[:layout] = layout if layout.present?
+      opts[:layout] = layout if layout.present? || layout == false
```
juanmanuelramallo, line 35: Nice one!

PR #316, opened 2020-04-20
Undefined method `counter_argument_present?` for collection

franks921:
### Summary

After installing 2.5.0 that I contributed to, I found after adding the counter methods to the `compile` method no collection rendering in my project was working. I'm not sure why the tests did not catch this, but reverting the change fixes it.

### Other Information

```slim
ul
  = render Collections::UserListComponent.with_collection(online_doctors)
```

```ruby
# app/components/collections/user_list_component.rb
class Collections::UserListComponent < ApplicationComponent

  with_collection_parameter :user

  attr_reader \
    :style,
    :target,
    :user

  def initialize(user:, style: nil, target: nil)
    @user   = user
    @style  = style
    @target = target
  end

end
```

```slim
/ app/components/collections/user_list_component.html.slim
li*{ data: { 'doctor-id': user.id, 'doctor-initials': user.initials, target: target }.compact }
  = render Molecules::ProfileComponent.new(user: user, style: style)
```


jonspalmer:
What's the spec we are missing that covers this change?

joelhawksley:
@franks921 thanks for raising this. Let's get a failing test and see if we can find a solution that preserves the compilation step ❤️

franks921:
It definitely has to do with moving those methods out of `collection.rb` and into `Base`. I don't have any automated testing for this but this is what I've done on my side that confirms this.

### Step 1
```
# = render Collections::UserListComponent.new(user: online_doctors.first)
= render Collections::UserListComponent.with_collection(online_doctors)
```
This raises an error `Undefined method `counter_argument_present?` for Collections::UserListComponent`.

### Step 2
```
= render Collections::UserListComponent.new(user: online_doctors.first)
# = render Collections::UserListComponent.with_collection(online_doctors)
```
This renders the component without any errors.

### Step 3
```
# = render Collections::UserListComponent.new(user: online_doctors.first)
= render Collections::UserListComponent.with_collection(online_doctors)
```
This renders the component collection without any errors.

I don't even know where to start writing a test for this as the tests we wrote are working fine, even my own specs in rspec using `render_inline` work, I'm really stumped at this point 😑

jonspalmer:
@franks921 Can you clarify what the difference between Case 1 and Case 3 i your examples?

franks921:
Sorry @jonspalmer that was the process I followed.

I commented out the single render and had the collection render uncommented, loaded the page and it broke, then I uncommented the single render and commented out the collection render reloaded the page and it worked on the single render and then lastly I commented out the single render and uncommented the collection render reloaded again and the collection render worked fine.

jonspalmer:
BTW My guess is that the bug is we need to call `compile!` at the start of `with_collection` and that the tests we have pass because we don't have a test case where `with_collection` is called before a regular render.

franks921:
That's exactly what I am checking now @jonspalmer, I will let you know once I'm done trying that

franks921:
And you were 100% on the money @jonspalmer. It is working perfectly now 🎉

jonspalmer:
Awesome. Let's still force a test case that fails as it's important we don't break this going forward. 

I would think we can add a test that checks a.component is not compiled and the immediately calls `with_collectuon` on it.

franks921:
@jonspalmer I tried everything I could think of to get a failing spec, but I just can't force it to fail or fail like it did in my project

franks921:
I've added a test for it, but it passes regardless if `compile!` is called in `with_collection` or not. I also renamed two files I missed in the last PR

761a0ff

franks921:
@joelhawksley can I close this PR then?

joelhawksley:
Resolved by https://github.com/github/view_component/pull/317. Thank you for filing this @franks921 ❤️

PR #317, opened 2020-04-21
Compile component before rendering collection

rainerborene:
After running our test suite today with the latest version of `view_component` I got the following message:

> ActionView::Template::Error: undefined method `counter_argument_present?' for Courses::Card:Class

This is an edge case. If you're running a test or development environment and never rendered that specific component individually, then you will face the same problem. So, this pull request fixes this problem. In other words, `Courses::Card` is only rendered through collections, and isn't compiled at this point yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # master
 
+# 2.5.1
+
+* Compile component before rendering collection.
+
+    Rainer Borene
+
 # v2.5.0
 
 * Add counter variables when rendering collections.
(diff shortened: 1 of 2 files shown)


joelhawksley:
@rainerborene thanks for the fix!

I think I prefer the `compile!` call site you selected compared to the proposed solution in https://github.com/github/view_component/pull/316.

Any chance you could provide a failing test? I know this is a tricky one to test for, so I understand if it's not possible.

I'll merge this in a couple hours if I don't hear back ❤️

jonspalmer:
I like this better too. The problem with the test is we have eager_load : true in the dummy app so you never hit this problem. I think we had a testing problem like this before "way back" in the day. I would guess we can test it "manually". I think the trick was to set eager_load to false, run the test suite , proof existing tests fail etc.

Issue #285, opened 2020-04-03
Pass an argument to content area


rosendi:
I have a component with the `form_for` helper:
```
class EditorComponent < ViewComponent::Base
  with_content_areas :form
end
```

And I want to pass the `f` into the form area.

```
<div class="editor">
  <div class="editor__panels">etc</div>

  <%= form_for ... do |f| %>
    <%= form(f) <!--- SOMETHING LIKE %>
  <% end %>
</div>
```

Then render it like:

```
<%= render(Editr.new) do |component| %>
  <% component.with(:form) do |f| <------- HERE %>
      <%= f.input_field ... %>
    <% end %>
<% end %>
```

How should I handle this case?

thomasbrus:
@rosendi I think it might be better to render the form itself in the view and perhaps extract some parts to the component, for example via a method `def build_form_url(opts ={})` and then call `form_for(url: component.build_form_url)` etc in the view. It's also an option to have a method `def form(&block)` on the component, you can then call `component.form do |f|` in the view. However you might run into issues with duplicate output in the HTML as described here https://github.com/github/view_component/pull/240.

rosendi:
@thomasbrus yep, currently I pass a complete form to `editor`.

jonspalmer:
@rosendi Could you perhaps provide a larger example of why you need this solution. In particular its not clear to me why you need to call `form_for` in the component's view rather than inside `<% component.with(:form)` when the component is used.

bbugh:
I think this is going to be a common requirement for reusable object-agnostic components. I have an example of this use case for a `NavComponent` in my feedback comment here: 

https://github.com/github/view_component/issues/15#issuecomment-623108834

rosendi:
@bbugh yep, I think VC should cover both cases.

rosendi:
@jonspalmer the original issue was to provide a custom form builder for the inplace editing component. The first iteration of InplaceEditingComponent can only update a single attribute:

```
class InplaceEditingComponent < ApplicationComponent
  attr_reader :model, :field, :placeholder, :hidden_fields, :form_options, :input_options 

  def initialize(model:, field:, placeholder:, hidden_fields: [], form_options: {}, input_options: {})
    @model = model
    @field = field
    @placeholder = placeholder
    @hidden_fields = hidden_fields
    @form_options = form_options
    @input_options = input_options
  end

  def value
    @value ||= Array.wrap(model).last.public_send(field)
  end
end
```

```
[...]

PR #322, opened 2020-04-30
sample solution for supplying attributes to content areas

andrewmcodes:
### Summary

Example solution for supplying attributes to content areas

Related Issues #285

### Other Information

Results of pairing with @joelhawksley

### DIFF
--- a/test/view_component/view_component_test.rb
+++ b/test/view_component/view_component_test.rb
@@ -506,7 +506,7 @@ class ViewComponentTest < ViewComponent::TestCase
 
   def test_content_area_named_attributes
     render_inline(CardComponent.new) do |component|
-      component.with(:header, classes: "card-header mt-4", data: {controller: "card-header-controller"}) { "Card Header" }
+      component.with(:header, classes: "card-header mt-4") { "Card Header" }
       component.with(:body, classes: "bg-primary") do
         "Have a great day."
       end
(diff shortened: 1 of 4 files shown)


andrewmcodes:
ping @jonspalmer

test/view_component/view_component_test.rb
```diff
@@ -504,6 +504,22 @@ def test_render_single_item_from_collection
     assert_selector("p", text: "On sale", count: 1)
   end
 
+  def test_content_area_named_attributes
+    render_inline(CardComponent.new) do |component|
+      component.with(:header, classes: "card-header mt-4", data: {controller: "card-header-controller"}) { "Card Header" }
```
muan, line 509:
```suggestion
      component.with(:header, classes: "card-header mt-4") { "Card Header" }
```

jonspalmer:
I'm a little confused about how this implementation solves the original problem in #285. This doesn't look like it allows arguments to be passed into the content areas. Instead it's provide a way to decorate those areas with other static properties. I don't think that's the same thing.

joelhawksley:
Thanks for doing this spike @andrewmcodes! We're going pursue a different approach to this change ❤️

andrewmcodes:
@joelhawksley sounds good! Sorry this change slipped off my radar due to some personal events but I am excited to see what everyone came up with! Thanks for working on this with me!

PR #323, opened 2020-05-02
Allow arguments to be passed to content areas #285

jonspalmer:
### Summary

Alternative solution than #322  for passing arguments to content areas as requested in #285 

### Other Information

I'm not 100% convinced we need this complexity. I'd love more example use cases. But assuming an agreed need I think this is the way we'd do it

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -118,11 +118,11 @@ module ViewComponent
         raise ArgumentError.new "Unknown content_area '#{area}' - expected one of '#{content_areas}'"
       end
 
-      if block_given?
-        content = view_context.capture(&block)
+      define_singleton_method(area) do |*args|
+        block = ->() { content } unless block_given?
+        view_context.capture(*args, &block)
       end
 
-      instance_variable_set("@#{area}".to_sym, content)
       nil
     end
 
@@ -236,7 +236,7 @@ module ViewComponent
         if areas.include?(:content)
           raise ArgumentError.new ":content is a reserved content area name. Please use another name, such as ':body'"
         end
-        attr_reader *areas
+        attr_reader *areas # provide a default implementation for all areas. Will be redifined by calls to .with.
         self.content_areas = areas
       end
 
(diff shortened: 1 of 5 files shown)


jonspalmer:
ping @andrewmcodes @rosendi @thomasbrus

test/view_component/view_component_test.rb
```diff
@@ -176,17 +176,6 @@ def test_renders_content_areas_template_replaces_content
...
-  def test_renders_content_areas_template_can_wrap_render_arguments
-    render_inline(ContentAreasComponent.new(title: "Hello!", footer: "Bye!")) do |component|
-      component.with(:title) { "<strong>#{component.title}</strong>".html_safe }
-      component.with(:body) { "Have a nice day." }
-    end
-
-    assert_selector(".title strong", text: "Hello!")
-    assert_selector(".body", text: "Have a nice day.")
-    assert_selector(".footer", text: "Bye!")
-  end
```
jonspalmer, line 188: With this change "wrapping" the render arguments results in a stack overflow. Its a weird use case anyway and not one that I'm that bothered about supporting. Seems like anyone rendering the component can choose one mechanism or the other.

lib/view_component/base.rb
```diff
@@ -118,11 +118,11 @@ def with(area, content = nil, &block)
         raise ArgumentError.new "Unknown content_area '#{area}' - expected one of '#{content_areas}'"
       end
 
-      if block_given?
-        content = view_context.capture(&block)
+      define_singleton_method(area) do |*args|
+        block = ->() { content } unless block_given?
+        view_context.capture(*args, &block)
```
jonspalmer, line 123: This is the key change. Rather than capturing the block immediately and storing the result in the attr_reader instance variable we redefine the method and capture the block when its called passing the arguments.

joelhawksley:
@jonspalmer thanks for putting this together. I vote that we let https://github.com/github/view_component/issues/325 play out before moving forward with any further changes to content areas ❤️

jonspalmer:
@joelhawksley Agreed. We're getting there on #325 I think the answer will be clear out of that

joelhawksley:
Closing as stale, feel free to reopen if you'd like ❤️

PR #327, opened 2020-05-04
Raise error when initializer omits with_collection_parameter

joelhawksley:
### Summary

It's currently possible to declare `with_collection_parameter`
without using the parameter in the initializer. I think we should
raise an error in this case so that developers can be aware of 
the potential bug introduced.

This change adds an error when the parameter set by
`with_collection_parameter` is not declared in the initializer.

cc @tclem


lib/view_component/base.rb
```diff
@@ -296,6 +301,20 @@ def templates
...
       end
 
+      def parameter_errors
+        @parameter_errors ||=
+          begin
+            errors = []
+
+            # If initializer omits with_collection_parameter
+            if @with_collection_parameter && !instance_method(:initialize).parameters.map(&:last).include?(@with_collection_parameter)
+              errors << "Collection parameter #{@with_collection_parameter} not used in component initializer."
```
jonspalmer, line 311: Nit Pik The use of the instance variable here feels a little icky to me. I wonder if we can't put an attr_reader or the like around it.

joelhawksley:
> I suspect we have to trap that error at the time that `with_collection` is called vs at compile time.

Correct. We don't know until runtime whether the default collection parameter is needed or not. I'll look into this approach ❤️

jonspalmer:
Compile time would be nice but I'd be ok with one piece of code at run time catching both cases. High level I think it's most important we give helpful error messages allowing the mistake to be fixed. I'd prioritize that over failing as early as possible.

joelhawksley:
@jonspalmer @tclem thanks for the feedback. 

I've updated my implementation to add a runtime check for the default iterator key if the component does not declare `with_collection_parameter`. 

I opted to leave the compile-time check in, as I'd like to set the example of doing _as much_ work as possible at compile time.

lib/view_component/collection.rb
```diff
@@ -35,5 +36,18 @@ def component_options(item, iterator)
 
       @options.merge(item_options)
     end
+
+    def ensure_initializer_accepts_iterator
```
jonspalmer, line 40: Rails convention might be to put a `! on the end of this method given that it will raise

lib/view_component/collection.rb
```diff
@@ -35,5 +36,18 @@ def component_options(item, iterator)
...
+      # If the component does not set a custom collection parameter,
+      # make sure the default parameter is accepted by the
+      # component initializer.
+      if !@component.with_collection_parameter_attr &&
+        !@component.instance_method(:initialize).parameters.map(&:last).include?(@component.collection_parameter_name)
+        raise ArgumentError.new(
+          "#{@component} initializer must accept " \
+          "`#{@component.collection_parameter_name}` collection parameter."
+        )
```
jonspalmer, line 49: Can we perhaps DRY up the implementation a little here? Can't we just make this a method on the component

```ruby
collection_param_error = @component.validate_collection_parameter!(default_only: true)
raise ArgumentError.new(collection_param_error) if collection_param_error
```

that way we can avoid the copy-past of the `instance_method(:initialize)` check and the construction of the error string

lib/view_component/base.rb
```diff
@@ -246,11 +248,37 @@ def with_collection_parameter(param)
...
+        )
       end
 
       private
 
+      # We use #with_collection_parameter as a setter,
+      # so this method is suffixed with _attr.
+      def with_collection_parameter_attr
+        @with_collection_parameter
+      end
```
jonspalmer, line 280: would it be better to just have the instance variable be `@collection_parameter` ?

lib/view_component/base.rb
```diff
@@ -169,17 +169,19 @@ def compiled?
...
       end
 
       # Compile templates to instance methods, assuming they haven't been compiled already.
       # We could in theory do this on app boot, at least in production environments.
       # Right now this just compiles the first time the component is rendered.
-      def compile(raise_template_errors: false)
+      def compile(raise_errors: false)
         return if compiled?
 
+        validate_collection_parameter! if raise_errors
```
jonspalmer, line 181: nice

PR #330, opened 2020-05-06
[326] Configurable preview route

### LINKED ISSUE #326: Make previews entry endpoint configurable
### Feature request
Allow users to configure the entry endpoint for previewing components.

### Motivation
Started here https://github.com/github/view_component/pull/324#discussion_r419091598.

Also several gems allow users to mount their views in a given endpoint, for instance Split, Flipper, Sidekiq:

```ruby
# config/routes.rb
mount Sidekiq::Web => '/sidekiq'
```

Although, we may not need to let the users to mount the previews like this, but instead we can just provide a configurable attribute for the base endpoint in the view components config, and use it if the previews are enabled.

juanmanuelramallo:
Resolves #326 

### Summary

- Added a config attribute to allow users to configure the endpoint to be used for previewing components
- Updated views to use a helper path method instead of hardcoded paths
- Updated the engine to use the new configuration attribute
- Removed unused "trim_result" method from test helper
- Added a test helper method to run assertions with a different preview route config
- Updated integration test

### Other Information

I'm not entirely convinced about naming this config `preview_route`. I hear suggestions.


app/controllers/view_components_controller.rb
```diff
@@ -59,4 +61,8 @@ def set_locale
   def prepend_application_view_paths
     prepend_view_path Rails.root.join("app/views") if defined?(Rails.root)
   end
+
+  def url_helpers
+    Rails.application.routes.url_helpers
+  end
```
jonspalmer, line 67: Don't all controllers have url_helpers already? I'm surprised we need this.

app/views/view_components/index.html.erb
```diff
@@ -1,8 +1,8 @@
 <% @previews.each do |preview| %>
-  <h3><%= link_to preview.preview_name.titleize, "/rails/view_components/#{preview.preview_name}" %></h3>
+  <h3><%= link_to preview.preview_name.titleize, preview_view_component_path(preview.preview_name) %></h3>
   <ul>
     <% preview.examples.each do |preview_example| %>
-      <li><%= link_to preview_example, "/rails/view_components/#{preview.preview_name}/#{preview_example}" %></li>
+      <li><%= link_to preview_example, preview_view_component_path("#{preview.preview_name}/#{preview_example}") %></li>
```
jonspalmer, line 5: nice

README.md
```diff
@@ -474,6 +474,15 @@ Preview classes live in `test/components/previews`, which can be configured usin
 config.view_component.preview_path = "#{Rails.root}/lib/component_previews"
 ```
 
+Previews are served from <http://localhost:3000/rails/view_components> by default. If you want to use a different endpoint you can configure the `preview_route` option:
+
+`config/application.rb`
+```ruby
+config.view_component.preview_route = "/previews"
```
jonspalmer, line 481: I think we'd call this `preview_path` ?

app/controllers/view_components_controller.rb
juanmanuelramallo: Surprisingly enough, Rails::ApplicationController (parent class) does not have the url helpers. 

Maybe this is a sign that we should start to decouple from rails internals, like the Rails::ApplicationController and start rolling our own standalone controller (?)

README.md
juanmanuelramallo: We already have that config for enabling users configure where the previews live in the codebase. Should we rename?

jonspalmer: oh right. hmm... I'd say `preview_source_path` and `preview_path` but that's a confusing breaking change. Let's debate the mount idea before we are forced to decide on this.

joelhawksley: I prefer this approach to the mounted engine implementation. I also think `preview_route` is fine 👍

app/controllers/view_components_controller.rb
joelhawksley, line 67: Do we need these helpers for anything else? If we don't, what if we just defined `preview_component_path` directly?

```suggestion
  def preview_component_path
    Rails.application.routes.url_helpers.preview_component_path
  end
```

juanmanuelramallo: Great call! Will update

jonspalmer: I'm still wondering if we do controllers as part of the engine can we get the URL helpers for free event if we go with this configuration mechanism vs mounting the engine?

juanmanuelramallo: As long as our controller inherits from Rails::ApplicationController the URL helpers will be needed

juanmanuelramallo: Hmmm just did some testing and turns out we don't need it, I wonder why I added it in the first place.

README.md
```diff
@@ -474,6 +474,15 @@ Preview classes live in `test/components/previews`, which can be configured usin
 config.view_component.preview_path = "#{Rails.root}/lib/component_previews"
 ```
 
+Previews are served from <http://localhost:3000/rails/view_components> by default. If you want to use a different endpoint you can configure the `preview_route` option:
```
joelhawksley, line 477:
```suggestion
Previews are served from <http://localhost:3000/rails/view_components> by default. To use a different endpoint, set the `preview_route` option:
```

lib/view_component/engine.rb
```diff
@@ -14,6 +14,7 @@ class Engine < Rails::Engine # :nodoc:
 
       if options.show_previews
         options.preview_path ||= defined?(Rails.root) ? "#{Rails.root}/test/components/previews" : nil
+        options.preview_route ||= "/rails/view_components"
```
joelhawksley, line 17:
```suggestion
```

lib/view_component/previewable.rb
```diff
@@ -20,6 +20,14 @@ module Previewable
...
       #
       mattr_accessor :show_previews, instance_writer: false
+
+      # Set the entry route for component previews through app configuration:
+      #
+      #     config.view_component.preview_route = "/previews"
+      #
+      # Defaults to +/rails/view_components+ when `show_previews' is enabled
+      #
+      mattr_accessor :preview_route, instance_writer: false
```
joelhawksley, line 30:
```suggestion
      mattr_accessor :preview_route, instance_writer: false, default: "/rails/view_components"
```

(see https://apidock.com/rails/Module/mattr_accessor)

lib/view_component/engine.rb
```diff
@@ -64,8 +65,8 @@ class Engine < Rails::Engine # :nodoc:
 
       if options.show_previews
         app.routes.prepend do
-          get "/rails/view_components"       => "view_components#index", :internal => true
-          get "/rails/view_components/*path" => "view_components#previews", :internal => true
+          get options.preview_route, to: "view_components#index", as: "preview_view_components", internal: true
+          get "#{options.preview_route}/*path", to: "view_components#previews", as: "preview_view_component", internal: true
```
joelhawksley, line 69:
```suggestion
          get options.preview_route, to: "view_components#index", as: :preview_view_components, internal: true
          get "#{options.preview_route}/*path", to: "view_components#previews", as: :preview_view_component, internal: true
```

lib/view_component/previewable.rb
juanmanuelramallo: Nice

juanmanuelramallo:
Rails 5.0.0 had a different implementation for the default in mattr_accessor it seems https://api.rubyonrails.org/v5.0.0/classes/Module.html#method-i-mattr_accessor 
It receives a block instead of a kwarg

juanmanuelramallo:
@jonspalmer @joelhawksley would you mind giving it another look? I added Joel's suggestions and added two commits, one to fix the engine configuration and the other to make it work on Rails <6.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # master
 
+* Add `config.view_component.preview_route` to set the endpoint for component previews. By default `/rails/view_component` is used.
```
```suggestion
* Add `config.view_component.preview_route` to set the endpoint for component previews. By default `/rails/view_components` is used.
```

Issue #331, opened 2020-05-06
[Design Discussion] Organizing Sidecar assets


jonspalmer:
# Introduction

There has been a lot of discussion on the right way to organize sidecar assets and ruby classes for ViewComponents: https://github.com/github/actionview-component/issues/15#issuecomment-565834219, #67, #168, #283.

We don't have a clean solution yet but we have some ideas

# What do we know?

1. ViewComponents are more than just a subclass of ViewComponent::Base and a view file. It's clear that the community has spoken for a need for sidecar javascript assets - Stimulus controllers, css, i18n yml, and also other Ruby files like Stimulus Reflex.
1. The current directory structure is flat and leads to very fat directories 
1. Namespaces help but don't solve the complete problem - they just subdivide it.
1. Ruby autoloader rules impose restrictions on how we organize things.

# The Problem today

Suppose we have four components: `AlertComponent`, `ButtonComponent`, `FormComponent` and `Form:TextFieldComponent` (note the Ruby namespace on that later - it's going to matter).  Today the directory structure would look like this:

```
app/components

├── alert.css
├── alert.html.erb
├── alert_component.rb.          => AlertComponent
├── alert_controller.js
├── application_component.rb     => ApplicationComponent
├── button.css
├── button.html.erb
├── button_component.rb.         => ButtonComponent
├── button_controller.js  
├── form
│   ├── text_field.css
│   ├── text_field.html.erb
│   ├── text_field_component.rb  => Form::TextFieldComponent
│   └── text_field_controller.js
├── form.css
├── form.html.erb
├── form_component.rb            => Form::FormComponent
└── form_controller.js
```

We can see app/components is already a 'fat' directory and we're only four components in.

# What would the ideal solution look like?

In an ideal world the solution would look much like organizing front end assets where we group each component and its sidecar file in their own directory:

```
app/components
├── alert
│   ├── alert.css
│   ├── alert.html.erb
│   ├── alert_component.rb
│   └── alert_controller.js
├── application_component.rb
├── button
│   ├── button.css
│   ├── button.html.erb
│   ├── button_component.rb
│   └── button_controller.js
└── form
    ├── form.css
    ├── form.html.erb
    ├── form_component.rb
    ├── form_controller.js
    └── text_field
        ├── text_field.css
        ├── text_field.html.erb
        ├── text_field_component.rb
        └── text_field_controller.js
```

# Solutions

## 1. Use Autoloader 

[...]

PR #335, opened 2020-05-08
Supporting sidecar view directory

jonspalmer:
### Summary
Simple implementation of option 2 from #331

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,6 +191,38 @@ class InlineVariantComponent < ViewComponent::Base
 end
 ```
 
+### Sidecar Assets
+
+ViewComponents supports two options for defining view files. 
+
+#### Sidecar view
+
+The simplest option is to place the view next to the Ruby component:
+
+```
+app/components
+├── ...
+├── test_component.rb
+├── test_component.html.erb
+├── ...
+```
+
+#### Sidecar directory
+
+As an alternative, views and other assets can be placed in a sidecar directory with the same name as the component, which can be useful for organizing views alongside other assets like Javascript and CSS.
+
+```
+app/components
+├── ...
+├── test_component.rb
+├── test_component
+|   ├── test_component.css
+|   ├── test_component.html.erb
+|   └── test_component.js
+├── ...
+
+```
+
 ### Conditional Rendering
 
 Components can implement a `#render?` method to be called after initialization to determine if the component should render.
--- a/test/view_component/view_component_test.rb
+++ b/test/view_component/view_component_test.rb
@@ -251,6 +251,12 @@ class ViewComponentTest < ViewComponent::TestCase
     assert_text("hello, world!")
   end
 
+  def test_renders_component_with_sidecar_directory
+    render_inline(SidecarDirectoryComponent.new)
+
+    assert_text("hello, world!")
+  end
+
   def test_renders_component_with_request_context
     render_inline(RequestComponent.new)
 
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -191,6 +191,39 @@ class InlineVariantComponent < ViewComponent::Base
...
+├── ...
+├── test_component.rb
+├── test_component.html.erb
+├── ...
+```
+
+#### Sidecar directory
+
+Alternatively views and other assets can be placed in a sidecar directory with the same name as the component.
+This configuration is particulary useful to organize views along side other assets like javascript and css.
```
joelhawksley, line 213:
```suggestion
As an alternative, views and other assets can be placed in a sidecar directory with the same name as the component, which can be useful for organizing views alongside other assets like Javascript and CSS.
```

test/view_component/view_component_test.rb
```diff
@@ -437,6 +443,16 @@ def test_raise_error_when_variant_template_file_and_inline_variant_call_exist
     assert_includes error.message, "Template file and inline render method found for variant 'phone' in VariantTemplateAndInlineVariantTemplateComponent."
   end
 
+  def test_raise_error_when_template_file_and_sidecar_directory_template_exist
+    error = assert_raises ViewComponent::TemplateError do
+      with_variant :phone do
+        render_inline(TemplateAndSidecarDirectoryTemplateComponent.new)
+      end
```
joelhawksley, line 450:
```suggestion
      render_inline(TemplateAndSidecarDirectoryTemplateComponent.new)
```

helphop:
I apologize if this is an inappropriate place or time to add my findings.  I have been able to achieve a successful organizing of components with their respected supporting files into one directory by following along this post: [https://fugu.dev/2020-04-16-working-with-view-component-and-stimulus-part-2/](url)

jaredcwhite:
@helphop I think namespacing in that manner seems perfectly reasonable. Probably the only downside is while editing those files you'll end up with lots of "component.rb" or "component.scss", etc. files open, and it might be tricky to switch between them. (Of course I have the same complaint with all the "index.html.erb" files of the world, etc.!)

cyupa:
Yep, I also had to add resolved_paths: ["app/components"] to my webpacker.yml. Should be added to the documentation. May I?

joelhawksley:
@cyupa by all means, please send a PR!

cyupa:
@joelhawksley Thanks for building this great tool and loved your RailsConf 2020 CE presentation - that's what convinced me to try it out and glad to hear it comes to Rails 6.1.

I'm a beginner when it comes to Rails and Stimulus and I found out what the issue was later that night.
It was *obviously* a problem with the "interface behind the keyboard" that was confused about how Stimulus maps namespaced controllers.

Opened this PR to extend the documentation:
https://github.com/github/view_component/pull/481

Thanks!

rickychilcott:
@cyupa - you didn't add the line `resolved_paths: ["app/components"]` to #481, but it appeared that I needed it in order for my static definition to work (i.e. `static targets = ["output"]`). I don't really understand webpack/webpacker enough to know why it's needed, but is there a reason you left it out of #481 after highlighting it here?

Thanks for your comment, by the way, I would have been lost otherwise.

cyupa:
@rickychilcott The pleasure is mine. To be honest, I'm a beginner myself so I don't comprehend how everything works behind the scenes and I happened to stumble upon `resolved_paths: ["app/components"]` to fix it but because I also had issues with the name of the Stimulus controller I assumed it was because of that.

Yesterday, when I submitted the documentation changes I did a check: I removed `resolved_paths: ["app/components"]` stopped the Rails server and restarted it to see if it still worked and it did.

Today I started working on a new component that had to integrate with Stimulus and guess what, the previous one didn't work anymore, so I went straight to my main suspect: `resolved_paths: ["app/components"]`.

After I added it back, it worked.
@joelhawksley I will try to open a pull request for this one as well.

rickychilcott:
Thanks @cyupa and @helphop (for the original write up).

I wonder if your new component had the `static` line in it for stimulus targets. Without `resolved_paths: ["app/components"]` and stimulus worked ok for me as long as I wasn't defining targets (i.e. `static targets = ["output"]`)

cyupa:
Yep, I think we're also using babel and it didn't compiled it into browser-compatible Javascript but webpack had to bundle it in the first place, I guess.

I opened https://github.com/github/view_component/pull/485.

PR #338, opened 2020-05-11
Add failing spec for preview with container element

robzolkos:
### Summary

This adds a failing test to cover a scenario [discussed](https://github.com/github/view_component/discussions/337) where a preview component may need (in the case of a table cell TD) or want (by the developer) to be wrapped in another parent element/container.  

eg (non-working)

```
class MyComponentPreview < ViewComponent::Preview
  def normal
    tag.table do 
      tag.tr do 
        render(MyTableCellComponent.new.....
      end
    end
  end
end
``` 

Currently previews only render the component itself or the component as a parent of content inside it.    

### Other Information

I'm happy to continue with this PR is someone could point me in the direction I should go.  I know that `tag.div` doesn't work in the context I'm using it but its a starting point.

### DIFF
--- a/test/test/components/previews/preview_component_preview.rb
+++ b/test/test/components/previews/preview_component_preview.rb
@@ -20,4 +20,10 @@ class PreviewComponentPreview < ViewComponent::Preview
   def with_params(cta: "Default CTA", title: "Default title")
     render(PreviewComponent.new(cta: cta, title: title))
   end
+
+  def wrapped_in_div
+    tag.div do
+      render(PreviewComponent.new(title: "Lorem Ipsum"))
+    end
+  end
 end
--- a/test/view_component/integration_test.rb
+++ b/test/view_component/integration_test.rb
@@ -209,6 +209,12 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     refute_includes response.body, "Label from params"
   end
 
+  test "renders preview component wrapped in div" do
+    get "/rails/view_components/preview_component/wrapped_in_div"
+
+    assert_includes response.body, '<div><div class="preview-component">Lorem Ipsum</div></div>'
+  end
+
   test "renders badge component open preview" do
     get "/rails/view_components/issues/badge_component/open"
 


joelhawksley:
@juanmanuelramallo do you have any ideas here?

juanmanuelramallo:
It must be related to the `render` method from the preview base class. 

The `render` method receives a component, args and a block. The component passed to the render method is what actually gets rendered in the view. 

I see two options:
1. Use an inline component as the first argument of the `render`. So that the html that wraps the component is an actual component, and it gets rendered in the preview endpoint. (no changes needed)
2. Make the preview object aware of the wrapper html code, and use it when rendering the preview, somehow. (changes needed) 

Do you see any other options available for this?

robzolkos:
> 1. Use an inline component as the first argument of the `render`

I don't mind this option.   But wondering if I can declare these in the same preview file.   For example I could create an inline `table` and `tr` component in the preview file and use them there (so they don't clutter up my components folder)

```
class TableComponent < ViewComponent 
  def call;  {some sort of magic here }; end
end

class RowComponent < ViewComponent
  def call;  {some sort of magic here }; end
end

class MyComponentPreview < ViewComponent::Preview
  def normal
    render(TableComponent.new) do 
      render(RowComponent.new) do 
        render(MyTableCellComponent.new.....
      end
    end
  end
end
```

I'm not even sure anything like this is possible - but it would solve the problem if it were.  May not be the most user friendly way forward.  Particularly for simple tags.

juanmanuelramallo:
Yeah I believe that would work. You can even define those classes inside the preview class, so that they are not defined in the global scope. 

Looking at your use case, it makes me wonder if we might need/want to update the Preview API so that it works seamlessly when the component to be previewed needs a certain html wrapper, like in this case.

@jonspalmer do you have any thoughts regarding this?

jonspalmer:
IMO a great way to fix this would be to support Previews that render a view file. We were inspired by ActionMailer where the preview is always an entire html document. However, as we learn with this example sometimes we have components that are html snippets that require rendering inside something else. 

Extra benefit would be that by supporting views we also open up the possibility of previews that involve multiple stories, would allow customized layouts or the inclusion of extra js or css etc. They'd be closer to the flexibility of stories in Storybook.

so Imagine it looked like this:

`test/components/previews/my_component_preview.rb`
```ruby
class MyComponentPreview < ViewComponent::Preview
  def normal
    render :normal
  end
end
```

`test/components/previews/my_component/preview/normal.html.erb`
```erb
<table>
  <tr>
    <%= render(MyTableCellComponent.new..... ) %>
  </tr>
</table>
```

There are some details to work out on how we'd arrange this and still support the simple version of rendering a component for a preview (its very much like a component rendering inline vs with a view). I'd also want to think a bit harder about how we'd organize the view files. Don't want to get into the same confusion we have with #331

What do you think?

robzolkos:
> What do you think?

Love it.  Hadn't thought of the need to be able to show multiple components on a page but solving that whilst solving this problem would be a big win.  eg Going to a Buttons preview and seeing all the varieties of button components.

joelhawksley:
@juanmanuelramallo I'd love your thoughts on possible solutions here ❤️

juanmanuelramallo:
I just created an experimental working branch following most of @jonspalmer 's ideas shared before, I just left the preview example methods empty when using an actual template.

Please have a look and share your thoughts #353

PR #340, opened 2020-05-13
Resolve Ruby debug warnings

joelhawksley:
### Summary

When running our test suite in Ruby debug mode, there was a _flood_ of warnings. I've done my best to mitigate most of them here.

I've also opened https://github.com/Shopify/better-html/pull/58 to address some of the warnings from `better-html` cc @vinistock ❤️


lib/view_component/base.rb
```diff
@@ -294,6 +309,8 @@ def validate_collection_parameter!(validate_default: false)
       private
 
       def provided_collection_parameter
+        @provided_collection_parameter ||= nil
+
         @provided_collection_parameter
```
nickh, line 314: Can we do this in a single statement?

```suggestion
```

Issue #304, opened 2020-04-15
RSpec and ViewComponent::TestHelpers


rmacklin:
Opening this issue based on the discussions started in https://github.com/github/view_component/issues/292#issuecomment-612737914

> Do we really need TestHelpers to have a dependency on Capybara?

> I think the reason for including `Capybara::Minitest::Assertions` was to give access to those assertions inside instances of `ViewComponent::TestCase` (those methods leverage the `page` definition we supply).

> makes sense in the Minitest case. How much does this logic help if you're using RSpec or some other testing library? The testing pattern if very different there and you wouldn't be inheriting from `ViewComponent::TestCase` and would be requiring `capybara/rspec` yourself.

> I think you're right that things are a bit murky w.r.t. rspec. We're including capybara's minitest assertions module in `ViewComponent::TestHelpers`:
> https://github.com/github/view_component/blob/d3952cb35c82734b5119bc2c173e99cc08219d5c/lib/view_component/test_helpers.rb#L7
> 
> 
> but we also have documentation suggesting that `ViewComponent::TestHelpers` should be included in rspec `component` configuration (where it seems like the minitest assertions would be out of place):
> https://github.com/github/view_component/blob/dc0c0c1671dc116a3c2b822a5031b4045c2eb569/README.md#L553-L562
> 
> 
> That said, there definitely are consumers of this library that use rspec and I haven't seen them raise an issue about this being a problem so far. I suspect they are just not using those methods.
> I don't have a lot of rspec experience, but I also noticed that the [capybara README section on rspec](https://github.com/teamcapybara/capybara/blob/3.32.1/README.md#using-capybara-with-rspec) mentions:
> 
> > Capybara matchers are also supported in view specs
> 
> I wonder if folks have been testing components in [view specs](https://relishapp.com/rspec/rspec-rails/v/4-0/docs/view-specs/view-spec) where they can already leverage capybara's rspec matchers.
> 
> Anyway, might be better to open a new issue to discuss the rspec stuff since this one about removing the capybara dependency has been closed.

and https://github.com/github/view_component/pull/299#discussion_r409022216:

> For RSpec the IMO the "best" solution is to separate `TestHelpers` into two pieces. RSpec users get value out of `render_inline` and the supporting code. We're using that today. What RSpec doesn't need is the Capybara Assertions etc. For RSpec you're going to pull in the rspec Capybara helpers explicitly. It would be entirely normal/expected to configure RSpec like this;
> 
> ```ruby
[...]

PR #344, opened 2020-05-15
Add rendered_component method to ViewComponent::TestHelpers

rmacklin:
### Summary

As discussed in https://github.com/github/view_component/issues/304#issuecomment-629468829, this PR renames `@raw` to `@rendered_component` and exposes a `rendered_component` reader.

With this method, it becomes possible to write the following expectation in an rspec-rails view spec:
```rb
expect(rendered_component).to have_css("header h1", text: "Buy milk")
```
which mirrors the example code from Capybara's README:
https://github.com/teamcapybara/capybara/blame/3.32.1/README.md#L235

I also backfilled some test coverage for the return value of `render_inline`. This seemed appropriate given that we started documenting it in the README in https://github.com/github/view_component/commit/18d4569721457a34ae4c2096cb2b0a0c36570637. (I should've added this test when I was working on https://github.com/github/view_component/pull/298.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ class MyComponentTest < ViewComponent::TestCase
 end
 ```
 
-In the absence of `capybara`, assertion against the return values of `render_inline`, which is an instance of `Nokogiri::HTML::DocumentFragment`:
+In the absence of `capybara`, assert against the return value of `render_inline`, which is an instance of `Nokogiri::HTML::DocumentFragment`:
 
 ```ruby
 test "render component" do
--- a/README.md
+++ b/README.md
@@ -397,6 +397,16 @@ test "render component" do
 end
 ```
 
+Alternatively, assert against the raw output of the component, which is exposed as `rendered_component`:
+
+```ruby
+test "render component" do
+  render_inline(TestComponent.new(title: "my title")) { "Hello, World!" }
+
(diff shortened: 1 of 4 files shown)


rmacklin:
GitHub Actions seems to have hit an error running this PR:

image

https://github.com/github/view_component/actions/runs/106062463

@joelhawksley Are you able to try rerunning the workflow manually?

rmacklin:
Hmm, [this](https://www.githubstatus.com/incidents/3q541nk49qmw) might explain it:
image

README.md
```diff
@@ -397,6 +397,16 @@ test "render component" do
 end
 ```
 
+Alternatively, assert against the raw output of the component, which is exposed as `rendered_component`:
```
joelhawksley, line 400: I'd be cool with replacing the example above (with `result = render_inline`) with this example. I think it's a much nicer approach 😄

rmacklin: The benefit of the other approach is that you can use CSS selectors to make more targeted assertions about certain elements in the rendered output (without relying on Capybara). So I still think there is some value in documenting it.

joelhawksley:
@rmacklin can you re-push to trigger a new build?

rmacklin:
@joelhawksley Done!

PR #348, opened 2020-05-20
Implement Slots as potential successor to Content Areas

joelhawksley:
This PR introduces Slots, an evolution of the Content Areas feature in ViewComponent.

## Problem statement

For the sake of explanation, I am going to use the [Box](https://primer.style/css/components/box) component from Primer CSS, GitHub's design system.

The Box component can have a single header, body, and footer, along with **multiple** rows.

An example Box might look something like:

```html
<div class="Box">
  <div class="Box-header">
    Box header
  </div>
  <div class="Box-body">
    <strong>Box body</strong>
  </div>
  <ul>
    <li class="Box-row">
      Box row one
    </li>
    <li class="Box-row">
      Box row two
    </li>
    <li class="Box-row">
      Box row three
    </li>
  </ul>
  <div class="Box-footer">
    Box footer
  </div>
</div>
``` 

With Content Areas, this could be implemented as:

`# box_component.rb`
```ruby
class BoxComponent < ViewComponent::Base
  content_areas :header, :body, :footer
  
  def initialize(rows: [])
    @rows = rows
  end
end
```

`# box_component.html.erb`
```erb
<div class="Box">
  <% if header %>
    <div class="Box-header">
      <%= header %>
    </div>
  <% end %>
  <% if body %>
    <div class="Box-body">
      <%= body %>
    </div>
  <% end %>
  <% if rows.any? %>
    <ul>
      <% rows.each do |row| %>
        <li class="Box-row <%= row[:class_names] %>">
          <%= row[:body] %>
        </li>
      <% end %>
    </ul>
  <% end %>
  <% if footer %>
    <div class="Box-footer">
      <%= footer %>
    </div>
  <% end %>
</div>
```

There are some shortcomings to this approach:

### Collections

Having to pass to pass in `rows` to `BoxComponent` has proven awkward. Such an approach forces us to implement features such as required and default arguments ourselves, patterns Ruby gives us for free in object initializers!

This issue has been [discussed in glorious length](https://github.com/github/view_component/discussions/325).

### Area-specific arguments

With content areas, any area-specific logic or configuration generally ends up being passed into the component initializer.

For example, if we wanted to allow for class names to be set on each content area, our initializer might look something like:

`# box_component.rb`
```ruby
class BoxComponent < ViewComponent::Base
  content_areas :header, :body, :footer
  
  def initialize(
    rows: [], 
    header_class_names: "", 
    body_class_names: "",
    footer_class_names: ""
  )
    @rows = rows
    # ...
  end
end
```

And if we needed to set different class names on each row, we would need to write code to handle default values and required arguments ourselves.

For more context, [see @rosendi's issue](https://github.com/github/view_component/issues/285).

## Existing solutions

@andrewmcodes and I implemented a method for passing arguments to Content Areas in https://github.com/github/view_component/pull/322, assigning the arguments to a hash named after the content area.

@jonspalmer also implemented a solution here: https://github.com/github/view_component/pull/323, using a block-based approach.

@jensljungblad's `elemental_components` has implementations of both area-specific arguments and collection areas, and has been held up as a desirable API.

## Proposed solution

I propose that we move to a new API that I'm calling Slots.

### Why the name “Slots”?

After shipping Content Areas, we received feedback that the feature was very similar to Slots [in Vue](https://vuejs.org/v2/guide/components-slots.html) and [in Web Components](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots).

As we will likely need to make breaking changes to the Content Areas API, I figured this was a good time to align the name of the feature with the greater ecosystem.

### Show me the code already!

Here's what our `BoxComponent` might look like with Slots:

`# box_component.rb`
```ruby
class BoxComponent < ViewComponent::Base
  with_slot :body, :footer
  with_slot :header, class_name: "Header"
  with_slot :row, class_name: "Row", collection: true
 
  class Header < ViewComponent::Slot
    def initialize(class_names: "")
      @class_names = class_names
    end
    
    def class_names
      "Box-header #{@class_names}"
    end
  end
  
  class Row < ViewComponent::CollectionSlot
    def initialize(theme: :gray)
      @theme = theme
    end
    
    def theme_class_name
      case @theme
      when :gray
        "Box-row--gray"
      when :hover_gray
        "Box-row--hover-gray"
      when :yellow
        "Box-row--yellow"
      when :blue
        "Box-row--blue"
      when :hover_blue
        "Box-row--hover-blue"
      else
        "Box-row--gray"
      end
    end  
  end
end
```

`# box_component.html.erb`
```erb
<div class="Box">
  <% if header %>
    <div class="<%= header.class_names %>">
      <%= header.content %>
    </div>
  <% end %>
  <% if body %>
    <div class="Box-body">
      <%= body.content %>
    </div>
  <% end %>
  <% if rows.any? %>
    <ul>
      <% rows.each do |row| %>
        <li class="Box-row <%= row.theme_class_name %>">
          <%= row.content %>
        </li>
      <% end %>
    </ul>
  <% end %>
  <% if footer %>
    <div class="Box-footer">
      <%= footer %>
    </div>
  <% end %>
</div>
```

`# index.html.erb`
```erb
<%= render(BoxComponent.new) do |component|
  <% component.slot(:header, class_names: "my-class-name") do %>
    This is my header!
  <% end %>
  <% component.slot(:body) do %>
    This is the body.
  <% end %>
  <% component.slot(:row) do %>
    Row one
  <% end %>
  <% component.slot(:row, theme: :yellow) do %>
    Yellow row
  <% end %>
  <% component.slot(:footer) do %>
    This is the footer.
  <% end %>
<% end %>
```

#### Slot types

Slots exist in two forms: normal slots and collection slots.

Normal slots (`:header`, `:body`, and `:footer` above) can be rendered once per component. They expose an accessor with the name of the slot (`#header`, `#body`, and `footer`) that returns an instance of `BoxComponent::Header`, etc.

Collection slots (`:row` above) can be rendered multiple times. They expose an accessor with the pluralized name of the slot (`#rows`), which is an Array of `BoxComponent::Item` instances.

#### Defining slots

Slots are defined by the `with_slot` macro:

`with_slot :header`

To define a collection slot, add `collection: true`:

`with_slot :row, collection: true`

To define a slot with a custom class, pass `class_name`:

`with_slot :body, class_name: 'BodySlot`

Slot classes should be subclasses of `ViewComponent::Slot`.

## Next steps

My intention with this PR is to codify a new API to ship as an experimental feature, with the goal of deprecating Content Areas if this approach proves successful.

I can't thank y'all enough for all of the [insightful discussion](https://github.com/github/view_component/discussions/325) that has happened around this initiative so far :heart:

Co-authored-by: Jon Palmer <jonspalmer@gmail.com>
Co-authored-by: Brian Bugh <brainbag@gmail.com>
Co-authored-by: Jens Ljungblad <jens.ljungblad@gmail.com>

cc @bbugh


lib/view_component/base.rb
```diff
@@ -135,6 +146,60 @@ def with(area, content = nil, &block)
...
+
+      slot_class = "#{self.class.name}::#{slot.to_s.capitalize}".constantize
+
+      # Instantiate Slot class, accommodating Slots that don't accept arguments
+      slot_instance =
+        if args.present?
+          slot_class.new(args)
+        else
+          slot_class.new
+        end
```
jonspalmer, line 175: Just a style nit but I'd probably do this as a one liner using the ternary operator

lib/view_component/base.rb
```diff
@@ -135,6 +146,60 @@ def with(area, content = nil, &block)
...
+
+      # Capture block and assign to slot_instance#content
+      slot_instance.content = view_context.capture(&block) if block_given?
+
+      if slot_class.superclass == ViewComponent::CollectionSlot
```
jonspalmer, line 180: Should we make this `.is_a?` vs superclass? Can we imagine situations where there are multiple classes in hierarchy?

jonspalmer, line 180: Should we make this `.is_a?` vs superclass? Can we imagine situations where there are multiple classes in hierarchy?

lib/view_component/base.rb
jonspalmer, line 175: Just a style nit but I'd probably do this as a one liner using the ternary operator

lib/view_component/base.rb
```diff
@@ -135,6 +146,60 @@ def with(area, content = nil, &block)
...
+          if instance_variable_defined?(accessor_name)
+            # If the slot collector has already been set,
+            # append to the Array.
+            instance_variable_get(accessor_name) << slot_instance
+          else
+            # Otherwise, build a new Array.
+            [slot_instance]
+          end
+
+        instance_variable_set(accessor_name, new_value)
```
jonspalmer, line 194:
```suggestion
        instance_variable_set(accessor_name, []) unless instance_variable_defined?(accessor_name)
        instance_variable_get(accessor_name) << slot_instance
```

lib/view_component/base.rb
```diff
@@ -284,6 +349,47 @@ def with_content_areas(*areas)
...
+      # support initalizing slots as:
+      #
+      # with_slots :header
+      def with_slots(*slots)
+        # Generate a new slot class based on symbol
+        # For example: :header => MyComponent::Header < ViewComponent::Slot
+        slots.each do |slot|
+          self.class_eval("class #{slot.to_s.capitalize} < ViewComponent::Slot; end")
+        end
+      end
```
jonspalmer, line 361: love this

lib/view_component/base.rb
```diff
@@ -284,6 +349,47 @@ def with_content_areas(*areas)
...
+      # support initalizing collection slots as:
+      #
+      # with_collection_slots :tab
+      def with_collection_slots(*slots)
+        # Generate a new slot class based on symbol
+        # For example: :tab => MyComponent::Tab < ViewComponent::CollectionSlot
+        slots.each do |slot|
+          self.class_eval("class #{slot.to_s.capitalize} < ViewComponent::CollectionSlot; end")
+        end
+      end
```
jonspalmer, line 372: and this

lib/view_component/slotable.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module ViewComponent # :nodoc:
+  module Slotable
+    extend ActiveSupport::Concern
+
+    included do
+      attr_accessor :content
+
+      class << self
+        def inherited(child)
```
jonspalmer, line 13: Wouldn't this normally be in `class_methods` ?

lib/view_component/base.rb
```diff
@@ -284,6 +349,47 @@ def with_content_areas(*areas)
...
+        if klass.name.demodulize.to_s == "Content"
+          raise ArgumentError.new ":content is a reserved slot name. Please use another name, such as ':body'"
+        end
+
+        if klass.superclass == ViewComponent::CollectionSlot
+          # If slot is a collection, pluralize the name of the slot (Item => items)
+          attr_accessor ActiveSupport::Inflector.pluralize(klass.name.demodulize.downcase).to_sym
+        else
+          attr_accessor klass.name.demodulize.downcase.to_sym
+        end
```
jonspalmer, line 387: It feels a little strange to have effectively a switch statement here vs it being OO. Couldn't the Slot classes declare this? That way consumers of the gem could create their own slot classes and override things like this.

lib/view_component/base.rb
jonspalmer: Similar commentary to elsewhere is making this dependent on the concrete classes important or could we make this `if slot_class.is_collection_slot?` now consumers could define their own slot classes. They'd just have to quack the right way?

bbugh:
I am thrilled ❤️. The API looks fantastic, very nice work. I love that it's standard Ruby syntax all around and very easy to understand at a glance.

One thing I noticed while giving it a quick look-over: it seems like there's undefined behavior when you have both a `with_slots` and a `Slot` class with arguments, with the same name:

```ruby
class BoxComponent < ViewComponent::Base
  with_slots :header # takes 0 arguments
  
  class Header < ViewComponent::Slot # now it takes 1 optional argument
    def initialize(class_names: "")
      @class_names = class_names
    end
  end

  with_slots :header # now it takes 0 arguments again
```

Since `with_slots` is creating a `Header` class internally, Ruby will overwrite the `initialize` with a different arity, but depending on the order of the `class` definition and `with_slots` it might or might not have arguments (Schrödinger's arity?). Maybe this could be a warning or error - *"Hey you already defined a slot class, don't do it manually :bug:"*

Also, I wonder if maybe the subclass should have a domain suffix like `HeaderSlot` to avoid any possible collisions? The only use case I can think of off the top of my head is extracting the slot object to a separate file (`header_slot.rb` vs `header.rb`), but it does mirror The Rails Way of suffixing everything except models.

Thanks for putting this together @joelhawksley. I'm excited to get some time to test it out soon!

lib/view_component/base.rb
bbugh, line 180: On a different topic about this block of code, I think that this late binding effect for `CollectionSlot` will be confusing for users. Presently, the default value for all unspecified slots is `nil`, but I wonder if the default value for a collection slot would be better as `[]`, whether the component consumer defined a block for it or not.

For example, the Box example in the documentation currently would raise an error if `row` isn't specified by the Box consumer, because it assumes that `rows` is an empty collection, not `nil`:

```erb
<% if rows.any? %>
```

Without specifying `row`, the component consumer will get a vague `NoMethodError: undefined method 'each' for nil:NilClass`.

One possible alternative is to improve the error message that would happen here, somehow, maybe?

```erb
<% render(BoxComponent.new) do  %>
  <!-- no rows -->
<% end %>
```

```
# current:
NoMethodError: undefined method `each' for nil:NilClass

# possible?
UnspecifiedSlotError: BoxComponent requires at least one `row` slot to be defined
```

bbugh:
Hey @jensljungblad, really appreciate you chiming in on this since you already have domain knowledge about the tribulations of designing a Rails component library!

> Is there precedence in Rails for defining methods automatically based on defined subclasses in the class?

Off the top of my head, the only place I know for sure in Rails that generates dynamic classes is `has_and_belongs_to_many` joins. I think RSpec is a good parallel in this case, they generate dynamic classes for each nested level of their DSL, e.g.:

```ruby
describe 'top level test' do
  self # => RSpec::ExampleGroups::TopLevelTest
  
  describe 'validations' do 
	self # => RSpec::ExampleGroups::TopLevelTest::Validations
  end
end
```

jensljungblad:
> Off the top of my head, the only place I know for sure in Rails that generates dynamic classes is `has_and_belongs_to_many` joins. I think RSpec is a good parallel in this case, they generate dynamic classes for each nested level of their DSL, e.g.:
> 
> ```ruby
> describe 'top level test' do
>   self # => RSpec::ExampleGroups::TopLevelTest
>   
>   describe 'validations' do 
> 	self # => RSpec::ExampleGroups::TopLevelTest::Validations
>   end
> end
> ```

That's a DSL though! I would expect that to generate methods.

I specifically meant generating methods from an explicit class declaration. Like how writing `class Header < ViewComponent::Slot; end` would actually generate methods on the parent class. In this case a `#header` method.

lib/view_component/base.rb
bbugh: I think the use of `class_eval` is a security issue. If a third party gem that exposes components gets a bad commit or a hostile takeover, this can be exploited for nefarious deeds:

```ruby
class MyEvilThirdPartyGemComponent
  with_slots :"<< self; do_something_nefarious; end; \#"
end
```

Which is then is class_evaluated as

```ruby
class << self; do_something_nefarious; end; # < ViewComponent::Slot; end
```

and calls `do_something_nefarious` which starts a background thread that steals your bitcoin wallet or something.

I think using `Class.new(ViewComponent::Slot)` would be better, but if I remember correctly it doesn't play nice with `included`, so that might need a different order of operations.

jensljungblad:
Another potential benefit with passing the class to `with_slot` could be re-use of the same class for multiple slots:

```ruby
class BoxComponent
  with_slot :header, Section
  with_slot :footer, Section

  class Section < Slot; end
end
```

That would mean a de-coupling the method name (i.e. `#header`) and the class name.

```ruby
# generate a header method pointing to an anonymous class
with_slot :header 

# generate a header method pointing to a custom class
with_slot :header, Header

# or perhaps more verbosely
with_slot :header, class: Header
```

lib/view_component/base.rb
```diff
@@ -284,6 +349,47 @@ def with_content_areas(*areas)
...
+
+        self.slot_names << klass.name.demodulize.downcase.to_sym
+        self.slots << klass
```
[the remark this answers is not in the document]
jonspalmer: This is a place where the default option to class_attributes works much better but its not available in earlier versions of Rails. The other fix I've used for this problem is to use the inherited hook. I'm pretty sure something like this works.

```ruby
module ViewComponent
  class Base < ActionView::Base

    class_attribute :slots
    self.slots = []

    # List of Slot names in symbol form (:title, etc.)
    class_attribute :slot_names
    self.slot_names = []

    def self.inherited(subclass)
      subclass.slots = self.slots.clone
      subclass.slot_names = self.slot_names.clone
    end
end
```

Has the advantage that subclasses can inherit the slots defined in their super classes.

lib/view_component/base.rb
jonspalmer: `[]` as the default makes sense to me.

jonspalmer:
Forgive me if this is a tangent but I wonder if we need the Slot classes. What if they could be any class and we do away with the auto registering?

In particular that would allow the slots to be ViewComponents themselves. Imagine your box component example like this:

`# row_component.rb`

```ruby
    def initialize(theme: :gray)
      @theme = theme
    end
    
    def theme_class_name
      case @theme
      when :gray
        "Box-row--gray"
      when :hover_gray
        "Box-row--hover-gray"
      when :yellow
        "Box-row--yellow"
      when :blue
        "Box-row--blue"
      when :hover_blue
        "Box-row--hover-blue"
      else
        "Box-row--gray"
      end
    end  
  end
```

`# row_component.html.erb`

```erb
<li class="Box-row <%= row.theme_class_name %>">
  <%= row.content %>
</li>
```

`# box_component.rb`

```ruby
class BoxComponent < ViewComponent::Base
  with_slots :body, :footer
  
  class Header 
    def initialize(class_names: "")
      @class_names = class_names
    end
    
    def class_names
      "Box-header #{@class_names}"
    end
  end
  
  with_slots Header
  with_collection_slots Row
end
```

`# box_component.html.erb`

```erb
<div class="Box">
  <% if header %>
    <div class="<%= header.class_names %>">
      <%= header.content %>
    </div>
  <% end %>
  <% if body %>
    <div class="Box-body">
      <%= body.content %>
    </div>
  <% end %>
  <% if rows.any? %>
    <ul>
      <% rows.each do |row| %>
        <%= render row %>
      <% end %>
    </ul>
  <% end %>
  <% if footer %>
    <div class="Box-footer">
      <%= footer %>
    </div>
  <% end %>
</div>
```

consumers could do things like this to swap out the Row implementation

```ruby
class MyCustomBoxComponent < BoxComponent
  with_collection_slots row:, MySpecialRowComponent
end
```

jonspalmer:
Another thought on the API.

Do we need `with_collection_slots :row` or could it simply be `with_slots :row,  collection: true` ?

jensljungblad:
> In particular that would allow the slots to be ViewComponents themselves

@jonspalmer What benefits would that bring? It's what I ended up with myself in https://github.com/jensljungblad/elemental_components but that was mostly because I wanted to re-use the attributes API I had already defined for my component class, which doesn't seem to be an issue here, unless nesting is desired (i.e. wanting to use `with_slot` in your slot class) but I'm not sure that's the case either!

(I think it's good to leave that up to the user though, so if the class can be anything and doesn't need to inherit from a specific superclass, that's nice)

jensljungblad:
I experimented a little myself in https://github.com/github/view_component/pull/354.

I've got one general question:

1. What do you think the expected behavior should be for calling on slots (regular and collection slots) when they haven't been populated yet? To me it makes sense that calling `#tabs` would return an empty array, and not `nil`, if `component.slot(:tab)` hasn't been called. For regular slots, `nil` seems reasonable? That way you could validate their existence etc. 

As for my experiment, two things that would need addressing:

1. Might make sense to make `class_name` or however that argument would be passed a string instead of a constant, similar to `ActiveRecord`, so that the classes could be defined at the bottom of the file, marked private etc.
2. Perhaps the classes could be picked up automatically as well, similar to `ActiveRecord`. That is, a `with_slot :footer` would look for a `SlotsComponent::Footer` constant and use that as the class if none was passed.
3. I kept the slots inheriting from a base `Slot` class only because of the `content` accessor. That could be dealt with in some other way so users can inherit from whatever base class they want.

jonspalmer:
@jensljungblad Benefit of allowing ViewComponents as the slot class would be that it would allow "bringing your own view" for the rendering of the slot and it would allow various options to swap out that implementation. I'm thinking about how you can vary the piece of things like [React Select](https://react-select.com/home) by passing it component classes.

> What do you think the expected behavior should be for calling on slots 
I definitely think collection slots should be an empty array.

>  similar to `ActiveRecord`,
What are the active record examples you're referencing here? 

I'm a big fan of the `with_slot` you describe in 3.1. This signature would be sweet:

```ruby
def self.with_slot(*slot_names, class: ViewComponent::Slot, collection: false)
```

In addition I'd add the wrinkle that `slot_names` would take symbols or classes. If you pass a class then it results in a slot with the name of that class. That would allow these combinations:

```ruby
  # These all results in slots for :header, :body, :footer but different classes used
  with_slot :header, :body, :footer
  with_slot :header, :body, :footer, class: CustomSlot
  with_slot Header, Body, Footer

  with_slot :subheader, class: HeaderSlot
  
  # These all results in an `items` collection slot but different classes used
  with_slot :item, collection: true
  with_slot Item, collection: true
  with_slot :item, class: Row, collection: true
```

jensljungblad:
> @jensljungblad Benefit of allowing ViewComponents as the slot class would be that it would allow "bringing your own view" for the rendering of the slot and it would allow various options to swap out that implementation. I'm thinking about how you can vary the piece of things like [React Select](https://react-select.com/home) by passing it component classes.

👍 

> What are the active record examples you're referencing here?

It actually relates to your suggestions.

Consider:

```ruby
class Box < Component
  with_slot Header
  class Header; end
end
```

This wont work, since the `Header` constant is referenced before being defined. You'd need to do:

```ruby
class Box < Component
  class Header; end
  with_slot Header
end
```

Maybe that's fine, but personally I'd probably prefer making these classes private and keep them at the bottom of the file. This is one reason `ActiveRecord` makes use of strings referencing class names instead of constants, for associations:

```ruby
has_many :things, class_name: "Thing"
```

That way the constant can be resolved at a later time.

If this should be a consideration at all, I don't think `with_slot "Header"` makes that much sense. Might as well go with `with_slot :header` and resolve the class automatically if it exists?

Something like:

```ruby
class Box < Component
  with_slot :header, :body, :footer # all anonymous
end

class Box < Component
  with_slot :header, :body, :footer

  # will be inferred and used by the :header slot, in a way similar to how AR infer associated
  # models where something like belongs_to :thing would resolve to a Thing class
  class Header; end associated 
end

class Box < Component
  with_slot :header, :body, :footer, class_name: "CustomSlot"

  # will be used by header, body and footer slots. so the class_name would be useful for sharing
  # the same slot class between slots.
  class CustomSlot; end associated 
end
```

bbugh:
I like the API where each line defines a slot(s) and can take options. It looks like `validates` in Rails and will probably feel familiar to people. Imagining something like a card header/footer that shares the same slot class:

```ruby
with_slot :header, :footer, class_name: "EndCap" 
with_slot :items, collection: true
```

On the other hand, I think it might be weird if it's inferred in some cases but not others:

```ruby
# is this inferred as class Header and Footer
with_slot :header, :footer

# but this is both EndCap?
with_slot :header, :footer, class_name: "EndCap"
```

Maybe it makes more sense for each line to be its own definition. I mean, realistically, even complex reusable components like [modals](https://bootstrap-vue.org/docs/components/modal#custom-rendering-with-slots) are going to max out at 4-6 slots or so.

---

> @jensljungblad: I experimented a little myself in #354.

I think this is a realistic evolution based on some of what we're discussing here. How would you see this progressing?

---

> @jensljungblad What do you think the expected behavior should be for calling on slots (regular and collection slots) when they haven't been populated yet? To me it makes sense that calling #tabs would return an empty array, and not nil, if component.slot(:tab) hasn't been called. For regular slots, nil seems reasonable? That way you could validate their existence etc.

We discussed this in the [pull request review here](https://github.com/github/view_component/pull/348/files#r428623851), the gist is that I (and @jonspalmer at the time) agree that `nil` for a regular slot and `[]` for a collection slot makes sense as defaults. 

---

> @jonspalmer: In particular that would allow the slots to be ViewComponents themselves.

I believe the reasons raised against this idea before was that ViewComponents require templates and can be created by `render`. I think that while their input API is similar, a Slot is just a lightweight domain-specific value object, while a Component is a heavier-weight behavior object. A slot is also something that has no contextual value outside of the parent component, while components can be composed to build larger things. 

> @jensljungblad: If this should be a consideration at all, I don't think with_slot "Header" makes that much sense. Might as well go with with_slot :header and resolve the class automatically if it exists?

Symbols do not allow easy namespace resolution, which is the usual reason for also allowing explicit class names, in addition to taking in string names to avoid load order issues. `:header` could be `header`, but how do you specify `Nested::Header`?

jensljungblad:
Here's an example of resolving the slot classes in the way I mentioned above: https://github.com/github/view_component/pull/354/commits/09779e1a1b9e0b5c563700ded18fc8e4449e0bb5

jensljungblad:
> Symbols do not allow easy namespace resolution, which is the usual reason for also allowing explicit class names, in addition to taking in string names to avoid load order issues. `:header` could be `header`, but how do you specify `Nested::Header`?

Yes to be clear I'd be in favor of something specifying the class name explicitly, but as a string to avoid load order issues. That makes namespace resolution easy, since you'd just do `class_name: "Nested::Header"` etc.

As for auto-resolving to an existing class if it exists, I don't know yet if it makes sense. Just thinking aloud :) I pushed a commit here that demonstrates it: https://github.com/github/view_component/commit/09779e1a1b9e0b5c563700ded18fc8e4449e0bb5.

jonspalmer:
I can see the letting slots be view_components might be confusing. Let's drop that for now until/if we have some concrete usecases to design around.

> Maybe it makes more sense for each line to be its own definition. I mean, realistically, even complex reusable components like modals are going to max out at 4-6 slots or so.

@bbugh I'm coming around to that idea too. 

@jensljungblad I like the idea of using strings to define the classes. One thing to consider is that in AR there are always classes. On the other hand we imagine simple cases where you should be able to do `with_slot :header` and get the default "Slot" class. You certainly wouldn't want the surprise of something weird happening if you happened to have a Header class in scope for some non-slot use.

Maybe for terseness we make all of these the resolve a slot with name `header?

```ruby
with_slot :header, class: "Header"
with_slot class: "Header" # infer the slot_name from the class is no slot_name is provided
with_slot class: "HeaderSlot" # by convention chomp "Slot" off the classname when inferring the slot_name
```

joelhawksley:
> Maybe it makes more sense for each line to be its own definition. 

I agree 100%, if we are going to take the API in the direction that is being suggested here.

I agree that `nil` and `[]` are good empty states.

> I believe the reasons raised against this idea before was that ViewComponents require templates and can be created by render. I think that while their input API is similar, a Slot is just a lightweight domain-specific value object, while a Component is a heavier-weight behavior object. A slot is also something that has no contextual value outside of the parent component, while components can be composed to build larger things. I'm personally in favor of keeping Slot and Component separated.

Agreed.

jensljungblad:
> Maybe it makes more sense for each line to be its own definition.
>> I agree 100%, if we are going to take the API in the direction that is being suggested here.

Depends on the amount of options and how much sense it would make to group several slots on one line I suppose. I could see it going either way! While I typically prefer keeping validations on separate lines, it's nice that `validates` accommodates both styles.

> @jonspalmer One thing to consider is that in AR there are always classes. You certainly wouldn't want the surprise of something weird happening if you happened to have a Header class in scope for some non-slot use.

Hm, good point. I guess the requirements are:

1. Keep it as simple as possible to create anonymous slots, and to create slots backed by a class.
2. Specifying a shared slot class isn't as common, so could be more verbose.

And some options:

1. Do not auto-resolve any classes. Always require specifying class if you want to provide your own

```ruby
with_slot :header, class_name: "Header"
class Header; end
```

2. Do auto-resolve classes, and ignore the fact that it might pick up a class not intended for slot-use (i.e. count on this being very uncommon)

```ruby
with_slot :header
class Header; end
```

3. Do auto-resolve classes, but add a convention that slot classes must have a `Slot`-suffix, making collisions very unlikely:

```ruby
with_slot :header
class HeaderSlot; end
```

@jonspalmer The `with_slot class: "Header"` example would only make sense if we go for a single `with_slot` line per slot, right?

jensljungblad:
Here's a commit demonstrating option 4) in the previous comment: https://github.com/github/view_component/pull/354/commits/614a4b8b6e532e7919fb0b032d4685703edae3e3.

jensljungblad:
@jonspalmer 

> ```
> with_slot class: "Header" # infer the slot_name from the class is no slot_name is provided`
> ```

I was going to say, what if the class is namespaced, as in `with_slot class: "Namespaced::Header`, what should the slot name be then? But that led to another thought:

Why would you ever have namespaced slot classes? If you extract a slot from a component, isn't that indicative of that being another component instead of a slot? Quoting from BEM again:

> [An element is] a part of a block that has no standalone meaning and is semantically tied to its block.

bbugh:
@jensljungblad I think that's a fair point. 

I'm curious what you guys think, but it seems like a good next step would be to write tests for all of the scenarios we've discussed and see how it feels/passes. 

Unfortunately, last time I tried to write tests, I ran into the issue that I requested changes for. As far as I know, that still exists, and it's not possible to create multiple components safely. Should we resolve that (even with the temporary fix I proposed) and then experiment with some examples?

lib/view_component/base.rb
joelhawksley: @bbugh that's a good point. I did some digging, but couldn't find a way to make `Class.new` play nicely with `included`. For now I'll add a check that restricts the content of slot names to letters, numbers and underscores.

jensljungblad:
After mulling this over, I think these are my main wants:

- One method to define slots (`with_slot`)
- De-couple slot name from class name

How this is done is of less importance, to me. Things I find less important:

- How classes are resolved, automatically or not (manual feels fine to me, I typically favor explicitness over saving a couple of keystrokes anyway)
- If you can assign multiple slots on one line or not
- If slots have to inherit from `ViewComponent::Slot` or not (content behavior could be mixed in)

So based on my earlier comments, the original example would look something like:

```ruby
class BoxComponent < ViewComponent::Base
  with_slot :header, -> { Header }
  with_slot :row, -> { Row }, collection: true
  with_slot :footer

  class Header < ViewComponent::Slot
    def initialize(class_names: "")
      @class_names = class_names
    end
  end
  
  class Row < ViewComponent::Slot
    def initialize(class_names: "")
      @class_names = class_names
    end
  end
end
```

So the detail of if it is `with_slot :header, -> { Header }` or `with_slot :header, class: "Header"` is less important than the fact the slot name and class name are de-coupled, in my opinion.

To me, this is more clear, and easier to follow, than having a mix of `with_slot` and class declarations defining slots, with class names determining the name of the slot method. I find it easier to map that `with_slot :header` will yield me a `#header` method than that a `class Header < ViewComponent::Slot` declaration would do the same.

Oh, also I think it would be fine to allow multiple slots being defined in one call, for those who prefer that style. The options (class and collection) would apply to all slots defined in the same call:

```ruby
with_slot :header, :body, :footer, class: "Section"
```

joelhawksley:
@jensljungblad after working on this branch for a couple of hours yesterday, I arrived at a similar conclusion. 

I agree that we should have one method to define slots, and that we should decouple the class name from the slot name. If I had to choose, I'd probably go with the `class: "Header"` syntax as it more closely matches ActiveRecord relations.

One idea I'm toying with is whether we should be exposing classes at all, given the debate on this PR. Defining classes inside other classes isn't a Rails convention. It's a bit more of a DSL, but perhaps we could go the route of something like:

```ruby
  with_slot :item, collection: true do
    def initialize(highlighted: false)
      @highlighted = highlighted
    end

    def class_names
      @highlighted ? "highlighted" : "normal"
    end
  end
```

which would dynamically generate a class under the hood.

bbugh:
Dang, after getting a taste for threaded discussions I really miss them!

> After mulling this over, I think these are my main wants:
> 
> * One method to define slots (`with_slot`)
> * De-couple slot name from class name
> 
> How this is done is of less importance, to me. Things I find less important:
> 
> * How classes are resolved, automatically or not (manual feels fine to me, I typically favor explicitness over saving a couple of keystrokes anyway)
> * If you can assign multiple slots on one line or not
> * If slots have to inherit from `ViewComponent::Slot` or not (content behavior could be mixed in)

@jensljungblad Thanks for summarizing the points. I agree with your assertions, with one minor caveat - I think slots having a consistent parent class (like `ViewComponent::Slot`) is quite important in order to have predictable behavior in templates (like `content` and null checks). 

> @joelhawksley 
> If I had to choose, I'd probably go with the class: "Header" syntax as it more closely matches ActiveRecord relations.

Agreed! 👍

> @joelhawksley: One idea I'm toying with is whether we should be exposing classes at all, given the debate on this PR. Defining classes inside other classes isn't a Rails convention. It's a bit more of a DSL, but perhaps we could go the route of something like:

This looks cool, but I don't think this is the way to go. I think the more we stick to standard Ruby syntax the more future-proof and "we didn't think of that" safe it will be.

For example, if we subclass the component, there's no simple way to override or add methods to the children slots. I think this is important when considering the inevitable third party component libraries. We've been dealing with this very issue of overriding children slots with a Vue component library this week, so this is fresh pain. 🤪

```ruby
# some subclass:
with_slot :item, collection: true do
  # ??? how does this behave? where's super go? 
  # ??? what about private? does it clobber or mix? etc.
end
```

vs

```ruby
# some subclass
with_slot :item, collection: true, class: "MyItemSubclass"

class MyItemSubclass < ItemSlot
  # Users will know exactly how this works, it's just ruby!
end
```

jensljungblad:
> @joelhawksley: One idea I'm toying with is whether we should be exposing classes at all, given the debate on this PR. Defining classes inside other classes isn't a Rails convention. It's a bit more of a DSL, but perhaps we could go the route of something like

It's pretty much was I suggested back here: https://github.com/github/view_component/pull/348#pullrequestreview-416316269 and also what I ended up with in https://github.com/jensljungblad/elemental_components. Here's the implementation: https://github.com/jensljungblad/elemental_components/blob/master/lib/elemental_components/element.rb#L31-L36.

The temptations are:
- No need to define class explicitly, avoiding any kind of namespace lookup issues or concerns. I also don't see why you would ever really need to reference the slot class for anything.
- No need to inherit from anything in particular
- More compact slot declaration: the name, config and slot class are all defined in the same call/close together

That being said, I kind of like the idea of explicitly defined classes, keeping it very plain-Ruby like. I'm also curious about this statement:

> Defining classes inside other classes isn't a Rails convention

The Rails codebase has a lot of nested classes, and it's been used plenty in the Rails apps I've been working on over the years as well. Seems like a pretty common Ruby technique to me! But I suppose none of the Rails frameworks *rely* on the *user* defining nested classes. Is that what you meant?

> @bbugh: For example, if we subclass the component, there's no simple way to override or add methods to the children slots.

I'm not sure I understand this, or the example.. Could you elaborate with a more complete example?

bbugh:
> > @bbugh: For example, if we subclass the component, there's no simple way to override or add methods to the children slots.
> 
> I'm not sure I understand this, or the example.. Could you elaborate with a more complete example?

Sure thing! This was RE: @joelhawksley's exploration of a psuedo-DSL for `with_slot` classes. A tale of two scenarios.

**Scenario 1: where `with_slot` is a psuedo-class DSL**

Given a third party list component and slots are defined with the DSL-like class creation:

```ruby
class ThirdPartyListComponent < ViewComponent::Base
  with_slot :item, collection: true do 
	def initialize(icon: '')
      @icon = icon   
    end 
  end
end
```

The usage of `with_slot` is ambiguous. It either acts as a re-opened class, OR it replaces the class. I think this is problematic, because has to be one or the other, which limits the API. For example:

```ruby
class CustomListComponent < ThirdPartyListComponent
  # Does this override the item slot, or extend the item slot? 👻
  with_slot :item, collection: true do
    def do_something; end
  end
end
```

**Scenario 2: where `with_slot` just takes a `class` descriptor**

Given a third party list component and slots are defined by just specifying the class:

```ruby
class ThirdPartyListComponent < ViewComponent::Base
  with_slot :item, collection: true, class: 'ItemSlot'
  
  class ItemSlot < ViewComponent::Slot
    # ...
  end
end
```

Extending or replacing the slot classes in a child component is just standard Ruby. You can extend the existing slot by reopening it:

```ruby
class CustomListComponent < ThirdPartyListComponent
  class ItemSlot < XXX
    def something; end
  end
end
```

Or you can replace the slot by specifying a different one:

```ruby
class CustomListComponent < ThirdPartyListComponent
  with_slot :item, collection: true, class: 'CustomItemSlot'

  class CustomItemSlot < ViewComponent::Slot
    # ... etc ...
  end
end
```

jensljungblad:
@bbugh Got it, that's an interesting question for sure. I've actually thought about extending components for a gem I'm working on so def. of interest.

However, have you verified that the example where you re-open the slot class works? I think that will just define a new class, `CustomListComponent::ItemSlot`, and not re-open the `ThirdPartyListComponent::ItemSlot` class, which the slot definition will still point to. I would think the only way to accomplish something like that would be to redefine the slot to point to your own class, and then have that class inherit from the third party component slot class.

joelhawksley:
> Seems like a pretty common Ruby technique to me! But I suppose none of the Rails frameworks rely on the user defining nested classes. Is that what you meant?

@jensljungblad that's correct.

joelhawksley:
@jensljungblad @jonspalmer @bbugh I think I'm all caught up on changes based on our discussion here. Would y'all mind giving another review?

README.md
```diff
@@ -160,6 +160,131 @@ Returning:
...
+##### Example ViewComponent with Slots
+
+`# box_component.rb`
+```ruby
+class BoxComponent < ViewComponent::Base
+  with_slot :body, :footer
+  with_slot :header, class_name: "Header"
+  with_slot :row, collection: true, class_name: "Row"
+
+  class Header < ViewComponent::Slot
```
BlakeWilliams, line 202: I wonder if it would be possible to have slots still inherit from `ViewComponent::Base`?

If we could do that, I think it'd be one less thing to remember and extracting full components is easier since you no longer have to change what it inherits from.

lib/view_component/base.rb
```diff
@@ -140,6 +145,55 @@ def with(area, content = nil, &block)
...
+    # For example:
+    # <%= render(SlotsComponent.new) do |component| %>
+    #   <% component.slot(:footer, class_names: "footer-class") do %>
+    #     <p>This is my footer!</p>
+    #   <% end %>
+    # <% end %>
+    #
+    def slot(slot_name, **args, &block)
+      # Raise ArgumentError if `slot` does not exist
+      unless slots.keys.include?(slot_name)
```
BlakeWilliams, line 164: Should this only raise in dev/test?

lib/view_component/base.rb
```diff
@@ -293,6 +351,65 @@ def with_content_areas(*areas)
...
       end
 
+      # support initalizing slots as:
+      #
+      # with_slot(
+      #   :header,
+      #   collection: true|false,
+      #   class_name: "Header" # class name string, used to instantiate Slot
+      # )
+      def with_slot(*slot_names, collection: false, class_name: nil)
```
BlakeWilliams, line 361: The `class_name` argument here actually confused me for a good minute or two since I was in the mindset of "this is for the view" and thought it was like HTML `class="foo bar"` and not a Ruby class name.

This almost makes me think that `with_slot` might be more clear as a DSL instead of a method that associates with another class?

eg:

```ruby
with_slot :header, collection: false do |arg1, arg2, arg3|
end
```

I think that may also have the added benefit of removing nested classes in components but would also keep all slot logic in a single place in the component.

That's probably a huge change and has its own challenges, but I thought it was worth bringing up the confusion I ran into here.

lib/view_component/base.rb
```diff
@@ -293,6 +351,65 @@ def with_content_areas(*areas)
...
+          else
+            attr_reader accessor_name
+          end
+
+          # Generate a Slot class unless one is provided.
+          # with_slot(:header) generates MyComponent::Header < ViewComponent::Slot
+          unless class_name.present?
+            self.const_set(slot_name.to_s.capitalize, Class.new(ViewComponent::Slot))
+            class_name = "#{self}::#{slot_name.to_s.capitalize}"
+          end
```
BlakeWilliams, line 402: Should we raise if slot doesn't inherit from `ViewComponent::Slot`?

lib/view_component/base.rb
```diff
@@ -293,6 +351,65 @@ def with_content_areas(*areas)
...
+              def #{accessor_name}
+                #{instance_variable_name} ||= []
+              end
+            RUBY
+          else
+            attr_reader accessor_name
+          end
+
+          # Generate a Slot class unless one is provided.
+          # with_slot(:header) generates MyComponent::Header < ViewComponent::Slot
```
BlakeWilliams, line 398:
```suggestion
          # `with_slot(:header)` generates MyComponent::Header < ViewComponent::Slot
```

Minor and feel free to ignore, but I feel like the backtics help separate code from comment here

lib/view_component/base.rb
bbugh: @BlakeWilliams this is constructing a dynamic class, and only gets called if no class name is specified!

BlakeWilliams: @bbugh Right, if it's not present. I tried putting the comment below to not associate it with the auto-generation but I don't think I did a good job of getting that across. :)

If we pass in a `class_name` argument that doesn't inherit from `ActionView::Slot`, maybe it's worth raising instead of relying on behavior failing? It would definitely be a development/help oriented raise + error message imo.

bbugh:
@joelhawksley I haven't had a chance to try it yet, but after reviewing the docs and code, they look excellent! I think you did a great job encompassing everything we discussed. I'll spend some more time on it tonight.

lib/view_component/base.rb
bbugh: Thanks for explaining! I agree that sounds useful, great catch.

README.md
joelhawksley: I could see that being possible, but my concern with that approach would be that it might imply that components could be used as Slots, when a Slot is just a simple accessor object. I do wonder if there's a simpler way to express the Slot interface though, as all we really care about is that the object has a `content` accessor.

lib/view_component/base.rb
joelhawksley: That's a fair question. My concern is that if this doesn't fail in prod, what should the behavior be in this case?

lib/view_component/base.rb
joelhawksley: @BlakeWilliams I share your confusion. The `class_name` API is meant to mimic ActiveRecord: https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-class-name. 

As you mentioned, there has been a lot of debates about the merits of a DSL approach vs. this Class-based approach. I could honestly go either way at this point, but I keep coming back to the fact that Slots are, in my mind, not meant to stand on their own conceptually, let alone be referenced/exist outside the scope of a single component.

@bbugh you seemed to have the strongest feelings about having the Class-based implementation. What's your feeling at this point?

lib/view_component/base.rb
BlakeWilliams: >  what should the behavior be in that situation?

Yeah, I think it's a bit tricky since raising is the best way to bring attention to the problem, but at the cost of a 500 when the slot might not be critical or was just a typo. 🤔

lib/view_component/base.rb
BlakeWilliams: > not meant to stand on their own conceptually, let alone be referenced/exist outside the scope of a single component.

That makes sense to me, and I think builds on your above comment about slots just having a simple `content` accessor.

I wonder if both of those points help make a stronger case for the DSL API over Ruby objects?

Personally I think the DSL style helps reinforce "this is not a component, it's just a smaller, isolated part of a component with its own encapsulated behavior and responsibilities".

lib/view_component/base.rb
joelhawksley: @BlakeWilliams what would you recommend the behavior to be? Have it not render?

lib/view_component/base.rb
bbugh: I agree that DSLs are nice, but I think it is very important not to use a DSL here and allow `class_name` to be set. I explained my reasoning here, let me know if you have more questions. https://github.com/github/view_component/pull/348#issuecomment-643839329

In short, using the DSL in this way means that the API must choose to either allow component subclasses to replace a slot OR to extend a slot, but it cannot support both. With regular Ruby classes, it doesn't limit the usage and any future use cases for third party libraries.

jensljungblad: @bbugh Did you see my comment here? https://github.com/github/view_component/pull/348#issuecomment-643862606 I tried it out and I'm pretty sure it wont work as you want it to regardless :/

jensljungblad: > The class_name argument here actually confused me for a good minute or two since I was in the mindset of "this is for the view" and thought it was like HTML class="foo bar" and not a Ruby class name.

Heh, you're right, good feedback..

BlakeWilliams: > In short, using the DSL in this way means that the API must choose to either allow component subclasses to replace a slot OR to extend a slot, but it cannot support both. With regular Ruby classes, it doesn't limit the usage and any future use cases for third party libraries.

That's a really good point, but I wonder if replacing slots wouldn't be good enough in most cases? I'm also curious how slots would be used and if they'll end up being complex enough that extending will be a common need/want.

If we did go down the DSL path, do we think it'd be able to address the extending problem if it became an issue by adding more DSL methods? e.g. `extend_slot :header`.  (I will admit, the `extend_slot` idea does get a bit into "reinventing the wheel" territory inheritance wise 😄)

I agree that we could go either way API wise, but I do feel like the DSL makes it clear that slots are special and is a bit more fun/clear to write in my opinion. I guess I'm not sure if it's worth trading API/UX for flexibility in this case, personally.

lib/view_component/base.rb
BlakeWilliams: I think not rendering would be ideal in most cases. Unless a slot is critical, I'd much rather render the page with a missing icon or some missing text than have the page be entirely unaccessible.

lib/view_component/base.rb
bbugh: @jensljungblad I'm not sure about the code not working, sorry! It sounds like maybe that's a limitation with the current setup for ViewComponent? I don't believe it's a Ruby limitation, this is a common pattern with Pundit's `Scope` nested classes. Here's code showing the different variations of what I mean:

```ruby
class Component < ViewComponent::Base
  class Subslot < ViewComponent::Slot
    def some_method
      "red"
    end
  end
end

class Child1 < Component
  # Subclass with no override
  # for demonstration purposes
  class Subslot < Subslot
    # some_method will be "red"
  end
end

class Child2 < Component
  # Subclass with override
  class Subslot < Subslot
    def some_method
      "blue"
    end
  end
end

class Child3 < Component
  # Use a different class altogether
  with_slot :subslot, class_name: 'SomeOtherSubslot'

  class SomeOtherSubslot < ViewComponent::Slot
    def some_method
      "fifty"
    end
  end
end
```

> @BlakeWilliams: That's a really good point, but I wonder if replacing slots wouldn't be good enough in most cases? I'm also curious how slots would be used and if they'll end up being complex enough that extending will be a common need/want.

Good question! To clarify, I'm not arguing this for the sake of complexity, if a slot gets complex, they can use a component! I'm advocating for `class_name` in order to be able to use standard Ruby code in components when you need to write standard Ruby code in subclasses (like the above example code).

For one example, imagine if somebody released a form library like `simple_form` made with `view_component`. In order to adopt it, you'd need to know that you could extend/replace parts as-needed. You might need to replace a slot, or you might need to extend a slot, both of which are easy to do if it's just Ruby, but with the DSL idea, `view_component` can only support one or the other.

For another example, let's say you need some of your custom `Slot`s in your component tree to descend from an `InstrumentedSlot`, because you want to instrument certain calls to `content` (like for caching). That's easy with plain Ruby, but impossible with the DSL idea.

One more example, let's say you're making a component that has multiple different kinds of slots, but they mostly share the same behavior, and you want to have a parent class for all of those slot types. That's easy with Ruby, but impossible with the DSL idea.

I made these up off the top of my head and could come up with more, but the point I'm wanting to make is that we can't predict how people might want to use it, so we shouldn't unnecessarily limit it.

I am all for DSLs, I think they're great! I think they're one of the best features of Ruby's culture. I also think it would be a mistake to do here, as I hope the above examples demonstrate. As far as I can tell, the `with_slot` class-block API doesn't give us anything except that it (arguably) looks a little nicer, but it takes away a lot of choice and power from the users. I think the best thing to do is to let people write `class Thing < ViewComponent::Slot`, specify the slot classes, put their code where they want, and be able to just write Ruby to get 'er done. ¢¢

BlakeWilliams: > I'm advocating for class_name in order to be able to use standard Ruby code in components when you need to write standard Ruby code in subclasses (like the above example code).

I see. Would the proposed `extend_slot` API idea also solve that problem though assuming it comes up as one once folks are actively using the API?

> For one example, imagine if somebody released a form library like simple_form made with view_component. In order to adopt it, you'd need to know that you could extend/replace parts as-needed. You might need to replace a slot, or you might need to extend a slot, both of which are easy to do if it's just Ruby, but with the DSL idea, view_component can only support one or the other.

I'm trying to imagine what a simple form library using slots would look like but feel like I'm struggling a bit. When I think of slots, I think slots are small parts of components that you need to change that are separate from the main content and may require a small bit of logic/arguments, but not anything too complex. Do you have any rough ideas what a flexible simple form slot library would look like at least in terms of usage?

> For another example, let's say you need some of your custom Slots in your component tree to descend from an InstrumentedSlot, because you want to instrument certain calls to content (like for caching). That's easy with plain Ruby, but impossible with the DSL idea.

> One more example, let's say you're making a component that has multiple different kinds of slots, but they mostly share the same behavior, and you want to have a parent class for all of those slot types. That's easy with Ruby, but impossible with the DSL idea.

I think those are really good, and valid points, but it does make me ask if class based slots are the correct abstraction for those problems.

> I am all for DSLs, I think they're great! I think they're one of the best features of Ruby's culture. I also think it would be a mistake to do here, as I hope the above examples demonstrate. As far as I can tell, the with_slot class-block API doesn't give us anything except that it (arguably) looks a little nicer, but it takes away a lot of choice and power from the users. I think the best thing to do is to let people write class Thing < ViewComponent::Slot, specify the slot classes, put their code where they want, and be able to just write Ruby to get 'er done. ¢¢

I've been thinking about slots a good bit in my free time and I'm honestly having a hard time differentiating 
slots from components with the class based API. We're trying to get extensibility and complexity out of them which components already give us. The primary difference seems to be that they lack a template and convention defines them *inside* of the component instead of in a separate file? (that does give me some file length and manageability concerns though)

Breaking that down, I guess my questions would be:

* When slots are class based, is there a benefit to using them over components? If so, are the trade-offs of having another concept, longer/more complex component files, etc. worth it?
* If we did go down the DSL path, would some of this be solved by being able to pass component instances to `component.slot` calls in the template?
* Would slots being simple/limited be a feature, and would it help differentiate them from components and when to use one over the other?

For context, I have written a lot of React so that may be influencing my thought process here a good bit. 😅 

One final thought on all of this, I wonder if it would make sense for these to be includes, or separate libraries you have to opt-in to before they're part of the stable API? It's really difficult to talk about these concepts concretely without having used them and experience the pros and cons with real world usage.

joelhawksley: > It's really difficult to talk about these concepts concretely without having used them and experience the pros and cons with real world usage.

This is why I've indicated that the Slots APIs are experimental in the README. They can and should change before the next major release (`v3.0.0`).

joelhawksley: > The primary difference seems to be that they lack a template and convention defines them inside of the component instead of in a separate file?

That's how I feel, too.

joelhawksley: @BlakeWilliams what do you think about supporting _both_ APIs? Perhaps we could have the DSL be the default, recommended path, but we could also support `class_names` for more complex use cases?

jensljungblad: @bbugh In your last example you are inheriting though, not re-opening the class! Are you saying then that you expect the `with_slot` declaration in the parent to resolve to the subclassed slot class? Is that intuitive?

jensljungblad: Could an experimental API just support both for now, then be tweaked once there are some actual problems/use-cases?

```ruby
with_slot :header

with_slot :body do
  def initialize(param:); end
end

with_slot :footer, class_name: "Footer" 
```

BlakeWilliams: > This is why I've indicated that the Slots APIs are experimental in the README. They can and should change before the next major release (`v3.0.0`).

I wonder if it would make sense to make it opt-in if it's experimental? I think that would make it more explicit that it will likely change and isn't part of the stable API? It would also let us version the API, so we could do something like `include ViewComponent::Slots::V2` which would allow us to support older version for a short time until we get to a stable API.

It's also possible I'm overthinking it. 😄 

> @BlakeWilliams what do you think about supporting _both_ APIs? Perhaps we could have the DSL be the default, recommended path, but we could also support `class_names` for more complex use cases?

I think that's a good idea! I do wonder if it might be too easy of a fallback initially and could push us towards using `class_name` instead of finding work arounds or other possible solutions DSL wise, but it still sounds like a solid plan. 👍

bbugh: If you do the API as an option at all it's going to cause these problems. 😌 That leaves it up to the component author and if they pick the save-one-line-DSL then end users can't decide for themselves how to use it. 

I'm sorry I'm not explaining this in a way that makes sense, the issues seem so obvious to me and I'm clearly failing to convey it well. I'll shut up about it, and hope that you guys encounter the problems yourselves before you launch it and it can't be undone.

jensljungblad: @bbugh I understand what you mean, and that it would be up to the component author to make it fully extensible. I just have a hard time thinking about this without actually implementing/trying it out.

I'd be okay to merge without the DSL first, just to have a starting point.

BlakeWilliams: > If you do the API as an option at all it's going to cause these problems. 😌 That leaves it up to the component author and if they pick the save-one-line-DSL then end users can't decide for themselves how to use it.

Let me know if I'm understanding or not, but is the issue that the API provider (who writes the component) using the DSL API means that consumers of that API can't extend slots?

> I'm sorry I'm not explaining this in a way that makes sense, the issues seem so obvious to me and I'm clearly failing to convey it well. I'll shut up about it, and hope that you guys encounter the problems yourselves before you launch it and it can't be undone.

This is a pretty complex issue that's not exactly concrete so talking about it can difficult (I know I've struggled a bit doing so 😄) but I super appreciate your raising concerns and discussing them here and hope you continue to do so if you'd like.

I think I understand the trade-offs, and what flexibility the DSL would lose assuming my first reply in this comment is correct (if not, I'd love to try and understand it). It could be just me, but slots and components seem to have so much overlap that a class based approach makes me think that it would be easier/clearer/more flexible if slots were passed components instead of being a separate/similar concept. To me, the DSL helps make it clear that slots aren't components and really makes the responsibility of a slot, and when to use a slot more clear. At least that's how I feel after reading, discussing, and thinking about it with my very React influenced point of view.

BlakeWilliams: > I'd be okay to merge without the DSL first, just to have a starting point.

Ultimately I think I ended up leaning towards a DSL being the way to go, but I agree that merging without the DSL as a starting point to get a feel for slots would also be fine and would allow us to have some more concrete conversations with specific, and possibly real-world examples. 👍

jensljungblad: > @bbugh: For one example, imagine if somebody released a form library like simple_form made with view_component. In order to adopt it, you'd need to know that you could extend/replace parts as-needed. You might need to replace a slot, or you might need to extend a slot, both of which are easy to do if it's just Ruby, but with the DSL idea, view_component can only support one or the other.

I've also done a bit of React in the past. In React there are plenty of third-party component libraries (bootstrap etc) that can be used by app developers. They typically solve these things with composition instead of inheritance: https://reactjs.org/docs/composition-vs-inheritance.html. I'd be interested in experimenting with some real world cases!

> @bbugh: For another example, let's say you need some of your custom Slots in your component tree to descend from an InstrumentedSlot, because you want to instrument certain calls to content (like for caching). That's easy with plain Ruby, but impossible with the DSL idea.

That should be possible still:

```ruby
class MyComponent
  with_slot :header do
    include(InstrumentedSlot)
  end
end

module InstrumentedSlot
  def content
    instrument { super }
  end
end
```

> @bbugh: One more example, let's say you're making a component that has multiple different kinds of slots, but they mostly share the same behavior, and you want to have a parent class for all of those slot types. That's easy with Ruby, but impossible with the DSL idea.

Should be possible with a module:

```ruby
class MyComponent
  with_slot :header do
    include(CommonBehaviour)
  end

  with_slot :body do
    include(CommonBehaviour)
  end
end
```

That being said, I still it'd be fine to start out with the class based approach and try it out!

jensljungblad: Maybe I'd also vote for `with_slot :header, -> { Header }` just to avoid confusion reg. Ruby class name and HTML class names.

rmacklin: I also prefer the proc syntax to using a string of the name of the class in the `class_name: 'Header'` syntax. With a proc, things like autocompletion in your editor will work whereas they likely won't if you're typing the class name in a string literal. Though I don't have experience with it, I would imagine the proc form also plays nicer with Sorbet/type checking than if it's all just strings.

joelhawksley: @rmacklin I can see how a proc would allow for simpler autocompletion, but I worry about going that route for a couple of reasons:

1) What if the proc ends up being request-dependent? Allowing an arbitrary block of code opens us up to it executing _whatever_ a developer wants, potentially making it difficult to precompile Slots.

2) Procs are known for being slow. They are far from "free".

3) Would it really be that bad to just reference the Slot class name directly, without a proc? (like `with_slot :header, Header`) If we're loading a component, I think it's safe to assume that we want to load all of its slots. What do you think @jensljungblad @bbugh?

rmacklin: > What if the proc ends up being request-dependent? Allowing an arbitrary block of code opens us up to it executing _whatever_ a developer wants, potentially making it difficult to precompile Slots.

I think we can/should prevent this. I had been thinking if we had the proc syntax, we'd call the procs to resolve the associated classes within the `compile` method. That'd ensure they're only executed once, which would prevent them from being request-dependent and would address the concern about them being slow.

Also I think the documentation could describe that the purpose of the proc is only to lazy-load the constant; if someone tries do so something crazy in the proc that would not supported and they're on their own. (FWIW, they can already execute whatever they want by overriding `const_get`, which I think is something we would not want to support.)

> Would it really be that bad to just reference the Slot class name directly, without a proc? (like `with_slot :header, Header`) If we're loading a component, I think it's safe to assume that we want to load all of its slots.

I think that forces `class Header` to be defined before you call `with_slot :header, Header`, which was considered undesirable in https://github.com/github/view_component/pull/348#issuecomment-633692353. If we're okay with that restriction, that syntax would also work well regarding autocompletion, type-checking, etc.

jensljungblad: Like @rmacklin said, that was my only concern with `with_slot :header, Header`, the fact that the class has to be defined at the top of the file. It's not the end of the world.

I didn't look very closely at the compile method but yes, my hope was that the resolved class could be cached somehow.

README.md
jonspalmer: I can't help scratch the itch for mr that slots could themselves be ViewComponents. I've mentioned this before and I don't yet have the concrete uses cases (though i have my suspicions) so I'm prepared to let this go for a future release.

But here is my thinking. If you stripped out some of the magic around class_name and generating classes we're providing in `with_slot` you could imagine that you had to use slots like this:

```erb
<% component.slot(:header, HeaderSlot.new(class_names: "my-class-name")) do %>
    This is my header!
  <% end %>
```
the requirement on HeaderSlot being that it provide a `content` property to put the captured block in. 

Under that light the `component.slot` is looking an awful lot like a `render` call on a ViewComponent.

Now imagine the guts of the `slot` method was something like this:

```ruby
    slot = slots[slot_name]

    if args.is_a?(Hash)
      # The class name of the Slot, such as Header
      slot_class = self.class.const_get(slot[:class_name])

      # Instantiate Slot class, accommodating Slots that don't accept arguments
      slot_instance = args.present? ? slot_class.new(args) : slot_class.new
    else
      # args were really a class
      slot_instance = args
    end

    if slot_instance.responds_to(:render_in)
      # RenderInSlotWrapper exposes a `content` property that is the result of `render_in`
      # I'd need to think it through more but I think we need some sort of wrapper that 'captures' the entirety of the rendered ViewComponent
      slot_instance = RenderInSlotWrapper.new(slot_instance).render_in(view_context, &block) 
    else
      # Capture block and assign to slot_instance#content
      slot_instance.content = view_context.capture(&block) if block_given?
    end
```

My spidey senses says there is something power here.

lib/view_component/base.rb
jonspalmer: My sense is it's powerful to have both options. I'm I'm the end user I'm likely to want to use the macros (I think of these as syntactic sugar/macros rather than a DSL). However, if I'm a Library writer I should use the API version and expose my slot classes IFF I want them to be extendable.

lib/view_component/base.rb
```diff
@@ -140,6 +145,55 @@ def with(area, content = nil, &block)
...
+      # The class name of the Slot, such as Header
+      slot_class = self.class.const_get(slot[:class_name])
+
+      # Instantiate Slot class, accommodating Slots that don't accept arguments
+      slot_instance = args.present? ? slot_class.new(args) : slot_class.new
+
+      # Capture block and assign to slot_instance#content
+      slot_instance.content = view_context.capture(&block) if block_given?
+
+      if slot[:collection]
```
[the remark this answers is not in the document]
jensljungblad: Such things could be fixed once the API is stable perhaps?

README.md
jensljungblad: @jonspalmer I disagree with the following:

```erb
<% component.slot(:header, HeaderSlot.new(class_names: "my-class-name")) do %>
  This is my header!
<% end %>
```

If the `HeaderSlot` is publicly accessible like that, I wouldn't view it as a slot anymore. Slots are internal to the component and shouldn't be usable outside of the component, imo. You can of course put a component in a slot, which would be a very common use case of component composition:

```erb
<%= render(MediaObjectComponent.new) do |media_object| %>
  <% media_object.slot(:left) do %>
    <%= render(AvatarComponent.new(image: comment.image)) %>
  <% end %>
  <% media_object.slot(:right) do %>
    <%= comment.body %>
  <% end %>
<% end %>
```

lib/view_component/base.rb
jonspalmer: I also have a pretty allergic reaction to using a Proc for the reasons @joelhawksley mentioned. I'm also concerned its not very "Rails" like I can't think of examples of this pattern. However, using strings for the names of classes is very common in AR and other places. It's not so convenient for autocompletion etc. but I'd take the safety and simplicity of the string over that.

README.md
jonspalmer: I agree with the pattern of using a component in a slot. I believe that will be powerful enough for many uses cases.

However, I don't really see the 'internal' piece of a component quite the same way. I see Slots as defining an interface between the component and the slots it uses. Its an Interface not a concrete implementation. As such a ViewComponent leveraging slots comes with so slot classes. IMO it would be powerful allow those classes to be extended. Its let about using those Slot classes outside the component its that I can extend them (subclassing most likely) to be used in the component.

For example consider the Row example from the docs 

```ruby
class Row < ViewComponent::Slot
    def initialize(theme: :gray)
      @theme = theme
    end

    def theme_class_name
      case @theme
      when :gray
        "Box-row--gray"
      when :hover_gray
        "Box-row--hover-gray"
      when :yellow
        "Box-row--yellow"
      when :blue
        "Box-row--blue"
      when :hover_blue
        "Box-row--hover-blue"
      else
        "Box-row--gray"
      end
    end
  end
``` 

The interface that the component relies on is that Row have a content getter/setter and it has a getter for `theme_class_name`.

No suppose I want to use this component but add some other classes to the header based on logic special to my use case. I'd do this:

```ruby
class MySpecialRow < Row

  def initialize(theme: :gray, special: :foo)
    super(theme)
    @special = special
  end

  def theme_class_name
     special_classes = some_other_logic(special)
     "#{super} #{special_classes}
  end
end
```

and then in the component usage

```erb
  <% component.slot(:row, MySpecialRow.new(theme: :hover_blue, special: :make_some_magic) do %>
    Special row
  <% end %>
```

thomasbrus:
@joelhawksley @jonspalmer Probably a bit late to chime in but I think all of this can be accomplished without introducing a new DSL.

```ruby
class BoxComponent < ViewComponent::Base
  def header(*args, &block)
    # Could also use content_tag with a .box__header class name here.
    # Point is that args / options are passed along
    helpers.render(BoxHeaderComponent.new(*args), &block)
  end

  def row(*args, &block)
    # ... renders .box__row
  end

  def footer(*args, &block)
    # ... renders .box__footer
  end
end
```

```slim
= render(BoxComponent.new) do |box|
  = box.header(class: 'bg-blue text-white') do
    | Title

  = box.row do
    | Row 1

  = box.row do
    | Row 2

  = box.footer do
    | Footer
```

Yes, it doesn't enforce the order of header / row / footer, like a `content_area` does. But I think that is fine. The flexibility comes in quite handy at times actually.

README.md
jensljungblad: Hm, couldn't you use composition instead? I'm just thinking composition is generally a better pattern in the view layer than overriding/extending. It's the only way you can build components in React, and it works pretty well there. But I admit I have a hard time seeing all the possibilities of 3rd party components and how people would want to extend them.

I guess to phrase it differently, let's say there's a 3rd party library with bootstrap components built with view components. Would you really want to modify them, or would you rather wrap them (composition) with your own components that hold the specialized behavior?

jensljungblad:
> @joelhawksley @jonspalmer Probably a bit late to chime in but I think all of this can be accomplished without introducing a new DSL.
> 
> Yes, it doesn't enforce the order of header / row / footer, like a `content_area` does. But I think that is fine. The flexibility comes in quite handy at times actually.

But you can still do this with or without slots, right? This should be possible with the current `content` block?

thomasbrus:
> But you can still do this with or without slots, right? This should be possible with the current `content` block?

Yes exactly! That's the point, it doesn't need a new API like slots. In this case the template would be:

```
.box
  = content
```

Another example is a Tabs component, it uses the pattern above, plus content areas:

```ruby
# app/components/tabs_compnent.rb
class TabsComponent < ApplicationComponent
  with_content_areas :navigation, :panels

  def item(*args, &block)
    # Renders a single tab (item)
    helpers.render(TabsItemComponent.new(*args), &block)
  end

  def panel(*args, &block)
    # Renders tab panel
    helpers.render(TabsPanelComponent.new(*args), &block)
  end
end
```

```slim
/ app/components/tabs_compnent.html.slim
.tabs(data-controller="tabs")
  .tabs__navigation
    = navigation

  .tabs__panels
    = panels
```

```
/ example usage
= render(TabsComponent.new(style: style)) do |tabs|
  - tabs.with(:navigation) do
    = tabs.item(id: :tab1, active: true)) do
      Tab 1
    = tabs.item(TabsItemComponent.new(id: :tab2) do
      Tab 2
    = tabs.item(TabsItemComponent.new(id: :tab3) do
      Tab 3

  - tabs.with(:panels) do
    = tabs.panel(id: :tab1, active: true) do
      First panel
    = tabs.panel(id: :tab2) do
      Second panel
    = tabs.panel(id: :tab3) do
      Third panel
```

I'm not saying Slots is a bad idea per se, it is just that the issue stated in the Problem Statement can already be solved quite nicely with what have now.

bbugh:
Sorry I'm way behind on all of the discussions here y'all, I've been tied up by some life stuff. I'll try to get some time to catch back up this weekend.

Hi Thomas! 👋 We have discussed that previously! Don't worry though, this is an insidiously more complex problem than it first appears and I'd guess that everyone who has participated in these discussions has gone through a long process of repeating "can't you just..." through all of the obvious-but-flawed solutions until we've ended up here together.

In your example, the issue I see that this proposal solves is that we would still have the need to make a lot of tiny tightly-coupled-but-loosely-grouped subcomponent files, which could be avoided with slots. I outlined scenarios for why slots are better here, take a look and see if that helps shine some light on how we've come to the present proposal: https://github.com/github/view_component/discussions/325#discussioncomment-12187

On the general topic of "why slots" -

In React, the "micro component" problem is easily solvable with a one-liner functional component: `export const Panel = (id) => <div>...` React embeds their templates in the code, so it's easy to group all of this together in the same file, and making a new component is super easy and cheap, since it's just another line of JavaScript. In `view_component`, this is not the case and it gets soupy quickly. We have `.rb`, `.html.erb/haml/slim`, then if you include sidecar files you also get `.js`, and `.css`, and the current design suggests putting all components in one big folder. Requiring that every little sub-section *also* be its own component files will quickly become an unpleasant mess.

In Vue (which Rails/ERB more closely resembles than React), the need for small bits of embedded templating is instead solved by using slots. You end up with the same kind of benefits, in that you can keep all of your logic and templates grouped together in the same place without spreading it across lots of files, but it's a very different solution and (imo) has more advantages. While slots resemble a "mini component", in practice they are quite different. They can be optional, you can have defaults, they easily pass data *up* the tree (important!), the "owner" component can easily refactor the internals without requiring consumers to change their code, they're lightweight, components are more self-contained which (imo) makes long-term maintenance easier, etc.

Ruby's way of building components much more closely resembles Vue, and I think taking inspiration from it over React to solve these issues will result in a better library. Ruby also has the advantage that its ERB templates has the full expressive power of the Ruby language without needing a DSL like Vue's directive/handlebars one, and the syntax for slots is built into the language.

thomasbrus:
@bbugh Hey, thanks for your lengthy reply!

> this proposal solves is that we would still have the need to make a lot of tiny tightly-coupled-but-loosely-grouped subcomponent files,

Not necessarily, it can just be a call to content_tag really ([see comment in `#header` method](https://github.com/github/view_component/pull/348#issuecomment-649850749))  :)

```ruby
class BoxComponent < ViewComponent::Base
  def row(**opts, &block)
    helpers.content_tag(:div, deep_merge_html_options(opts.fetch(:html, {}), class: 'box__row'), &block)
  end
end
```

In general I agree about that it is not desirable to have tons of tiny subcomponents. I think in the case of a `TabsItemComponent` and `TabsPanelComponent` it is justified because these are actually somewhat complex in how they behave and are styled. They can be tested separately. (Btw, in our codebase we happen to have to convention to not define classes inline.)

I agree about that in React it is much easier to locally scope styling and subcomponents, and other assets. Which is great. The "rails" way is to just have a folder for each type.. `app/controllers`, `app/components`, `app/assets` etc. But for a view layer that is component-based, within a Rails application, a nested approach might be desirable yes.

I understand the idea of looking at other frameworks/libraries to port some of the "proven" ideas that worked for them and to see if we can apply them here. It is actually a great approach 🚀 

Purely looking at the Problem Statement at the start of the issue, I think a plain Ruby approach suffices ("my" proposal, for lack of a better name). Not to say Slots don't provide additional benefits still.

lib/view_component/base.rb
bbugh: @jensljungblad the examples that you gave are not solutions to the problem I'm trying to explain. I will try at some point with real code to demonstrate.

lib/view_component/base.rb
joelhawksley: Resolved.

joelhawksley:
@bbugh @jensljungblad @jonspalmer @BlakeWilliams @thomasbrus thank you all for your feedback and discussion!

I've made the changes I think we should make ahead of landing this PR, including moving Slots to `ViewComponent::Slotable` so that folks must opt-in to using Slots for now.

I think we've exhausted the benefits of iterating on this PR and should focus our energy on testing what we have here in our applications. Could you provide a final review before I merge?

bbugh:
I've been hearing the Shots song in my head while reviewing this. SLOTS SLOTS SLOTS SLOTS SLOTS! SLOTS SLOTS SLOTS SLOTS SLOTS!

This has really come together, great work! 🎉 I agree that any further discussion on this PR will have diminishing returns, and a chance to do some concrete usage will be very beneficial to further exploration. Can't wait to try it out!

joelhawksley:
@bbugh for some reason I didn't see your review until now! Will address in a quick follow-up PR.

lib/view_component/base.rb
```diff
@@ -181,6 +186,10 @@ def inherited(child)
         # has been re-defined by the consuming application, likely in ApplicationComponent.
         child.source_location = caller_locations(1, 10).reject { |l| l.label == "inherited" }[0].absolute_path
 
+        # Clone slot configuration into child class
+        # see #test_slots_pollution
+        child.slots = self.slots.clone
```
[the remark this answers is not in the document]
joelhawksley: In theory, I agree. I avoided doing so as I didn't want to re-define `included` in the module.

PR #353, opened 2020-05-23
[338] Previewing examples with contextual HTML

juanmanuelramallo:
Fixes #338 (or at least trying to)

### Summary
- Updated preview object to know about preview example templates
- Updated controller to render the actual preview example template if any
- Added specs

An example of usage would be:

`test/components/previews/cell_component_preview.rb`
```ruby
class CellComponentPreview < ViewComponent::Preview
  def normal
    render_with_template
  end
end
```

`test/components/previews/cell_component_preview/normal.html.erb`
```erb
<table class="table">
  <%= render CellComponent.new %>
</table>
```

Then the CellComponent preview would be available at the same path, but it will actually render the normal template. 

### To-do

- [x] Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # master
 
-* Add support for templates as preview examples
+* Add support for templates as ViewComponent::Preview examples.
 
     *Juan Manuel Ramallo
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # master
 
-* Add support to use templates as preview examples
+* Add support for templates as preview examples
 
     *Juan Manuel Ramallo
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Add support to use templates as preview examples
+
+    *Juan Manuel Ramallo
+
 # 2.14.1
 
 * Allow using `render_inline` in test when the render monkey patch is disabled with `config.view_component.render_monkey_patch_enabled = false` in versions of Rails < 6.1.
--- a/README.md
+++ b/README.md
@@ -681,6 +681,7 @@ config.view_component.preview_route = "/previews"
 This example will make the previews available from <http://localhost:3000/previews>.
 
 #### Adding contextual HTML for previews
+
 Given the preview component lives in `test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`.
 
 `test/components/previews/cell_component_preview.rb`
--- a/README.md
+++ b/README.md
@@ -704,7 +704,7 @@ end
 </div>
 ```
 
-If you'd rather use a different location for your custom preview templates, you can pass a path in the `template` argument
+To use a different location for preview templates, pass the `template` argument:
 (the path should be relative to your `config.view_component.preview_path`):
 
 `test/components/previews/cell_component_preview.rb`
--- a/README.md
+++ b/README.md
@@ -687,6 +687,7 @@ Given the preview component lives in `test/components/previews/cell_component_pr
 ```ruby
 class CellComponentPreview < ViewComponent::Preview
   def default
+    render_with_template
   end
 end
 ```
@@ -701,6 +702,34 @@ end
   </tbody>
 </div>
 ```
+
+If you'd rather use a different location for your custom preview templates, you can pass a path in the `template` argument
+(the path should be relative to your `config.view_component.preview_path`):
+
+`test/components/previews/cell_component_preview.rb`
+```ruby
+class CellComponentPreview < ViewComponent::Preview
+  def default
+    render_with_template(template: 'custom_cell_component_preview/my_preview_template')
+  end
+end
+```
+
+To use dynamic values from the params you can use the `locals` argument:
+
+`test/components/previews/cell_component_preview.rb`
+```ruby
+class CellComponentPreview < ViewComponent::Preview
+  def default(title: "Default title", subtitle: "A subtitle")
+    render_with_template(locals: {
+      title: title,
+      subtitle: subtitle
+    })
+  end
+end
+```
+
+Which enables passing in a value with <http://localhost:3000/rails/components/cell_component/default?title=Custom+title&subtitle=Another+subtitle>.
 
 #### Configuring TestController
 
--- a/README.md
+++ b/README.md
@@ -682,7 +682,7 @@ This example will make the previews available from <http://localhost:3000/previe
 
 #### Adding contextual HTML for previews
 
-Given the preview component lives in `test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`.
+Given a preview component`test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`:
 
 `test/components/previews/cell_component_preview.rb`
 ```ruby
--- a/README.md
+++ b/README.md
@@ -688,7 +688,6 @@ Given a preview component`test/components/previews/cell_component_preview.rb`, t
 ```ruby
 class CellComponentPreview < ViewComponent::Preview
   def default
-    render_with_template
   end
 end
 ```
--- a/README.md
+++ b/README.md
@@ -680,9 +680,9 @@ config.view_component.preview_route = "/previews"
 
 This example will make the previews available from <http://localhost:3000/previews>.
 
-#### Adding contextual HTML for previews
+#### Preview templates
 
-Given a preview component`test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`:
+Given a preview `test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`:
 
 `test/components/previews/cell_component_preview.rb`
 ```ruby
--- a/README.md
+++ b/README.md
@@ -680,6 +680,28 @@ config.view_component.preview_route = "/previews"
 
 This example will make the previews available from <http://localhost:3000/previews>.
 
+#### Adding contextual HTML for previews
+Given the preview component lives in `test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`.
+
+`test/components/previews/cell_component_preview.rb`
+```ruby
+class CellComponentPreview < ViewComponent::Preview
+  def default
+  end
+end
+```
+
+`test/components/previews/cell_component_preview/default.html.erb`
+```erb
+<table class="table">
+  <tbody>
+    <tr>
+      <%= render CellComponent.new %>
+    </tr>
+  </tbody>
+</div>
+```
+
 #### Configuring TestController
 
 Component tests and previews assume the existence of an `ApplicationController` class, which be can be configured using the `test_controller` option:
--- a/README.md
+++ b/README.md
@@ -716,7 +716,7 @@ class CellComponentPreview < ViewComponent::Preview
 end
 ```
 
-To use dynamic values from the params you can use the `locals` argument:
+Values from `params` can be accessed through `locals`:
 
 `test/components/previews/cell_component_preview.rb`
 ```ruby
--- a/README.md
+++ b/README.md
@@ -705,7 +705,7 @@ end
 ```
 
 To use a different location for preview templates, pass the `template` argument:
-(the path should be relative to your `config.view_component.preview_path`):
+(the path should be relative to `config.view_component.preview_path`):
 
 `test/components/previews/cell_component_preview.rb`
 ```ruby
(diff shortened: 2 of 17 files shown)


lib/view_component/preview.rb
```diff
@@ -24,7 +24,7 @@ def render_args(example, params: {})
         provided_params = params.slice(*example_params_names).to_h.symbolize_keys
         result = provided_params.empty? ? new.public_send(example) : new.public_send(example, **provided_params)
         @layout = nil unless defined?(@layout)
-        result.merge(layout: @layout)
+        { layout: @layout }.tap { |args| args.merge!(result) if result }
```
jonspalmer, line 27: This implementation feels unexpected. `tap` to call `merge!` because it as side effects is hard to read and argue about. Why do we need to handle the case where `result` is empty?

lib/view_component/preview.rb
```diff
@@ -62,6 +62,21 @@ def layout(layout_name)
         @layout = layout_name
       end
 
+      # Returns +true+ if the preview example has a template
+      def preview_example_template?(example)
+        return false unless example_exists?(example)
+
+        preview_example_template_path(example).present?
+      end
```
jonspalmer, line 70: Do we need both methods? We're always going to call them in combination. What if this method handled the defaulting we could simplify the whole thing.

lib/view_component/preview.rb
juanmanuelramallo: Because result can be actually nil, instead of empty. 

I guess we can rewrite this as:
```suggestion
        result ||= {}
        result.merge(layout: @layout)
```

lib/view_component/preview.rb
juanmanuelramallo: Ok sounds good, we can do that!

juanmanuelramallo:
Thanks @jonspalmer, I'm gonna add some specs and clean up the code as per your comments. 

You can check the usage example in the PR description.

lib/view_component/preview.rb
jonspalmer: I'm trying to understand how result can be empty. Does your change introduce that case or has it always been like this? If the former can't we fix it if the later how come it hasn't come up before?

app/controllers/view_components_controller.rb
```diff
@@ -56,7 +56,20 @@ def set_locale
     end
   end
 
+  def template_path
+    if (path = @preview.preview_example_template_path(@example_name))
+      prepend_preview_examples_view_path
+      path
+    else
+      "view_components/preview"
+    end
```
jonspalmer, line 65: Its perhaps a personal style issue but I find assignment in an if condition a bit hard to read

```suggestion
    path = @preview.preview_example_template_path(@example_name)
    if path.presence?
      prepend_preview_examples_view_path
    else
      path = "view_components/preview"
    end
    path
```

lib/view_component/preview.rb
juanmanuelramallo: My change introduce that case, because the example method is now empty, it's just the mere definition of the example method:
```ruby
class MyComponentPreview < ViewComponent::Preview
  def default # body-less method
  end
end
```

Hence, this is why the result can be `nil`, because `new.public_send` will always return nothing (in these cases where a template is used)

I guess we can refactor this still a bit further and avoid calling `new.public_send` when using a template no? And maybe even pass the template path as part of the return of this `render_args` method? Wdyt?

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # master
 
+* Add support to use erb templates as preview examples
```
```suggestion
* Add support for ERB templates as preview examples.
```

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # master
 
+* Add support for ERB templates as preview examples.
```
joelhawksley, line 3: Do we want to support all template handlers, not just ERB?

README.md
```diff
@@ -525,6 +525,28 @@ config.view_component.preview_route = "/previews"
 
 This example will make the previews available from <http://localhost:3000/previews>.
 
+#### Adding contextual HTML for previews
+Given the preview component lives in `test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`.
```
joelhawksley, line 529:
```suggestion
Given a preview component in `test/components/previews/cell_component_preview.rb`, template files can be defined in `test/components/previews/cell_component_preview/`.
```

README.md
```diff
@@ -525,6 +525,28 @@ config.view_component.preview_route = "/previews"
...
+
+`test/components/previews/cell_component_preview.rb`
+```ruby
+class CellComponentPreview < ViewComponent::Preview
+  def default
```
joelhawksley, line 534: Rails controllers allow templates to be rendered without defining a method in the controller. Might we consider supporting similar behavior, not requiring `#default` to be defined in this case where a template is also provided?

CHANGELOG.md
juanmanuelramallo: Yeah, given we are just calling render in the controller I think we are good to say that we support any handle, no?

README.md
juanmanuelramallo: That will involve changing the concept of an "example" in the Preview object. Currently one example is just one method in the preview object. The example might call render in its body (initial previewing approach) or have an empty body and use a template (new approach).

But we definitely can support this. We would just need to update the preview object.

lib/view_component/preview.rb
```diff
@@ -62,6 +63,16 @@ def layout(layout_name)
         @layout = layout_name
       end
 
+      # Returns the relative path (from preview_path) to the preview example template if the example and template exists
+      def preview_example_template_path(example)
+        return nil unless example_exists?(example)
+
+        path = Dir["#{preview_path}/#{preview_name}_preview/#{example}.html.erb"].first
```
joelhawksley, line 70: @juanmanuelramallo doesn't this line couple us to ERB specficially?

juanmanuelramallo: Hmm you are right, I completely forgot about this sorry. 

Then I think we should refactor this a bit more and make the examples with templates not use the body-less methods from the preview object, and just look into the correct folder for any template there. If there's any then show those examples (with whatever handler they might have). 

Does that make sense?

joelhawksley: Makes sense to me 👍

jonspalmer:
Do we have other things that line up with Rails actions? Like if I set instance variables in the preview method do they become available to the preview?

I think perhaps where this is leading is that we'd really like previews to just be controller with actions. This would make a lot of layout issues simpler too. Getting the routing right might be challenging but with a bit of smarts I bet we can handle it via subclassed hook.

juanmanuelramallo:
Hey @joelhawksley @jonspalmer, I'm catching up with this again. 

Would you mind reviewing https://github.com/github/view_component/pull/353/commits/95101543e6ef3b3312ea1ad150828112dc1a0143 ?

tomasc:
This would be very helpful, and would probably be an alternative solution to #390.

I agree with @jonspalmer that previews could be very close to controllers. Similarly, as controllers, their actions could render to string or through a template. And I wonder whether these controllers should not simply reside in `app/controllers`. Having them in test folder initially sort of makes sense, however, I would argue these previews are more often than not used in the development env and sometimes (as in my case) in staging or even production as well.

Or … maybe the previews are good as they are (without this PR) and more complex setups that require contextual HTML can be solved per-app through standard Rails controllers.

?

juanmanuelramallo:
I like that @tomasc, and I agree with @jonspalmer too about just using controllers for these type of previews. But that would be responsibility of the end user then. 

I'm ok with closing this PR if you all agree with that. 

This PR is just an experiment on how to solve the issue brought up by #337 and #390.

README.md
juanmanuelramallo: Well I went a bit different about this, just with the intention to keep the implementation simple. I introduced a specific method to render a preview with a template and make the preview object have all the examples defined in there as methods, as we currently do.

robzolkos:
> I like that @tomasc, and I agree with @jonspalmer too about just using controllers for these type of previews. But that would be responsibility of the end user then.
> 
> I'm ok with closing this PR if you all agree with that.
> 
> This PR is just an experiment on how to solve the issue brought up by #337 and #390.

Before this is closed - we should make sure there is documentation on how to move forward for peeps that need this functionality (I created #338)   

My reading is that preview code would go into `/app` folder, which is not consistent with mail previews (or any other test related code).   It isn't "application" code after all.

CHANGELOG.md
joelhawksley: Might we update this message to reflect that all handlers are supported?

README.md
```diff
@@ -680,6 +680,57 @@ config.view_component.preview_route = "/previews"
 
 This example will make the previews available from <http://localhost:3000/previews>.
 
+#### Adding contextual HTML for previews
```
joelhawksley, line 683:
```suggestion
#### Adding contextual HTML for previews

```

README.md
```diff
@@ -680,6 +680,57 @@ config.view_component.preview_route = "/previews"
...
+Given the preview component lives in `test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`.
```
joelhawksley, line 684:
```suggestion
Given a preview component`test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`:
```

README.md
```diff
@@ -680,6 +680,57 @@ config.view_component.preview_route = "/previews"
...
+<table class="table">
+  <tbody>
+    <tr>
+      <%= render CellComponent.new %>
+    </tr>
+  </tbody>
+</div>
+```
+
+If you'd rather use a different location for your custom preview templates, you can pass a path in the `template` argument
```
joelhawksley, line 706:
```suggestion
To use a different location for preview templates, pass the `template` argument:
```

README.md
```diff
@@ -680,6 +680,57 @@ config.view_component.preview_route = "/previews"
...
+(the path should be relative to your `config.view_component.preview_path`):
```
joelhawksley, line 707:
```suggestion
(the path should be relative to `config.view_component.preview_path`):
```

README.md
```diff
@@ -680,6 +680,57 @@ config.view_component.preview_route = "/previews"
...
+`test/components/previews/cell_component_preview.rb`
+```ruby
+class CellComponentPreview < ViewComponent::Preview
+  def default
+    render_with_template(template: 'custom_cell_component_preview/my_preview_template')
+  end
+end
+```
+
+To use dynamic values from the params you can use the `locals` argument:
```
joelhawksley, line 718:
```suggestion
Values from `params` can be accessed through `locals`:
```

test/test/components/previews/inline_component_preview.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+class InlineComponentPreview < ViewComponent::Preview
+  def default
+    render(InlineComponent.new)
+  end
+
+  def inside_form
+    render_with_template
```
joelhawksley, line 9: It'd be _really_ great if we could avoid having to call `render_with_template` here. Rails controllers allow for implicit rendering, and I think we could do the same.

juanmanuelramallo: Gotcha

juanmanuelramallo: https://github.com/github/view_component/pull/353/commits/2edcaf622a5212ca5788b8c41c16689380b75cad
Would something like this work?

juanmanuelramallo: I know we initially talked about not having not even the method definition in the preview object, but I feel that would introduce some unnecesary complexity to the codebase. I know it's kinda strange to have body-less methods, but since this lives in the test land,I would encourage for explicitness rather than the other way around. 

It's just a thinking, if you feel that it would be better to avoid the method definition anyways I can give it a try.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # master
 
+* Add support for templates as preview examples
```
```suggestion
* Add support for templates as ViewComponent::Preview examples.
```

README.md
joelhawksley, line 683:
```suggestion
#### Preview templates
```

README.md
```diff
@@ -680,6 +680,57 @@ config.view_component.preview_route = "/previews"
 
 This example will make the previews available from <http://localhost:3000/previews>.
 
+#### Adding contextual HTML for previews
+
+Given a preview component`test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`:
```
joelhawksley, line 685:
```suggestion
Given a preview `test/components/previews/cell_component_preview.rb`, template files can be defined at `test/components/previews/cell_component_preview/`:
```

test/test/components/previews/inline_component_preview.rb
```diff
@@ -0,0 +1,39 @@
...
+  def with_params(form_title: "Default Form Title")
+    render_with_template(locals: { form_title: form_title })
+  end
+
+  def with_non_standard_template
+    render_with_template(template: "non_standard_path/with_non_standard_template")
+  end
+
+  def with_haml
+    render_with_template
```
joelhawksley, line 23:
```suggestion
```

robzolkos:
Thank you @juanmanuelramallo @jonspalmer   You rock ❤️

Issue #346, opened 2020-05-16
Helpers and view_context are nil


mickenorlen:
Helpers and view_context are nil in `user_component.rb`. Am I doing something wrong?

### Steps to reproduce

I'm following docs: https://github.com/github/view_component#using-helpers

```RUBY
class UserComponent < ViewComponent::Base
  def initialize
    puts 'USERCOMP'
    puts helpers
    puts @view_context
    puts @helpers
  end
end

```

`application.html.erb`
```RUBY
<%= UserComponent.new %>
```

### Expected behavior
Output view context

### Actual behavior
`logs`
```RUBY
"USERCOMP"
nil
nil
nil
```

### System configuration
**Rails version**:
6.0.3

**Ruby version**:
ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux]

**Gem version**:
2.5.1

joelhawksley:
@mickenorlen thank you for filing this issue!

First, an explanation:

When passing a component instance to `render`, Rails calls `render_in` on the instance, passing in the view context, which on our case happens in `Base`:

https://github.com/github/view_component/blob/master/lib/view_component/base.rb#L43

We then use that view context to set the `view_context` and `helpers` accessors. 

In other words, those properties are coupled to the rendering, not instantiation of components, so they are not available on component instances.

What were you hoping to accomplish by accessing these properties on a component instance?

mickenorlen:
Thanks for a quick reply (even in the weekend! Hope you get some good rest too :))

Ok that makes sense, I guess i just got confused :) In the end everything is working when called from render!

PR #357, opened 2020-05-27
Add `before_render`, deprecating `before_render_check`

joelhawksley:
`before_render_check`, an API we never documented,
is a confusing name for this lifecycle method.

`before_render` matches the naming conventions
for Rails callbacks.

Related issue: https://github.com/github/view_component/issues/346

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -66,7 +66,7 @@ module ViewComponent
       # Assign captured content passed to component as a block to @content
       @content = view_context.capture(self, &block) if block_given?
 
-      before_render_check
+      before_render
 
       if render?
         send(self.class.call_method_name(@variant))
@@ -75,6 +75,10 @@ module ViewComponent
       end
     ensure
       @current_template = old_current_template
+    end
+
+    def before_render
+      before_render_check
     end
 
     def before_render_check
@@ -203,6 +207,12 @@ module ViewComponent
         if template_errors.present?
           raise ViewComponent::TemplateError.new(template_errors) if raise_errors
           return false
+        end
+
+        if instance_methods(false).include?(:before_render_check)
+          ActiveSupport::Deprecation.warn(
+            "`before_render_check` will be removed in v3.0.0. Use `before_render` instead."
+          )
         end
 
         # Remove any existing singleton methods,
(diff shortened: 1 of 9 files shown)


lib/view_component/base.rb
```diff
@@ -66,7 +66,7 @@ def render_in(view_context, &block)
       # Assign captured content passed to component as a block to @content
       @content = view_context.capture(self, &block) if block_given?
 
-      before_render_check
+      before_render
```
BlakeWilliams, line 69: I'm wondering if `before_render` should happen before we assign `@content`?

Slightly related, I wonder if we should only call `view_context.capture(self, &block)` if `render?` is true?

test/app/components/old_validations_component.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# TODO: Remove in v3.0.0
+class OldValidationsComponent < ViewComponent::Base
+  include ActiveModel::Validations
+
+  validates :content, presence: true
+
+  def before_render_check
```
kenyonj, line 9: should this be `before_render` ?
```suggestion
  def before_render
```

lib/view_component/base.rb
joelhawksley: Good question. I can see benefits to it being called in both orders of execution.

When I was writing this PR, I was intending `before_render` to mean "once ViewComponent has done all the magic to get the component to render, but before it actually renders."

Could imagine someone using this hook to set an instance variable based on whether `@content` is present or not, for example.

Similarly, I do think that `render?` could execute before `before_render`, but I'd be hesitant to move it to before capturing the provided block, as in theory folks could be using `render?` to check for the presence of `@content`.

BlakeWilliams: Yeah, I was thinking a bit about it and I think that makes sense.

I think being able to check the presence of `@content` to ensure a block was passed would make sense, but I wonder if it could lead to people checking the contents of `@content` which I could see being either (or both) a feature and an anti-pattern.

test/app/components/old_validations_component.rb
joelhawksley: I purposefully left this component in place using `before_render_check` to ensure backward compatibility until we delete `before_render_check` ❤️

Issue #288, opened 2020-04-06
coexistence with wicked_pdf


ocarreterom:
### Steps to reproduce

Add `wicked_pdf` to your Gemfile.
```rb
gem 'wicked_pdf'
```

### Expected behavior

Work well.

### Actual behavior

I get an error:
```sh
20:07:18 web.1   |  Completed 500 Internal Server Error in 425ms (ActiveRecord: 48.6ms | Allocations: 173487)
20:07:18 web.1   |  FrozenError (can't modify frozen fatal)
```

### System configuration
**Rails version**:  6.0.2.2

**Ruby version**: 2.6.5

**Gem version**: 2.2.1

joelhawksley:
@ocarreterom thank you for the report!

This was also being discussed here: https://github.com/github/view_component/issues/257

Do you have a proposed solution in mind? Unfortunately the conflict is due to competing monkey patches of `render`. 

I could see having an option to turn off the ViewComponent monkey patch, potentially allowing some sort of alternate render method such as `render_component`. What do you think @jonspalmer?

ocarreterom:
This monkey patch is due to the backport and it might will be fixed in Rails 6.1?

ocarreterom:
Also, `render_component` looks good to me.

<img src="https://media3.giphy.com/media/tIeCLkB8geYtW/giphy.gif" width="200">

joelhawksley:
@ocarreterom correct. If you're running `6.1`, the monkey patch isn't used.

jonspalmer:
@joelhawksley Providing a view helper like `render_component` and then a config option to prevent the monkey patch seems very sensible and simple to implement. I'd agree that seems like the simplest approach. 

I do wonder if we implemented the monkey patch like wicker_pdf could they coexist: https://github.com/mileszs/wicked_pdf/blob/08a104c2115147410b8008079cba08c5d8a5f226/lib/wicked_pdf/pdf_helper.rb#L37

There is explicit support there for alias_chaining there. We could try that pattern. If it worked we'd have zero config compatibility. While Rails 6.1 isn't going to have this problem I wonder if we should make the effort to build it "the right way" for older Rails versions.

joelhawksley:
@jonspalmer oh, interesting. It would certainly be worth a shot! 

@ocarreterom would you be interested in giving that approach a try? Our CI builds against Rails 5, 6, and master in a matrix, so we should be able to be fairly confident in test feedback here.

jonspalmer:
[...]

PR #358, opened 2020-05-28
288 support wicked pdf

johannesengl:
### Summary

This PR aims to solve conflicts with other gems also monkey-patching the `ActionView` and `ActionController` `render` method by giving users an option config to disable the monkey patch and instead of using the `render` method in view component templates to use the `render_component` helper. 

In < Rails 6.1 the method `render_component` gets included in `ActionView` and users can render view components in the following way: 
```
<%= render_component Component.new(message: "bar") %>
```

The method `render_component` and `render_component_to_string` also get included in `ActionController`

Related Issue
https://github.com/github/view_component/issues/288


test/view_component/integration_test.rb
```diff
@@ -280,4 +280,31 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_select("p", text: "Radio clock counter: 1")
     assert_select("p", text: "Mints counter: 2")
   end
+
+  if Rails.version.to_f < 6.1
+    test "does not include render monkey patches if render_monkey_patch_enabled config is set to false" do
+      begin
+        old_value = Rails.application.config.view_component.render_monkey_patch_enabled
+
+        Rails.application.config.view_component.render_monkey_patch_enabled = false
```
johannesengl, line 289: Setting this option here to false does not work since the initializers of `ViewComponent::Engine` only run once when the rails app gets booted.  Still not sure how to solve this.

joelhawksley:
@johannesengl thanks for the PR! I've marked it as ready for review. Would you be willing to resolve the merge conflicts?

test/view_component/integration_test.rb
johannesengl: Resolving this since the code moved to a different location. Will add this comment to the new location.

test/test_helper.rb
```diff
@@ -30,3 +30,15 @@ def with_preview_route(new_value)
   Rails.application.config.view_component.preview_route = old_value
   app.reloader.reload!
 end
+
+def without_render_monkey_patch
+  begin
+    old_value = Rails.application.config.view_component.render_monkey_patch_enabled
+    Rails.application.config.view_component.render_monkey_patch_enabled = false
```
johannesengl, line 37: Setting this option here to `false` does not work since the initializers of `ViewComponent::Engine` only run once when the rails app gets booted.  Still not sure how to solve this. Recommendations are welcome. 

Found this thread on StackOverflow giving different options of how we could solve this: 
https://stackoverflow.com/questions/37815826/how-to-tell-rspec-to-restart-rails-before-an-example-is-run

johannesengl:
> @johannesengl thanks for the PR! I've marked it as ready for review. Would you be willing to resolve the merge conflicts?

Sure no problem. I just resolved the conflicts. 

When should I open a PR in draft mode and when should it be set to ready for review? The current PR introduces two new tests that are failing and I opened the PR so we could discuss how to solve this issue:  
https://github.com/github/view_component/pull/358#discussion_r433748970

lib/view_component/engine.rb
```diff
@@ -10,6 +10,7 @@ class Engine < Rails::Engine # :nodoc:
     initializer "view_component.set_configs" do |app|
       options = app.config.view_component
 
+      options.render_monkey_patch_enabled = true if options.render_monkey_patch_enabled.nil?
```
joelhawksley, line 13: What if `render_monkey_patch_enabled` is set to true?

lib/view_component/base.rb
```diff
@@ -16,6 +16,8 @@ class Base < ActionView::Base
     class_attribute :content_areas
     self.content_areas = [] # class_attribute:default doesn't work until Rails 5.2
 
+    mattr_accessor :render_monkey_patch_enabled, instance_writer: false
```
joelhawksley, line 19: Might we group this with the other `mattr_accessor` in this file?

And can we set a default value to true?

lib/view_component/engine.rb
```diff
@@ -42,21 +43,35 @@ class Engine < Rails::Engine # :nodoc:
       end
     end
 
-    initializer "view_component.monkey_patch_render" do
+    initializer "view_component.monkey_patch_render" do |app|
+      next if Rails.version.to_f >= 6.1 || !app.config.view_component.render_monkey_patch_enabled
```
joelhawksley, line 47:
```suggestion
      return if Rails.version.to_f >= 6.1 || !app.config.view_component.render_monkey_patch_enabled
```

lib/view_component/engine.rb
```diff
@@ -42,21 +43,35 @@ class Engine < Rails::Engine # :nodoc:
...
-        end
+        require "view_component/rendering_monkey_patch"
+        require "view_component/render_to_string_monkey_patch"
+        ActionController::Base.prepend ViewComponent::RenderingMonkeyPatch
+        ActionController::Base.prepend ViewComponent::RenderToStringMonkeyPatch
+      end
+    end
+
+    initializer "view_component.include_render_component" do |app|
+      next if Rails.version.to_f >= 6.1 || app.config.view_component.render_monkey_patch_enabled
```
joelhawksley, line 63:
```suggestion
      return if Rails.version.to_f >= 6.1 || app.config.view_component.render_monkey_patch_enabled
```

test/view_component/integration_test.rb
```diff
@@ -293,4 +293,21 @@ class IntegrationTest < ActionDispatch::IntegrationTest
       assert_select "title", "Component Previews for preview_component"
     end
   end
+
+  if Rails.version.to_f < 6.1
```
joelhawksley, line 297: Can we add test cases for `render_component_to_string` and `render_component` from a controller?

test/test_helper.rb
joelhawksley: @johannesengl I wouldn't be opposed to setting up a second Rails app in the test suite for this compatibility mode.

lib/view_component/engine.rb
rmacklin: I think this can't be changed to `return` because we are not inside a method (so it'd result in a `LocalJumpError` exception.

rmacklin:
_Disclaimer: This is totally not a big deal and I'm happy to stick with Joel's preference for this project._

IMHO, the draft PR state indicates "not ready for merge" (which GitHub helpfully enforces!) but doesn't make a claim about whether it can/should be reviewed; since it's possible to explicitly request a review on a draft PR, _that action_ is the indication that you'd like a review on the PR as it currently stands.

lib/view_component/base.rb
johannesengl: Good idea! Done in 80f80090b6cae95a95af2407d0854f490808d7de

lib/view_component/engine.rb
johannesengl: The default should be `true`. If it is set to `true` in will include the monkey patches. If set to `false` `render_component` gets included in `ActionView` and `ActionController`.

lib/view_component/base.rb
```diff
@@ -152,6 +152,14 @@ def request
...
     @@test_controller = "ApplicationController"
 
+    # Option to configure if render monkey patches should be inluded or not.
+    # If set to true and rails version < 6.1 render monkey patches will be included
+    # If set to false render moneky patches will not be included but instead the method
+    # render_component will be included in ActionView and ActionController which can be used
+    # for rendering components.
+    mattr_accessor :render_monkey_patch_enabled, instance_writer: false, default: true
+
+
```
joelhawksley, line 162:
```suggestion
```

test/app/views/integration_examples/render_component.html.erb
```diff
@@ -0,0 +1 @@
+<%= render_component ErbComponent.new(message: "bar") %>
```
joelhawksley, line 1:
```suggestion
<%= render_component ErbComponent.new(message: "bar") %>

```

lib/view_component/base.rb
johannesengl: Done in 10bdc944593c8b52f209630571c887df8e8a0bc8

test/app/views/integration_examples/render_component.html.erb
johannesengl: Done in 10bdc944593c8b52f209630571c887df8e8a0bc8

test/test_helper.rb
johannesengl: @joelhawksley 
I played around with the test setup for a while and tried two approaches:

**Second test dummy app**
In my option creating a new dummy application each time we want to test the behavior of config options changing the initializing process of the engine is not scalable.

**Swapping the `RAILS_ENV` in the rake test task**
I like the approach of swapping the RAILS_ENV more since we don't have to maintain two full dummy applications. 

What is your opinion regarding this? Any idea how we could solve this in a more elegant way? 

Done in d6f9fa91f0dda9930df119306028e2c0092b0a3a

test/view_component/integration_test.rb
johannesengl: Done in c3cb1be1f48dfcaca5a5509396326934114bd2d1

johannesengl:
> Looks good! Can you add an entry to CHANGELOG and update the README to include your info?

Sure. Just updated the CHANGELOG and README: 685d40273a9178b17ff3f915a2b664a601475407

coverage/coverage.txt
```diff
@@ -2,9 +2,4 @@
 Incomplete test coverage
 --------------------------------------------------------------------------------
 
-/app/controllers/view_components_controller.rb: 97.22% (missed: 49)
-/lib/view_component/base.rb: 97.7% (missed: 165,244,272,322)
-/lib/view_component/preview.rb: 92.31% (missed: 32,42,78)
-/lib/view_component/render_to_string_monkey_patch.rb: 83.33% (missed: 9)
-/lib/view_component/test_helpers.rb: 95.45% (missed: 17)
-/test/view_component/integration_test.rb: 97.16% (missed: 14,15,16,18,20)
+100% test coverage. You're all set, friend!
```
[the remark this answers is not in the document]
johannesengl: @joelhawksley I think this is connected to the changes I made to the test setup. I would love to get some advice on how to setup the second dummy app or swap the test environment in order to test a different initializing flow on the engine. 
The way I solved it also brings an issue when running tests and passing a test file name `rake test TEST=...` since I hardcoded in one rake task the file that should be executed on the other environment. Left another comment regarding this here: https://github.com/github/view_component/pull/358#discussion_r441476057

test/test_helper.rb
joelhawksley: @johannesengl I think a second dummy app is the way to go here, given the downsides you've identified with the `RAILS_ENV` approach. I don't see us having many other cases where this kind of configuration change will need to be tested.

johannesengl:
@joelhawksley 

I would love to discuss a bit more in detail how we would like to setup a second dummy app in our test suit.

Do we want to have only one process running our test suite and then in the test_helper initialize two dummy apps? 
I tried this out and pushed a new branch: https://github.com/johannesengl/view_component/tree/second-dummy-app-same-process/test

This approach fails at the point of initializing the second rails application. When executing `Dummy2::Application.initialize!` I get an error `wrong argument type NilClass (expected Proc/Method/UnboundMethod) (TypeError)`

Another option would be to [invoke two rake tasks](https://github.com/johannesengl/view_component/blob/second-dummy-app/Rakefile) and have two folders for reach app including test files. I tried this approach and pushed it to this branch: 
https://github.com/johannesengl/view_component/tree/second-dummy-app/test

This issue here is that we also have unexpected behavior when running tests and passing a specific path to a test file. Furthermore, the second run will override the first test coverage reports. 

Also happy to jump on a call to discuss this further.

rmacklin:
> Furthermore, the second run will override the first test coverage reports.

SimpleCov declares support for merging of coverage across test suites. Have we tried this setup:
https://github.com/colszowka/simplecov/blob/v0.18.5/README.md#using-simplecov-for-centralized-config

joelhawksley:
@johannesengl this is a tricky one! I paired with my colleague @seejohnrun on this, and he recommended that we look into using the `build_app` pattern Rails uses to test similar functionality:

https://github.com/rails/rails/blob/e4b6c719cd1df3dd2bf16645dbe7780dd541ec78/railties/test/isolation/abstract_unit.rb#L99

Perhaps you could look into going that route? Let me know what you think.

Just a note: I'm going to be away from my computer for the next week, so I'll plan to catch up on this PR on or around June 29th 😄

johannesengl:
> @johannesengl this is a tricky one! I paired with my colleague @seejohnrun on this, and he recommended that we look into using the `build_app` pattern Rails uses to test similar functionality:
> 
> https://github.com/rails/rails/blob/e4b6c719cd1df3dd2bf16645dbe7780dd541ec78/railties/test/isolation/abstract_unit.rb#L99
> 
> Perhaps you could look into going that route? Let me know what you think.
> 
> Just a note: I'm going to be away from my computer for the next week, so I'll plan to catch up on this PR on or around June 29th 😄

@joelhawksley Welcome back. I hope you had a nice break :) Lets maybe align on what we want to archive and also rethink different options. 

**Option always include render_component**
We decided to only add the `render_component` helper if `render_monkey_patch_enabled = false`. If we would be willing to add `render_component` in any case we could test the functionality without the need for a second dummy app. 
However, we would still not be able to write a test making sure that the render monkey patches are not included if render_monkey_patch_enabled config is set to false. What do you think regarding this option? 

**Option several rake tasks**
Current issues: 
- Combine simplecov reports
- Find fix for making it still possible on passing a `TEST` parameter to test rake task

**Option Booting up new rails app based on abstract_unit.rb used in rails testing**
Just to clarify. We want to boot up a second rails app in the same process using the `build_app` pattern right? 
I played around with the `abstract_unit` code. What it does is it boots up rails from an empty state by including some modules in `ActiveSupport::TestCase`. This opposes from what we have since we already initialized one dummy rails app. 
The `build_app` method actually invokes the rails new generator. I assume more fitting for us would be the `make_basic_app ` method. I was able to archive that the view component engine got executed a second time with the new config options which led to including `render_component`. However the `render` monkey patches were still included in `ActionView` and `ActionCntroller`. 

I was able to run the following test case in isolation successfully. When running the entire test suit other tests however failed.

```
  test "rendering component using the render_component helper" do
    make_basic_app do |app|
      app.config.view_component.render_monkey_patch_enabled = false    
    end

    assert !Rails.application.config.view_component.render_monkey_patch_enabled

    get "/render_component"
    assert_includes last_response.body, "bar"

    get "/controller_inline_render_component"
    assert_includes last_response.body, "bar"

    get "/controller_to_string_render_component"
    assert_includes last_response.body, "bar"
  end
```

I would love to maybe pair with someone on this to move this PR forward.

rmacklin:
> Option always include render_component

I would support always providing `render_component`. Being explicit and using `render_component` would offer a minor optimization since it jumps straight to https://github.com/rails/rails/blob/cf1881237a3b40c6232cb321bed4f03ccc897cf5/actionview/lib/action_view/helpers/rendering_helper.rb#L43 without evaluating the `case` or the `if`, so it could be useful even in Rails 6.1.

joelhawksley:
> I would support always providing render_component. [...]

I'm wary of going this route. I care a lot about the APIs we provide by default, and I'd much prefer that we use the Rails render call stack in 6.1. 

> Option Booting up new rails app based on abstract_unit.rb used in rails testing

This looks directionally correct. Let's pair on getting this resolved ❤️

johannesengl:
> > I would support always providing render_component. [...]
> 
> I'm wary of going this route. I care a lot about the APIs we provide by default, and I'd much prefer that we use the Rails render call stack in 6.1.
> 
> > Option Booting up new rails app based on abstract_unit.rb used in rails testing
> 
> This looks directionally correct. Let's pair on getting this resolved ❤️

After having paired on the Option of booting up new rails app based on abstract_unit.rb with @joelhawksley we decided not to further pursue this direction but rather always include `render_component` in < rails 6.1. 
The option of rendering components via `render_component` is only given for better backward compatibility and therefore does not justify are more complex test setup using the `build_app` pattern from rails.

.gitignore
```diff
@@ -54,3 +54,5 @@ build-iPhoneSimulator/
 
 # unless supporting rvm < 1.11.0 or doing something fancy, ignore this:
 .rvmrc
+
+.editorconfig
```
joelhawksley, line 58:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # master
 
+# 2.13.0
+
+* Add the ability to disable the render monkey patch on < Rails 6.1 and include the `render_component` method in `ActionView` and `ActionController` which can be used for rendering components instead of `render`. Furthermore the method `render_component_to_string` is included in `ActionController` for rendering a component as string.
```
```suggestion
* Add the ability to disable the render monkey patch with `config.view_component.render_monkey_patch_enabled`. In versions of Rails < 6.1, add `render_component` and `render_component_to_string` methods which can be used for rendering components instead of `render`.
```

README.md
```diff
@@ -711,6 +711,22 @@ To use component previews:
 config.view_component.preview_path = "#{Rails.root}/spec/components/previews"
 ```
 
+### Disabling the render monkey patch on < Rails 6.1
```
joelhawksley, line 714:
```suggestion
### Disabling the render monkey patch (Rails < 6.1)
```

README.md
```diff
@@ -711,6 +711,22 @@ To use component previews:
...
+
+In order to [resolve conflicts](https://github.com/github/view_component/issues/288) on `< Rails 6.1` between ViewComponent other gems also monkey patching the `render` method it is possible to configure ViewComponent to not include the render monkey patch using the following config option:
```
joelhawksley, line 716:
```suggestion
In order to [resolve conflicts](https://github.com/github/view_component/issues/288) on `Rails < 6.1` between ViewComponent other gems also monkey patching the `render` method, it is possible to configure ViewComponent to not include the render monkey patch:
```

README.md
```diff
@@ -711,6 +711,22 @@ To use component previews:
...
+
+`config.view_component.render_monkey_patch_enabled = false # defaults to true`
+
+If the option is set to `false` ViewComponent will *NOT* monkey patch the render method and it is not possible to render view components using `render`. Instead the method `render_component` will be included in `ActionView` and `ActionController` in order to render view components:
+
+
```
joelhawksley, line 722:
```suggestion
```

README.md
```diff
@@ -711,6 +711,22 @@ To use component previews:
...
+```
+<%= render_component Component.new(message: "bar") %>
+```
+
+The method `render_component_to_string` also gets included in `ActionController`
+
+
```
joelhawksley, line 729:
```suggestion
```

README.md
```diff
@@ -711,6 +711,22 @@ To use component previews:
 config.view_component.preview_path = "#{Rails.root}/spec/components/previews"
 ```
 
+### Disabling the render monkey patch on < Rails 6.1
+
+In order to [resolve conflicts](https://github.com/github/view_component/issues/288) on `< Rails 6.1` between ViewComponent other gems also monkey patching the `render` method it is possible to configure ViewComponent to not include the render monkey patch using the following config option:
+
+`config.view_component.render_monkey_patch_enabled = false # defaults to true`
+
+If the option is set to `false` ViewComponent will *NOT* monkey patch the render method and it is not possible to render view components using `render`. Instead the method `render_component` will be included in `ActionView` and `ActionController` in order to render view components:
```
joelhawksley, line 720:
```suggestion
If the option is set to `false` ViewComponent will *NOT* monkey patch the render method and it is not possible to render view components using `render`. Use `render_component` (or  `render_component_to_string`) instead:
```

README.md
```diff
@@ -711,6 +711,22 @@ To use component previews:
...
+
+
+```
+<%= render_component Component.new(message: "bar") %>
+```
+
+The method `render_component_to_string` also gets included in `ActionController`
```
joelhawksley, line 727:
```suggestion
```

README.md
```diff
@@ -917,6 +933,11 @@ ViewComponent is built by:
 |@maxbeizer|@franco|@tbroad-ramsey|@jensljungblad|@bbugh|
 |Nashville, TN|Switzerland|Spring Hill, TN|New York, NY|Austin, TX|
 
+|<img src="https://avatars.githubusercontent.com/johannesengl?s=256" alt="johannesengl" width="128" />|||||
+|:---:|:---:|:---:|:---:|:---:|
+|@johannesengl|||||
+|Berlin, Germany|||||
```
joelhawksley, line 939:
```suggestion
|<img src="https://avatars.githubusercontent.com/johannesengl?s=256" alt="johannesengl" width="128" />|
|:---:|
|@johannesengl|
|Berlin, Germany|
```

joelhawksley:
@johannesengl thank you SO much for your work on this PR! I look forward to your next PR :)

johannesengl:
> @johannesengl thank you SO much for your work on this PR! I look forward to your next PR :)

@joelhawksley It was my pleasure. I really enjoyed the collaboration with you. Thanks for making this a great first open-source experience! :)

PR #361, opened 2020-06-01
Revert counter to start at 0 instead of 1

frznk-s:
### Summary

Revert counter to start at 0 instead of 1

### Other Information

https://github.com/github/view_component/discussions/359


joelhawksley:
What level of SemVer change should this be? This change could be potentially disruptive.

Thoughts @tclem @jonspalmer?

tclem:
I can see this being a breaking change for any user of the library that's rendering collections and using `_counter` which would be considered part of the public api. Any logic that might be based on 1-indexing will break. I think that means it's a major [semver](https://semver.org/) change: _"Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the public API. "_

joelhawksley:
@franks921 can you write a changelog entry under a `v3.0.0` header?

 I'm hoping we can resolve the [Slots work](https://github.com/github/view_component/pull/348) as part of `3.0`, so I'm thinking we should get this PR ready to go and plan to land `3.0` in the next month or so.

franks921:
Done @joelhawksley

joelhawksley:
@franks921 would you be up for putting this change behind a configuration option that we can default to `false` for now and change to `true` in 3.0?

franks921:
Hi @joelhawksley 

Apologies for the late reply. What kind of configuration option did you have in mind? Something like `collection_counter_starts_at_zero = false`?

joelhawksley:
@franks921 yes, that sounds perfect!

franks921:
Ok cool @joelhawksley. I'm just a little swamped with work for the next few days then I will get onto this 👍🏼

PR #363, opened 2020-06-04
DOC: Add example of testing for with_content_areas

jordanbrock:
CHANGES
Just added a small documentation change to the testing section of README.md. Provided an example of testing a component that uses `with_content_areas`. I just grabbed the example from the actual tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,6 +439,21 @@ test "render component" do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


joelhawksley:
Thanks @jordanbrock ❤️

PR #364, opened 2020-06-04
Ruby parse errors throw confusing error

maxbeizer:
### Summary

I ran into this fun today when I left a trailing comma on an `attr_reader`

If the Ruby class is invalid, the library should not fail with `initializer must accept ...` ArgumentErrors. Instead it should throw an error better explaining the issue.

Open question: is it valid to initialize a class for a view component without an initialize parameter? It seems like it should be 🤔 .

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -307,7 +307,14 @@ module ViewComponent
         parameter = validate_default ? collection_parameter : provided_collection_parameter
 
         return unless parameter
-        return if instance_method(:initialize).parameters.map(&:last).include?(parameter)
+        return if initialize_parameters.map(&:last).include?(parameter)
+
+        # See https://github.com/github/view_component/pull/364
+        if initialize_parameters.empty?
+          raise ArgumentError.new(
+            "#{self} initializer is empty or invalid."
+          )
+        end
 
         raise ArgumentError.new(
           "#{self} initializer must accept " \
@@ -316,6 +323,10 @@ module ViewComponent
       end
 
       private
+
+      def initialize_parameters
+        instance_method(:initialize).parameters
+      end
 
       def provided_collection_parameter
         @provided_collection_parameter ||= nil
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -309,7 +309,9 @@ module ViewComponent
         return unless parameter
         return if initialize_parameters.map(&:last).include?(parameter)
 
-        # See https://github.com/github/view_component/pull/364
+        # If Ruby cannot parse the component class, then the initalize
+        # parameters will be empty and ViewComponent will not be able to render
+        # the component.
         if initialize_parameters.empty?
           raise ArgumentError.new(
             "#{self} initializer is empty or invalid."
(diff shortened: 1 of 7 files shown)


lib/view_component/base.rb
```diff
@@ -308,7 +308,14 @@ def validate_collection_parameter!(validate_default: false)
         parameter = validate_default ? collection_parameter : provided_collection_parameter
 
         return unless parameter
-        return if instance_method(:initialize).parameters.map(&:last).include?(parameter)
+        return if initialize_parameters.map(&:last).include?(parameter)
+
+        # See https://github.com/github/view_component/pull/364
```
joelhawksley, line 313: Would you mind adding a brief description of the reason for this check as a code comment here, in place of a link to this PR?

maxbeizer:
@joelhawksley requested changes made 💖 

> Open question: is it valid to initialize a class for a view component without an initialize parameter?

What do you think about that? Sort of reminds me of stateless components in React

test/view_component/view_component_test.rb
```diff
@@ -523,4 +523,27 @@ def test_collection_component_missing_default_parameter_name
...
+  def modify_file(file, content)
+    filename = Rails.root.join(file)
+    old_content = File.read(filename)
+    begin
+      File.open(filename, "wb+") { |f| f.write(content) }
+      yield
+    ensure
+      File.open(filename, "wb+") { |f| f.write(old_content) }
+    end
+  end
```
joelhawksley, line 548:
```suggestion
```

test/app/components/product_reader_oops_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<li>
+  <h1>Product</h1>
+  <h2><%= product.name %></h2>
+  <p><%= notice %></p>
+</li>
```
joelhawksley, line 5:
```suggestion
<div>Hello, World!</div>
```

test/view_component/view_component_test.rb
```diff
@@ -523,4 +523,14 @@ def test_collection_component_missing_default_parameter_name
 
     assert_match(/MissingDefaultCollectionParameterComponent initializer must accept `missing_default_collection_parameter` collection parameter/, exception.message)
   end
+
+  def test_collection_component_with_trailing_comma_attr_reader
+    exception = assert_raises ArgumentError do
+      render_inline(
+        ProductReaderOopsComponent.with_collection([OpenStruct.new(name: "Mints")])
```
joelhawksley, line 530:
```suggestion
        ProductReaderOopsComponent.with_collection(["foo"])
```

test/app/components/product_reader_oops_component.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class ProductReaderOopsComponent < ViewComponent::Base
+  attr_reader :product,
+              :notice,
+
+  def initialize(product_reader_oops:, notice:)
+    @product = product_reader_oops
+    @notice  = notice
+  end
```
joelhawksley, line 10:
```suggestion
  def initialize(product_reader_oops:)
  end
```

Issue #345, opened 2020-05-16
View component is significantly slower than partials


bunnymatic:
`ViewComponent` is slow - roughly 3x slower than using plain rails partials (see spot benchmarking below).   In a recent project I worked on, which heavily leveraged `ViewComponents`, we're finding that the app is very slow.  We've been looking at other optimizations but the template rendering seems to be where most of the time is spent (on pages that have many components) and we'd love to not have to replace `ViewComponents` with something else if we don't have to.

### Steps to reproduce

I built a little Rails app to test this in isolation.  It includes 4 pages each which render a very simple chunk of html that includes the filename of the template.  These pages take a `count` parameter so you can easily try rendering lots of partials (not unlike what I've seen in the aforementioned project).  The 4 rendering methods are:
* `ViewComponent`
* `Rcode5::Component` - this is a hand rolled equivalent to `ViewComponent` which was mostly an exercise to see how simple one could make a component system.
* standard Rails partials
* standard Rails partials leveraging the `collection:` parameter.

To see the speed difference between these methods:
* clone the app https://github.com/bunnymatic/rails-playtime/ and use branch `chores/view-component-benchmarking` (or PR https://github.com/bunnymatic/rails-playtime/pull/8)
* `bin/setup`
* `bundle exec rails s`
* open http://localhost:3000?count=100
* click through the nav and notice the timing (from `rack_mini_profiler`) in the top left corner on each page.
* you should see that the timing compared for these methods is roughly 70:35:25:2.5 (in the same order as the list above) meaning `ViewComponent` renders 2x slower than my home grown version, ~3x slower than simple partials, and ~30x slower than when you render partials and leverage the `:collection` param.

### Expected behavior

ViewComponent should be at least as fast as my dumb homegrown version or at least closer to the rendering speed of Rails partials.

### Actual behavior

I ran the app locally (MacBookPro 2015 2.5 GHz i7 16GB) and pull the following from my Rails development logs.

Using a simple partial in a loop
```
Started GET "/partials?count=400" for ::1
Completed 200 OK in 225ms (Views: 224.0ms | ActiveRecord: 0.0ms)
Completed 200 OK in 233ms (Views: 232.8ms | ActiveRecord: 0.0ms)
Completed 200 OK in 222ms (Views: 221.9ms | ActiveRecord: 0.0ms)
```

Using `render partial:` with `collection:`
```
Started GET "/partials_collection?count=400"
[...]

PR #372, opened 2020-06-11
Enabling caching in development, and discard on each request.

fsateler:
This is the same strategy as followed by ActionView.

Closes: #345

### Summary

ViewComponent will on development never cache templates, and thus they are always recompiled. If a given component is rendered in a loop (imagine a component that is rendered on a table), the recompile time starts to add up.

OTOH, ActionView caches, but when ActionView::Resolver.caching? is false, it registers a callback to clear the cache on each request:

https://github.com/rails/rails/blob/3cd1b19ff2e49642191c598d620b87ffe5ec6cda/actionview/lib/action_view/railtie.rb#L96-L99

Thus, the optimal solution would be to do something similar: invalidate the cache on each request.

### Other Information

I'm creating as draft because:

1. ~I have yet to test this.~
2. There is a test failure:

```
Failure:
ViewComponentTest#test_compiles_unreferenced_component [test/view_component/view_component_test.rb:387]:
Failed assertion, no message given.
```

I'm not sure why it would be expected that an unreferenced component is compiled

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Cache components per-request in development, preventing unnecessary recompilation during a single request.
+
+    *Felipe Sateler*
+
 # 2.8.0
 
 * Add `before_render`, deprecating `before_render_check`.
--- a/README.md
+++ b/README.md
@@ -780,7 +780,7 @@ ViewComponent is built by:
 
 |<img src="https://avatars.githubusercontent.com/simonrand?s=256" alt="simonrand" width="128" />|<img src="https://avatars.githubusercontent.com/fugufish?s=256" alt="fugufish" width="128" />|<img src="https://avatars.githubusercontent.com/cover?s=256" alt="cover" width="128" />|<img src="https://avatars.githubusercontent.com/franks921?s=256" alt="franks921" width="128" />|<img src="https://avatars.githubusercontent.com/fsateler?s=256" alt="fsateler" width="128" />|
 |:---:|:---:|:---:|:---:|:---:|
-|@simonrand|@fugufish|@cover|@franks921|fsateler|
+|@simonrand|@fugufish|@cover|@franks921|@fsateler|
 |Dublin, Ireland|Salt Lake City, Utah|Barcelona|South Africa|Chile|
 
 ## License
--- a/README.md
+++ b/README.md
@@ -778,10 +778,10 @@ ViewComponent is built by:
 |@blakewilliams|@seanpdoyle|@tclem|@nashby|@jaredcwhite|
 |Boston, MA|New York, NY|San Francisco, CA|Minsk|Portland, OR|
 
-|<img src="https://avatars.githubusercontent.com/simonrand?s=256" alt="simonrand" width="128" />|<img src="https://avatars.githubusercontent.com/fugufish?s=256" alt="fugufish" width="128" />|<img src="https://avatars.githubusercontent.com/cover?s=256" alt="cover" width="128" />|<img src="https://avatars.githubusercontent.com/franks921?s=256" alt="franks921" width="128" />|
-|:---:|:---:|:---:|:---:|
-|@simonrand|@fugufish|@cover|@franks921|
-|Dublin, Ireland|Salt Lake City, Utah|Barcelona|South Africa|
+|<img src="https://avatars.githubusercontent.com/simonrand?s=256" alt="simonrand" width="128" />|<img src="https://avatars.githubusercontent.com/fugufish?s=256" alt="fugufish" width="128" />|<img src="https://avatars.githubusercontent.com/cover?s=256" alt="cover" width="128" />|<img src="https://avatars.githubusercontent.com/franks921?s=256" alt="franks921" width="128" />|<img src="https://avatars.githubusercontent.com/fsateler?s=256" alt="fsateler" width="128" />|
+|:---:|:---:|:---:|:---:|:---:|
+|@simonrand|@fugufish|@cover|@franks921|fsateler|
+|Dublin, Ireland|Salt Lake City, Utah|Barcelona|South Africa|Chile|
 
 ## License
 
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -3,6 +3,7 @@
 require "action_view"
 require "active_support/configurable"
 require "view_component/collection"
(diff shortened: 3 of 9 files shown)


fsateler:
I have tested this code:

```erb
  <% Rack::MiniProfiler.step('component') do %>
  <% 10000.times do %>
    <%= render SpecialInput::CheckboxComponent.new(form: f, name: 'test', title: 'asdf') %>
    <br/>
  <% end %>
  <% end %>
```

Before this change:

image

7.6 seconds in rendering 10_000 components.

After:

image

0.6 seconds rendering the same 10_000 components.

Content does change after a change in the template

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # master
 
+* Enable caching in development, but discard on each request. Prevents unnecessary recompiles during the same request.
```
```suggestion
* Cache components per-request in development, preventing unnecessary recompilation during a single request.
```

lib/view_component/base.rb
```diff
@@ -156,6 +156,25 @@ def request
...
     @@test_controller = "ApplicationController"
 
+    module CompileCache
+      mattr_reader :cache, instance_reader: false, instance_accessor: false do
+        Set.new
+      end
+      module_function
+
+      def register(clazz)
+        cache << clazz
```
joelhawksley, line 166:
```suggestion
      def register(klass)
        cache << klass
```

lib/view_component/base.rb
```diff
@@ -156,6 +156,25 @@ def request
...
+      end
+
+      def compiled?(clazz)
+        cache.include? clazz
```
joelhawksley, line 170:
```suggestion
      def compiled?(klass)
        cache.include? klass
```

lib/view_component/base.rb
```diff
@@ -156,6 +156,25 @@ def request
     mattr_accessor :test_controller
     @@test_controller = "ApplicationController"
 
+    module CompileCache
```
joelhawksley, line 159: `base.rb` is getting pretty big! What do you think about putting this in a separate file and including it?

test/view_component/view_component_test.rb
```diff
@@ -290,29 +288,24 @@ def test_template_changes_are_not_reflected_in_production
...
     render_inline(MyComponent.new)
-
-    ActionView::Base.cache_template_loading = old_value
   end
 
-  def test_template_changes_are_reflected_outside_production
-    old_value = ActionView::Base.cache_template_loading
-    ActionView::Base.cache_template_loading = false
-
+  def test_template_changes_are_reflected_after_invalidate
```
joelhawksley, line 293: What if we moved this test `integration_test.rb` and reloaded the page instead? Then we wouldn't have to call `ViewComponent::Base::CompileCache.invalidate!`, and our test would more closely resemble the behavior we're trying to implement.

fsateler:
All done!

I had to configure `cache_template_loading` in the test app config though for this to work.

fsateler:
`test_compiles_unrendered_component` is failing on CI but I can't reproduce locally. How can I test the exact rails version?

joelhawksley:
@fsateler if you prefix your commands with `RAILS_VERSION=5.2.3`  you should be able to reproduce the issue.

https://github.com/github/view_component/blob/2c8f4f83a8ef8ca5f172090d36efbd8e6e0bbd54/Gemfile#L6

fsateler:
Weird. I still can't reproduce this locally.  What other difference could there be between CI and a local clone?

simonrand:
@fsateler nice work with this! I was only thinking the other day this patch would be helpful for development performance.. 

FWIW I think the test failures are a test run ordering issue, if `IntegrationTest` is run before `ViewComponentTest` it will have invalidated the cache which will lead to.

I'd think we should be running tests with `config.action_view.cache_template_loading = true` so we're testing with a more production-like setup - but then your integration spec testing the reload behaviour will need to change https://github.com/github/view_component/pull/372/files#diff-008155519288d348c3cc3bbfba09ea30R41-R55

fsateler:
@simonrand good catch!

I don't think that works though, since independent of how I invalidate the cache, should IntegrationTest run before, ViewComponentTest will fail. More changes are needed. I made the `cache` module property writable so we can swap it and not alter the global state.

test/view_component/integration_test.rb
```diff
@@ -38,6 +38,31 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_includes inline_response, baseline_response
   end
 
+  test "template changes are reflected on new request on development" do
```
joelhawksley, line 41: 👏 nicely done! Would you mind adding a similar integration test proving that template changes are _not_ reflected when `cache_template_loading` is `true`?

README.md
```diff
@@ -778,10 +778,10 @@ ViewComponent is built by:
 |@blakewilliams|@seanpdoyle|@tclem|@nashby|@jaredcwhite|
 |Boston, MA|New York, NY|San Francisco, CA|Minsk|Portland, OR|
 
-|<img src="https://avatars.githubusercontent.com/simonrand?s=256" alt="simonrand" width="128" />|<img src="https://avatars.githubusercontent.com/fugufish?s=256" alt="fugufish" width="128" />|<img src="https://avatars.githubusercontent.com/cover?s=256" alt="cover" width="128" />|<img src="https://avatars.githubusercontent.com/franks921?s=256" alt="franks921" width="128" />|
-|:---:|:---:|:---:|:---:|
-|@simonrand|@fugufish|@cover|@franks921|
-|Dublin, Ireland|Salt Lake City, Utah|Barcelona|South Africa|
+|<img src="https://avatars.githubusercontent.com/simonrand?s=256" alt="simonrand" width="128" />|<img src="https://avatars.githubusercontent.com/fugufish?s=256" alt="fugufish" width="128" />|<img src="https://avatars.githubusercontent.com/cover?s=256" alt="cover" width="128" />|<img src="https://avatars.githubusercontent.com/franks921?s=256" alt="franks921" width="128" />|<img src="https://avatars.githubusercontent.com/fsateler?s=256" alt="fsateler" width="128" />|
+|:---:|:---:|:---:|:---:|:---:|
+|@simonrand|@fugufish|@cover|@franks921|fsateler|
```
joelhawksley, line 783:
```suggestion
|@simonrand|@fugufish|@cover|@franks921|@fsateler|
```

PR #381, opened 2020-06-17
Fix Ruby 2.7 warnings

joelhawksley:
### Summary

This PR makes a couple private changes to address warnings when running the test suite in Ruby 2.7.

It also sets our default ruby version for local development to `2.7.1`.


joelhawksley:
@kytrinyx I was seeing the warnings in the test output for the library on my local machine. I _think_ that should be sufficient validation for these changes, right?

kytrinyx:
> I was seeing the warnings in the test output for the library on my local machine.

Oh nice! in that case I think we're all good. Thank you 💖

PR #384, opened 2020-06-20
add support for multiple preview paths

tomasc:
### Summary

Change `config.preview_path` to `config.preview_paths` to support multiple locations of component preview files.
This is helpful for example when loading previews from multiple engines or sub-gems in a mono-repo.

### Other Information

Unfortunately, this would be a breaking change. Of course, there would be a way to support both `config.preview_path` and `config.preview_paths`, but I am not sure whether it's worth the weight added by the additional code?


jonspalmer:
I don't think we could/should release this as a breaking change.

Can we talk a little about the motivation for this change? You mention a mono-repo but how is it structured? If you're mono repo includes multiple engines I'd kind of expect them to have separate preview pages. 

The other pattern I could imaging is allowing previews to 'include' entire preview suites from a gem. Storybook is/has added this feature and it's powerful. I'd encourage a different mechanism for that where you would register the gem for previews and in an. ideal world it would autodiscover the right preview paths etc. Such support would be some new feature for view-components and allow us to keep a singular preview_path.

tomasc:
Hi @jonspalmer, thanks for looking at this!

My use case is that I have a mono-repo, and on top of it I have a separate engine (sort of my own Storybook), that collects the previews, and organizes them by namespace. Instead of the default `index` view, I have my own interface that loads the previews (via AJAX), with basic `knobs` support.

I think having some sort of mechanism that would allow us to scope the previews per engine (and preview them there, scoped, as you suggest) would be a good start. My hunch though, looking at how the `ViewComponent::Engine` is currently implemented (and tied to `Rails.configuration`), is that refactoring the Engine might introduce breaking changes.

Including multiple preview suites is a great idea (that's what in the end after), however, I think the `view_component` gem should only make it possible (not actually implement it). One could then use Storybook or their own solution (as I do) to collect all previews from variety of locations.

Changing the `preview_path` to `preview_paths` however to me still makes sense. It more closely mirrors other Rails' path configurations (f.e. `view_paths` or `autoload_paths`) and would be a quick win making the gem more flexible. That being said I absolutely understand your concerns about breaking change. I could try to see how to support both `preview_paths` and `preview_path` for backward compatibility.

What do you think?

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # master
 
+# Unreleased
+
```
joelhawksley, line 4:
```suggestion
```

joelhawksley:
> Changing the preview_path to preview_paths however to me still makes sense. It more closely mirrors other Rails' path configurations (f.e. view_paths or autoload_paths) and would be a quick win making the gem more flexible. That being said I absolutely understand your concerns about breaking change. I could try to see how to support both preview_paths and preview_path for backward compatibility.

I'm OK with that approach. Let's move forward with that approach, deprecating `preview_path` for removal in `v3.0.0`.

tomasc:
@jonspalmer @juanmanuelramallo @joelhawksley the PR has been updated to preserve the `config.preview_path` option, but deprecate it in favor of `config.preview_paths`.

When specified, the deprecated `config.preview_path` will be added on top of `config.preview_paths`.

Please take a look and let me know in case you have any suggestions for changes. Thanks!

PR #393, opened 2020-07-03
Avoid printing the missing capybara dependency warning by default

rmacklin:
### Summary

https://github.com/github/view_component/issues/304#issuecomment-651419797 mentioned being annoyed by the warning about a missing capybara dependency in an application which uses view_component but not capybara.

My first instinct was to simply remove the warning. However, it still seems helpful for users who want to use the capybara assertions but are getting errors because they don't have a capybara dependency.

So, in this PR, I've adjusted it such that the warning is silenced unless the `DEBUG` environment variable is set. This seemed like a good compromise, since it seems reasonable that running tests with `DEBUG=1` may result in additional output for debugging purposes.

That said, I realize we aren't using the `DEBUG` environment variable elsewhere in the gem, so if we'd rather avoid it and simply remove the warning altogether, that's fine too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Only print warning about a missing capybara dependency if the `DEBUG` environment variable is set.
+
+    *Richard Macklin*
+
 # 2.13.0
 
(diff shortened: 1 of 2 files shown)


joelhawksley:
@rmacklin thanks for the fix! Mind adding a CHANGELOG entry?

rmacklin:
Done! (Didn't do it initially because I wasn't sure if this was worth an entry)

Issue #391, opened 2020-06-30
`render` in controller should respect layout


fsateler:
## Feature request

When using `render MyComponent.new` in a controller, the layout should be respected. See also https://github.com/github/view_component/issues/38#issue-487741087, which explicitly set the layout option.

### Motivation

I would like to stop using ActionView templates altogether. It is easier to work with a single view library than multiple ones, especially as "you need to switch from ActionView to ViewComponent" tends to be annoying when adding a small thing. Therefore, I'd like to have Page Components (eg, Users::ShowComponent, etc), and render them from the controller without going into an ActionView file first.

### More information

A simple solution like the one posted on https://github.com/github/view_component/issues/38#issue-487741087 is possible, but unfortunately uses a different view_context than for the layout itself. Therefore, sharing `content_for` with the layout is not possible.

joelhawksley:
@fsateler I agree that this should be possible. Might you be willing to open a PR with failing tests for the APIs you think should work?

mollerhoj:
Working on a relatively small project, this would be my exact use case as well. Would you consider allowing for a different name than "Component". It seems like a strange name of objects that represent entire views - Something along the lines of ViewModels seem like a better fit. Perhaps even just allowing placing the objects in the `view/` folder, so that the structure becomes

```
app/views/users/show.html.erb
app/views/users/show.rb    #replacing Users::ShowComponent with Users::Show
```

joelhawksley:
@mollerhoj I hear you! 

We started with building ViewComponent to leverage the `views` folder, but it became pretty messy to make things work cleanly with Rails. I can see why it would make sense to include them in the `views` folder though, so maybe that's a Rails feature that could be added?

mollerhoj:
I've created a new Issue to avoid hijacking this issue.

fsateler:
This issue will be worked in upstream rails: https://github.com/rails/rails/pull/39869

gap777:
@joelhawksley Is this issue really closed/resolved? In Rails 7.2.2.1, `render MyViewComponent.new(...)` seems to NOT include layout. 
I could use the workaround mentioned in https://github.com/ViewComponent/view_component/issues/38#issuecomment-1162287527, but really.... that shouldn't be necessary.
Thoughts?

PR #395, opened 2020-07-03
Add failing tests for rendering with layouts

fsateler:
### Summary

This is a merge request for discussing a possible API for implementing #391 

I would have liked to have layouts included by default with `render`. I would understand if that is deemed an API break and a alternative needs to be determined.

Additionally, I have added a test that checks the view context is shared between the layout and the component.


joelhawksley:
@fsateler would you be up for pairing on this one? Email me: joelhawksley@github.com

joelhawksley:
@fsateler would you be up for updating this PR?

test/sandbox/app/controllers/layouts_controller.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+# Tests for render with/without layouts
+class LayoutsController < ActionController::Base
+  layout "global_for_action", only: :global_for_action
```
fsateler, line 5: For some reason `"application"` layout is not searched... maybe this is new in rails, and we need an ApplicationController?

Spone: You were right :) Adding an `ApplicationController` makes the tests pass!

See https://api.rubyonrails.org/classes/ActionView/Layouts.html#module-ActionView::Layouts-label-Layout+assignment

Spone:
Can we merge this @joelhawksley?

Issue #396, opened 2020-07-07
Testing with render_inline fails when render_monkey_patch_enabled is set to false


czj:
We are running Rails < 6.1 and wanted to use the `render_component` syntax for clarity, but also speed and search-and-replace-ability in our codebase.

### Steps to reproduce

- Create a Rails 6.0.3.2 application
- Define that you don't want the `render` method monkey-patch

```ruby
Rails.application.config.view_component.render_monkey_patch_enabled = true
```

- Create a component like this one:

```ruby
# frozen_string_literal: true

module Eshop
  class ShowroomClosureAlertComponent < ApplicationComponent
    def initialize(showroom_closed:)
      @showroom_closed = showroom_closed
    end
  end
end
```

```erb
<% if showroom_closed %>
  <div class="alert alert-info text-center">
    Something in French from our website.
  </div>
<% end %>
```

- Create associated test like this one:

```ruby
require "test_helper"

module Ecommerce
  class ShowroomClosureAlertComponentTest < ViewComponent::TestCase
    test "component renders something useful" do
      node = render_inline ShowroomClosureAlertComponent.new(showroom_closed: false)
      assert_empty node.inner_html
    end
  end
end
```

### Expected behavior

The method `render_inline`is supposed to use `render_component` because we have disabled `render` monkey-patch.

### Actual behavior

The tests raise an exception because `render_inline` tries to use `render` :

```
ArgumentError: '#<Ecommerce::ShowroomClosureAlertComponent:0x00000001236774a0 @showroom_closed=true>' is not an ActiveModel-compatible object. It must implement :to_partial_path.
```

### System configuration

- Rails 6.0.3.2
- Ruby 2.6.6
- Gem 3.1.4
- Bundler 2.1.4

czj:
To me it looks like a fix would require the test helper https://github.com/github/view_component/blob/master/lib/view_component/test_helpers.rb to define `render_inline` differently if the monkey patch is not applied.

### Actual

```ruby
    def render_inline(component, **args, &block)
      @rendered_component = controller.view_context.render(component, args, &block)

      Nokogiri::HTML.fragment(@rendered_component)
    end
```

### Without the monkey patch

```ruby
def render_inline(component, **args, &block)
  @rendered_component = controller.view_context.render_component(component, args, &block)

  Nokogiri::HTML.fragment(@rendered_component)
end
```

joelhawksley:
@czj that looks great to me! Mind opening a PR?

czj:
[...]

PR #398, opened 2020-07-08
Use render_component in render_inline test helper

czj:
### Summary

Using `render_component` in the the `render_inline` test helper instead of `render` allows tests to pass on any version of Rails (>= 6.1 and < 6.1 with or without monkey patch always pass.

Without this fix, when running Rails < 6.1 without the monkey patch, using `render_inline`would result in tests failure.

Closes https://github.com/github/view_component/issues/396

### DIFF
--- a/README.md
+++ b/README.md
@@ -929,10 +929,10 @@ ViewComponent is built by:
 |@maxbeizer|@franco|@tbroad-ramsey|@jensljungblad|@bbugh|
 |Nashville, TN|Switzerland|Spring Hill, TN|New York, NY|Austin, TX|
 
-|<img src="https://avatars.githubusercontent.com/johannesengl?s=256" alt="johannesengl" width="128" />|
+|<img src="https://avatars.githubusercontent.com/johannesengl?s=256" alt="johannesengl" width="128" />|<img src="https://avatars.githubusercontent.com/czj?s=256" alt="czj" width="128" />|
 |:---:|
-|@johannesengl|
-|Berlin, Germany|
+|@johannesengl|@czj|
+|Berlin, Germany|Paris, France|
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -930,7 +930,7 @@ ViewComponent is built by:
 |Nashville, TN|Switzerland|Spring Hill, TN|New York, NY|Austin, TX|
 
 |<img src="https://avatars.githubusercontent.com/johannesengl?s=256" alt="johannesengl" width="128" />|<img src="https://avatars.githubusercontent.com/czj?s=256" alt="czj" width="128" />|
-|:---:|
+|:---:|:---:|
 |@johannesengl|@czj|
 |Berlin, Germany|Paris, France|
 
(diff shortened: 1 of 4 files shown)


lib/view_component/test_helpers.rb
```diff
@@ -20,7 +20,7 @@ def refute_component_rendered
     attr_reader :rendered_component
 
     def render_inline(component, **args, &block)
-      @rendered_component = controller.view_context.render(component, args, &block)
+      @rendered_component = controller.view_context.render_component(component, &block)
```
joelhawksley, line 23:
```suggestion
      @rendered_component = 
        if if Rails.version.to_f >= 6.1
          controller.view_context.render(component, &block)
        else
          controller.view_context.render_component(component, &block)
        end
```

joelhawksley:
@czj one suggested change so that this will work on 6.1, where `render_component` isn't available. 

Can you add a CHANGELOG entry and your name to the README?

lib/view_component/test_helpers.rb
krzysiek1507: Wouldn't it be better to check `Rails::VERSION::MAJOR >= 6 && Rails::VERSION::MINOR >= 1`?

czj:
Thanks @joelhawksley, changes have been applied.

I have one last question : is there a reason why the `render_component` method is not made available for Rails >= 6.1 ? I searched into commits and issues but did not find one.

For me, this method would be useful in any version of Rails, for two reasons :

- you can search your code for any ViewComponent render by using /render_component/ instead of doing a less reliable search for /Component\./
- using `render_component` should be faster and more memory efficient : you skip all the complexity of Rails' "all-in-one" `render` method (pass less arguments, instantiate less objects in memory, go through less method calls)

That seems like a sane alternative to me.

joelhawksley:
@czj https://github.com/github/view_component/pull/358#issuecomment-652882373 mentions the decision, but doesn't give a reason.

@johannesengl and I debated it a bit while pairing, and came back to the [design philosophy](https://github.com/github/view_component#design-philosophy) for the library, which is to integrate with Rails as cleanly as possible. I don't feel that adding an explicit `render_component` method is in line with that philosophy.

That being said, I'd be open to adding a configuration option to enable `render_component` in Rails 6.1+. Would you like to implement that in another PR?

README.md
```diff
@@ -929,10 +929,10 @@ ViewComponent is built by:
 |@maxbeizer|@franco|@tbroad-ramsey|@jensljungblad|@bbugh|
 |Nashville, TN|Switzerland|Spring Hill, TN|New York, NY|Austin, TX|
 
-|<img src="https://avatars.githubusercontent.com/johannesengl?s=256" alt="johannesengl" width="128" />|
+|<img src="https://avatars.githubusercontent.com/johannesengl?s=256" alt="johannesengl" width="128" />|<img src="https://avatars.githubusercontent.com/czj?s=256" alt="czj" width="128" />|
 |:---:|
```
joelhawksley, line 933:
```suggestion
|:---:|:---:|
```

PR #414, opened 2020-07-21
Slots return stripped HTML

joelhawksley:
While building a component that leveraged Slots,
we ran into an issue where it was difficult to not introduce
unintended whitespace in the rendered HTML for the component.

This change strips the rendered HTML for slots, so that
we don't end up with unintentional whitespace in the rendered
output.

Co-authored-by: Jason Long <jasonlong@github.com>

### DIFF
--- a/lib/view_component/slotable.rb
+++ b/lib/view_component/slotable.rb
@@ -98,7 +98,7 @@ module ViewComponent
       slot_instance = args.present? ? slot_class.new(**args) : slot_class.new
 
       # Capture block and assign to slot_instance#content
-      slot_instance.content = view_context.capture(&block) if block_given?
+      slot_instance.content = view_context.capture(&block).strip.html_safe if block_given?
 
       if slot[:collection]
         # Initialize instance variable as an empty array
--- a/test/view_component/integration_test.rb
+++ b/test/view_component/integration_test.rb
@@ -359,6 +359,11 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_select(".item.normal", count: 2)
 
     assert_select(".footer.text-blue h3", text: "This is the footer")
+
+    title_node = Nokogiri::HTML.fragment(response.body).css(".title").to_html
+    expected_title_html = "<div class=\"title\">\n    <p>This is my title!</p>\n  </div>"
+
+    assert_equal(title_node, expected_title_html)
   end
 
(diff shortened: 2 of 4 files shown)


lib/view_component/slotable.rb
```diff
@@ -98,7 +98,7 @@ def slot(slot_name, **args, &block)
       slot_instance = args.present? ? slot_class.new(**args) : slot_class.new
 
       # Capture block and assign to slot_instance#content
-      slot_instance.content = view_context.capture(&block) if block_given?
+      slot_instance.content = view_context.capture(&block).strip.html_safe if block_given?
```
BlakeWilliams, line 101: Will `.capture` always return `html_safe?` content?

I'm curious if it'd make sense to check if the captured content is `html_safe?` and only apply `html_safe` if it was already marked as HTML safe.

joelhawksley: @BlakeWilliams `capture` returns an `ActiveSupport::SafeBuffer`, so I think we're fine to assume that it is safe.

test/view_component/integration_test.rb
```diff
@@ -359,6 +359,11 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_select(".item.normal", count: 2)
 
     assert_select(".footer.text-blue h3", text: "This is the footer")
+
+    title_node = Nokogiri::HTML.fragment(response.body).css(".title").to_html
+    expected_title_html = "<div class=\"title\">\n    <p>This is my title!</p>\n  </div>"
```
jonrohan, line 364: What would we have to do to get this to succeed?

```suggestion
    expected_title_html = "<div class=\"title\"><p>This is my title!</p></div>"
```

joelhawksley: @jonrohan we'd have to strip the rendered output for the component itself. It's certainly doable 😄

jonrohan: The browser wouldn't treat these differently, but they are 10 bytes in size difference.

jduff: I recently made some changes to `asset_dom_equal` so that it ignores white space between elements. I’ve been using it to test the output of my view components and it has been working well. It’s in master but not a released version of the gem, but might be something to look at for this https://github.com/rails/rails-dom-testing/pull/84

PR #416, opened 2020-07-21
Strip whitespace between html tags

jonrohan:
### Summary

This addresses the feature brought up in https://github.com/github/view_component/pull/414/files#r458362252

This will strip out any whitespace characters between html tags in the rendered output. Browsers don't treat these differently when rendering them, but having the whitespace in the delivered html increases the amount of bytes downloaded. These savings are extremely small, but when you think about a giant page rendered entirely out of ViewComponents, this savings can add up.

**Before**

```html
<p>
    <a href="foo/">Foo</a>
</p>
```

**After**

```html
<p><a href="foo/">Foo</a></p>
```

**Note this will not remove space between text and tags**

```html
<strong>
    Hello
</strong>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # master
 
+* Remove whitespace between rendered html tags.
+
+    *Jon Rohan*
+
 # 2.17.0
 
 * Slots return stripped HTML, removing leading and trailing whitespace.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # master
 
-* Remove whitespace between rendered html tags.
+* Remove whitespace between rendered HTML tags.
 
     *Jon Rohan*
 
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -75,7 +75,7 @@ module ViewComponent
       before_render
 
       if render?
-        send(self.class.call_method_name(@variant)).gsub(/>[\n\s]+</, "><")
+        send(self.class.call_method_name(@variant)).gsub(/>[\n\s]+</, "><").html_safe
       else
         ""
       end
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -75,7 +75,7 @@ module ViewComponent
       before_render
 
       if render?
-        send(self.class.call_method_name(@variant))
+        send(self.class.call_method_name(@variant)).gsub(/>[\n\s]+</, "><")
       else
         ""
       end
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -75,7 +75,7 @@ module ViewComponent
       before_render
 
       if render?
-        send(self.class.call_method_name(@variant)).gsub(/>[\n\s]+</, "><").html_safe
+        send(self.class.call_method_name(@variant)).gsub(/>[\n\s]+</, "><").strip.html_safe
       else
         ""
       end
--- a/test/view_component/integration_test.rb
+++ b/test/view_component/integration_test.rb
@@ -360,7 +360,7 @@ class IntegrationTest < ActionDispatch::IntegrationTest
 
     assert_select(".footer.text-blue h3", text: "This is the footer")
 
-    title_node = Nokogiri::HTML.fragment(response.body).css(".title").to_html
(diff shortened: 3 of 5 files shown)


test/view_component/integration_test.rb
```diff
@@ -27,7 +27,7 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_select("div", "bar")
     assert_response :success
 
-    baseline_response = response.body
+    baseline_response = response.body.gsub(/>[\s\n]+</, "><").strip
```
jonrohan, line 30: These baseline controller renders included the whitespace and were failing the test because they weren't equal.

lib/view_component/base.rb
```diff
@@ -75,7 +75,7 @@ def render_in(view_context, &block)
       before_render
 
       if render?
-        send(self.class.call_method_name(@variant))
+        send(self.class.call_method_name(@variant)).gsub(/>[\n\s]+</, "><").strip.html_safe
```
jonrohan, line 78: I'm not sure if this is the right place or we should extract it?

test/view_component/integration_test.rb
```diff
@@ -360,10 +360,10 @@ class IntegrationTest < ActionDispatch::IntegrationTest
 
     assert_select(".footer.text-blue h3", text: "This is the footer")
 
-    title_node = Nokogiri::HTML.fragment(response.body).css(".title").to_html
-    expected_title_html = "<div class=\"title\">\n    <p>This is my title!</p>\n  </div>"
+    title_node = Nokogiri::HTML.fragment(response.body).css(".title").to_html.strip
```
jonrohan, line 363: for some reason here I was ending up with a trailing `\n"`

test/view_component/integration_test.rb
```diff
@@ -360,10 +360,10 @@ class IntegrationTest < ActionDispatch::IntegrationTest
...
+    expected_title_html = "<div class=\"title\"><p>This is my title!</p></div>"
 
-    assert_equal(title_node, expected_title_html)
+    assert_equal(expected_title_html, title_node)
```
jonrohan, line 366: Swapped these, `assert_equal(expected, actual)` the test suite was calling them this way and it was confusing.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # master
 
+* Remove whitespace between rendered html tags.
```
```suggestion
* Remove whitespace between rendered HTML tags.
```

lib/view_component/base.rb
joelhawksley: This is probably fine. Not being a _regex-pert_ myself, do you think we need any more test cases for this replacement pattern?

jonrohan: This should remove spaces and newlines. which is covered by this test https://github.com/github/view_component/pull/416/files#diff-c7432794f4e04debce14f00887227594R648-R654 (the test partial has both)

jduff:
Wouldn't gzipping responses already give you most of the benefits of this?

If you really want to compact your html would it make more sense to do that at the rails render level or with rack (I think there is a deflate middleware in rack that might do this) and pay the cost once instead of with each component render?

Maybe this `gsub` and the `strip` that was added in a previous commit isn't too costly, but I imagine it would add up if you have a lot of components being rendered.

I'm new to using this library so please ignore me if this has already been considered :).

jonrohan:
> Wouldn't gzipping responses already give you most of the benefits of this?

You could be right, It's a cloudy area for me as to what I've heard about the area. Mozilla has a article that mentions stripping whitespace https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Author_fast-loading_HTML_pages#Reduce_page_weight 

In another life, I wrote django and there's this tag `spaceless` https://docs.djangoproject.com/en/3.0/ref/templates/builtins/#spaceless that would do this. And I've often wished I could bring this functionality to github.com. I have no idea if rails has a solution, but would be rad if it did. 

> Maybe this `gsub` and the `strip` that was added in a previous commit isn't too costly, but I imagine it would add up if you have a lot of components being rendered.

Yeah this could be a concern also. Until we can prove that this does or doesn't increase rendering time, It's probably best to let it sit or close it for now. For the maintainers here, don't worry about closing this, but I would love if we could think about how to deliver the smallest fasted html for components. ⚡

joelhawksley:
Thanks @jonrohan! Closing for now ❤️

PR #421, opened 2020-07-27
Fix auto-loading of previews

mattbrictson:
### Summary

Currently, the Rails server has to be manually restarted in order to pick up changes to view_component preview files. This is a bug. Auto-loading for previews worked before, but I believe it was inadvertently broken in e45338f.

This PR restores the auto-loading behavior for previews. Now if you make a change to a preview `.rb` file, the changes take effect without a server restart.


BlakeWilliams:
Thanks for the contribution!

PR #424, opened 2020-07-27
100% Test coverage

joelhawksley:
### Summary

This PR get us to 100% test coverage 🎊 by adding some missing tests and perhaps more importantly, collating our coverage results from all of our matrix CI builds.


lib/view_component/preview.rb
```diff
@@ -42,21 +42,11 @@ def render_args(example, params: {})
         result.merge(layout: @layout)
       end
 
-      # Returns the component object class associated to the preview.
-      def component
```
BlakeWilliams, line 46: Could this be being used by apps using previews? 🤔

joelhawksley: @juanmanuelramallo Can you have a look at this PR? There was some unused/untested preview code.

lib/view_component/base.rb
```diff
@@ -128,7 +128,9 @@ def view_cache_dependencies
 
     # For caching, such as #cache_if
     def format
+      # :nocov:
       @variant
+      # :nocov:
```
jonspalmer, line 133: Why nocov here? Seems like this is just a gap.

lib/view_component/base.rb
```diff
@@ -175,7 +177,10 @@ def with_collection(collection, **args)
 
       # Provide identifier for ActionView template annotations
       def short_identifier
+        # This is only called in 6.1, which we don't run coverage measurements against
```
jonspalmer, line 180: Why aren't we running coverage against 6.1?

lib/view_component/base.rb
```diff
@@ -293,7 +299,9 @@ def type
       end
 
       def format
+        # :nocov:
         :html
+        # :nocov:
```
jonspalmer, line 304: Why nocov here isn't this just a gap?

test/app/controllers/integration_examples_controller.rb
```diff
@@ -16,6 +16,9 @@ def controller_inline_baseline
   end
 
   def controller_to_string
+    # ensure render_to_string isn't broken by monkey patch
+    render_to_string("integration_examples/_controller_inline", locals: { message: "bar" })
```
jonspalmer, line 20: I'm not sure I understand what this code does. Can you explain a little

test/view_component/integration_test.rb
```diff
@@ -374,12 +376,14 @@ class IntegrationTest < ActionDispatch::IntegrationTest
...
 
   if Rails.version.to_f >= 6.1
+    # :nocov:
     test "rendering component using the render_component helper raises an error" do
       error = assert_raises ActionView::Template::Error do
         get "/render_component"
       end
       assert_match /undefined method `render_component'/, error.message
     end
+    # :nocov:
```
jonspalmer, line 386: why nocov here?

lib/view_component/preview.rb
juanmanuelramallo: Just looked over the production code of previews and it's indeed not being used. Safe to remove

test/app/controllers/integration_examples_controller.rb
joelhawksley: @jonspalmer I've added a comment 😄

joelhawksley:
@jonspalmer I've reworked this PR to instead combine coverage data from all of our builds. Mind taking another look?

Issue #426, opened 2020-07-28
Haml error: ActionView::Template::Error (undefined method `capture_haml' ...)


spdawson:
### Steps to reproduce

Minimal reproduction [repository](https://github.com/spdawson/view_component_haml_error)

### Expected behavior

Component should render without incident, when using Haml template.

### Actual behavior

Error is raised:

```
ActionView::Template::Error (undefined method `capture_haml' for #<#<Class:0x00007fc4fcb73338>:0x00007fc4fcaca440>
Did you mean?  capture_with_haml):
```

### System configuration
**Rails version**: 6.0.3.2

**Ruby version**: 2.7.1

**Gem version**: 2.18.0

joelhawksley:
@spdawson thank you for the report! We definitely want to make sure `haml` works with ViewComponent. Would you be willing to write a PR with a test case for this issue?

spdawson:
@joelhawksley I've added a test case in #428

joelhawksley:
Thanks @spdawson! Would you be up for pairing with me on this one? Send me an email: joelhawksley@github.com

dmarcoux:
I would be willing to help if it's needed. I plan to use view components with Haml at work. If Haml doesn't work with view components right now, it's a blocker. Let me know how I can help.

joelhawksley:
@dmarcoux send me your pairing availability for the next week to joelhawksley@github.com ❤️

dmarcoux:
> @dmarcoux send me your pairing availability for the next week to [joelhawksley@github.com](mailto:joelhawksley@github.com) heart

Done :slightly_smiling_face:

manuelpuyol:
hey @spdawson @dmarcoux 
I was taking a look at this issue and the [repository provided](https://github.com/spdawson/view_component_haml_error), I noticed it is using `helpers.link_to`.
As https://github.com/github/view_component/pull/428#discussion_r468713628 explained, `helpers.` is unnecessary in this situation and removing it solved the issue (at least in the provided repo)

Could you confirm that removing `helpers.` fixes the issue? Also, is there any specific reason as why you are using `helpers.`?

spdawson:
> Could you confirm that removing `helpers.` fixes the issue? Also, is there any specific reason as why you are using `helpers.`?

Yes, removing the `helpers.` scope _does_ resolve this issue for me; thanks @dmarcoux!

morgler:
[...]

PR #428, opened 2020-07-29
Add a test case for a Haml template that uses a block

spdawson:
### Summary

Add a test case for #426.

### DIFF
--- a/test/app/components/haml_component.html.haml
+++ b/test/app/components/haml_component.html.haml
@@ -1,3 +1,8 @@
 %div
   = content
   = @message
+
+- if @url
+  %div
+    = helpers.link_to(@url) do
+      hello!
--- a/test/view_component/view_component_test.rb
+++ b/test/view_component/view_component_test.rb
@@ -61,6 +61,14 @@ class ViewComponentTest < ViewComponent::TestCase
     assert_text("bar")
   end
 
+  def test_renders_haml_template_that_uses_a_block
+    render_inline(HamlComponent.new(message: "bar", url: "http://example.com")) { "foo" }
+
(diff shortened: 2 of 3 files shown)


test/view_component/view_component_test.rb
```diff
@@ -61,6 +61,14 @@ def test_renders_haml_template
     assert_text("bar")
   end
 
+  def test_renders_haml_template_that_uses_a_block
+    render_inline(HamlComponent.new(message: "bar", url: "http://example.com")) { "foo" }
+
+    assert_selector("a[href='http://example.com']", "hello!")
```
dmarcoux, line 67:
```suggestion
    assert_selector("a[href='http://example.com']", text: "hello!")
```

Once my previous suggestion is applied, this line will cause the following error:

```
Error:
ViewComponentTest#test_renders_haml_template_that_uses_a_block:
ArgumentError: Unused parameters passed to Capybara::Queries::SelectorQuery : ["hello!"]
```

This suggestion fixes the issue.

test/app/components/haml_component.html.haml
```diff
@@ -1,3 +1,8 @@
 %div
   = content
   = @message
+
+- if @url
+  %div
+    = helpers.link_to(@url) do
```
dmarcoux, line 7:
```suggestion
    = link_to(@url) do
```

Using `helpers` here takes us to the view context of this template's controller. See the definition of [helpers in lib/view_component/base.rb](https://github.com/github/view_component/blob/c0129a93fbd9d7f37e0c85c82984a5d05b9f3fc1/lib/view_component/base.rb#L114-L117). It's unneeded for the `link_to` helper. It's already available in the template's context.

PR #431, opened 2020-07-30
Fix default_preview_layout configuration example

eebs:
The configuration example for `default_preview_layout` was missing the `view_component` namespace.

Bumped into this when configuring previews. Thanks for a wonderful gem!


joelhawksley:
@eebs thanks for the PR! Keep 'em coming 😉

PR #433, opened 2020-08-01
Previews with render_component fail when render_monkey_patch_enabled is set to false

mixergtz:
We are running Rails **5.2.4** and recently tried to updated the gem to `2.18`. When changing all `render` occurrences to `render_component` we realized Previews were raising a `NoMethodError` for `render_component`

### Steps to reproduce
- Create a Rails < 6.1 application (in our case is 5.2.4.1)
- Define that you don't want the `render` method monkey-patch

`Rails.application.config.view_component.render_monkey_patch_enabled = false`

- Create simple component like this one:
```ruby
class TestComponent < ViewComponent::Base
  def initialize(title:)
    @title = title
  end
end
```

```ruby
<span title="<%= @title %>"><%= content %></span>
```

- Create associated Preview like this one
```ruby
def with_default_title
  render_component(TestComponent.new(title: "Test component default")) { "Hello" }
end
```

- Go to 
```
http://localhost:3000/rails/view_components/test_component/with_default_title
```

### Expected
The method `render_component` should render the component because we have disabled `render` monkey-patch

### Actual behavior
The preview raises an exception because `render_component` isn't defined in ViewComponent::Preview

Can be easily reproduced by using https://github.com/mixergtz/view_component_fix/


mixergtz:
@joelhawksley This is a first approximation to solve this bug. I'm more than happy to work on feedback 😄

lib/view_component/preview.rb
```diff
@@ -17,6 +17,15 @@ def render(component, **args, &block)
       }
     end
 
+    def render_component(component, &block)
```
joelhawksley, line 20: Should this accept `args`?

joelhawksley:
@juanmanuelramallo I'd love to get your ✅  on this PR, as you have owned a lot of the Previews work ❤️

lib/view_component/preview.rb
juanmanuelramallo: Yes, otherwise [this feature](https://github.com/github/view_component#:~:text=It's%20also%20possible%20to%20set%20dynamic%20values%20from%20the%20params%20by%20setting%20them%20as%20arguments) won't work

lib/view_component/preview.rb
```diff
@@ -17,6 +17,15 @@ def render(component, **args, &block)
...
+      {
+        block: block,
+        component: component,
+        locals: {},
+        template: "view_components/preview",
+      }
+    end
```
juanmanuelramallo, line 27: Can we make this method an alias of `ViewComponent::Preview#render`?

test/view_component/integration_test.rb
```diff
@@ -397,6 +397,14 @@ class IntegrationTest < ActionDispatch::IntegrationTest
       get "/controller_to_string_render_component"
       assert_includes response.body, "bar"
     end
+
+    test "rendering component in preview using #render_component and monkey patch disabled" do
+      old_default = ViewComponent::Base.render_monkey_patch_enabled
+      ViewComponent::Base.render_monkey_patch_enabled = false
```
juanmanuelramallo, line 403: nitpick: What do you think about creating a helper method for overriding this config?

```ruby
# test_helper.rb

...

def with_render_monkey_patch_config(enabled)
  old_default = ViewComponent::Base.render_monkey_patch_enabled
  ViewComponent::Base.render_monkey_patch_enabled = enabled
  yield
ensure
  ViewComponent::Base.render_monkey_patch_enabled = old_default
end
```

And then use it here like:

```ruby
# test/view_component/integration_test.rb

...

test "rendering component ..." do
  with_render_monkey_patch_config(false) do
    get "/rails/view_components/monkey_patch_disabled_component/default"
    assert_includes response.body, "<div>hello,world!</div>"
  end
end

**Comment by mixergtz** [comment]
@juanmanuelramallo Thanks a lot for your feedback! Sorry for the delay, changes are ready :)

File: CHANGELOG.md [hunk 4]
```diff
@@ -1,5 +1,9 @@
 # master
 
+* Fix bug in previews when disabling render monkey patch
```
```suggestion
* Fix bug where previews didn't work when monkey patch was disabled.
```

joelhawksley:
@juanmanuelramallo let me know when this looks good to you and I'll merge 😄

app/views/view_components/preview.html.erb
```diff
@@ -1 +1,5 @@
-<%= render(@render_args[:component], @render_args[:args], &@render_args[:block])%>
+<% if ViewComponent::Base.render_monkey_patch_enabled || Rails.version.to_f >= 6.1 %>
+  <%= render(@render_args[:component], @render_args[:args], &@render_args[:block])%>
+<% else %>
+  <%= render_component(@render_args[:component], &@render_args[:block])%>
```
juanmanuelramallo, line 4: Don't we need to pass the args in here too?

```suggestion
  <%= render_component(@render_args[:component], @render_args[:args], &@render_args[:block])%>
```

juanmanuelramallo:
@joelhawksley Looking good! just left a non-blocking question

app/views/view_components/preview.html.erb
mixergtz: That `render_component` is `ViewComponent::RenderComponentHelper#render_component` which doesn't support `args` so an error will be raised.

I checked the dynamic values feature and it works as expected

juanmanuelramallo: no need, this answer is great! Thank you for checking!

orendon:
🎉

PR #438, opened 2020-08-05
Update bundler version to 2.1.4

dmarcoux:
### Summary

This updates bundler to the latest version.

### Other Information

I don't believe this should be in the `CHANGELOG.md`.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -209,4 +209,4 @@ DEPENDENCIES
   view_component!
 
 BUNDLED WITH
-   1.17.3
+   2.1.4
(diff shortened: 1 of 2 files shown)


dmarcoux:
The CI runs for `rails 5.0.0` depends on `bundler < 2.0`, so this prevents using a more modern version of bundler.

joelhawksley:
@dmarcoux could we potentially load a different version of bundler depending on the rails version?

dmarcoux:
Oh right, I might have been a bit too strict with my versioning in the gemspec.

dmarcoux:
I'll have to try in another PR since I cannot reopen this PR. Whoopsie!

PR #439, opened 2020-08-05
Update bundler version to 2.1.4

dmarcoux:
### Summary

This updates bundler to the latest version.

### Other Information

I don't believe this should be in the `CHANGELOG.md`.


dmarcoux:
Oh well, just like #438, it doesn't work. My [previous comment](https://github.com/github/view_component/pull/438#issuecomment-669296411) still holds.

PR #441, opened 2020-08-07
[Idea] Allow using component inside sidecar folder 

mixergtz:
Hi again! @joelhawksley 

We currently have ~40 components and growing! and we were wondering if there is a way to place the main component class inside the sidecar folder. That would help us to keep a tidy `/components` folder and search for component files faster (due to folders being listed first than files)

I tried the following structure:
```
components
  |--> test_component
          |--> test_component.rb
          |--> test_component.html.erb
```
But it raises an error for `TestComponent` for being defined as a `Module` and as a `Class`

Defining a different name for the class can be a bit confusing and also causes a disconnection between the Component and its sidecar assets. 

I did a small PoC using `base` as a default name for the main file inside the sidecar folder. It is working but the test is failing randomly. 

Before continuing I'd like to know if this is something worth pursuing? What are your thoughts about this use case? or maybe I missed something in the docs?

Happy to hear from you 😄

### DIFF
--- a/coverage/coverage.txt
+++ b/coverage/coverage.txt
@@ -3,7 +3,7 @@ Incomplete test coverage
 --------------------------------------------------------------------------------
 
 /app/controllers/view_components_controller.rb: 98.18% (missed: 59)
-/lib/view_component/base.rb: 97.87% (missed: 178,268,296,359)
+/lib/view_component/base.rb: 97.91% (missed: 178,268,296,359)
 /lib/view_component/engine.rb: 96.43% (missed: 22,25)
 /lib/view_component/preview.rb: 94.12% (missed: 49,59,109)
 /lib/view_component/render_to_string_monkey_patch.rb: 83.33% (missed: 9)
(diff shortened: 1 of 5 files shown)


jonspalmer:
We're all very supportive of the desire to have support for sidecar components. Sadly the constraints of Ruby, namespacing and autoloading don't give us a lot of options.

It's worth noting that the pattern you describe here is already possible if you call your components component vs base. See for example this write up: https://fugu.dev/2020-04-16-working-with-view-component-and-stimulus-part-2/

For some folks this pattern works. Completely personally for me I don't want 40 component.rb files. I'd expect to find MyComponent to be in a class called MyComponent etc. That said its reasonable for us to support other patterns and let the consumer of this gem choose what works for them. Does using component vs base solve your problem or would some code be needed still?

mixergtz:
Thanks a lot for your answer @jonspalmer! I totally understand the limitations we have. 

Just tried out the `Test::Component` approach and it worked without problems. I think it should work for our use case.

Do you think this is something we can note in the Readme for other folks getting into the same issue? I'm happy to add it if you want

joelhawksley:
@mixergtz sure! A note in the README sounds good to me ❤️

PR #447, opened 2020-08-18
Fix Slot example template

tysongach:
### Summary

This was returning the footer object rather than its content.


joelhawksley:
Thanks! ❤️

PR #450, opened 2020-08-21
Action text support

helphop:
### Summary

The pull request add support for ActionText and the rich_text_area_tag in a ViewComponents that use it.
 
To Do:

1. Add integration tests
2. Add Unit tests

### Other Information

For this to work in Rails the  default_url_options[:host] must be set.

I put this in my config/environments/development.rb file:
Rails.application.routes.default_url_options[:host] = "localhost"

But it can be set any number of ways.


test/config/database.yml
```diff
@@ -0,0 +1,9 @@
+default: &default
+  adapter: sqlite3
+  pool: <%= ENV.fetch("RAILS_MAX_THREADS") { 5 } %>
+  timeout: 5000
+
+test:
+  <<: *default
+  database: db/test.sqlite3
+
```
jonspalmer, line 9: Do we have to add a db to make the test cases work? We could have a model without it being backed by a db

lib/view_component/base.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
-
+require "action_text/engine"
```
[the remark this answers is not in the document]
helphop: Originally I added these lines of code to the ViewComponent I was working on in my rails application.  Then I thought it would be helpful if this functionality was available in the ViewComponent gem out of the box so that I would not have to wire things each time I needed this functionality.  As ViewComponent will be added to the next Rails release I also thought this might be something that other users would benefit from.  
That being said I will look into being how this could be added conditionally if ActionText is enabled.  Would that work?

lib/view_component/base.rb
```diff
@@ -22,6 +22,11 @@ class Base < ActionView::Base
     class_attribute :slots
     self.slots = {}
 
+    #required for rich_text_area
+    def main_app
+      Rails.application.class.routes.url_helpers
```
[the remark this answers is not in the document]
helphop: I think you are correct and I will do more research to see how I can remove this dependency if possible.

test/config/database.yml
helphop: The reason for the db was to get the tests to pass. ActionText requires a db.  I agree that we could just provide instructions for each user to add this functionality.  Since Rails 6.1 is supposed to be released incorporating ViewComponents would there be a way to add an installer that would add the required code to the ApplicationComponent should they require ActionText in ViewComponents?

helphop: As per your suggestion I tried out creating an ApplicationComponent as follows:

`require "action_text/engine"
  include ActionText::Engine.helpers

  def main_app
     Rails.application.class.routes.url_helpers
  end
end`

Then my components inherit from this class.  This works.  However, there is still an issue with getting ViewComponents  to work with ActionText outside of Rails. ActionText has a file called tag_helper.rb which references rails:

```
    def rich_text_area_tag(name, value = nil, options = {})
      options = options.symbolize_keys

      options[:input] ||= "trix_input_#{ActionText::TagHelper.id += 1}"
      options[:class] ||= "trix-content"

      options[:data] ||= {}
      options[:data][:direct_upload_url] = main_app.rails_direct_uploads_url
      options[:data][:blob_url_template] = main_app.rails_service_blob_url(":signed_id", ":filename")

      editor_tag = content_tag("trix-editor", "", options)
      input_tag = hidden_field_tag(name, value, id: options[:input])

      input_tag + editor_tag
    end
  end
```

So I don't see how to get ActionText working with ViewComponent without the Rails dependency.

lib/view_component/base.rb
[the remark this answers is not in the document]
helphop: @rmacklin please see my comments in the following thread.  ActionText has a method rich_text_area_tag that relies specifically on Rails.

lib/view_component/base.rb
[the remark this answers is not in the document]
joelhawksley: > I will look into being how this could be added conditionally if ActionText is enabled 

I think that's the right way to go.

> As ViewComponent will be added to the next Rails 

As noted in https://github.com/github/view_component/issues/206, this is no longer on the roadmap. _However_, we do [prioritize drop-in compatibility](https://github.com/github/view_component#design-philosophy) with Rails.

test/config/database.yml
jonspalmer: I'm not disputing that you need Rails to use ViewComponent with or without ActionText. My question is what's the minimum we need with this change to support ActionText and to test it effectively. Maybe it will become more clear once you have fleshed out the test cases.

lib/view_component/base.rb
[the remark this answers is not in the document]
jonspalmer: I wonder if a way to solve the opt in would be for ViewComponent to encapsulate this logic in a Module/Concern that could be included. Either automatically if we can detect ActionText or just allow consumers to include it in their ApplicationComponent.

test/config/database.yml
helphop: @jonspalmer I just added conditionals to check if ActionText is loaded and if it is loaded then load the necessary libraries to support it.

 I don't know if they way I have done this is best practices, I'm just trying to find a solution and will gladly follow any advice.  I understand that the minimal changes to ViewComponent required is the goal.

helphop: I moved the support for action text to a module that checks if action text has been loaded.

helphop:
I'm not sure I can get tests to pass for ActionText in a component according to [https://github.com/rails/rails/issues/35200#issuecomment-476349869](url)

It needs a  Rails app to have it function.

joelhawksley:
@helphop thanks for your continued work on this PR! Would you be willing to resolve the merge conflicts?

helphop:
I've tried to pull and rebase and find the conflicts but I can't seem to find any.

helphop:
I'm looking for some assistance.  Github is complaining about conflicts but I can't find any conflicts in the file listed.  I can't find any conflicts in any of the files in this Pull Request.  

Can anyone point me in the right direction on how to resolve this issue?

helphop:
Github still says there are conflicts but I can't find the conflicts.  Any help appreciated?

joelhawksley:
@helphop you can remove `coverage.txt` entirely. It's no longer needed.

joelhawksley:
@helphop @rmacklin might y'all be able to have another look here when you have some time? ❤️

lib/view_component/action_textable.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module ActionTextable
+    if require("action_text")
```
rmacklin, line 5: This doesn't look like the correct `if` predicate. This is going to unconditionally require `action_text`, which returns `true` if it is the first time it's being required and `false` otherwise. This would have the undesired effect of requiring `action_text` even in apps that don't use (and don't want to load) it.

Instead, I think we'd want to check for a constant that is only defined if the application has loaded `action_test/engine` - maybe `if defined?(ActionText::Engine)` but I haven't tested that myself. (You'd want to test in an app that has ActionText and in one that doesn't, and make sure it behaves correctly.)

helphop:
> @helphop @rmacklin might y'all be able to have another look here when you have some time? ❤️

I did some work on this today.  For some reason now Github says that my commits do not belong to any of the branches.  I have no idea why that would be since I just  pushed my changes to this branch.

helphop:
Thanks @rmacklin. I will review this and run some tests.

On September 21, 2020, Nathan Jones <notifications@github.com> wrote:
> @rmacklin commented on this pull request.
>

> In lib/view_component/action_textable.rb
> <https://github.com/github/view_component/pull/450#discussion_r492236962>:
>

> @@ -0,0 +1,14 @@ +# frozen_string_literal: true + +module
ViewComponent + module ActionTextable + if require("action_text")
> This doesn't look like the correct if predicate. This is going to
> unconditionally require action_text, which returns true if it is the

> first time it's being required and false otherwise. This would have
> the undesired effect of requiring action_text even in apps that don't

> use (and don't want to load) it.
>
> Instead, I think we'd want to check for a constant that is only
> defined if the application has loaded action_test/engine - maybe if
> defined?(ActionText::Engine) but I haven't tested that myself. (You'd

> want to test in an app that has ActionText and in one that doesn't,
> and make sure it behaves correctly.)
>

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/github/view_component/pull/450#pullrequestreview-
492842625>, or unsubscribe
<https://github.com/notifications/unsubscribe-
auth/AAFY245BW3RF6U6RVMIKHSTSG6GB5ANCNFSM4QHWYLZA>.

joelhawksley:
> @helphop @rmacklin might y'all be able to have another look here when you have some time? 

Whoops! I meant to ping @jonspalmer, not @helphop  there! My apologies ❤️

helphop:
As per @rmacklin suggestion I changed the check for ActionText to check if the ActionText const is defined as well as the ActionText::Engine is defined.   

I tested it with an application that has ActionText and one without ActionText and it seems to work.

joelhawksley:
@helphop would you be up for bringing this PR up to date? Sorry this fell through the cracks ❤️

helphop:
I would be willing.  Unfortunately I have very limited time as I'm about to take my family on a vacation.  I still have trouble getting the tests to pass and seem to be at a impasse for my skill level. :(

joelhawksley:
Closing as stale. Please reopen if you'd like to continue working on this!

PR #451, opened 2020-08-23
Multiple templates per component, allowing arguments

### LINKED ISSUE #387: Support multiple views
## Feature request

Allow refactoring small view parts into a separate view, and be able to `render` them into the template. These views should not be available outside the component.

```
app/components/my_component/my_component.html.erb
app/components/my_component/a_part.html.erb
```

```erb
<%# my_component.html.erb %>
<%= render 'a_part' %>
```

### Motivation

Sometimes it is useful to separate a sub-component from the component. For example, if I want to render a popover with extra info about a particular model, I would want to extract that popover into a separate view. 

Alternative Solutions:

1. Extract a new component. Because the popover is not a generic one to be used in other parts of the application, it doesn't make much sense to create a (now public) component for it.
2. Use `content_for`. This is not suitable if the partial needs to take arguments. In my example, it wouldn't work if I'm displaying a list of models and I want to show the popover for each model.

Prior art:
ActionView does support partials, of course.
Cells does allow multiple views per cell, and allows using `render view: 'a_part'`.

### Possible solutions I see

1. Provide a method that prefixes the component dir and then goes on to call `render template: ` with the full path. I would suspect this is relatively slow given that `render` is slow in ActionView. 
2. Precompile all the templates into methods with a reserved prefix, and allow calling them from the view. For example, using `extra_views_{view_name}`.

fsateler:
### Summary

Fixes #387 

This PR builds on the branch by @joelhawksley and @fermion, and adds the ability to specify arguments for the generated functions.

The logic being that one may want to use templates to extract sections, without wanting to build a full component. For example, I could have a Table component, and I would want to have the row definition in a sidecar template. This requires being able to pass in each item for every iteration:

```ruby
  def initialize(models)
    @models = models
  end
  template_arguments :row, :model
end
```

```erb
<%# table_component.html.erb %>
<table>
  <% @models.each do |model| %>
    <%= call_row model: model %>
  <% end %>
</table>
```
```erb
<%# row.html.erb %>
<tr>
  <%# I can use `model` here %>
  <td><%= model.name  %></td>
</tr>
```

Conceptually, this is very similar to ActionView's usage of `locals:`. However, this way has the following benefits:

1. Performance is increased as the template needs to be compiled only once.
2. We don't need to `instance_exec` things in a custom-build binding, further improving performance.
3. Ruby checks our argument list.
4. We force the user to specify the locals, thus forcing them to specify the interface.

### Other Information

I'm submitting as a draft, because there are some important API issues that need discussing:

1. Should we always use keyword arguments? Maybe we should use positional arguments? (I happen to like keyword arguments, but it may not be everyone's cup of tea).
2. This doesn't allow default values for any parameter. Is this a problem that needs addressing?

### DIFF
--- a/lib/view_component/compiler.rb
+++ b/lib/view_component/compiler.rb
@@ -77,9 +77,12 @@ module ViewComponent
           redefinition_lock.synchronize do
             component_class.silence_redefinition_of_method(method_name)
             # rubocop:disable Style/EvalWithLocation
-            component_class.class_eval <<-RUBY, template[:path], 0
-            def #{method_name}
+            component_class.class_eval <<-RUBY, template[:path], -1
+            #{visibility}def #{method_name}(#{method_args})
+              old_buffer = @output_buffer if defined? @output_buffer
               #{compiled_template(template[:path])}
+            ensure
+              @output_buffer = old_buffer
             end
             RUBY
             # rubocop:enable Style/EvalWithLocation


BlakeWilliams:
Thanks for the PR! We're starting to think about slots v2 and this is a use-case we should definitely take into account. Would you be willing to share your thoughts on this discussion https://github.com/github/view_component/discussions/454?

jonspalmer:
I'm with @BlakeWilliams that this feels like a use case that Slots could/should support. 

If this doesn't get baked into slots I'd encourage us to think hard about the rendering API and making it "Rails like". I agree with the four points you mention - they seem consistent with ViewComponents strong APIs and cached templates. However, the `call_row` api feels awkward and not very Rails like. What I love about the ViewComponent implementation is that you call 'render' in a view just like you would for a partial etc. Can we make this similar. Could it be `render template: :row` or `render_template :row, ....` ?

fsateler:
Hi!

I remember discussing this with @joelhawksley , and he also said that slots v2 could take this role. Maybe it can, but I have not yet seen how that works. Discussion #454 doesn't mention slot templates, for example. I have posted my further thoughts there.

I think we first need to address this bigger point before tackling the one raised by @jonspalmer (although I do agree that `call_foo` is very ugly).

fsateler:
Hi!

Friendly ping to ask about this feature. Do you consider Slots V2 to cover this use case? AFAICT, it could be, but it is rather cumbersome:

1. Using a Component Slot requires creating a new component + view, and requires creating a new object for each slot item (when it `renders_many`).
2. Using a Delegate Slot does not allow using ERB syntax.
3. A pass through slot is not applicable.

All three options also expose the slot to the user, which is not desired.

If splitting templates is still a desired feature, I'd like to move this forward, and also give it a nicer interface than `call_<template_name>`

BlakeWilliams:
> 1. Using a Component Slot requires creating a new component + view, and requires creating a new object for each slot item (when it `renders_many`).
> 2. Using a Delegate Slot does not allow using ERB syntax.
> 3. A pass through slot is not applicable.
> 
> All three options also expose the slot to the user, which is not desired.

I agree 2 and 3 are mostly incompatible with this goal (2 could work imo, depending on the use case). I do think option 1 could work though. Your point about the slots being exposed to the user is completely valid, but I wonder if that could be resolved by adding a `private: true` option to slots if this becomes a proven/valuable use of slots?

fsateler:
Hi @BlakeWilliams !

I'm going to create an example. Imagine a Table Component:

```ruby
class TableComponent < ApplicationComponent
  renders_many :columns, "ColumnComponent"
  def initialize(rows:)
    @rows = rows
  end

  class ColumnComponent < ApplicationComponent # This one actually doesn't need to be a component :)
     def initialize(attribute:)
        @attribute = attribute
     end

    def render_value model
       model.send(@attribute) # for brevity, we don't implement other ways of rendering the value
    end
  end
end
```

```erb
<%# table_component.html.erb %>
<table>
  <% @rows.each do |row| %>
    <tr>
      <% columns.each do |col| %>
        <td><%= col.render_value(row) %></td>
      <% end %>
   </tr>
  <% end %>
</table>
```

Now, imagine that I add a bunch of options, and thus creating each row is a bit more complicated. Thus, I want to extract a method to render the row. I can use option 1 above:


```ruby
class RowComponent 
  def initialize(item:, columns:, **some_more_options)
    @item = item
    @columns = columns
    @options = some_more_options
  end

  def row_options
     # something based on @options
  end
end
```

```erb
<%# row_component.html.erb %>
<%= content_tag :tr, row_options do %>
    <% @columns.each do |col| %>
      <td><%= col.render_value(@item) %></td>
    <% end %>
<% end %>
```


```ruby
class TableComponent < ApplicationComponent
  renders_many :columns, "ColumnComponent"
  # @api private
  renders_many :rows, -> (item) do
    RowComponent.new(item: item, columns: columns, **some_options)
  end

  def initialize(rows:)
    @rows = rows
  end

  def before_render
    rows(@rows.map{|item| {item: item, columns: columns} })
  end
end
```

```erb
<%# table_component.html.erb %>
<table>
  <% rows.each do |row| %>
    <%= row %>
  <% end %>
</table>
```

The problem with this solution is that:

1. RowComponent is now public. This can be worked around with some documentation though.
2. It creates a new Hash and a new RowComponent for each Row, each in one. This is actually the deadly problem, since creating large arrays is very problematic for the ruby GC and can create large pauses. A simple example with 6000 rows means passing 10% of the time in GC (as measured by miniprofiler flamegraph) vs 0% in the first version.
3. Actually, I also find the solution to be a bit ugly. I need to remember to populate the rows slots in before_render, with a weird map in between.

joelhawksley:
@fsateler are you still interested in building out this functionality? If so, perhaps opening a PR with and ADR would be a good next step?

fsateler:
Hi @joelhawksley ! Indeed I might. I have two questions though:

1. What changed since the last comment? I'm happy to update and move this forward but the previous messages here suggesteds slots as a way for this problem. I'd like to know what changed that something like this would be acceptable for you.
2. What is an ADR? Any pointers on how to create one?

fsateler:
Hi! I have rebased the branch and added the ADR.

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,52 @@
...
+
+1. It creates a new public class.
+2. If new component is invoked repeatedly (eg, for a list of items), this creates
+   GC pressure by creating lots of intermediate objects.
+
+## Decision
+
+We will allow having multiple templates in the sidecar asset. Each asset will be compiled to
+it's own method `call_<template_name>`. In order to allow the compiled method to receive arguments,
+the component must define them via a `template_arguments :template_name, :argument1, :argument2`.
```
joelhawksley, line 32: 👋🏻 thank you for taking the time to update this PR with docs and the ADR ❤️ 

This bit is perhaps the part I'm the least comfortable with. I'm wary of re-implementing Ruby method parameters via a DSL.

I've been trying to rack my brain on how we could pull this off without a DSL, and I think this might work: we could compile additional templates to the end of methods defined in a specific format on a component. Let me try to explain (😅 it's the end of the day for me)

```rb
  def render_other(message:)
  end
end
```

```erb
<%# my_component/other.html.erb %>

<%= message %>
```

In ViewComponent, we would then compile the template and append it to the bottom of the `render_other` method body, giving us something like:

```rb
def render_other(message:)
  @output_buffer.safe_append message
end
```

Basically, if you defined a template `other`, we'd look for a `render_other` method and compile it into that method. This approach is for sure inspired by @camertron's [rux](https://github.com/camertron/rux) ❤️ 

What do you think?

fsateler: Hmm, interesting idea. Looks doable, and has the benefit of eliminating the only-kwargs restriction. However, it feels a bit more magical, and has a few corner cases that I'm not sure how to handle:

### Should VC just overwrite the method? 

It might be confusing if I have: 
```ruby
def render_other(message)
  message.squish!
end
```

What should the resulting method be? (Is it even possible to append to a method?) Should we error out to avoid confusion? If the body is ignored, did we actually gain more by reusing the ruby method definition?

### What happens if `render_other` is defined but there is no template?

Should we error out? or just let it be?

### What happens if there is no `render_other` method?

Should we just generate an argument-less method?

lib/view_component/compiler.rb
```diff
@@ -64,10 +64,27 @@ def #{method_name}
       CompileCache.register(component_class)
     end
 
+    def add_template_arguments(template, *args)
+      template = template.to_s
+      raise ArgumentError, "Arguments already defined for template #{template}" if template_arguments.key?(template)
+
+      template_exists = templates.any? { |t| t[:base_name].split(".").first == template }
+      raise ArgumentError, "Template does not exist: #{template}" unless template_exists
```
fsateler, line 72: I'm thinking this check should also be the other way around: we should only compile methods for the templates we have described (either via the DSL or the `def render_foo`). WDYT?

camertron:
Hey @fsateler, I'm a latecomer to this PR and discussion, but I hope you'll allow me to state my opinion anyway. I'm concerned rendering partials in components breaks the component abstraction. One of the benefits of view_component is that it avoids many of the problems inherent in traditional Rails views, key among them being the problem of shared state. While there's nothing stopping you from rendering normal Rails view partials inside components today, I'm not sure it's a good idea to bake it into the view_component framework like this, since doing so further encourages using partials that should probably be their own components.

Is there something preventing you from converting your partials into components?

fsateler:
@camertron I think it is best if you comment directly on the parts of the ADR that don't make sense to you. I don't know what is missing there to answer your question.

fsateler:
Btw, this merge explicitly is not about invoking unrelated partials. Its about being able to extract some parts of a large method into several smaller methods

camertron:
@fsateler ah I used the word "partial" when I meant to say "additional sidecar asset." Correct me if I'm wrong, but a big part of your proposal is providing additional .html.erb files scoped to the component (i.e. sidecar assets) that get rendered from another sidecar asset. Perhaps my earlier comment makes more sense with the right terminology 😅  I'll make additional comments on the ADR 😄

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,52 @@
...
+An option would be to extract another Component for the extracted section, but that
+has the following drawbacks:
+
+1. It creates a new public class.
+2. If new component is invoked repeatedly (eg, for a list of items), this creates
+   GC pressure by creating lots of intermediate objects.
+
+## Decision
+
+We will allow having multiple templates in the sidecar asset. Each asset will be compiled to
```
camertron, line 30: I've read this a number of times, and my brain keeps short-circuiting to "this is what slots are for." Slots take arguments and are invoked in a specific place in the template. To me, they have identical behavior to what you're proposing 🤔 

I do see your point about wanting to render sub-templates without having to turn them into a full component though. What about something like this:

```ruby
# app/components/my_component.rb

  # notice we're not passing any component class, string, or lambda here
  renders_many :posts
end
```

```erb
<%# app/components/my_component.html.erb %>
<%= render(MyComponent.new) do |c| %>
  <%# renders_many implicitly makes all these kwargs available to the sub template. %>
  <%# The sub template's filename matches the singular name of the slot. %>
  <% c.post(post: p) %>
<% end %>
```

```erb
<%# app/components/my_component/_post.html.erb %>
<%= post.title %>
```

However, even something like this makes me a little uncomfortable. The idea behind view components is to use, well, _components_. In my view, you might as well use regular 'ol actionview if you're not breaking view code up into components 😄

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,52 @@
...
+may be passed into the partial via ivars or locals, and it is impossible to know
+which without actually opening up the file. Additionally, partials are globally 
+invocable, thus making it difficult to detect if a given partial is in use or not,
+and who are its users.
+
+An option would be to extract another Component for the extracted section, but that
+has the following drawbacks:
+
+1. It creates a new public class.
+2. If new component is invoked repeatedly (eg, for a list of items), this creates
```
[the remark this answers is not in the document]
fsateler: I posted something here: https://github.com/github/view_component/pull/451#issuecomment-762516144 . I'll come back with better numbers. I don't think lambdas are an option since they don't allow ERB.

adr/0003-multiple-templates-per-component.md
fsateler: > I've read this a number of times, and my brain keeps short-circuiting to "this is what slots are for." Slots take arguments and are invoked in a specific place in the template. To me, they have identical behavior to what you're proposing thinking

No, slots are a primarily a way to *inject* content into components (eg, the message in an Alert component). This is all about internal organization of the component.

> I do see your point about wanting to render sub-templates without having to turn them into a full component though. What about something like this:
> <snip code>

I don't understand the code. Where do the posts come from and where is the loop? Is MyComponent recursiveness intended?

One comment though: my proposal avoids the magical variables you pass via kwargs. Your proposal makes the interface between the partial and the main view invisible (ie, you would need to go read the view, and the main component, to be able to know which variables I need to pass). My proposal makes such variables explicit, and @joelhawksley's proposal improves on that to allow all ruby argument forms.

> The idea behind view components is to use, well, components. In my view, you might as well use regular 'ol actionview if you're not breaking view code up into components 

Well, not all components are the same size. If the component is relatively large, splitting out sub-methods is a reasonable thing to do. All this merge does is allow you to write that method in ERB. Reasonable people might disagree, but I don't regard `RowComponent` a component: its coupling with the `TableComponent` is so tight that they are effectively a large super-component. I'd prefer to not have the row be a public interface.

fsateler: > > No, slots are a primarily a way to inject content into components (eg, the message in an Alert component). This proposal is all about internal organization of the component.
> 
> I see, so you're saying part of the benefit of rendering sidecar templates is that it allows you to organize the component's constituent pieces and keep them "private," i.e. without opening up a slot? That's a really interesting idea I hadn't considered.

Exactly! Please help me retitle/re-describe the issue, apparently it is not clear.

Do note that what I want is not exactly "private" as is understood in other languages. I very much want derived components to be able to override the erb files, if they need to.  See use case 2 here: https://github.com/github/view_component/issues/387#issuecomment-663785026

> 
> > I don't understand the code. Where do the posts come from and where is the loop? Is MyComponent recursiveness intended?
> 
> Ah sorry, the recursiveness is not intended, I mistakenly rendered `MyComponent` in the wrong file. I have updated the comment and added the missing loop.

Thanks. I still think this PR would improve things, since it would simplify the `MyComponent` interface to:

```erb
<%= render(MyComponent.new(posts: Post.all)) %>
```

> 
> > Your proposal makes the interface between the partial and the main view invisible (ie, you would need to go read the view, and the main component, to be able to know which variables I need to pass). My proposal makes such variables explicit.
> 
> That's definitely a valid concern. Could be mitigated by passing a lambda and rendering the ERB inside it, eg:
> 
> ```ruby
> # app/components/my_component.rb
> 
>   renders_many :posts, lambda do |p:|
>     render_erb '_sub_component', post: p
>   end
> end
> ```

But then you need something like this PR that allows extra ERB files :wink: 

I'm open to alternative interfaces. As I indicated on https://github.com/github/view_component/pull/451#issuecomment-681172299, I agree the current interface is very ugly, and we should strive for a better one. 

> 
> > All this merge does is allow you to write that method in ERB.
> 
> I do think it's potentially a good idea to allow rendering additional, component-scoped sidecar assets.

:smile:

BlakeWilliams: > If the component is relatively large, splitting out sub-methods is a reasonable thing to do. All this merge does is allow you to write that method in ERB. Reasonable people might disagree, but I don't regard `RowComponent` a component: its coupling with the `TableComponent` is so tight that they are effectively a large super-component. I'd prefer to not have the row be a public interface.

One of the biggest benefits of adopting components is that you can look at a minimal number of files and understand the component's behavior and its responsibility (components are very focused, by design). When thinking in terms of components, if a row has its own behavior it's a good sign that it should be its own component. I can understand the desire to keep it coupled to Table and not extract a component, but that  introduces other concerns like the table component losing focus, doing too much, as well as some testing concerns. It's not enforced, but it could always be undocumented or left as a private API though documentation/comments.

> > I see, so you're saying part of the benefit of rendering sidecar templates is that it allows you to organize the component's constituent pieces and keep them "private," i.e. without opening up a slot? That's a really interesting idea I hadn't considered.
> 
> Exactly! Please help me retitle/re-describe the issue, apparently it is not clear.

I think I understand the proposal, but I'm not sure I see the benefits of extracting separate ERB files. Similar to the above, I think it allows components to be broken apart in ways that I'm concerned would allow components to grow too large or take on behaviors that would be better encapsulated in separate object(s)/component(s). You also lose the benefit of being able to open two files in your editor and see the entire behavior of the component.

If we do accept this proposal, we'll end up with three different ways to render content in the framework, as well as the methods provided by Rails, like partials.

* Components (encapsulates behavior)
* Slots (composes components)
* Component partials (renders more HTML in the scope of a component?)

I think it would help to come up with a name for the abstraction that describes it well. I also have some other questions, like does it still go through `render`, are we bypassing Rails completely, is there parts of Rails we can take advantage of etc.?

That's a lot, but I did have one more question. Does any other component oriented framework have prior art we could take inspiration from here? I'm interested in seeing this behavior elsewhere and how it works and fits in a component based architecture.

adr/0003-multiple-templates-per-component.md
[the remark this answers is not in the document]
fsateler: I have created a [benchmark](https://github.com/fsateler/view_component/commit/54a8ba44)

The result is:

```
               slot:   139.000  i/100ms
            partial:   299.000  i/100ms
               slot:      1.536k (± 8.9%) i/s -     15.290k in  10.038060s
            partial:      2.935k (± 6.8%) i/s -     29.302k in  10.032277s

Comparison:
            partial::     2935.2 i/s
               slot::     1536.0 i/s - 1.91x  (± 0.00) slower
```

Increasing the number of objects to iterate over to 10.000:

```
               slot:     1.000  i/100ms
            partial:     1.000  i/100ms
               slot:      1.431  (± 0.0%) i/s -     15.000  in  10.580122s
            partial:      3.183  (± 0.0%) i/s -     33.000  in  10.430793s

Comparison:
            partial::        3.2 i/s
               slot::        1.4 i/s - 2.22x  (± 0.00) slower
```

This is with ruby 2.7.3. results are similar with 3.0.1.

adr/0003-multiple-templates-per-component.md
fsateler: > > If the component is relatively large, splitting out sub-methods is a reasonable thing to do. All this merge does is allow you to write that method in ERB. Reasonable people might disagree, but I don't regard `RowComponent` a component: its coupling with the `TableComponent` is so tight that they are effectively a large super-component. I'd prefer to not have the row be a public interface.
> 
> One of the biggest benefits of adopting components is that you can look at a minimal number of files and understand the component's behavior and its responsibility (components are very focused, by design). 

I'm not very convinced by this. The way I see it, adopting components introduces encapsulation into the previous wild west of views. There's nothing that says a given component is small or large. The entire view could be a component!

> When thinking in terms of components, if a row has its own behavior it's a good sign that it should be its own component. I can understand the desire to keep it coupled to Table and not extract a component, but that introduces other concerns like the table component losing focus, doing too much, as well as some testing concerns. It's not enforced, but it could always be undocumented or left as a private API though documentation/comments.

Introducing generic-in-name-only concepts hurts readability IMO. If I'm designing the `UserProfileComponent`, do I really need `UserProfilePictureComponent`? Will it be used anywhere else? If the answer is no, I believe it should just be part of the `UserProfileComponent`. Otherwise, you end up navigating a very deep tree of components.

Plus, in this example now you have to look at 4 files instead of 3 :wink: 

> 
> > > I see, so you're saying part of the benefit of rendering sidecar templates is that it allows you to organize the component's constituent pieces and keep them "private," i.e. without opening up a slot? That's a really interesting idea I hadn't considered.
> > 
> > 
> > Exactly! Please help me retitle/re-describe the issue, apparently it is not clear.
> 
> I think I understand the proposal, but I'm not sure I see the benefits of extracting separate ERB files. Similar to the above, I think it allows components to be broken apart in ways that I'm concerned would allow components to grow too large or take on behaviors that would be better encapsulated in separate object(s)/component(s). You also lose the benefit of being able to open two files in your editor and see the entire behavior of the component.

I encourage you to think of this the other way around. This PR is only needed because we have no way to define multiple "methods" in ERB. If ERB had a native way to pack multiple templates into a single file, would you have ViewComponent actively block that ability? Because ERB does not have that facility, this is the second best. I think inventing yet another file format to pack multiple templates is a bad idea.

> 
> If we do accept this proposal, we'll end up with three different ways to render content in the framework, as well as the methods provided by Rails, like partials.
> 
> * Components (encapsulates behavior)
> * Slots (composes components)
> * Component partials (renders more HTML in the scope of a component?)

Do note that component partials are not meant to be externally available. Perhaps I can make that explicit by marking them `private` or `protected`?

> 
> I think it would help to come up with a name for the abstraction that describes it well. I also have some other questions, like does it still go through `render`, are we bypassing Rails completely, is there parts of Rails we can take advantage of etc.?

I'm not sure what you are aiming at here, but this does not use `render`. It should be a pure function taking some inputs and returning some html.

> That's a lot, but I did have one more question. Does any other component oriented framework have prior art we could take inspiration from here? I'm interested in seeing this behavior elsewhere and how it works and fits in a component based architecture.

I'm not sure it counts, but React/JSX allows creating arbitrary html wherever you want, so you can have methods returning sub-components:

```jsx
class Test extends React.Component {
  render() {
    return <h1>Hello, {this.someFunc()}</h1>;
  }
  someFunc() {
    return <small>{this.props.name}</small>;
  }
}
ReactDOM.render(<Test name="Felipe" />, document.getElementById('root'));
```

This would print

```html
<h1>Hello, <small>Felipe</small></h1>
```

lib/view_component/compiler.rb
```diff
@@ -41,16 +41,16 @@ def compile(raise_errors: false)
...
-        method_name = call_method_name(template[:variant])
-
         if component_class.instance_methods.include?(method_name.to_sym)
           component_class.send(:undef_method, method_name.to_sym)
         end
 
         component_class.class_eval <<-RUBY, template[:path], -1
-          def #{method_name}
+          def #{method_name}(#{method_args.join(", ")})
             @output_buffer = ActionView::OutputBuffer.new
```
fsateler, line 54: I'm missing tests! This clobbers `@output_buffer`

adr/0003-multiple-templates-per-component.md
Spone: > Does any other component oriented framework have prior art we could take inspiration from here? I'm interested in seeing this behavior elsewhere and how it works and fits in a component based architecture.

Komponent allows [component partials](https://github.com/komposable/komponent#component-partials)

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,52 @@
...
+   GC pressure by creating lots of intermediate objects.
+
```
BlakeWilliams, line 27:
```suggestion
## Considered Options

* Introduce component partials to components
* Keep components as-is

### Component partials

Quick description...

**Pros:**
* Better performance due to lack of GC pressure and object creation

**Cons:**
* Another concept for users of ViewComponent to learn and understand.
* Components are no longer the only way to encapsulate behavior.
* Partials aren't reusable, and belong to a specific component.

### Keeping components as-is

**Pros:**
* The API remains simple and components are the only way to encapsulate behavior.

**Cons:**
* Extracting a component results in more GC and intermediate objects.
```

I've used this format in the past to help with ADR's, thoughts on adding this to help us get a quick understanding of each option? I haven't completely filled out each pro/con section fwiw, but I think that leaves room to suggest changes and discuss individual points. (e.g. your garbage collection/object concern is a very valid, and very good point)

adr/0003-multiple-templates-per-component.md
BlakeWilliams: > I'm not sure it counts, but React/JSX allows creating arbitrary html wherever you want, so you can have methods returning sub-components:

That's possible today in ViewComponent via `content_tag` right? The only thing that's missing is a way to render erb. Not saying it shouldn't, but it is possible to implement that functionality today without it being a library feature. The `_after_compile` hook would allow some optimizations too, I believe.

> Komponent allows [component partials](https://github.com/komposable/komponent#component-partials) (but the example in the documentation is overly simplistic)

Thanks! I'll take a peek at that. I'm curious, did you utilize component partials often?

adr/0003-multiple-templates-per-component.md
fsateler: I like the format! I have omitted the "Partials aren't reusable, and belong to a specific component.", since I actually think that is a Pro, not a Con. Perhaps we can start a new thread to discuss that?

adr/0003-multiple-templates-per-component.md
camertron: @joelhawksley and I talked about this during our 1:1 today, and we came to the conclusion that there are really two feature requests here: 1) the ability to render arbitrary templates from slots, and 2) private slots.

As I said earlier, slots are the mechanism the view_component library has for composing a component from constituent parts. I think a lot of us maintainers are very hesitant to add another way to render templates because of the confusion we already see around slots, and the complexity of the approaches we've discussed here in this issue feel fairly prohibitive. In our 1:1, Joel and I tried to come up with a solution based on the one he proposed in an earlier comment and realized after ~30 minutes that we were just reimplementing actionview 😅  

The alternative we settled on was to be able to specify a template path in the slot definition. This is pretty similar to my last comment, although makes the template explicit. The other necessary feature is private slots. A private slot would only be usable by templates rendered "inside" the component. Trying to use them outside the component would raise an error.

Let me try to rewrite your table component using this technique.

```erb
<%# app/views/foo/index.html.erb %>
<%= render(TableComponent.new(rows: @rows)) do |component| %>
  <%# this raises an error %>
  <% c.column %>
<% end %>
```

```ruby
# app/components/table_component.rb
class TableComponent < ApplicationComponent
  renders_many :columns, template: 'column.html.erb', public: false

  def initialize(rows:)
    @rows = rows
  end
end
```

```erb
<%# app/components/table_component.html.erb %>
<table>
  <% @rows.each do |row| %>
    <tr>
      <% column(row: row, attribute: :name) %>
   </tr>
  <% end %>
</table>
```

```erb
<%# app/components/table_component/column.html.erb %>
<td><%= row.send(attribute) %></td>
```

There are a number of nice benefits to this approach:

1. Leverages the existing slots API.
1. Does not require instantiating components, so should be relatively performant.
1. Does what you want (I think?)

fsateler: I think the example does not fit my problem. In your example, the user specifies the columns (good!). But what I wanted was to extract the `row`, and not expose that to the user. How would I achieve that with this proposal?

Additionally, there is no specifying of the arguments, which:
1. Makes the API implicit rather than explicit
2.  We would need to invoke the template via `instance_exec` / `binding`, and that is slower.

I do like that there are fewer pieces though.

fsateler: Maybe we could pair to brainstorm how this could end up looking like?

camertron: Hmm I see what you mean. What about this?

```ruby
# app/components/table_component.rb
class TableComponent < ApplicationComponent
  renders_many :rows, template: 'row.html.erb', public: false
  renders_many :columns, template: 'column.html.erb', public: false

  def initialize(rows:)
    @rows = rows
  end
end
```

```erb
<%# app/components/table_component.html.erb %>
<table>
  <% @rows.each do |row| %>
    <% row(row: row) %>
  <% end %>
</table>
```

```erb
<%# app/components/row.html.erb %>
<tr>
  <% column(row: row, attribute: :name) %>
</tr>
```

```erb
<%# app/components/table_component/column.html.erb %>
<td><%= row.send(attribute) %></td>
```

Although now that I write this out, it feels a little weird to have rows and columns at the same level.

> 1. Makes the API implicit rather than explicit

That's true, although I don't know if that's avoidable given that we're actively trying to remove `RowComponent` and `ColumnComponent`, which would have provided it.

> 2. We would need to invoke the template via `instance_exec` / `binding`, and that is slower.

Unless I'm misunderstanding what you mean, there should be no need for either of these. Slots are already implemented by dynamically defining methods on the component class. We could do the same thing with these private slots that render additional templates.

@fsateler I want to thank you for your patience as we try to figure this out. This issue has prompted a huge amount of discussion internally! I think it would make sense for all of us to hop on a call at some point to brainstorm. Feel free to DM me [on Twitter](https://twitter.com/camertron).

fsateler: > Although now that I write this out, it feels a little weird to have rows and columns at the same level.

Indeed, but in your example, `columns` should be the public slot for the user:

```erb
<%= render TableComponent.new(rows: @rows) do |c| %>
  <%# we want a table with name and age columns %>
  <% c.column(:name) %>
  <% c.column(:age) %>
<% end %>
```

> > 2. We would need to invoke the template via instance_exec / binding, and that is slower.
> Unless I'm misunderstanding what you mean, there should be no need for either of these.

Point 2 is a consequence of point 1. You need to somehow make the dynamically passed variables (in this example, `row` and `attribute`) as available in the invoked method binding. Where would be the canonical list to define the method parameters? (This is what the original PR tries to solve with the `template_arguments` method).

Spone: > I'm curious, did you utilize component partials often?

@BlakeWilliams it's used 3 times in the largest codebase I work on (about ~100 components). It's mainly used to extract part of the template to another file, when the component template grows very big, but no part of the template would make sense as a reusable separate component.

Spone:
> Hmm, any hint on how to investigate pvc test? I don't see how this change could cause a accessibility failure..

I re-ran the pvc test and it passed. It can be a bit flaky, so never mind :)

joelhawksley (commented):
Very nice! I'll have another round of feedback around docs/ADR once these items are resolved ❤️

docs/guide/templates.md
```diff
@@ -131,3 +131,74 @@ class MyComponent < ViewComponent::Base
   strip_trailing_whitespace(false)
 end
 ```
+
+## Multiple templates
```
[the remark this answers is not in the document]
Spone: What about "Partials"?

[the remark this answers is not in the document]
Spone: Or maybe "Sidecar partials" to remove ambiguity?

docs/guide/templates.md
```diff
@@ -131,3 +131,74 @@ class MyComponent < ViewComponent::Base
...
+app/components
+├── ...
+├── test_component.rb
+├── test_component
+|   ├── list.html.erb
+|   └── summary.html.erb
+├── ...
+```
+
+Templates are compiled to methods in the format `render_template_#{template_basename}`, which can then be called in the component.
```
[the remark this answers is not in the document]
fsateler: It looks better indeed. It may make it a bit harder to grep for them but that shouldn't be a problem because they should be mostly private methods.

fsateler:
I have applied (almost) all suggestions. Thanks for the review. I liked the new method name pattern. I did not apply @Spone 's suggestion for the new name in the documentation, but that could be applied easily if there is consensus about that. IMO, the name `partial` could cause confusion, but that is just a gut feeling, I have no data to back this up.

docs/guide/templates.md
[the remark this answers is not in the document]
joelhawksley: I like the term in theory, but in practice I think it's already "taken" by Rails' usage of it. As this feature is quite different from Rails partials, I'm hesitant to use the term.

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Status
+
+Proposed.
+
+## Context
+
+As components become larger (for example, because you are implementing a whole page), it becomes
+useful to be able to extract sections of the view to a different file. ActionView has
+partials, and ViewComponent lacks a similar mechanism.
```
joelhawksley, line 13:
```suggestion
As views become larger (such as an entire page), it becomes useful to be able to extract sections of the view to a different file. In ActionView this is done with partials, but ViewComponent lacks a similar mechanism.
```

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+
+ActionView partials have the problem that their interface isn't introspectable. Data
+may be passed into the partial via ivars or locals, and it is impossible to know
+which without actually opening up the file. Additionally, partials are globally
+invocable, thus making it difficult to detect if a given partial is in use or not,
+and who are its users.
```
joelhawksley, line 19:
```suggestion
The interface of ActionView partials cannot be introspected. Because data may be passed into the partial via ivars or locals it is impossible to know which without reading the file. Partials are also globally accessible, making it difficult to determine if a given partial is in use or not.
```

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Considered Options
+
+* Introduce component partials to components
```
joelhawksley, line 23:
```suggestion
* Introduce sub-templates to components
```

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+* Keep components as-is
```
joelhawksley, line 24:
```suggestion
* Do nothing
```

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Component partials
```
joelhawksley, line 26:
```suggestion
### Sub-templates
```

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+
+Introduce an API to have multiple ERB templates available within the component,
+and make it possible to invoke them from the main view. This API should be as
+explicit as possible, in particular it should explicitly list the arguments it
+receives (if any). This allows a single method to be compiled, and invoked
+directly, without having to dance around with `binding` as ActionView partials
+do.
```
joelhawksley, line 33:
```suggestion
Introduce support for multiple ERB templates within a single component and make it possible to invoke them from the main view, explicitly listing the arguments the additional templates accept. This allows a single method to be compiled and invoked directly.
```

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+* Better performance due to lack of GC pressure and object creation
+* Reduces the number of components needed to express a more complex view.
+* Extracted sections aren't exposed outside the component, thus reducing component library API surface.
+
+**Cons:**
+
+* Another concept for users of ViewComponent to learn and understand.
+* Components are no longer the only way to encapsulate behavior.
+
+### Keeping components as-is
```
joelhawksley, line 46:
```suggestion
### Do nothing
```

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+
+We will allow having multiple templates in the sidecar asset. Each asset will be compiled to
+it's own method `render_template_<template_name>`. The main entry point will continue to be called
+`call`. To allow the compiled method to receive arguments,
+the template must define them using the same syntax as
+[Rails Strict Locals](https://edgeguides.rubyonrails.org/action_view_overview.html#strict-locals).
+Because we don't have backwards compatibility issues, and we want to avoid the `binding` dance
+to add arbitrary locals, a difference is made: a missing strict locals tag means the template
+takes no arguments (that is, is equivalent to `locals: ()`).
+This will create required keyword arguments to the `call_<template_name>` method.
```
joelhawksley, line 69:
```suggestion
Support multiple sidecar templates. Compile each template into its own method `render_<template_name>_template`. To allow the compiled method to receive arguments,
the template must define them using the same syntax as [Rails' Strict Locals](https://edgeguides.rubyonrails.org/action_view_overview.html#strict-locals), with one difference: a missing strict locals tag means the template takes no arguments (equivalent to `locals: ()`).
```

docs/adrs/0005-allow-sub-templates.md
```diff
@@ -0,0 +1,86 @@
...
+
+There is no provision at the moment to allow `render(*)` to render a partial. This could be
+added later if necessary and it becomes desirable.
+
+The generated methods are only invokable via keyword arguments, inheriting the limitation
+from ActionView.
+
+The generated methods are public, and thus could be invoked by a third party. There is
+no pressing need to make the methods private, and we avoid introducing new concepts
+into ViewComponent.
```
joelhawksley, line 86:
```suggestion
This implementation has better performance characteristics over both an extracted component
and ActionView partials because it avoids creating intermediate objects and the overhead of
creating bindings and `instance_exec`. Having explicit arguments makes the interface explicit.

There is no provision at the moment to allow `render(*)` to render a sub template. This could be
added later if necessary and it becomes desirable.

The generated methods are only invokable via keyword arguments, inheriting the limitation
from ActionView.

The generated methods are public, and thus could be invoked by a third party. There is
no pressing need to make the methods private, and we avoid introducing new concepts
into ViewComponent.
```

lib/view_component/compiler.rb
```diff
@@ -70,18 +72,21 @@ def compile(raise_errors: false, force: false)
...
           if component_class.instance_methods.include?(method_name.to_sym)
             component_class.send(:undef_method, method_name.to_sym)
           end
 
           # rubocop:disable Style/EvalWithLocation
-          component_class.class_eval <<-RUBY, template[:path], 0
-          def #{method_name}
+          component_class.class_eval <<-RUBY, template[:path], -1
+          def #{method_name}(#{method_args})
+            old_buffer = @output_buffer if defined? @output_buffer
```
joelhawksley, line 86: I couldn't find a good place to leave this comment, so I'll leave it here: What do you think about making the sub-template methods private, at least to start? I really want to emphasize that sub-templates are only internally available.

docs/CHANGELOG.md
```diff
@@ -30,6 +30,10 @@ nav_order: 5
 
     *Anton Prins*
 
+* Add support for sub-templates.
```
joelhawksley, line 33:
```suggestion
* Add experimental support for sub-templates.
```

docs/guide/templates.md
```diff
@@ -131,3 +131,73 @@ class MyComponent < ViewComponent::Base
...
+app/components
+├── ...
+├── test_component.rb
+├── test_component
+|   ├── list.html.erb
+|   └── summary.html.erb
+├── ...
+```
+
+Templates are compiled to methods in the format `render_#{template_basename}_template`, which can then be called in the component.
```
joelhawksley, line 155:
```suggestion
Templates are compiled to private methods in the format `render_#{template_basename}_template`:
```

docs/guide/templates.md
```diff
@@ -131,3 +131,73 @@ class MyComponent < ViewComponent::Base
...
+      render_summary_template
+    end
+  end
+end
+```
+
+To define which parameters a template accepts, use the
+[Rails Strict Locals](https://edgeguides.rubyonrails.org/action_view_overview.html#strict-locals) syntax.
+Unlike Rails, a missing `locals` declaration means the template takes no arguments.
+This option only affects sub-templates.
```
joelhawksley, line 177:
```suggestion
To define which parameters a sub-template accepts, use the [Rails Strict Locals](https://edgeguides.rubyonrails.org/action_view_overview.html#strict-locals) syntax.

_Note: Unlike Rails, a missing `locals` declaration means the template takes no arguments._
```

docs/guide/templates.md
```diff
@@ -131,3 +131,73 @@ class MyComponent < ViewComponent::Base
...
+    end
+  end
+end
+```
+
+```erb
+<%# list.html.erb %>
+<%# locals: (multiple:) %->
+The parameter is <%= multiple %>
+```
```
joelhawksley, line 203: Might we put this example block above the Ruby? It's the more interesting bit IMO

lib/view_component/compiler.rb
```diff
@@ -265,12 +292,16 @@ def compiled_template(file_path)
...
-        "call"
-      end
+    def call_method_name(template, variant)
+      name =
+        if template.blank? || template == component_class.name.demodulize.underscore
+          "call"
+        else
+          "render"
+        end
+      name += "_#{variant}" if variant.present? && variants.include?(variant)
```
joelhawksley, line 302: Might we add a test for sub-template variants?

joelhawksley (commented):
Another round, I think we're getting pretty close ❤️

docs/CHANGELOG.md
BlakeWilliams: Is the plan to make this an opt-in feature like we have with other experimental features?

I'm still hesitant about the tradeoffs being worth introducing yet another rendering concept to VC, especially since slots aren't as well understood as they could be. Happy to be overruled, but if we do plan on adding this feature I'd love to see it as an opt-in experiment like we've done with other features before committing to it.

lib/view_component/compiler.rb
fsateler: Ouch, this makes things a bit more complicated... but OK. Private by default it is.

docs/adrs/0005-allow-sub-templates.md
fsateler: This suggestion mentions the methods are public, I'll change it to private as per the comment below.

lib/view_component/compiler.rb
fsateler: Hmm, this comment brings up an interesting point. The API effectively outsources variant handling to the user, because there is no dispatch to the correct variant automatically. That is, I would need to do something like:

```ruby
if variant == 'some_variant'
  render_list_template_some_variant
else
  render_list_template
end
```

This is not very user-friendly. I'm wondering what a better API would look like? Maybe the compiler should compile to `__render_<foo>_template` (that is, add a `__` prefix), and then `render_<foo>_template` would do the variant checking?

```ruby
def render_list_template elements:
   case @__vc_variant
   when 'some_variant'
      __render_list_template_some_variant elements: elements
   else
      __render_list_template elements: elements
   end
end
```

joelhawksley: Hmm. What happens if the variants have different Strict Locals?

fsateler: That sounds unusual, but we could use argument forwarding (`**args`) to support that

joelhawksley: I would prefer that we just generate a method for each variant. It's simpler. We already do this for the `call_` methods that can be defined by a consumer.

joelhawksley:
Looking at this during office hours today, I think we're in a tough spot. We now have a situation where some ViewComponent templates support variants and some don't, which I think would be quite confusing to consumers. (Although I'm of the opinion generally that variants shouldn't exist at all). This makes me want to lean more on @BlakeWilliams's inline templates proposal.

fsateler:
Hi!

> We now have a situation where some ViewComponent templates support variants and some don't, which I think would be quite confusing to consumers

I don't understand this point. AFAICT, you can have variants in this PR. Perhaps you mean you need to know which variant to render? That may be addressable with something like I proposed earlier: https://github.com/ViewComponent/view_component/pull/451#discussion_r988396530

```ruby
def render_foo_template **args
   case @__vc_variant
   when 'some_variant'
      render_list_template_some_variant **args
   else
      render_list_template_default_variant **args
   end
end
```

This would allow rendering the current template forcing a variant (via the `_some_variant` suffix) or via the current variant (via the method without suffix).

fsateler:
Hi @joelhawksley ! I'm still interested in moving this forward, but I'd need to understand better this objection:

> We now have a situation where some ViewComponent templates support variants and some don't, which I think would be quite confusing to consumers.

You *can* have variants in this PR, although they are not very ergonomic. Do you want me to think about improving that? Or is something else you have in mind?

joelhawksley:
@fsateler I haven't forgotten about this PR ❤️ 

We're iterating on https://github.com/ViewComponent/view_component/pull/1556 and want to see where we end up API-wise to inform our decisions on your PR here.

barmintor:
Reading over this (trying to get the full context of the direction of variants), I 

> It creates a new Hash and a new RowComponent for each Row, each in one. This is actually the deadly problem, since creating large arrays is very problematic for the ruby GC and can create large pauses. A simple example with 6000 rows means passing 10% of the time in GC (as measured by miniprofiler flamegraph) vs 0% in the first version.

Isn't this exactly the problem lazy enumerators are supposed to solve?

fsateler:
> Reading over this (trying to get the full context of the direction of variants), I
> 
> > It creates a new Hash and a new RowComponent for each Row, each in one. This is actually the deadly problem, since creating large arrays is very problematic for the ruby GC and can create large pauses. A simple example with 6000 rows means passing 10% of the time in GC (as measured by miniprofiler flamegraph) vs 0% in the first version.
> 
> Isn't this exactly the problem lazy enumerators are supposed to solve?

Not really.  Lots of transformations are more memory efficient in lazy enumerators. For example, `some_list.select(&:something?).reject(&:other?).take(10)` will be a lot more efficient with lazy enumerators, because you avoid all the intermediate arrays. 

In this case, however, we are doing a `map`. So, while we would be avoiding a new array allocation, it doesn't avoid allocating a new Hash and Component for each item, even if the objects are then inmediately discarded.

joelhawksley:
@fsateler responding to https://github.com/ViewComponent/view_component/issues/387#issuecomment-1664553704 here ❤️ 

Revisiting this with @camertron and @BlakeWilliams, we're wondering if it'd make sense to first land hooks to the compiler that would allow a gem to add this behavior. I don't think it would be too bad to do so. We're still ultimately feeling hesitant to confuse folks' use of Slots, which are already quite complex.

I also wonder if we might want to consider a Rails core contribution to add named render helper methods for views such as `render_users_show`. I might take a look into this idea over the holidays here!

fsateler:
@joelhawksley awesome! if there is anything that I can help with I'm happy to do so.

lomefin:
Hi @joelhawksley any chance this feature could be released?

Urilope:
> ### Summary
> Fixes #387
> 
> This PR builds on the branch by @joelhawksley and @fermion, and adds the ability to specify arguments for the generated functions.
> 
> The logic being that one may want to use templates to extract sections, without wanting to build a full component. For example, I could have a Table component, and I would want to have the row definition in a sidecar template. This requires being able to pass in each item for every iteration:
> 
> ```ruby
> class TableComponent < ViewComponent::Base
>   def initialize(models)
>     @models = models
>   end
>   template_arguments :row, :model
> end
> ```
> 
> ```
> <%# table_component.html.erb %>
> <table>
>   <% @models.each do |model| %>
>     <%= call_row model: model %>
>   <% end %>
> </table>
> ```
> 
> ```
> <%# row.html.erb %>
> <tr>
>   <%# I can use `model` here %>
>   <td><%= model.name  %></td>
> </tr>
> ```
> 
> Conceptually, this is very similar to ActionView's usage of `locals:`. However, this way has the following benefits:
> 
> 1. Performance is increased as the template needs to be compiled only once.
> 2. We don't need to `instance_exec` things in a custom-build binding, further improving performance.
> 3. Ruby checks our argument list.
> 4. We force the user to specify the locals, thus forcing them to specify the interface.
> 
> ### Other Information
> I'm submitting as a draft, because there are some important API issues that need discussing:
> 
> 1. Should we always use keyword arguments? Maybe we should use positional arguments? (I happen to like keyword arguments, but it may not be everyone's cup of tea).
> 2. This doesn't allow default values for any parameter. Is this a problem that needs addressing?

joelhawksley:
@lomefin I don't have any plans to land this at this point. If someone would like to make it possible to add this feature in a gem, I'm all for that approach 👍🏻

fsateler:
Hi Joel! Thanks for making a decision on this. I understand that the vision does not match.

> @lomefin I don't have any plans to land this at this point. If someone would like to make it possible to add this feature in a gem, I'm all for that approach 👍🏻

Does this mean you would be amenable to changes that enable such a gem to be built? I'm looking at the current code (VC is quite different from what it was when the first patch was made!!), and I think we could piggyback on the Template machinery to achieve our goals. In particular, I'm thinking about making [gather_templates](https://github.com/ViewComponent/view_component/blob/1ad9df2c67fe771598ed53018b44f63fa4a62153/lib/view_component/compiler.rb#L169) extendable, but that would require exposing the Template class as public API for plugins.

joelhawksley:
@fsateler that's exactly what I'm referring to! I'd recommend working against the `v4` feature branch ❤️

PR #458, opened 2020-08-31
Update stale.yml to use stale label instead of wontfix

BlakeWilliams:
Migrates from `wontfix` to `stale` label

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -7,7 +7,7 @@ exemptLabels:
   - pinned
   - security
 # Label to use when marking an issue as stale
-staleLabel: wontfix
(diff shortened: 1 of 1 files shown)


rmacklin:
Nice!

When I first saw that various issues had been labeled "wontfix" despite there being comments in support of fixing them, it seemed odd, and potentially could've been perceived as uninviting which is counter to the welcoming vibes I'd seen before in this project. So I'm very happy to see this change!

joelhawksley:
@rmacklin agreed! It was an honest mistake - the `stale.yml` we originally used was the default configuration. Perhaps the defaults should be more welcoming!

PR #459, opened 2020-08-31
lint before running tests

joelhawksley:
This PR changes our CI workflow to run rubocop before
running our test matrices, as:

1) It's unnecessary to run `rubocop` for each of our
FOURTEEN matrices!

2) If a workflow is going to fail due to linters, it's
a nicer developer experience to have that failure happen as quickly
as possible.

cc @manuelpuyol

### DIFF
--- a/.github/workflows/ruby_on_rails.yml
+++ b/.github/workflows/ruby_on_rails.yml
@@ -1,4 +1,4 @@
-name: Ruby on Rails
+name: CI
 
 on: [push, pull_request]
 
@@ -17,7 +17,7 @@ jobs:
         bundle update
         bundle install --jobs 4 --retry 3
         bundle exec rubocop
-  build:
+  test:
     needs: lint
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/ruby_on_rails.yml
```diff
@@ -1,9 +1,24 @@
...
-  build:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Setup Ruby
+      uses: actions/setup-ruby@v1
+      with:
+        ruby-version: 2.7.x
+    - name: Build and test with Rake
```
rmacklin, line 14: Should we rename this?
```suggestion
    - name: Lint with Rubocop
```

rmacklin:
Hmm, looks like it's running jobs twice (once for the "pull_request" event and once for the "push" event):
image

https://github.com/github/view_component/pull/459/checks?check_run_id=1052308514
https://github.com/github/view_component/pull/459/checks?check_run_id=1052308645

I don't think that was happening before. Is that a bug or expected behavior when you open a PR into a repo that has GitHub Actions set up to run on both those events?

joelhawksley:
@rmacklin I've been seeing the duplicate builds for a while now. It certainly would be nice to only get one set of them!

rmacklin:
Per my edit above, I think in practice it'll only happen for PRs where the branch is inside https://github.com/github/view_component rather than a fork.

It seems like it's behaving as expected, but I don't know if there's a different way to set up the workflow file such that it'll avoid duplicate runs on PRs and still run CI for branch pushes if the branch isn't associated with a PR (which I suspect you want to keep doing).

PR #460, opened 2020-08-31
cache bundler vendor directory in CI

joelhawksley:
This PR attempts to configure GitHub Actions
to cache installed gems between builds in CI,
with the goal of speeding up CI builds.

It also moves us to building against Rails `master`
by default, as we're approaching the release of 6.1.

cc @rmacklin @manuelpuyol


joelhawksley:
From the check logs (https://github.com/github/view_component/pull/460/checks?check_run_id=1052418926), it looks like caching is working! The 2.7.x builds are the first to benefit, as they could reuse the cache from the `lint` build:

Screen Shot 2020-08-31 at 11 46 15 AM

.github/workflows/ci.yml
```diff
@@ -11,10 +11,17 @@ jobs:
...
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-ruby2.7.x-${{ hashFiles('**/Gemfile.lock') }}
+        restore-keys: |
+          ${{ runner.os }}-gems-
     - name: Lint with Rubocop
       run: |
         gem install bundler:1.17.3
         bundle update
+        bundle config path vendor/bundle
```
rmacklin, line 24: Should we be running `bundle config path vendor/bundle` before `bundle update`? 
```suggestion
        bundle config path vendor/bundle
        bundle update
```
Although, I'm not clear on the background why we were calling `bundle update` and then `bundle install`...

.github/workflows/ci.yml
```diff
@@ -39,10 +46,17 @@ jobs:
       if: startsWith(matrix.ruby_version, '2.4')
       run: |
         gem update --system --no-document
+    - uses: actions/cache@v2
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-ruby${{ matrix.ruby_version }}-${{ hashFiles('**/Gemfile.lock') }}
```
rmacklin: Would we also want to include `matrix.rails_version` in the cache key, since that effects which gems we use?

.rubocop.yml
```diff
@@ -5,4 +5,5 @@ inherit_gem:
 
 AllCops:
   Exclude:
-    - "test/tmp/**/*"
+    - test/tmp/**/*
+    - vendor/bundle/**/*
```
rmacklin, line 9: Not sure if it was intentional, but it looks like the trailing newline was removed:

image

.github/workflows/ci.yml
```diff
@@ -11,10 +11,17 @@ jobs:
       uses: actions/setup-ruby@v1
       with:
         ruby-version: 2.7.x
+    - uses: actions/cache@v2
+      with:
+        path: vendor/bundle
+        key: ${{ runner.os }}-gems-ruby2.7.x-${{ hashFiles('**/Gemfile.lock') }}
```
rmacklin, line 17:
```suggestion
        key: ${{ runner.os }}-gems-railsmaster-ruby2.7.x-${{ hashFiles('**/Gemfile.lock') }}
```

.github/workflows/ci.yml
rmacklin: Looks like you pushed c0889138fdf0389c53d79878bca5358c6474c3fc before I submitted this review! So this comment is stale, but added a related comment here: https://github.com/github/view_component/pull/460#pullrequestreview-478832486

manuelpuyol:
a nit:

I see we have two caches

1. `gems-analysis-${{ hashFiles('**/Gemfile.lock') }}`
2. `gems-build-rails-${{ matrix.rails_version }}-ruby-${{ matrix.ruby_version }}-${{ hashFiles('**/Gemfile.lock') }}`

couldn't we make the first match the second so we have a single cache?

`gems-build-rails-master-ruby-2.7.x-${{ hashFiles('**/Gemfile.lock') }}`

.github/workflows/ci.yml
```diff
@@ -11,9 +11,14 @@ jobs:
...
+    - uses: actions/cache@v2
+      with:
+        path: vendor/bundle
+        key: gems-analysis-${{ hashFiles('**/Gemfile.lock') }}
     - name: Lint with Rubocop
       run: |
         gem install bundler:1.17.3
+        bundle config path vendor/bundle
         bundle update
         bundle install --jobs 4 --retry 3
```
rmacklin, line 23:
```suggestion
        bundle config path vendor/bundle
        bundle install --jobs 4 --retry 3
```

.github/workflows/ci.yml
```diff
@@ -39,9 +44,14 @@ jobs:
...
+    - uses: actions/cache@v2
+      with:
+        path: vendor/bundle
+        key: gems-build-rails-${{ matrix.rails_version }}-ruby-${{ matrix.ruby_version }}-${{ hashFiles('**/Gemfile.lock') }}
     - name: Build and test with Rake
       run: |
         gem install bundler:1.17.3
+        bundle config path vendor/bundle
         bundle update
         bundle install --jobs 4 --retry 3
```
rmacklin, line 56:
```suggestion
        bundle config path vendor/bundle
        bundle update
```

.github/workflows/ci.yml
```diff
@@ -64,9 +74,14 @@ jobs:
...
+    - uses: actions/cache@v2
+      with:
+        path: vendor/bundle
+        key: gems-analysis-${{ hashFiles('**/Gemfile.lock') }}
     - name: Collate simplecov
       run: |
         gem install bundler:1.17.3
+        bundle config path vendor/bundle
         bundle update
         bundle install --jobs 4 --retry 3
```
rmacklin, line 86:
```suggestion
        bundle config path vendor/bundle
        bundle install --jobs 4 --retry 3
```

rmacklin:
Yeah, I'm wondering the same. I see that they were split into separate caches in f28f49ea85bcb957203df1fac211897898a07e69 but I'm wondering if that change became unnecessary by 7b0ba4c50052fc21c552d2199d70a3e3de347b38 & f1ea57ad21bd13269f22f7fd63ff8eb9b6cca86d?

PR #462, opened 2020-09-03
Introduce component template inheritance

BlakeWilliams:
This change introduces template inheritance to components. Previously
subclassed components would have to define, or re-define a template
file. With this change, component templates are compiled when they're
inherited so the subclass will also inherit the `call_*` template methods.

This functionality already existed for `inline` templates defined via
`call`, but this change introduces a test to ensure we have coverage of
that intended behavior.

related to https://github.com/github/view_component/discussions/455


joelhawksley:
@DiegoOrejuela would you be willing to provide the first round of code review here?

lib/view_component/base.rb
```diff
@@ -179,6 +179,10 @@ def short_identifier
       end
 
       def inherited(child)
+        # Compile so child will inherit compiled `call_*` template methods that
+        # `compile` defines
+        compile
```
fsateler, line 184: Shouldn't compilation be triggered at compile time? That is, when a child compononent is compiled, it should first ensure its parent is compiled.

During development, this would cause more compilations than necessary.

BlakeWilliams: @fsateler That's a really great question! If I had to guess, the performance impact would be relatively small. I plan on trying to dig into some numbers to see what that impact may be, then make an informed decision from there.

DiegoOrejuela:
> @DiegoOrejuela would you be willing to provide the first round of code review here?

@joelhawksley On the weekend I make an example of what I expect. I feel like this development somehow covers a case that I hadn't thought of. Thank you @BlakeWilliams

lib/view_component/base.rb
fsateler: Well, I figure there's a reason why we delay compilation for components in general :wink:

BlakeWilliams: I checked the numbers and compiles look like they're super fast. I was seeing the slowest compile take around 1ms on my machine with the majority of compiles being much faster than that. I think that seems like a fine trade-off (at least for now) instead of adding more complexity here.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # master
 
+* Subclassed components now inherit their parent's templates
```
```suggestion
* Subclassed components inherit their parent's templates.
```

joelhawksley (approved):

joelhawksley:
@BlakeWilliams looks good. Mind adding a note about this functionality to the readme?

README.md
```diff
@@ -316,6 +316,18 @@ class InlineVariantComponent < ViewComponent::Base
 end
 ```
 
+### Template Inheritance
+
+Components that subclass another component automatically inherit their parent's
+template if they don't define their own template.
+
+```ruby
+# If `my_link_component.html.erb` is not defined so the component will fall back
```
```suggestion
# If `my_link_component.html.erb` is not defined, the component will use
```

README.md
```diff
@@ -316,6 +316,18 @@ class InlineVariantComponent < ViewComponent::Base
 end
 ```
 
+### Template Inheritance
+
+Components that subclass another component automatically inherit their parent's
```
joelhawksley, line 338:
```suggestion
Components that subclass another component inherit the parent component's
```

joelhawksley (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master
 
+* Subclassed components inherit their parent's templates.
```
```suggestion
* Subclassed components inherit templates from parent.
```

PR #463, opened 2020-09-03
Raise errors if helpers/controller are called before render

jules2689:
As we only set view context during render, we cannot call into any view-context
objects like helpers or controllers.

Unfortunately the error is really obtuse, so this adds a proper exception to tell
the developer exactly what is going on

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -116,7 +116,7 @@ module ViewComponent
 
     # Provides a proxy to access helper methods from the context of the current controller
     def helpers
-      raise ViewContextCalledBeforeRenderError, "`helpers` can only be called at render time." if controller.view_context.nil?
+      raise ViewContextCalledBeforeRenderError, "`helpers` can only be called at render time." if view_context.nil?
       @helpers ||= controller.view_context
     end
 
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -75,7 +75,6 @@ module ViewComponent
       @content = view_context.capture(self, &block) if block_given?
 
       before_render
-      @render_in_called = true
 
       if render?
         send(self.class.call_method_name(@variant))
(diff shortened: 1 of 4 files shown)


lib/view_component/base.rb
```diff
@@ -108,11 +111,13 @@ def render(options = {}, args = {}, &block)
     end
 
     def controller
+      raise ViewContextCalledBeforeRenderError, "Cannot call `controller` from initialize. Must be called during render" unless @render_in_called
```
joelhawksley, line 114: Two things:

1) Do you think we could simplify this check to see if `view_context` is defined or not, instead of using `@render_in_called`? 
2) I wonder if we could be more specific with the error message:

```suggestion
      raise ViewContextCalledBeforeRenderError, "`controller` can only be called at render time.
```

joelhawksley (commented):
I like this change. Anything to make these cases easier to understand ❤️ 

Mind adding a CHANGELOG?

lib/view_component/base.rb
jules2689: > Do you think we could simplify this check to see if view_context is defined or not, instead of using @render_in_called?

I considered this approach, but the ivar seemed more direct instead of implicit. I'm cool either way, WDYT?

>  wonder if we could be more specific with the error message:

Totally!

jules2689:
@joelhawksley updated!

lib/view_component/base.rb
```diff
@@ -108,11 +110,13 @@ def render(options = {}, args = {}, &block)
     end
 
     def controller
+      raise ViewContextCalledBeforeRenderError, "`controller` can only be called at render time." if view_context.nil?
       @controller ||= view_context.controller
     end
 
     # Provides a proxy to access helper methods from the context of the current controller
     def helpers
+      raise ViewContextCalledBeforeRenderError, "`helpers` can only be called at render time." if view_context.nil?
```
joelhawksley, line 119:
```suggestion
      raise ViewContextCalledBeforeRenderError, "`helpers` can only be called at render time." if controller.view_context.nil?
```

lib/view_component/base.rb
```diff
@@ -108,11 +110,13 @@ def render(options = {}, args = {}, &block)
     end
 
     def controller
+      raise ViewContextCalledBeforeRenderError, "`controller` can only be called at render time." if view_context.nil?
       @controller ||= view_context.controller
     end
 
     # Provides a proxy to access helper methods from the context of the current controller
     def helpers
+      raise ViewContextCalledBeforeRenderError, "`helpers` can only be called at render time." if controller.view_context.nil?
```
joelhawksley, line 119:
```suggestion
      raise ViewContextCalledBeforeRenderError, "`helpers` can only be called at render time." if view_context.nil?
```

joelhawksley (approved):

PR #465, opened 2020-09-07
Allow customization of the controller used in component tests

### LINKED ISSUE #440: Allow `test_controller` to be customizable per component
## Feature request

Right now, there is no way to customize the `test_controller` on a per-component-basis, which has made it difficult for us to test different groups of components.

`Base.test_controller` is a `mattr_accessor`, which doesn't allow customizing on a per-component basis. Additionally, `TestHelpers#controller` always uses `Base.test_controller`, rather than using the `test_controller` of the component currently under test.

If `Base.test_controller` was changed to a `class_attribute`, and `TestHelpers#controller` was changed to use the more specific one, the default scenario of everyone using `ApplicationController` would continue to work, while those who want to isolated different namespaces of an application's components would be able to do so!

### Motivation

We have 3 different namespaces within our application, each with its own set of view components (and each with their own base controller class, i.e. `Foo::BaseController` and `Bar::BaseController`). We'd love to be able to mirror that hierarchy in our component classes, having `Foo::BaseComponent` that all of our `Foo`ey components inherit from, and then be able to inherit from something other than `ApplicationController` in those component's tests.

agrobbin:
Sorry this took me so long to get back to @joelhawksley! Finally was able to spend some time here during the long weekend.

### Summary

If you have different namespaces within your application, or possibly public and authenticated controllers, which change component functionality in specific ways, you can now configure a different controller in component tests.

Closes #440.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -936,14 +936,34 @@ Previews can be extended to allow users to add authentication, authorization, be
 config.view_component.preview_controller = "MyPreviewController"
 ```
 
-#### Configuring TestController
+### Configuring the controller used in tests
 
-Component tests assume the existence of an `ApplicationController` class, which be can be configured using the `test_controller` option:
-
-`config/application.rb`
+Component tests assume the existence of an `ApplicationController` class, which can be configured globally using the `test_controller` option:
 
 ```ruby
 config.view_component.test_controller = "BaseController"
+```
+
+To configure the controller used for a test case, use `with_controller_class` from `ViewComponent::TestHelpers`.
+
+```ruby
+class ExampleComponentTest < ViewComponent::TestCase
+  def test_component_in_public_controller
+    with_controller_class PublicController do
+      render_inline ExampleComponent.new
+
+      assert_text "foo"
+    end
+  end
+
+  def test_component_in_authenticated_controller
+    with_controller_class AuthenticatedController do
+      render_inline ExampleComponent.new
+
+      assert_text "bar"
+    end
+  end
+end
 ```
 
 ### Setting up RSpec
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -193,8 +193,9 @@ module ViewComponent
     end
 
     # The controller used for testing components.
-    # Defaults to ApplicationController. This should be set early
-    # in the initialization process and should be set to a string.
+    # Defaults to ApplicationController, but can be configured
+    # on a per-test basis using `with_controller_class`.
+    # This should be set early in the initialization process and should be a string.
     mattr_accessor :test_controller
     @@test_controller = "ApplicationController"
 
--- a/lib/view_component/test_helpers.rb
+++ b/lib/view_component/test_helpers.rb
@@ -35,7 +35,7 @@ module ViewComponent
     end
 
     def controller
-      @controller ||= Base.test_controller.constantize.new.tap { |c| c.request = request }.extend(Rails.application.routes.url_helpers)
+      @controller ||= build_controller(Base.test_controller.constantize)
     end
 
     def request
@@ -47,7 +47,21 @@ module ViewComponent
 
       controller.view_context.lookup_context.variants = variant
       yield
+    ensure
       controller.view_context.lookup_context.variants = old_variants
+    end
+
+    def with_controller_class(klass)
+      old_controller = defined?(@controller) && @controller
+
+      @controller = build_controller(klass)
+      yield
+    ensure
+      @controller = old_controller
+    end
+
+    def build_controller(klass)
+      klass.new.tap { |c| c.request = request }.extend(Rails.application.routes.url_helpers)
     end
   end
 end
--- a/test/app/components/custom_test_controller_component.rb
+++ b/test/app/components/custom_test_controller_component.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+class CustomTestControllerComponent < ViewComponent::Base
+  def call
+    helpers.foo
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Allow customization of the controller used in component tests.
+
+    *Alex Robbin*
+
 * Generate preview at overridden path if one exists when using `--preview` flag.
 
     *Nishiki Liu*
--- a/test/app/controllers/custom_test_controller_controller.rb
+++ b/test/app/controllers/custom_test_controller_controller.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+class CustomTestControllerController < ActionController::Base
+  helper_method :foo
+
+  private
+
+  def foo
+    "foo"
+  end
+end
--- a/test/view_component/view_component_test.rb
+++ b/test/view_component/view_component_test.rb
@@ -478,6 +478,14 @@ class ViewComponentTest < ViewComponent::TestCase
     assert_includes error.message, "More than one template found for TemplateAndSidecarDirectoryTemplateComponent."
   end
 
+  def test_with_custom_test_controller
+    with_controller_class CustomTestControllerController do
+      render_inline(CustomTestControllerComponent.new)
+
+      assert_text("foo")
+    end
+  end
+
   def test_backtrace_returns_correct_file_and_line_number
     error = assert_raises NameError do
       render_inline(ExceptionInTemplateComponent.new)


lib/view_component/test_helpers.rb
```diff
@@ -21,9 +21,11 @@ def refute_component_rendered
       # :nocov:
     end
 
-    attr_reader :rendered_component
+    attr_reader :controller, :variant, :rendered_component
```
jonspalmer, line 24: In this new factoring is `controller` used outside of the `render_inline` method? If not why not make it just a variable in that method?

lib/view_component/test_helpers.rb
```diff
@@ -34,20 +36,22 @@ def render_inline(component, **args, &block)
...
-    def with_variant(variant)
-      old_variants = controller.view_context.lookup_context.variants
+    def with_variant(new_variant)
+      old_variant = variant
 
-      controller.view_context.lookup_context.variants = variant
+      @variant = new_variant
       yield
-      controller.view_context.lookup_context.variants = old_variants
+      @variant = old_variant
```
jonspalmer, line 48: Are there strong tests on this change? Its subtly different than the previous implementation in that the "old" version would be read from the controller lookup_context (which I assume could have other defaults etc)

jonspalmer: as an aside reading this code again I'm sort of confused why we have this pattern. Its strikes me it would be much simpler to pass `variant` as an argument to `render_inline` vs this wrapper. @joelhawksley any thoughts on why the API is like this?

lib/view_component/test_helpers.rb
```diff
@@ -34,20 +36,22 @@ def render_inline(component, **args, &block)
...
+    end
+
+    private
+
+    def build_controller(component)
+      component.test_controller.constantize.new.tap { |c| c.request = request }.extend(Rails.application.routes.url_helpers)
```
jonspalmer, line 54: I'm imagine that the `constantize` is slightly expensive and we aren't memoizing it per component. I wonder if we can arrange that?

test/app/components/custom_test_controller_component.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class CustomTestControllerComponent < ViewComponent::Base
+  self.test_controller = "CustomTestControllerController"
+
+  def call
+    helpers.foo
+  end
+end
```
jonspalmer, line 9: I think we need some more test cases that make sure the behavior of setting `test_controller` is as expected. I'm pretty sure this test could/would still pass if test_controller was a global parameter vs scoped to the component class. I think a few other things are needed like.

- setting the test_controller doesn't impact other components
- Subclassing a component with test_controller set inherits the test_controller
- Subclassing a component and overriding the test_controller of the super class that set test_controller

test/app/components/custom_test_controller_component.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class CustomTestControllerComponent < ViewComponent::Base
+  self.test_controller = "CustomTestControllerController"
```
jonspalmer, line 4: A lot of rails apis would wrap class_attributes to provide a method so you can do `test_controller("CustomTestControllerController")` or the like. 

Also is it necessary for this to be a string vs a class? I believe we want a string version as something that can get set in application.rb or initializers without having to worry about the class being autoloaded. However, I wonder at this point if we can't just use classes.

jonspalmer (commented):
I like the feature this PR adds and I can see why it's powerful. 

A few comments and questions on the implementation details.

One overall comment though is it feels strange to pollute the "production" code for view components with details about how they are tested. That `test_controller` is a property of the Component code vs some declaration in the testing code is somewhat unexpected. I understand why it's simpler this way but I wonder if we can't find a way to factor this support differently to leave the "production" component code alone.

lib/view_component/test_helpers.rb
agrobbin: At some point during my work here, I did run into some test failures due to me not correctly modifying `controller.view_context.lookup_context.variants`, but I'm going to defer to the maintainers here for sure! If you think there's some test coverage missing I'm happy to add some.

lib/view_component/test_helpers.rb
agrobbin: Definitely can't hurt, done!

test/app/components/custom_test_controller_component.rb
agrobbin: Personally, I like the less DSL-y `self.test_controller=` interface over the more magical `test_controller 'CustomTestControllerController'`, but regardless, I tried not to change the interface here to leave things backwards compatible and not introduce more changes that I needed to!

test/app/components/custom_test_controller_component.rb
agrobbin: Done, done, and done @jonspalmer! I've added a new `test_test_controller_as_class_attribute` case to `view_component_test.rb` that should cover these scenarios.

agrobbin:
Just checking in on this @joelhawksley @jonspalmer. Would love to get this into a future version of ViewComponent!

joelhawksley:
@jonspalmer mind having another look here?

mrbongiolo:
Shouldn't the test be responsible for setting up the controller to use? It seems kinda odd to setup something related to testing directly in the component. 
Couldn't we provide `test_controller = MyController` with the TestHelpers? So that it could be changed by the tests that require a different controller.
I do something like this on my project to test the view "objects" and view_components.

agrobbin:
@mrbongiolo sorry for the delay here! I didn't want to change the interface here, requiring a deprecation cycle or anything of the sort. Leaving `test_controller` as a method on the component class itself keeps everything working like it did before, with the ability to customize it per class.

agrobbin:
@joelhawksley @jonspalmer just wondering if this is something that could get another peak!

joelhawksley (commented):
👋🏻 Thanks for your work on this PR @agrobbin! I totally see the need to have the ability to change the test controller, and think it's something we should add to the framework.

After reading this PR over with @BlakeWilliams this morning, I think we should consider implementing this functionality such that we configure the `test_controller` inside test files, not the component itself. I could see us doing this two ways:

### `with_test_controller`

It might be nice to have a helper to change the test controller on a case-by-case basis. For example:

```ruby
with_test_controller(MyTestController) do
  render_inline(MyComponent.new)
  #...
end
```

### `test_controller` method/macro on test class

For a broader solution, I think we could lean on a per-test-file configuration method. Perhaps it could be something like:

```ruby
class MyTestCase
  # Avoid test_ prefix to prevent conflict w/ minitest 
  def view_component_test_controller
    MyTestController
  end
end
```

or, as a macro:

```ruby
class MyTestCase
  test_controller MyTestController
end
```

---

Overall, the reason I'm suggesting we make this change is that I'm wary of introducing test-specific behavior to the ViewComponent::Base API itself. I'm more than open to being told this is a bad idea, so please don't hesitate to push back if you think this is the wrong direction to take ❤️

agrobbin:
Hi @joelhawksley and @BlakeWilliams! First off, really sorry it's taken me so long to get back to this PR. I've been unable to find the time until this weekend. Also, thanks for the detailed follow-ups and suggestions.

I don't disagree at all with your assessment that `test_controller` being defined on the component itself is not ideal. It definitely feels like a better option to have this available as a test-specific option.

I've just pushed an update here that introduces a `test_controller` macro to `TestHelpers`, that I believe does what you're looking for. I left `Base.test_controller`, allowing that to be the way to change the global default (and maintaining backwards compatibility).

Let me know what you think!

jonspalmer:
I like the improvements. I strongly believe we should also add a `with_test_controller` helper and document the differences. This allows more obvious/flexible testing patterns and support testing frameworks like RSpec more cleanly.

My challenge is that setting `test_controller` the way we are has side effects. Consider changing the tests just a little:

```ruby
class CustomTestControllerTest < ViewComponent::TestCase
  
  def test_custom_test_controller
    self.class.test_controller CustomTestControllerController
    render_inline(CustomTestControllerComponent.new)

    assert_text("foo")
  end
  
  def test_no_custom_test_controller
    assert_not_equal(CustomTestControllerController, self.class.test_controller)
  end
end
```

The second test will pass if it's run first and fail if it's run second. 

It's certainly fair to say the macro should only be used at the top level not inside the individual test cases but that's a somewhat limiting API. A `with_test_controller` method would make this pattern much more obvious.

```ruby
class CustomTestControllerTest < ViewComponent::TestCase
  
  def test_custom_test_controller
    with_test_controller(CustomTestControllerController) do
      render_inline(CustomTestControllerComponent.new)

      assert_text("foo")
    end
  end
  
  def test_no_custom_test_controller
    assert_not_equal(CustomTestControllerController, self.class.test_controller)
  end
end
```


This turns out to be powerful in RSpec which has concepts like nested contexts. To explore this suppose my app has public pages and pages that need authentication. We have `PublicController` and `AuthenticatedController` classes that enforce that logic. Suppose I have an `AnalyticsComponent` the fires of visitor or user metrics to a service like Google Analytics and it varies its behavior based the controller it's being rendered by.  (Aside - I have no idea if this setup architecturally is actually a "good idea" but its close enough to real work examples that its a pattern we should support).

To test this need two sets of tests one for public and one for authenticated.  

I suspect in MiniTest (for which I'm less familiar) you would have separate test files:

```ruby
class AnalyticsComponentPublicControllerTest < ViewComponent::TestCase
  test_component PublicController
  
  def test_foo do ... end
  def test_bar do ... end
end

class AnalyticsComponentAuthenticatedControllerTest < ViewComponent::TestCase
  test_component AuthenticatedController
  
  def test_fizz do ... end
  def test_buzz do ... end
end
```

However RSpec you'd write these tests in one file something like this:

```ruby
describe AnalyticsComponent do
  context "with public controller" do
    around do |example|
      with_test_controller(PublicController) { example.run }
    end
    
    it "test foo" do ... end
    it "test bar" do ... end
  end
  
  context "with authenticated controller" do
    around do |example|
      with_test_controller(AuthenticatedController) { example.run }
    end
    
    it "test fizz" do ... end
    it "test buzz" do ... end
  end
end
```        

Should be a simple enough addition to pattern match after `with_variant`

jonspalmer:
@BlakeWilliams & @joelhawksley Can we revisit the PVC tests #525. This failure is kinda weird: https://github.com/github/view_component/runs/1943909274?check_suite_focus=true

It could be a bug in this PR that the `ActiveSupport::Concern` `class_methods` aren't working. But it could also point to a gap in how the PVC tests are run by ViewComponent or how PVC is using TestHelpers. Thoughts?

agrobbin:
@jonspalmer I just pushed up another pass at this, including both `test_controller` and `with_test_controller`. Let me know what you think!

jonspalmer:
I like it. Seems clean. One question looking at this again with slightly fresher eyes. I know why we called the config setting `test_controller` but in this new APi the `test` feels redundant. Would it be more natural to call the new methods  `controller_class` and `with_controller`?

lib/view_component/test_helpers.rb
```diff
@@ -49,5 +61,17 @@ def with_variant(variant)
...
       controller.view_context.lookup_context.variants = old_variants
     end
+
+    def with_test_controller(klass)
+      old_controller = defined?(@controller) && @controller
+
+      @controller = build_controller(klass)
+      yield
+      @controller = old_controller
+    end
```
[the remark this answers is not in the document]
agrobbin: @BlakeWilliams this follows the same pattern as `with_variant`, which does not use `ensure`. I believe RSpec gracefully takes care of this without needing `ensure`, with the way it captures expectation failures, but I don't know if other test frameworks like Minitest do a similar thing.

If you still think we should do this, should I update `with_variant` as well while I'm in the neighborhood?

lib/view_component/test_helpers.rb
```diff
@@ -35,7 +47,7 @@ def render_inline(component, **args, &block)
     end
 
     def controller
-      @controller ||= Base.test_controller.constantize.new.tap { |c| c.request = request }.extend(Rails.application.routes.url_helpers)
+      @controller ||= build_controller(self.class.test_controller)
```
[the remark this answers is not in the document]
agrobbin: I'm not sure I follow your suggestion @BlakeWilliams. `self.class.test_controller` here is referring to [this method](https://github.com/github/view_component/pull/465/files#diff-d66a35a69a27f26d54c6db1034c0a6b61e5df8df60e12c7030d97fe28faae9e2R27-R33), while your suggested change removes the ability to write a component test that sets a different controller for the entire test.

Or are you suggesting that we *only* allow overriding the controller via `with_controller`? I'm personally not in favor of that, as the original goal of this functionality was to allow an app with multiple application namespaces with similarly namespaced components to easily change the controller for all of its tests, which would become less convenient by removing that functionality.

agrobbin:
@jonspalmer I just pushed an update here that renames things. I went with `controller_class` and `with_controller_class`, for symmetry. Let me know what you think!

lib/view_component/test_helpers.rb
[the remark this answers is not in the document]
BlakeWilliams: > If you still think we should do this, should I update with_variant as well while I'm in the neighborhood?

Yes please, if you don't mind. ❤️

lib/view_component/test_helpers.rb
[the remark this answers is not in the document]
BlakeWilliams: > Or are you suggesting that we _only_ allow overriding the controller via `with_controller`?

I think for the first iteration of this API `with_controller` is a huge improvement and would love to see how that's used before adding a public method for overriding the controller per-test.

I think you could still use the private API to override controllers globally via something like:

```ruby
# minitest
def setup
  @controller = build_controller(MyController)
end

# might break with_controller, but also works?
def controller
  @_controller ||= MyController.new
end
```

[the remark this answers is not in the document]
agrobbin: OK I just pushed an update here that strips this down to just `with_controller_class`!

docs/index.md
```diff
@@ -936,16 +936,60 @@ Previews can be extended to allow users to add authentication, authorization, be
...
-Component tests assume the existence of an `ApplicationController` class, which be can be configured using the `test_controller` option:
-
-`config/application.rb`
+Component tests assume the existence of an `ApplicationController` class, which can be configured globally using the `test_controller` option:
 
 ```ruby
 config.view_component.test_controller = "BaseController"
 ```
 
+You can configure the controller used for a test case using the `with_controller_class` helper from `ViewComponent::TestHelpers`.
```
joelhawksley, line 947:
```suggestion
To configure the controller used for a specific test case, use `with_controller_class` from `ViewComponent::TestHelpers`:
```

docs/index.md
```diff
@@ -936,16 +936,60 @@ Previews can be extended to allow users to add authentication, authorization, be
...
+  context "with authenticated controller" do
+    around { |example| with_controller_class(AuthenticatedController, &example) }
+
+    it "shows bar" do
+      render_inline ExampleComponent.new
+
+      expect(page).to have_text("bar")
+    end
+  end
+end
```
joelhawksley, line 990: While we happily support RSpec, I'm hesitant to have examples from multiple test frameworks in the documentation.

```suggestion
```

lib/view_component/base.rb
```diff
@@ -192,7 +192,7 @@ def content_evaluated?
       @_content_evaluated
     end
 
-    # The controller used for testing components.
+    # The default controller used for testing components.
     # Defaults to ApplicationController. This should be set early
     # in the initialization process and should be set to a string.
```
joelhawksley, line 197:
```suggestion
    # The controller used for testing components.
    # Defaults to ApplicationController, but can be configured 
    # on a per-test basis using `with_controller_class`.  
    # This should be set early in the initialization process and should be a string.
```

lib/view_component/test_helpers.rb
```diff
@@ -2,6 +2,8 @@
 
 module ViewComponent
   module TestHelpers
+    extend ActiveSupport::Concern
```
joelhawksley, line 5: I'm curious- What necessitated this change?

joelhawksley (commented):
Looking good! Thank for all of the work you've put into getting this PR together ❤️

lib/view_component/test_helpers.rb
agrobbin: Whoops, this was leftover from when `controller_class` was available as a top-level macro method. Removed!

agrobbin:
Thanks @joelhawksley, just pushed an update!

BlakeWilliams (approved):
This looks great!

Once we resolve the CHANGELOG conflict, we can merge this in. 👍

agrobbin:
@BlakeWilliams rebased!

BlakeWilliams (approved):

BlakeWilliams:
Thank you for the contribution! ❤️

PR #476, opened 2020-09-10
[WIP] Add support for fragment caching

dkniffin:


pinzonjulian:
Oh @dkniffin  this test was so simple to write! I'm glad you got around to do it! 🤘🏽

test/view_component/integration_test.rb
```diff
@@ -475,4 +475,16 @@ class IntegrationTest < ActionDispatch::IntegrationTest
...
     end
   end
+
+  test "Rails fragment cache gets flushed correctly when the component template changes" do
+    get "/fragment_cache"
+    assert_response :success
+    assert_select("div", "Foo\n  bar")
+    assert_select("div.changed", false)
+
+    modify_file "app/components/erb_component.html.erb", "<div class='changed'>Hello World</div>" do
```
juanmanuelramallo, line 485: I don't think this is correct in the context of fragment caching. 

The template is cached using a key, when the key changes the cache block is executed again and cached under the new key. The old key will be flushed, in time, depending on the eviction algorithm using the cache itself.

This example is rather showing the scenario where a developer changes the production code. Which is a different problem, one should decide to either update the cache key construction before releasing the changes or maybe using an Active Record object that responds to cache_key_with_version.

Moreover I think ViewComponent already play nice with Rails' fragment caching, see this example https://github.com/github/view_component/blob/master/test/view_component/integration_test.rb#L159 using Rails' `cache_if`.

juanmanuelramallo: Well nevermind, I just realized this is a valid case when the cache entry can't be manually invalidated or changed. 

I'll try to lend a hand into investigating how to solve the issue.

Matt-Yorkley:
Alright! There are some hard problems here but I feel like the mechanisms which allow fragment caching to work with regular views also face very similar problems (specifically in relation to the digest) and already solve them, right?

So for example, `ActionView::Digestor` performs some tasks with regular templates to ensure that the digest for any given template will be updated when it should be, which includes (roughly):
- Taking an MD5 of the each tempate file
- Building a quick list of _dependencies_ for each template (like if a view also renders another view inside it, the digest needs to change if the _child_ view changes)
- Making a final digest based on the content of the template file _and_ it's dependencies

And after that's done, there's a simple lookup that takes the path to a given template and return that digest.

So what's needed is something that looks a lot like `ActionView::Digestor` and performs roughly the same tasks, but for components, right? But with a couple of caveats related to the key differences between regular templates and components, like:
- components could have _includes_ (like helpers), and you'd want to include those in the "dependencies"
- components can have inheritance (like `ActionComponent`, but potentially others), and you'd also want to factor that in when building the digest

joelhawksley:
@dkniffin thanks for taking this work on! Are you cool with closing this in favor of @Matt-Yorkley's work?

dkniffin:
Yep! 👍

PR #484, opened 2020-09-16
Bug report + failing test for using @variant in a component

dbackeus:
### Summary

When attempting to use an instance variable named `@variant` in a component it's magically changed to `nil` when rendered instead of retaining its original value.

In some twist of fate this hit me on my very first attempt at making a ViewComponent which was going to be a wrapper for Bootstrap's Alert component. I looked at React Bootstrap for naming inspiration and stole the idea of [using the "variant" prop](https://react-bootstrap.github.io/components/alerts/#alert-props) to define which style the alert should have and that was that.


test/app/components/variant_ivar_component.rb
```diff
@@ -0,0 +1,9 @@
+class VariantIvarComponent < ViewComponent::Base
```
joelhawksley, line 1:
```suggestion
# frozen_string_literal: true

class VariantIvarComponent < ViewComponent::Base
```

BlakeWilliams (approved):

PR #488, opened 2020-09-21
use def test_method_name syntax for tests

joelhawksley:
### Summary

This PR standardizes our test syntax to use method names.


jonspalmer:
Still makes me sad we're not using rspec but this cleanup makes sense 😄

manuelpuyol (approved):
lgtm

PR #491, opened 2020-09-28
Add `before_render` DSL method

BlakeWilliams:
This adds a new DSL method, `before_render` that allows components to
define a block that is called before rendering. This block uses
`ActiveSupport::Callbacks` under the hood, so it can prevent a component
from rendering via `throw :abort`.

* Include `ActiveSupport::Callbacks` in `ViewComponent::Base`
* Define callbacks for `render`
* Add public `before_render` method to `ViewComponent::Base` that allows
  components to define a method that is called before rendering and
  prevent `render` from executing, allowing us to eventually replace
  `render?` and `#before_render`.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -179,8 +179,16 @@ module ViewComponent
     class << self
       attr_accessor :source_location, :virtual_path
 
-      def before_render(&block)
-        set_callback :render, :before, block
+      def before_render(method_name = nil, &block)
+        if block_given?
+          set_callback :render, :before, block
+        else
+          set_callback :render, :before, method_name
+        end
+      end
+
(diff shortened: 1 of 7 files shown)


BlakeWilliams:
At least locally, it looks like there is a small performance penalty when using `ActiveSupport::Callbacks`.

Pre-callback benchmark: ~10x faster than partials
Post-callback benchmark: ~9.5x faster than partials

jonspalmer:
We've been down a path like this before with #192 and ended up abandoning it. Can you share some of the motivation and some of the use cases where we'd need this support (and prepared to deal with the performance overhead for it).

BlakeWilliams:
Closing this for now

PR #492, opened 2020-09-29
Fix duplicate filenames in template search

suan:
### Summary

Within a codebase which has `.rb` added as a template handler (eg via `ActionView::Template.register_template_handler(:rb, :source.to_proc)`), `#matching_views_in_source_location` wrongly returns duplicate filepaths for erb view component templates, because "rb" is a sbstring of "erb", and the lookup uses a naive `Dir` glob algo with the registered template extensions.

Fix this by doing `.uniq` after each search

With the current behavior, the above scenario ultimately results in a bogus "There can only be one default template file per component" being raised at render-time

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -393,14 +393,14 @@ module ViewComponent
         extensions = ActionView::Template.template_handler_extensions.join(",")
 
         # view files in the same directory as the component
-        sidecar_files = Dir["#{location_without_extension}.*{#{extensions}}"]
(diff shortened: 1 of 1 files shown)


BlakeWilliams:
👋 @suan Thanks for the contribution! Could you add a test for this case?

nielsslot:
With #671 merged I can't imagine this being an issue anymore. @suan, any chance you could tell us if your issue is fixed in current main?

joelhawksley:
Closing as resolved by #671. Thanks for pointing this out @nielsslot!

PR #493, opened 2020-10-01
Allow preview controller to be customized via config options

### LINKED ISSUE #434: Make previews controller extensible
## Feature request

Make the previews controller to be extensible by allowing users to add authentication, authorization, before actions, or anything that the end user would need to meet their needs. 

### Motivation
It all started in discussion #412 
@jonspalmer came up with a good suggestion, which is to copy Devise's "parent class" feature. It will enable devs to set up a parent class if needed. 

https://github.com/github/view_component/discussions/412#discussioncomment-41084

swanson:
Resolves #434

### Summary

Adds a `preview_base_controller` configuration option to allow users to override the base controller used for previews.


swanson:
I'm up for trying to add tests case if desired. I tried for a bit but was having problems:

- I used the same approach as the  `with_preview_route` in the test helpers but it didn't seem to pick up the configuration changes I was making
- I didn't have any great ideas on what to assert as the original issue did not have any specific use-cases on what kinds of customization would be expected

Open to ideas / feedback

scottbarrow:
Would love this ability, thank you.
I'm trying to preview a component that I built backed by a controller right now bodging the controller inline in the preview template is a lot of work
I'm looking forward to seeing this merged

lib/view_component/base.rb
```diff
@@ -160,6 +160,12 @@ def request
 
     attr_reader :content, :view_context
 
+    # The parent controller used for previewing components.
+    # Defaults to Rails::ApplicationController. This should be set early
+    # in the initialization process and should be set to a string.
+    mattr_accessor :preview_base_controller
+    @@preview_base_controller = "Rails::ApplicationController"
+
```
juanmanuelramallo, line 168: Maybe we can place this in the previewable module (?) https://github.com/github/view_component/blob/master/lib/view_component/previewable.rb

README.md
```diff
@@ -826,6 +826,16 @@ end
...
 Which enables passing in a value with <http://localhost:3000/rails/components/cell_component/default?title=Custom+title&subtitle=Another+subtitle>.
 
+#### Configuring preview controller
+
+Previews can be extended to allow users to add authentication, authorization, before actions, or anything that the end user would need to meet their needs using the `preview_base_controller` option:
+
+`config/application.rb`
+
+```ruby
+config.view_component.test_controller = "MyPreviewController"
```
juanmanuelramallo, line 836:
```suggestion
config.view_component.preview_base_controller = "MyPreviewController"
```

juanmanuelramallo (commented):
Maybe we can do something like this https://github.com/github/view_component/blob/master/test/test_helper.rb#L24 to test this feature?

swanson:
> Maybe we can do something like this https://github.com/github/view_component/blob/master/test/test_helper.rb#L24 to test this feature?

Thanks, I tried a similar approach but even when reloading, it didn't seem to pick up the configuration change. I will investigate further.

test/test_helper.rb
```diff
@@ -30,6 +30,15 @@ def with_preview_route(new_value)
...
 
+def with_preview_base_controller(new_value)
+  old_value = Rails.application.config.view_component.preview_base_controller
+  Rails.application.config.view_component.preview_base_controller = new_value
+  app.reloader.reload!
+  yield
+  Rails.application.config.view_component.preview_base_controller = old_value
+  app.reloader.reload!
+end
+
```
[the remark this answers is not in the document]
juanmanuelramallo: hmmm ok so it seems that the reloader doesn't actually reloads the previews controller hierarchy.

I'm fine with leaving this feature without tests for the time being then, until we come up with a way to cover it.

One shady suggestion might be to remove the definition of the controller `Object.send(:remove_const, 'ViewComponentsController')` and then require the file manually by doing `require 'app/controllers/view_components_controller`. I'm not sure that would work, but I'm sure it would look ugly :see_no_evil:

juanmanuelramallo (commented):
I'm not sure how @joelhawksley would like to proceed. IMHO we can extract the failing test out of this PR so that we can find a good way to test it.

app/controllers/view_components_controller.rb
```diff
@@ -2,7 +2,7 @@
 
 require "rails/application_controller"
 
-class ViewComponentsController < Rails::ApplicationController # :nodoc:
+class ViewComponentsController < ViewComponent::Base.preview_base_controller.constantize # :nodoc:
```
BlakeWilliams, line 5: I wonder if it would make sense to reverse the inheritance here?

To reverse it, I think we could:

* Always subclass `Rails::ApplicationController`, and custom controllers would be subclasses of `ViewComponentsController`
* Define a `config.view_component.preview_controller` instead of `config.view_component.preview_base_controller`

Then a custom controller could potentially be implemented like this:

```ruby
class MyViewComponentController < ViewComponentsController
  # Define things here
end
```

Then your configuration would look like:

```ruby
config.view_component.preview_base_controller = "MyViewComponentController"
```

I think that change may also make testing this easier.

Thoughts? Would that still solve the problems the existing approach solves?

juanmanuelramallo: I like this idea! And yes I think it would make testing easier as well.

swanson: That makes sense to me. I will work on that approach.

swanson:
@BlakeWilliams @juanmanuelramallo I've switched to having users subclass the controller vs providing the base class. But I'm still stuck on getting the tests working. Not sure if maybe the `test_controller` option is conflicting?

Symptoms: using the `test_helper` with `reload` doesn't seem to pick up the change OR maybe the `IntegrationExamplesController` is always taking over? If I remove the configuration for `test_controller` and add `preview_controller`, things work as expected, but can't seem to get it to swap out on the fly just for the one test case.

lib/view_component/engine.rb
```diff
@@ -87,9 +88,11 @@ class Engine < Rails::Engine # :nodoc:
       options = app.config.view_component
 
       if options.show_previews
+        preview_controller = options.preview_controller.sub(/Controller$/, "").underscore
+
         app.routes.prepend do
-          get options.preview_route, to: "view_components#index", as: :preview_view_components, internal: true
-          get "#{options.preview_route}/*path", to: "view_components#previews", as: :preview_view_component, internal: true
+          get options.preview_route, to: "#{preview_controller}#index", as: :preview_view_components, internal: true
+          get "#{options.preview_route}/*path", to: "#{preview_controller}#previews", as: :preview_view_component, internal: true
```
juanmanuelramallo, line 95: Shouldn't these two use the underscored version of the controller name? (and without the "Controller" substring)

swanson: Yes, they are -- see above.

swanson:
I had no success swapping out the config in a single test case, but instead I switched the test application to use a custom preview controller that passes through to the default controller -- unless a specific query string parameter is sent, in which case it returns a known hard-coded response. 

Super hacky, but it does pass. Open to other options/ideas!

lib/view_component/engine.rb
juanmanuelramallo: Nice one!

juanmanuelramallo (commented):
Hey I think I found the issue why the test helper method you added wasn't working. For some reason this variable https://github.com/github/view_component/pull/493/files#diff-28535a42af4bc1dff3bc550d0e400775R91 always contains "view_components". So I tried moving that inside the route definition block, and it worked as expected. 

It might be related to the fact that the rails app has already been initialized hence the variable remains the same. 

However, when we place that variable inside the route definition block (I mean here https://github.com/github/view_component/pull/493/files#diff-28535a42af4bc1dff3bc550d0e400775R93) and the test helper reloads the app that block is re-run. Ending up in the result we expect. 

Does this make sense?

swanson:
@juanmanuelramallo Wow! That was tricky, I don't think I would ever have thought to test that, thanks!

juanmanuelramallo (approved):
Neat!

BlakeWilliams:
@swanson This looks great! Can you add a changelog entry for this improvement?

once that's complete, @juanmanuelramallo would you like to merge this PR?

juanmanuelramallo:
@BlakeWilliams sure thing! Is there anything in particular I should have in mind? (it'd be my first time merging a PR here)

BlakeWilliams:
@juanmanuelramallo I generally follow the steps listed here, https://github.com/github/view_component/blob/master/CONTRIBUTING.md#submitting-a-pull-request

Things like ensuring that folks add themselves as a contributor, changelog entries if applicable, etc.

juanmanuelramallo:
Noted, thanks!

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## master
 
+* Add `preview_controller` option to overriding the controller used for previewing components
```
```suggestion
* Add `preview_controller` option to override the controller used for component previews.
```

juanmanuelramallo:
Hey @swanson sorry for the back and forth, but would you mind adding yourself in the list of contributors at the bottom of the readme?

Step 6 in this guide https://github.com/github/view_component/blob/master/CONTRIBUTING.md#submitting-a-pull-request and don't forget about the first part of step 8 :see_no_evil:

swanson:
@juanmanuelramallo no problem, done!

juanmanuelramallo:
Thank you @swanson!

PR #499, opened 2020-10-08
Don't add /test/components/previews preview path unless the directory exists

AMHOL:
This prevents a non-existent directory from being added to `ActiveSupport::Dependencies.autoload_paths` when show_previews is enabled, which causes the following seemingly unrelated issue:

```
** ERROR: directory is already being watched! **

Directory: [omitted]

is already being watched through: [omitted]

MORE INFO: https://github.com/guard/listen/wiki/Duplicate-directory-errors
[...]
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master
 
+* Don't add `/test/components/previews` to preview_paths if directory doesn't exist.
+
+    *Andy Holland*
+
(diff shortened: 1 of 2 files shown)


AMHOL:
Tests pass locally for me under `2.6.6`

joelhawksley (approved):

joelhawksley:
Thanks @AMHOL! ❤️

PR #501, opened 2020-10-11
Extract template compiler from base

BlakeWilliams:
### Summary

This extracts the template compiling logic from the base class to help reduce
`ViewComponent::Base`s public API. This should hopefully make it easier to
modify the template compilation logic by separating it from the base class
logic.

### Other Information

There may be a breaking change or two hidden in this PR, such as
`ViewComponent::Base` no longer having `variants` defined on it. I don't think
that many components would have relied on that method being callable, but worth
discussing if it should be supported.

### DIFF
--- a/lib/view_component/compiler.rb
+++ b/lib/view_component/compiler.rb
@@ -1,0 +1,177 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  class Compiler
+    def initialize(component_class)
+      @component_class = component_class
+    end
+
+    def compile(raise_errors: false)
+      if template_errors.present?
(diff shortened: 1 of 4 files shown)


lib/view_component/compiler.rb
```diff
@@ -0,0 +1,214 @@
...
+      define_render_template_for
+
+      CompileCache.register(component_class)
+    end
+
+    private
+
+    attr_reader :component_class
+
+    def define_render_template_for
```
BlakeWilliams, line 74: I think this might be the only introduction of new code here, but this gives us a small performance boost by letting us avoid having to call `send` to render the correct template in the component.

joelhawksley (approved):

PR #503, opened 2020-10-13
Slots V2

BlakeWilliams:
### Summary

This PR introduces the second iteration of slots, `Slotable::V2`. This takes
feedback we've received about the original iteration of slots as well as the
experiences we've had with the original iteration of slots and builds on them.

Here's an overview of the changes:

* `with_slot` becomes `renders_one`
* `with_slot` collections become `renders_many`
* Slots no longer accept a `class_name` argument, and instead accept a component or a lambda that allow you to extend the slot.
* Slots now accept positional arguments in addition to keyword arguments
* Slots no longer use `content` in templates to render their content, they instead use `to_s`. e.g. `<%= item.content %>` becomes `<%= item %>`
* Components no longer set slot values with the `#slot` method, instead they use the slot method directly. e.g. `c.slot(:row, theme: :yellow)`becomes `c.row(theme: yellow)`

### V1 example

Defining a component
```ruby
  with_slot :row, collection: true, class_name: "Row"

  class Row < ViewComponent::Slot
    def initialize(theme: :gray)
      @theme = theme
    end

    def theme_class_name
      case @theme
      when :gray
        "Box-row--gray"
      when :hover_gray
        "Box-row--hover-gray"
      when :yellow
        "Box-row--yellow"
      when :blue
        "Box-row--blue"
      when :hover_blue
        "Box-row--hover-blue"
      else
        "Box-row--gray"
      end
    end
  end
end
```

Defining a component's template
```erb
<%= if rows.any? %>
  <% rows.each do |row| %>
    <%= row.content %>
  <% end %>
<% end %>
```
Using a component
```erb
<% component.slot(:row) do %>
  Row one
<% end %>
<% component.slot(:row, theme: :yellow) do %>
  Yellow row
<% end %>
```

### V2 equivalent of the above example

Defining a component
```ruby
  renders_many :rows, -> (theme: :gray) { Row.new(theme: theme) }

    def initialize(theme:)
      @theme = theme
    end

    def theme_class_name
      case @theme
      when :gray
        "Box-row--gray"
      when :hover_gray
        "Box-row--hover-gray"
      when :yellow
        "Box-row--yellow"
      when :blue
        "Box-row--blue"
      when :hover_blue
        "Box-row--hover-blue"
      else
        "Box-row--gray"
      end
    end
  end
end
```

Defining a component's template
```erb
<%= if rows.any? %>
  <% rows.each do |row| %>
    <%= row %>
  <% end %>
<% end %>
```
Defining the `Row` component's template:

```erb
<div class="<%= theme_class_name %>">
  <%= content %>
</div>
```

Using a component
```erb
<% component.row do %>
  Row one
<% end %>
<% component.row(theme: :yellow) do %>
  Yellow row
<% end %>
```


lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,219 @@
...
+        # helper method with the same name as the slot. For collection
+        # components, the method can be called multiple times to append to the
+        # slot.
+        #
+        #   <%= render_inline(MyComponent.new) do |component| %>
+        #     <%= component.header(classes: "Foo") do %>
+        #       <p>Bar</p>
+        #     <% end %>
+        #   <% end %>
+        def renders_one(slot_name, &block)
```
BlakeWilliams, line 55: The new API accepts a block to achieve the flexibility of defining a class, but I wonder if we could achieve the same flexibility with a different approach. The existing block approach also ends up causing consumers of the `renders_*` API to define initializers for slots if they want to accept arguments. They also have to  expose those arguments somehow (likely via methods or `attr_reader`). This means slots still end up being verbose to implement.

I have a few ideas that I'm kicking around, but I wanted to get thoughts on this idea:

* `renders_*` methods accept a labmda and automatically expose any defined instance variables as method names
* If the return value of `renders_*` return a component, we render it

The first example would look something like:

```ruby
renders_many :items, -> (classes:) { @classes = classes } # callable via `items.classes` in the template
```

and component usage could look like this:

```ruby
renders_many :items, -> (item, classes:) { MyProductComponent.new(item, classes: classes) }
```

I think this gives us a few benefits:

* Simple slots are now more succinct, and can more easily accept arguments.
* More complex slots are now easier to create by using components
* Any existing component can now become a slot in another component
* Common slot behavior can be encapsulated in components

BlakeWilliams: Building on that thought, I think we could also have a shortcut syntax for the second example.

`renders_many :items, MyProductComponent`

and "content area slots" like `renders_many :items` would still work with this API.

swanson:
I really like the `<% component.row do %>` short-hand, but I prefer the old `with_slot` DSL for defining slots. I can see how the new language is more Rails-y, but my sense is that "slots" term is winning the mindshare (with the two closest alternatives to view_component to me being Vue, the headless-ui work from the Tailwind team that use "Slots").

bradgessler:
I like this API much better, including the terminology "renders" over "slots". It actually reminds me quite a bit of ActiveRecord `has_many` blocks.

I am curious, what would it look like if somebody wanted to reverse the order of what's rendered from the components `before_render` methods? There's a lot of other use cases that may want to programmatically alter the state of the view component or collections from component code (not the view).

bbugh:
Thanks for all of your hard work on this! The `slot` less usage looks great.

One of the things I am most excited about with `view_component` are libraries like `bootstrap-view-component` and `simple-form-view-component` equivalents. I was disappointed to see the DSL come back after all of the discussion in the first thread, but I don't have anything new to say and won't rehash it.

How will the DSL behave when a subclassed component needs to override the slot? 

```
class ThirdPartyGemComponent < ViewComponent::Base
  renders_one :thing do
    # ...
  end
end
```

```
class MyAppComponent < ThirdPartyGemComponent
  renders_one :thing do
    # Does this replace the other slot? Extend it? 
  end
end
```

What about templates? If `MyAppComponent` defines a template, will it override the `ThirdPartyGemComponent` one?

I'm SO THRILLED y'all are working so hard on making this an important part of Rails. I didn't like the Rails view/helper system ten years and I like it even less now. Component-based setups are really awesome way to shore up some of the issues with a viewmodel-less MVC scenario.

BlakeWilliams:
> How will the DSL behave when a subclassed component needs to override the slot?

That's a great question. Right now you would you would call `renders_one` or `renders_many` and the slot would be overwritten. There is no inheriting slots, but I think the changes in v2 will help promote composition. React has some really great documentation on it that I feel applies to view components as well: https://reactjs.org/docs/composition-vs-inheritance.html

That being said, this PR is trending away from slots being their own user defined classes, and more towards being functions that can return HTML or a component instance that will be rendered on the page. I think being able to delegate to an existing component provides an escape hatch for inheritance if it's necessary.

> ```
> class ThirdPartyGemComponent < ViewComponent::Base
>   renders_one :thing do
>     # ...
>   end
> end
> ```
> 
> ```
> class MyAppComponent < ThirdPartyGemComponent
>   renders_one :thing do
>     # Does this replace the other slot? Extend it? 
>   end
> end
> ```
> 

In this example, `renders_one :thing` would replace the parent implementation. We should definitely add test coverage though.

> What about templates? If `MyAppComponent` defines a template, will it override the `ThirdPartyGemComponent` one?

Yep! Exactly. I think that should be the existing behavior too.

> I'm SO THRILLED y'all are working so hard on making this an important part of Rails. I didn't like the Rails view/helper system ten years and I like it even less now. Component-based setups are really awesome way to shore up some of the issues with a viewmodel-less MVC scenario.

Thank you! That means a lot. ❤️

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+        #     def initialize(name:)
+        #       @name = name
+        #     end
+        #   end
+        #
+        # = Rendering slot content
+        #
+        # The component's sidecar template can access the slot by calling a
+        # helper method with the same name as the slot (pluralized if the slot
+        # is a collection).
```
bbugh, line 84: I think this is out of date? It seems like using `render_many` with a pluraly name removes the need for automatic plurality? 

```suggestion
        # helper method with the same name as the slot.
```

(This text is also in the `renders_one` doc.)

BlakeWilliams: Good catch, thanks!

lib/view_component/compiler.rb
```diff
@@ -146,18 +146,16 @@ def matching_views_in_source_location
       source_location = component_class.source_location
       return [] unless source_location
 
-      location_without_extension = source_location.chomp(File.extname(source_location))
```
BlakeWilliams, line 149: This is technically a breaking change and now ignores the component's file name in template lookups. This is so we can define components inside of components.

There is work to be done on how to find the nested component's template though.

e.g.

```
class ModalComponent < ViewComponent
  class FooterComponent < ViewComponent
  end
end
```

Where do we look for the `FooterComponent` template files? My first thought is a subdirectory, so `modal_component/footer_component.html.erb`

joelhawksley: Yes, let's go with the subdirectory 👍🏻

view_component.gemspec
```diff
@@ -41,4 +41,5 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rubocop-github", "~> 0.13.0"
   spec.add_development_dependency "simplecov", "~> 0.18.0"
   spec.add_development_dependency "simplecov-console", "~> 0.7.2"
+  spec.add_development_dependency "pry", "~> 0.13.1"
```
joelhawksley, line 44: Might we make this dependency less precise?

lib/view_component/base.rb
```diff
@@ -132,7 +129,9 @@ def view_cache_dependencies
 
     # For caching, such as #cache_if
     def format
-      @variant
+      if defined?(@variant)
```
joelhawksley, line 132: TODO: Add note about 2.6/2.7 difference.

lib/view_component/slot.rb
```diff
@@ -2,6 +2,28 @@
...
+
+    def to_s
+      if defined?(@_component_instance)
+        parent.render(@_component_instance) { content }
+      else
+        content
+      end
+    end
+
+    def method_missing(symbol, *args, &block)
```
joelhawksley, line 21: TODO: Add comments.

lib/view_component/slot.rb
```diff
@@ -2,6 +2,28 @@
...
-    attr_accessor :content
+    attr_accessor :content, :parent
+    attr_writer :_component_instance
+
+    # Parent must be `nil` for v1
+    def initialize(parent = nil)
+      @parent = parent
+    end
+
+    def to_s
```
joelhawksley, line 13: TODO: Add comments.

lib/view_component/slot.rb
```diff
@@ -2,6 +2,28 @@
...
+      else
+        content
+      end
+    end
+
+    def method_missing(symbol, *args, &block)
+      if defined?(@_component_instance)
+        @_component_instance.public_send(symbol, *args, &block)
+      else
+        parent.send(symbol, *args, &block)
```
joelhawksley, line 25: Let's leave this out, so sub-components are required to explicitly reference the parent.

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+        self.registered_slots = {}
+      end
+
+      class_methods do
+        ##
+        # Registers a slot on the component.
+        #
+        # = Example
+        #
+        #   renders_one :header do
```
joelhawksley, line 27: TODO: update

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+        #     def initialize(classes:)
+        #       @name = name
+        #     end
+        #   end
```
joelhawksley, line 31: Might this example include a slot template and usage of the `classes` argument?

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+        # helper method with the same name as the slot. For collection
+        # components, the method can be called multiple times to append to the
+        # slot.
+        #
+        #   <%= render_inline(MyComponent.new) do |component| %>
+        #     <%= component.header(classes: "Foo") do %>
+        #       <p>Bar</p>
+        #     <% end %>
+        #   <% end %>
+        def renders_one(slot_name, callable = nil)
```
joelhawksley, line 55: Might we name `callable` to be more domain-specific to rendering? How about `renderable`?

```suggestion
        def renders_one(slot_name, renderable = nil)
```

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+        #       <p>two</p>
+        #     <% end %>
+        #   <% end %>
+        def renders_many(slot_name, callable = nil)
+          validate_slot_name(slot_name)
+
+          singular_name = ActiveSupport::Inflector.singularize(slot_name)
+
+          # Define setter for singular names
+          # e.g. `with_slot :tab, collection: true` allows fetching all tabs with
```
joelhawksley, line 112: Needs updated.

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+          # `component.tabs` and setting a tab with `component.tab`
+          define_method singular_name do |*args, **kwargs, &block|
+            # TODO raise here if attempting to get a collection slot using a singular method name?
```
joelhawksley, line 115: Yes, we should raise. (And suggest the plural getter)

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+            # e.g. `component.item` with `with_slot :item, collection: true`
+            set_slot(slot_name, *args, **kwargs, &block)
+          end
+
+          # Instantiates and and adds multiple slots forwarding the first
+          # argument to each slot constructor
+          define_method slot_name do |*args, **kwargs, &block|
+            if args.empty? && kwargs.empty? && block.nil?
```
joelhawksley, line 123: Let's investigate implementing a collection setter in this PR: `component.items(@items)`, as it may affect this method.

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+            # TODO remove? This breaks overriding slots when slots are inherited
+            raise ArgumentError.new("#{slot_name} slot declared multiple times")
+          end
+        end
+      end
+
+      def get_slot(slot_name)
+        # :nocov:
+        unless self.class.registered_slots.keys.include?(slot_name)
+          raise ArgumentError.new "Unknown slot '#{slot_name}' - expected one of '#{self.class.registered_slots.keys}'"
```
joelhawksley, line 164: TODO: Add test coverage _or_ more detailed comment on this exception.

joelhawksley: _Or_, remove this code.

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+
+        slot = self.class.registered_slots[slot_name]
+        slot_instance_variable_name = slot[:instance_variable_name]
+
+        slot_instance = Slot.new(self)
+        slot_instance.content = view_context.capture(&block)
+
+        if slot[:callable].is_a?(Class) && slot[:callable] < ViewComponent::Base
+          result = slot[:callable].new(*args, **kwargs)
+          slot_instance._component_instance = result
```
joelhawksley, line 197:
```suggestion
          slot_instance._component_instance = slot[:callable].new(*args, **kwargs)
```

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+        end
+        # :nocov:
+
+        slot = self.class.registered_slots[slot_name]
+        slot_instance_variable_name = slot[:instance_variable_name]
+
+        slot_instance = Slot.new(self)
+        slot_instance.content = view_context.capture(&block)
+
+        if slot[:callable].is_a?(Class) && slot[:callable] < ViewComponent::Base
```
joelhawksley, line 195:
```suggestion
        # If is a Component
        if slot[:callable].is_a?(Class) && slot[:callable] < ViewComponent::Base
```

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+          result = slot[:callable].new(*args, **kwargs)
+          slot_instance._component_instance = result
+        elsif slot[:callable]
```
joelhawksley, line 198:
```suggestion
        # If is a lambda
        elsif slot[:callable]
```

lib/view_component/slotable_v2.rb
```diff
@@ -0,0 +1,222 @@
...
+          result = slot_instance.instance_exec(*args, **kwargs, &slot[:callable])
```
joelhawksley, line 199: Nit: could we use a more specific name than `result`?

test/app/components/slots_v2_component.html.erb
```diff
@@ -0,0 +1,33 @@
+<div class="card <%= @class_names %>">
+  <div class="title">
+    <%= title %>
+  </div>
+  <div class="subtitle">
+    <%= subtitle %>
+  </div>
+  <div class="nav">
+    <%= tabs.inspect %>
```
joelhawksley, line 9:
```suggestion
```

test/app/components/slots_v2_component.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class SlotsV2Component < ViewComponent::Base
+  include ViewComponent::Slotable::V2
+
+  renders_one :title
+  renders_one :subtitle
+  renders_one :footer, -> (class_names: "") do
+    content_tag :footer, class: "footer text-blue" do
```
joelhawksley, line 9: Can we use the `classes` argument in this content_tag?

test/app/components/slots_v2_component.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class SlotsV2Component < ViewComponent::Base
+  include ViewComponent::Slotable::V2
+
+  renders_one :title
+  renders_one :subtitle
+  renders_one :footer, -> (class_names: "") do
```
joelhawksley, line 8:
```suggestion
  renders_one :footer, -> (classes: "") do
```

test/app/components/slots_v2_component.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  renders_many :tabs
+
+  renders_many :items, -> (highlighted: false) do
+    MyHighlightComponent.new(highlighted: highlighted)
+  end
+
+  def initialize(class_names: "")
+    @class_names = class_names
+  end
```
joelhawksley, line 22:
```suggestion
  def initialize(classes: "")
    @class_names = classes
  end
```

test/view_component/slotable_v2_test.rb
```diff
@@ -0,0 +1,172 @@
...
+    assert_selector(".item.highlighted", count: 1)
+    assert_selector(".item.normal", count: 2)
+  end
+
+  # In a previous implementation of slots,
+  # the list of slots registered to a component
+  # was accidentally assigned to all components!
+  def test_slots_pollution
+    new_component_class = Class.new(ViewComponent::Base)
+    new_component_class.include(ViewComponent::Slotable)
```
joelhawksley, line 167:
```suggestion
    new_component_class.include(ViewComponent::SlotableV2)
```

lib/view_component/compiler.rb
```diff
@@ -146,21 +146,26 @@ def matching_views_in_source_location
...
       directory = File.dirname(source_location)
       filename = File.basename(source_location, ".rb")
       component_name = component_class.name.demodulize.underscore
 
+      sub_component_files = if component_class.name.include?("::")
+        subcomponent_path = component_class.name.split("::")[0...-1].map(&:underscore).join("/")
+        Dir["#{directory}/#{subcomponent_path}/#{component_name}.*{#{extensions}}"]
+      else
+        []
+      end
```
jonspalmer, line 161: Can you clarity what is needed here? Not obvious to me what sub_component_files are in this context.

lib/view_component/compiler.rb
```diff
@@ -146,21 +146,26 @@ def matching_views_in_source_location
...
-      # view files in the same directory as the component
-      sidecar_files = Dir["#{location_without_extension}.*{#{extensions}}"]
-
       # view files in a directory named like the component
       directory = File.dirname(source_location)
       filename = File.basename(source_location, ".rb")
       component_name = component_class.name.demodulize.underscore
 
+      sub_component_files = if component_class.name.include?("::")
+        subcomponent_path = component_class.name.split("::")[0...-1].map(&:underscore).join("/")
```
jonspalmer, line 157:
```suggestion
        subcomponent_path = component_class.name.deconstantize.underscore`
```

but given that we combine `subcomponent_path` with `component_name` that combination is just `component_class.name.underscore`?

lib/view_component/sub_component_wrapper.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  class SubComponentWrapper
+    attr_writer :_component_instance, :_content_block, :_content
```
jonspalmer, line 5: Why the underscore convention?

lib/view_component/sub_component_wrapper.rb
```diff
@@ -0,0 +1,60 @@
...
+    #
+    # class HeaderComponent < ViewComponent::Base
+    #   def name
+    #     @name
+    #   end
+    # end
+    #
+    def method_missing(symbol, *args, &block)
+      @_component_instance.public_send(symbol, *args, &block)
+    end
```
jonspalmer, line 58: should we be defining `respond_to_missing?` as well as `method_missing`?

jonspalmer (commented):
great work a couple of questions about details

lib/view_component/compiler.rb
BlakeWilliams: This is much better, thank you for the suggestion!

lib/view_component/sub_component_wrapper.rb
BlakeWilliams: Good call, added in https://github.com/github/view_component/pull/503/commits/123e7da6bf7a545772d368e959960f9471afccf2

lib/view_component/compiler.rb
BlakeWilliams: I'll try to add a comment to help clarify in the code, but the gist of it is that this adds support for defining a component inside of another component. 

e.g.

```ruby
# frozen_string_literal: true

class SubComponentComponent < ViewComponent::Base
  include ViewComponent::SubComponents

  renders_many :items, -> (highlighted: false) do
    MyHighlightComponent.new(highlighted: highlighted)
  end

  def initialize(classes: "")
    @classes = classes
  end

  class MyHighlightComponent < ViewComponent::Base
    def initialize(highlighted: false)
      @highlighted = highlighted
    end

    def classes
      @highlighted ? "highlighted" : "normal"
    end
  end
end

```

The current behavior is that components nested inside of other components within the same file will assume that the parent component's template is also its template, so defining a template for the nested component doesn't work and adding a `call` method raises.

lib/view_component/sub_component_wrapper.rb
BlakeWilliams: The thought here was to try and limit the possible collisions that this class and the underlying `_component_instance` might have since this class is essentially a delegator.

joelhawksley:
TODO: collection slot rendering:

`component.items(@items...)`

Gemfile.lock
```diff
@@ -1,84 +1,84 @@
 GIT
   remote: git://github.com/rails/rails.git
-  revision: 26fd55e02a19111d9b7785ca262e95dde6645f1a
```
joelhawksley, line 3: Let's leave the gemfile as-is.

Gemfile.lock
```diff
@@ -194,26 +201,28 @@ GEM
...
 
 DEPENDENCIES
   benchmark-ips (~> 2.8.2)
+  benchmark-memory (~> 0.1.2)
   better_html (~> 1)
   bundler (~> 1.14)
   capybara (~> 3)
   haml (~> 5)
   minitest (= 5.6.0)
+  pry (~> 0.13)
```
joelhawksley, line 225: Actually, let's include pry 😄

lib/view_component/base.rb
joelhawksley: Still needed.

lib/view_component/compiler.rb
```diff
@@ -146,21 +146,36 @@ def matching_views_in_source_location
...
 
-      # view files in the same directory as the component
-      sidecar_files = Dir["#{location_without_extension}.*{#{extensions}}"]
-
       # view files in a directory named like the component
       directory = File.dirname(source_location)
       filename = File.basename(source_location, ".rb")
       component_name = component_class.name.demodulize.underscore
 
+      # This adds support for nested components defined in the same file.
```
joelhawksley, line 156:
```suggestion
      # Add support for nested components defined in the same file.
```

lib/view_component/sub_component_wrapper.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  class SubComponentWrapper
+    attr_writer :_component_instance, :_content_block, :_content
+
+    # Parent must be `nil` for v1
```
joelhawksley, line 7:
```suggestion
    # Parent must be `nil` for v1
```

Why?

lib/view_component/sub_component_wrapper.rb
```diff
@@ -0,0 +1,64 @@
...
+    #
+    # If the subcomponent renderable is a component, the string class name of a
+    # component, or a function that returns a component, we render that
+    # component instance, returning the string.
+    #
+    # If the subcomponent renderable is a function and returns a string, it is
+    # set as `@_content` and is returned directly.
+    #
+    # If there is no subcomponent renderable, we evaluate the block passed to
+    # the subcomponent and return it (content area style)
```
joelhawksley, line 24:
```suggestion
    # the subcomponent and return it.
```

lib/view_component/sub_component_wrapper.rb
```diff
@@ -0,0 +1,64 @@
...
+          &@_content_block
+        )
+      elsif defined?(@_content)
+        @_content
+      elsif defined?(@_content_block)
+        @_content_block.call
+      end
+    end
+
+    # This allows access to public component methods via the wrapper
```
joelhawksley, line 39:
```suggestion
    # Allow access to public component methods via the wrapper
```

lib/view_component/sub_component_wrapper.rb
```diff
@@ -0,0 +1,64 @@
...
+    #
+    # e.g.
+    #
+    # calling `header.name` (where `header` is a subcomponent) will call `name`
+    # on the `HeaderComponent` instance
+    #
+    # Where the component may includes:
```
joelhawksley, line 46:
```suggestion
    # Where the component may include:
```

lib/view_component/sub_component_wrapper.rb
```diff
@@ -0,0 +1,64 @@
...
+        @_content_block.call
+      end
+    end
+
+    # This allows access to public component methods via the wrapper
+    #
+    # e.g.
+    #
+    # calling `header.name` (where `header` is a subcomponent) will call `name`
+    # on the `HeaderComponent` instance
```
joelhawksley, line 44:
```suggestion
    # on the `HeaderComponent` instance.
```

lib/view_component/sub_component_wrapper.rb
joelhawksley: Maybe this example could be fleshed out a bit more?

joelhawksley:
TODO: Add to `docs/index.md`

lib/view_component/sub_components.rb
```diff
@@ -0,0 +1,242 @@
+# frozen_string_literal: true
+
+require "active_support/concern"
+
+require "view_component/slot"
+
+module ViewComponent
+  module SubComponents
```
joelhawksley, line 8: If we wanted to use `Slot` for our name, that's fine 👌🏻

lib/view_component/sub_components.rb
```diff
@@ -0,0 +1,242 @@
...
+      #
+      # = Example
+      #
+      #   renders_one :header -> (classes:) do
+      #     HeaderComponent.new(classes: classes)
+      #   end
+      #
+      #   # OR
+      #
+      #   renders_on :header, HeaderComponent
```
joelhawksley, line 30:
```suggestion
      #   renders_one :header, HeaderComponent
```

lib/view_component/sub_components.rb
```diff
@@ -0,0 +1,242 @@
...
+      # 2. Since we have to pass block content to components when calling
+      # `render`, evaluating the block here would require us to call
+      # `view_context.capture` twice, which is slower
+      sub_component._content_block = block if block_given?
+
+      # If class
+      if slot[:renderable]
+        sub_component._component_instance = slot[:renderable].new(*args, **kwargs)
+      # If class name as a string
+      elsif slot[:renderable_class_name]
```
joelhawksley, line 216: Let's make sure we have public docs for this case.

lib/view_component/sub_components.rb
```diff
@@ -0,0 +1,242 @@
...
+        sub_component._component_instance = self.class.const_get(slot[:renderable_class_name]).new(*args, **kwargs)
+      # If passed a lambda
+      elsif slot[:renderable_function]
+        renderable_value = slot[:renderable_function].bind(self).call(*args, **kwargs, &block)
```
joelhawksley, line 220: Let's add a comment about `bind` here ❤️

performance/benchmark.rb
```diff
@@ -9,7 +9,9 @@
 ENV["RAILS_ENV"] = "production"
```
joelhawksley, line 9: TODO: Move this new benchmark to separate file.

joelhawksley:
TODO: Add test for subcomponent with `call` method instead of template.

test/app/components/inherited_slots_v2_component.rb
```diff
@@ -0,0 +1,4 @@
+# frozen_string_literal: true
+
+class InheritedSubComponentComponent < SubComponentComponent
```
joelhawksley, line 3: Class name does not match filename.

joelhawksley:
TODO: Note that slots V1 is deprecated.

test/app/components/sub_component_component.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+class SubComponentComponent < ViewComponent::Base
+  include ViewComponent::SubComponents
+
+  renders_one :title
+  renders_one :subtitle
+  renders_one :footer, -> (classes: "", &block) do
+    content_tag :footer, class: "footer #{classes}" do
+      block.call if block
```
joelhawksley, line 10: Perhaps we could come up with a helper/abstraction to avoid this `block.call if block` situation? This is a leak in the API to the developer.

test/app/components/sub_component_component/my_highlight_component.html.erb
```diff
@@ -0,0 +1 @@
+<%= content %>
```
joelhawksley, line 1: Can we use `highlighted` here?

test/app/components/sub_component_delegate_component.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class SubComponentDelegateComponent < ViewComponent::Base
```
joelhawksley, line 3: Add comment about this component using a component from another file as a subcomponent/slot?

test/app/components/sub_component_with_pos_arg_component.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  class Item < ViewComponent::Base
+    attr_reader :title, :classes
+
+    def initialize(title, classes:)
+      @title = title
+      @classes = classes
+    end
+
+    def call
```
joelhawksley, line 16: Looks like we're good on the `call` case ❤️

test/view_component/sub_components_test.rb
```diff
@@ -0,0 +1,185 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class SubComponentsTest < ViewComponent::TestCase
+  def test_renders_slots
+    c = SubComponentComponent.new(classes: "mt-4")
+    render_inline(c) do |component|
```
joelhawksley, line 8:
```suggestion
    render_inline(SubComponentComponent.new(classes: "mt-4")) do |component|
```

docs/index.md
```diff
@@ -167,9 +167,140 @@ Returning:
 </div>
 ```
 
-#### Slots (experimental)
 
-_Slots are currently under development as a successor to Content Areas. The Slot APIs should be considered unfinished and subject to breaking changes in non-major releases of ViewComponent._
+#### Slots
```
BlakeWilliams, line 171: This is a rough first draft of docs, feedback welcome!

docs/index.md
```diff
@@ -167,9 +167,141 @@ Returning:
 </div>
 ```
 
-#### Slots (experimental)
+#### Slots
 
-_Slots are currently under development as a successor to Content Areas. The Slot APIs should be considered unfinished and subject to breaking changes in non-major releases of ViewComponent._
+_Slots now refer to V2 of the slot implementation, which is the successor to Slots V1 and Content Areas._
```
halo, line 172: Is it correct to assume that V2 is also subject to breaking changes? If so, it would be great if that could be noted here. 

Otherwise it's not clear to me when it would be a good time to adopt this API (or the whole gem for that matter). Personally, I'm still waiting for the slots API to settle so I can avoid unnecessary refactoring.

docs/index.md
```diff
@@ -167,9 +167,141 @@ Returning:
...
+
+Slots enable multiple blocks of content to be passed to a single ViewComponent.
+
+Slots expose two helper methods for defining slots, `renders_one` and
+`renders_many`.
+
+`renders_one` allows you to define a slot that will be rendered at most once
+per-page. e.g. `renders_one :header`
```
halo, line 180: > at most once per-page

Just for clarity, I *could* render it twice if I wanted to? Or is it technically really not possible/intended? 

Or did you mean this?

```suggestion
per component. e.g. `renders_one :header`
```

docs/index.md
```diff
@@ -167,9 +167,141 @@ Returning:
...
+`renders_many` allows you to define a slot that can be rendered multiple times
+per-component. e.g. `renders_many :blog_posts`
+
+#### Defining Slots
+
+There's three logical varieties of slots:
+
+- [Delegate Slots](#delegate-slots) - Used to render other components.
+- [Lambda Slots](#lambda-slots) - Used to render strings or initialized
+- components.
```
halo, line 191:
```suggestion
- [Lambda Slots](#lambda-slots) - Used to render strings or initialized components.
```

docs/index.md
```diff
@@ -167,9 +167,141 @@ Returning:
...
+  renders_one :header
+end
+```
+
+`# blog_component.html.erb`
+
+```ruby
+<div>
+  <%= render BlogComponent.new do |c| %>
+    <%= c.title do %>
```
halo, line 295:
```suggestion
    <%= c.header do %>
```
Right?

docs/index.md
```diff
@@ -167,9 +167,141 @@ Returning:
...
+    <% end %>
+  <% end %>
+</div>
+```
+
+#### Slots V1 (deprecated)
+
+_Slots V1 are now deprecated and will be removed in 3.0. Please migrate to [Slots V2](#slots)_
 
 Slots enable multiple blocks of content to be passed to a single ViewComponent, reducing the need for sub-components (e.g. ModalHeader, ModalBody).
```
halo, line 306: Question: should this section really be in the main README in full length? Or could we move it to something like `UPGRADING_FROM_SLOTS_V1.md` to reduce confusion.

halo (commented):
Thank you for your hard work on this. This is great.

docs/index.md
```diff
@@ -167,9 +167,141 @@ Returning:
...
+
+`# blog_component.rb`
+
+```ruby
+class BlogComponnt < ViewComponent::Base
+  renders_one :header
+end
+```
+
+`# blog_component.html.erb`
```
halo, line 290:
```suggestion
`# index.html.erb`
```
Right?

docs/index.md
BlakeWilliams: I definitely meant per-component 😄 

Although re-reading it, I think saying `can be defined at most once per-component render` might be more accurate/correct?

docs/index.md
BlakeWilliams: Good question. I was thinking that it would make more sense as a separate docs file, but not sure how that would work with viewcomponent.org

docs/index.md
BlakeWilliams: > Is it correct to assume that V2 is also subject to breaking changes?

I don't believe so, unless some major issue with their design comes up.

> Otherwise it's not clear to me when it would be a good time to adopt this API (or the whole gem for that matter). Personally, I'm still waiting for the slots API to settle so I can avoid unnecessary refactoring.

I feel like view component has been pretty stable for a while now, with slots being marked as experimental.

halo: Thats encouraging and reasonable. 😄 

I would like to chime in that if V2 is not experimental any more and it is the natural successor of content areas, then this whole `Slots` section should move up above the `Content Areas` section. And the `Content Areas` section should be marked as deprecated.

Maybe I'm missing something, but that's how it looks to me :)

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+```ruby
+class BlogComponent < ViewComponent::Base
+  # `HeaderComponent` is nested inside of this component, we have to
+  # define it as a string instead of a class name.
+  renders_one :header, "HeaderComponent"
 
-_Note: Slot classes must be subclasses of `ViewComponent::Slot`._
+  # `PostComponent` is defined in another file, so we can refer to it by class
+  # name without using a string.
+  renders_many :post, PostComponent
```
joelhawksley, line 208: `post`, or `posts`?

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+  <% end %>
+<% end %>
+```
+
+##### Labmda Slots
 
-    def classes
-      "Box-header #{@classes}"
+Lambda slots render their return value on the page. This is useful for creating
+small slots via helpers like `content_tag` or using returning components with
```
joelhawksley, line 254:
```suggestion
small slots via helpers like `content_tag` or using components with
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+    Blog every day.
+  <% end %>
+<% end %>
+```
+
+##### Labmda Slots
 
-    def classes
-      "Box-header #{@classes}"
+Lambda slots render their return value on the page. This is useful for creating
```
joelhawksley, line 253:
```suggestion
Lambda slots render their return value on the page. Lambda slots are useful for creating
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+small slots via helpers like `content_tag` or using returning components with
+parent-component specific default values.
+
+```ruby
+class Blogcomponent < ViewComponent::Base
+  # Renders the returned string
+  renders_one :header, -> (title:) do
```
joelhawksley, line 260: We should probably use `title` in this example.

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+      link_to "My Blog", root_path
     end
   end
 
-  class Row < ViewComponent::Slot
-    def initialize(theme: :gray)
-      @theme = theme
-    end
+  # Returns a component that will be rendered in that slot with a default
+  # argument.
```
joelhawksley, line 267:
```suggestion
  # Returns a component that will be rendered in that slot with a default argument.
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
-      when :blue
-        "Box-row--blue"
-      when :hover_blue
-        "Box-row--hover-blue"
-      else
-        "Box-row--gray"
-      end
+##### Pass Through Slots
+
+Pass through slots simply render the content given the block.
```
joelhawksley, line 276:
```suggestion
Pass through slots capture content passed with a block.
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
-      when :yellow
-        "Box-row--yellow"
-      when :blue
-        "Box-row--blue"
-      when :hover_blue
-        "Box-row--hover-blue"
-      else
-        "Box-row--gray"
-      end
+##### Pass Through Slots
```
joelhawksley, line 274: GitHub style is to only capitalize the first letter:

```suggestion
##### Pass through slots
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+
+Pass through slots simply render the content given the block.
+
+Pass through slots can be defined by omitting the second argument to
+`renders_one` and `renders_many`.
```
joelhawksley, line 279:
```suggestion
Define a pass through slot by omitting the second argument to `renders_one` and `renders_many`:
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+
+`# blog_component.rb`
+
+```ruby
```
joelhawksley, line 283: IMO we should move to having the filename as a comment inside the main fenced block, for a cleaner look:

```suggestion
```ruby
# blog_component.rb
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+class BlogComponnt < ViewComponent::Base
+  renders_one :header
+end
+```
```
joelhawksley, line 287: Do we need an example of the component's template here?

I'm starting to wonder if we should have a standard way to present components in the docs, perhaps with the Ruby and Template files side-by-side in a two-column table?

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+    <%= c.header do %>
+      <%= link_to "My blog", root_path %>
+    <% end %>
+  <% end %>
+</div>
+```
+
+##### Rendering Collections
+
+Slots defined with `renders_many` can be used to render a collection.
```
joelhawksley, line 303:
```suggestion
Collection slots (declared with `renders_many`) can also be passed a collection.
```

joelhawksley: While we cover it in the section on delegate slots, I think we should explicitly document the single item setter (`c.link`) for collection slots somewhere.

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
 
-_Slots are currently under development as a successor to Content Areas. The Slot APIs should be considered unfinished and subject to breaking changes in non-major releases of ViewComponent._
+_Slots now refer to V2 of the slot implementation, which is the successor to Slots V1 and Content Areas._
 
-Slots enable multiple blocks of content to be passed to a single ViewComponent, reducing the need for sub-components (e.g. ModalHeader, ModalBody).
+Slots enable multiple blocks of content to be passed to a single ViewComponent.
 
-By default, slots can be rendered once per component. They provide an accessor with the name of the slot (`#header`) that returns an instance of `ViewComponent::Slot`, etc.
+Slots expose two helper methods for defining slots, `renders_one` and
+`renders_many`.
```
joelhawksley, line 177:
```suggestion
Slots expose two helper methods for defining slots, `renders_one` and `renders_many`.
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
 
-Slots declared with `collection: true` can be rendered multiple times. They provide an accessor with the pluralized name of the slot (`#rows`), which is an Array of `ViewComponent::Slot` instances.
+`renders_one` allows you to define a slot that will be rendered at most once
+per component. e.g. `renders_one :header`
```
joelhawksley, line 180:
```suggestion
`renders_one` allows you to define a slot that will be rendered at most once per component. e.g. `renders_one :header`
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
 
-To learn more about the design of the Slots API, see [#348](https://github.com/github/view_component/pull/348) and [#325](https://github.com/github/view_component/discussions/325).
+`renders_many` allows you to define a slot that can be rendered multiple times
+per-component. e.g. `renders_many :blog_posts`
```
joelhawksley, line 183:
```suggestion
`renders_many` allows you to define a slot that can be rendered multiple times per-component. e.g. `renders_many :blog_posts`
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
 
-##### Defining Slots
+#### Defining Slots
```
joelhawksley, line 185:
```suggestion
#### Defining slots
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
 
-Slots are defined by `with_slot`:
+There's three logical varieties of slots:
```
joelhawksley, line 187:
```suggestion
Slots come in three forms:
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
 
-`with_slot :header`
+- [Delegate Slots](#delegate-slots) - Used to render other components.
+- [Lambda Slots](#lambda-slots) - Used to render strings or initialized components.
+- [Pass through Slots](#pass-through-slots) - Used to pass content directly to a
+    component.
 
-To define a collection slot, add `collection: true`:
+##### Delegate Slots
```
joelhawksley, line 194:
```suggestion
##### Delegate slots
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+#### Defining Slots
 
-Slots are defined by `with_slot`:
+There's three logical varieties of slots:
 
-`with_slot :header`
+- [Delegate Slots](#delegate-slots) - Used to render other components.
+- [Lambda Slots](#lambda-slots) - Used to render strings or initialized components.
+- [Pass through Slots](#pass-through-slots) - Used to pass content directly to a
+    component.
```
joelhawksley, line 192:
```suggestion
- [Pass through slots](#pass-through-slots)  pass content directly to another component.
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
 
-##### Defining Slots
+#### Defining Slots
 
-Slots are defined by `with_slot`:
+There's three logical varieties of slots:
 
-`with_slot :header`
+- [Delegate Slots](#delegate-slots) - Used to render other components.
+- [Lambda Slots](#lambda-slots) - Used to render strings or initialized components.
```
joelhawksley, line 190:
```suggestion
- [Lambda slots](#lambda-slots) render strings or initialized components.
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+per-component. e.g. `renders_many :blog_posts`
 
-##### Defining Slots
+#### Defining Slots
 
-Slots are defined by `with_slot`:
+There's three logical varieties of slots:
 
-`with_slot :header`
+- [Delegate Slots](#delegate-slots) - Used to render other components.
```
joelhawksley, line 189:
```suggestion
- [Delegate slots](#delegate-slots) render other components.
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+Delegate slots are simple slot definitions that delegate to another component.
 
-To define a slot with a custom Ruby class, pass `class_name`:
+`# blog_component.rb`
 
-`with_slot :body, class_name: 'BodySlot'`
+```ruby
+class BlogComponent < ViewComponent::Base
+  # `HeaderComponent` is nested inside of this component, we have to
+  # define it as a string instead of a class name.
```
joelhawksley, line 203:
```suggestion
  # Since `HeaderComponent` is nested inside of this component, we have to
  # reference it as a string instead of a class name.
```

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+  renders_one :header, "HeaderComponent"
 
-_Note: Slot classes must be subclasses of `ViewComponent::Slot`._
+  # `PostComponent` is defined in another file, so we can refer to it by class
+  # name without using a string.
```
joelhawksley, line 207:
```suggestion
  # `PostComponent` is defined in another file, so we can refer to it by class name.
```

docs/slots_v1.md
```diff
@@ -0,0 +1,127 @@
+#### Slots V1 (deprecated)
+
+_Slots V1 are now deprecated and will be removed in 3.0. Please migrate to [Slots V2](#slots)_
```
joelhawksley, line 3:
```suggestion
_Slots V1 is now deprecated and will be removed in 3.0. Please migrate to [Slots V2](#slots)_
```

Will this Slots V2 link work correctly?

docs/index.md
```diff
@@ -167,131 +167,177 @@ Returning:
...
+Lambda slots render their return value on the page. This is useful for creating
+small slots via helpers like `content_tag` or using returning components with
+parent-component specific default values.
+
+```ruby
+class Blogcomponent < ViewComponent::Base
+  # Renders the returned string
+  renders_one :header, -> (title:) do
+    content_tag :h1 do
+      link_to "My Blog", root_path
```
BlakeWilliams, line 257:
```suggestion
      link_to title, root_path
```

docs/index.md
```diff
@@ -167,131 +167,172 @@ Returning:
...
-`with_slot :body, class_name: 'BodySlot'`
+```ruby
+class BlogComponent < ViewComponent::Base
+  # Since `HeaderComponent` is nested inside of this component, we have to
+  # reference it as a string instead of a class name.
+  renders_one :header, "HeaderComponent"
 
-_Note: Slot classes must be subclasses of `ViewComponent::Slot`._
+  # `PostComponent` is defined in another file, so we can refer to it by class name.
+  renders_many :post, PostComponent
```
BlakeWilliams, line 203:
```suggestion
  renders_many :posts, PostComponent
```

docs/index.md
BlakeWilliams: Should be `posts`, good catch

docs/slots_v1.md
```diff
@@ -0,0 +1,127 @@
+#### Slots V1 (deprecated)
+
+_Slots V1 is now deprecated and will be removed in 3.0. Please migrate to [Slots V2](#slots)_
```
BlakeWilliams, line 3: TODO: Link to `docs/index#slots`

docs/index.md
```diff
@@ -167,131 +167,169 @@ Returning:
 </div>
 ```
 
-#### Slots (experimental)
+#### Slots
 
-_Slots are currently under development as a successor to Content Areas. The Slot APIs should be considered unfinished and subject to breaking changes in non-major releases of ViewComponent._
+_Slots now refer to V2 of the slot implementation, which is the successor to Slots V1 and Content Areas._
```
BlakeWilliams, line 172: TODO link to v1 docs

docs/index.md
```diff
@@ -167,131 +167,169 @@ Returning:
 </div>
 ```
 
-#### Slots (experimental)
+#### Slots
```
BlakeWilliams, line 170:
```suggestion
#### Slots (experimental)
```

docs/index.md
```diff
@@ -167,131 +167,169 @@ Returning:
...
 
-_Slots are currently under development as a successor to Content Areas. The Slot APIs should be considered unfinished and subject to breaking changes in non-major releases of ViewComponent._
+_Slots now refer to V2 of the slot implementation, which is the successor to Slots V1 and Content Areas._
 
-Slots enable multiple blocks of content to be passed to a single ViewComponent, reducing the need for sub-components (e.g. ModalHeader, ModalBody).
+Slots enable multiple blocks of content to be passed to a single ViewComponent.
```
BlakeWilliams, line 174:
```suggestion
Slots enable multiple blocks of content to be passed to a single ViewComponent, improving the ergonomics of complex components.
```

docs/index.md
```diff
@@ -169,129 +169,167 @@ Returning:
 
 #### Slots (experimental)
 
-_Slots are currently under development as a successor to Content Areas. The Slot APIs should be considered unfinished and subject to breaking changes in non-major releases of ViewComponent._
+_Slots now refer to V2 of the slot implementation, which is the successor to Slots V1 and Content Areas._
```
BlakeWilliams, line 172:
```suggestion
_Slots are currently under development as the successor to Content Areas. The Slot APIs should be considered unfinished (it's already in its second iteration, [see the original API](/slots-v1).) and subject to breaking changes in non-major releases of ViewComponent._
```

docs/index.md
BlakeWilliams: This was a good call out, I just marked v2 as experimental as well.

joelhawksley (approved):
🙇🏻 incredible work here, @BlakeWilliams !

docs/index.md
```diff
@@ -169,129 +169,167 @@ Returning:
...
-      end
+##### Pass through slots
+
+Pass through slots capture content passed with a block.
+
+Define a pass through slot by omitting the second argument to `renders_one` and `renders_many`:
+
+```ruby
+# blog_component.rb
+class BlogComponnt < ViewComponent::Base
```
BlakeWilliams, line 276:
```suggestion
class BlogComponent < ViewComponent::Base
```

BlakeWilliams:
Just pushed up a PR that should resolve the last bit of docs feedback.

docs/index.md
```diff
@@ -169,129 +169,182 @@ Returning:
...
 #### Slots (experimental)
 
-_Slots are currently under development as a successor to Content Areas. The Slot APIs should be considered unfinished and subject to breaking changes in non-major releases of ViewComponent._
+_Slots are currently under development as the successor to Content Areas. The Slot APIs should be considered unfinished (it's already in its second iteration, [see the original API](/slots-v1).) and subject to breaking changes in non-major releases of ViewComponent._
 
-Slots enable multiple blocks of content to be passed to a single ViewComponent, reducing the need for sub-components (e.g. ModalHeader, ModalBody).
+Slots enable multiple blocks of content to be passed to a single ViewComponent, improving the ergonomics of complex components.
 
-By default, slots can be rendered once per component. They provide an accessor with the name of the slot (`#header`) that returns an instance of `ViewComponent::Slot`, etc.
+Slots expose two helper methods for defining slots, `renders_one` and `renders_many`.
```
joelhawksley, line 176:
```suggestion
Slots are defined with with `renders_one` and `renders_many`:
```

docs/index.md
```diff
@@ -169,129 +169,182 @@ Returning:
...
 
-Slots declared with `collection: true` can be rendered multiple times. They provide an accessor with the pluralized name of the slot (`#rows`), which is an Array of `ViewComponent::Slot` instances.
+`renders_one` allows you to define a slot that will be rendered at most once per component. e.g. `renders_one :header`
```
joelhawksley, line 178:
```suggestion
`renders_one` defines a slot that will be rendered at most once per component: `renders_one :header`
```

docs/index.md
```diff
@@ -169,129 +169,182 @@ Returning:
...
 
-To learn more about the design of the Slots API, see [#348](https://github.com/github/view_component/pull/348) and [#325](https://github.com/github/view_component/discussions/325).
+`renders_many` allows you to define a slot that can be rendered multiple times per-component. e.g. `renders_many :blog_posts`
```
joelhawksley, line 180:
```suggestion
`renders_many` defines a slot that can be rendered multiple times per-component: `renders_many :blog_posts`
```

docs/index.md
```diff
@@ -169,129 +169,182 @@ Returning:
...
-`with_slot :header`
+- [Delegate slots](#delegate-slots) render other components.
+- [Lambda slots](#lambda-slots) render strings or initialized components.
+- [Pass through slots](#pass-through-slots)  pass content directly to another component.
 
-To define a collection slot, add `collection: true`:
+##### Delegate slots
 
-`with_slot :row, collection: true`
+Delegate slots are simple slot definitions that delegate to another component.
```
joelhawksley, line 192:
```suggestion
Delegate slots delegate to another component:
```

docs/index.md
```diff
@@ -169,129 +169,182 @@ Returning:
...
+    Blog every day.
+  <% end %>
+<% end %>
+```
+
+##### Labmda Slots
+
+Lambda slots render their return value on the page. Lambda slots are useful for creating
+small slots via helpers like `content_tag` or using components with
+parent-component specific default values.
```
joelhawksley, line 250:
```suggestion
Lambda slots render their return value. Lambda slots are useful for working with helpers like `content_tag` or as wrappers for another component with specific default values.
```

joelhawksley (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
 
 ## master
 
+* Add `ActionView::SlotableV2`
+  * `with_slot` becomes `renders_one`.
+  * `with_slot collection: true` becomes `renders_many`.
+  * Slot definitions now accept either a component class, component class
+    name, or a lambda instead of a `class_name:` keyword argument.
```
joelhawksley, line 9:
```suggestion
  * Slot definitions now accept either a component class, component class name, or a lambda instead of a `class_name:` keyword argument.
```

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
...
+  * Slots now support positional arguments.
+  * Slots no longer use the `content` attribute to render content, instead
+    relying on `to_s`. e.g. `<%= my_slot %>`
```
joelhawksley, line 12:
```suggestion
  * Slots no longer use the `content` attribute to render content, instead relying on `to_s`. e.g. `<%= my_slot %>`.
```

CHANGELOG.md
```diff
@@ -2,6 +2,19 @@
...
+  * Slot values are no longer set via the `slot` method, and instead use
+    the name of the slot.
```
joelhawksley, line 14:
```suggestion
  * Slot values are no longer set via the `slot` method, and instead use the name of the slot.
```

joelhawksley (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 
+* Add `ActionView::SlotableV2`
+  * `with_slot` becomes `renders_one`.
+  * `with_slot collection: true` becomes `renders_many`.
+  * Slot definitions now accept either a component class, component class name, or a lambda instead of a `class_name:` keyword argument.
+  * Slots now support positional arguments.
+  * Slots no longer use the `content` attribute to render content, instead relying on `to_s`. e.g. `<%= my_slot %>`.
+  * Slot values are no longer set via the `slot` method, and instead use the name of the slot.
+
+  * Blake Williams
```
joelhawksley, line 13:
```suggestion
  *Blake Williams*
```

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 
+* Add `ActionView::SlotableV2`
+  * `with_slot` becomes `renders_one`.
+  * `with_slot collection: true` becomes `renders_many`.
+  * Slot definitions now accept either a component class, component class name, or a lambda instead of a `class_name:` keyword argument.
+  * Slots now support positional arguments.
+  * Slots no longer use the `content` attribute to render content, instead relying on `to_s`. e.g. `<%= my_slot %>`.
+  * Slot values are no longer set via the `slot` method, and instead use the name of the slot.
+
+  *Blake Williams*
```
BlakeWilliams, line 13:
```suggestion
    *Blake Williams*
```

cheshire137:
Is this available in the view_component gem yet? The latest version I'm seeing is [2.22.1](https://rubygems.org/gems/view_component/versions/2.22.1) from Nov 10. 🙇‍♀️

cheshire137:
Ah, [2.23.0](https://rubygems.org/gems/view_component/versions/2.23.0) has it, thanks @BlakeWilliams!

PR #506, opened 2020-10-15
Update to work with latest Rails master

joelhawksley:
This PR updates our Rails `master` dependency, removing an accommodation we had for template annotations that is no longer necessary 🎊


BlakeWilliams (approved):
:shipit: !

PR #515, opened 2020-10-28
Allow compiler buffer instance to be customized

jaredcwhite:
I've been working on implementing support for ViewComponent outside of a full-blown Rails app, as well adding a template language similar to ERB but slightly different (a larger story for another time), and I ran into just a couple of slight issues related to each. This PR contains the following fixes:

* I wanted to be able to control the output buffer object and not have it hardcoded to `ActionView::OutputBuffer`. Now there's a component class method which will return a buffer instance.

* Because my template language ends in `.serb`, the matching views array ended up with duplicate file paths because `erb` matches `serb` and `serb` itself also matches `serb`. By calling `uniq` on the array when its returned, we ensure there's no possibility of duplicate file paths.

I don't think there needs to be a new test for the first change as if any existing tests pass at all, we know it works. :) I'd be happy to write a test for the second change but I'm not sure the best way to go about that.

Overall though, I'm quite impressed that ViewComponent is working well thus far only pulling in ActionView and not all of Rails. There might be a certain feature here or there I haven't tried out yet that is "broken", but if I do run into anything that seems like it could be mitigated fairly easily, I'll file another PR.

### DIFF
--- a/lib/view_component/compiler.rb
+++ b/lib/view_component/compiler.rb
@@ -56,7 +56,7 @@ module ViewComponent
 
         component_class.class_eval <<-RUBY, template[:path], -1
           def #{method_name}
-            @output_buffer = ActionView::OutputBuffer.new
+            @output_buffer = self.class.template_buffer_instance
             #{compiled_template(template[:path])}
           end
         RUBY
@@ -160,7 +160,7 @@ module ViewComponent
 
       sidecar_directory_files = Dir["#{directory}/#{component_name}/#{filename}.*{#{extensions}}"]
 
-      (sidecar_files - [source_location] + sidecar_directory_files)
+      (sidecar_files - [source_location] + sidecar_directory_files).uniq
     end
 
     def inline_calls
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -224,6 +224,11 @@ module ViewComponent
         @_template_compiler ||= Compiler.new(self)
       end
 
+      # Returns a new instance of the string buffer used by the template compiler
+      def template_buffer_instance
+        ActionView::OutputBuffer.new
+      end
+
       # we'll eventually want to update this to support other types
       def type
         "text/html"


jaredcwhite:
Haha, I just realized #492 addresses one of the same problems as this, so let me know if I should remove the `uniq` here and let the other PR handle that.

lib/view_component/compiler.rb
```diff
@@ -56,7 +56,7 @@ def compile(raise_errors: false)
 
         component_class.class_eval <<-RUBY, template[:path], -1
           def #{method_name}
-            @output_buffer = ActionView::OutputBuffer.new
+            @output_buffer = self.class.template_buffer_instance
```
BlakeWilliams, line 59: Thoughts on removing this line, and initializing `@output_buffer` in the `ViewComponent::Base#render_in?` I think that may help your use-case by de-coupling the compiled method from `ActionView::OutputBuffer.new`

jaredcwhite: @BlakeWilliams I think that could work, but do we want the base class to be concerned with setting up template/rendering minutia, vs the template compiler?

jaredcwhite:
I'd like to see this move forward. Is there anything I can do to help in the review process?

joelhawksley:
@jaredcwhite sorry for the delay! @BlakeWilliams and I will have a look at this PR during our Monday pairing session ❤️

lib/view_component/compiler.rb
BlakeWilliams: Sorry for the delay on this. My current thinking is that `render_in` is a Rails specific API. Given that, it could make sense that `render_in` should be responsible for setting up the correct `@output_buffer` instance, if it's supposed to be configurable.

BlakeWilliams: I was curious how HAML handles this, and it looks like it brings its own output buffer instead of using `@output_buffer`. 

Here's the generated template code expanded a bit:

```ruby
  begin
    extend Haml::Helpers
    _hamlout = @haml_buffer = Haml::Buffer.new(haml_buffer, {})
    _erbout = _hamlout.buffer
    @output_buffer = output_buffer ||= ActionView::OutputBuffer.new if defined?(ActionView::OutputBuffer)
    _hamlout.buffer << ("<div>\n".freeze)

    _hamlout.buffer << ((
      ::Haml::Helpers.html_escape((( content
                                   )).to_s)
    ).to_s)

    _hamlout.buffer << ("\n".freeze)

    _hamlout.buffer << ((::Haml::Helpers.html_escape((( @message
                                                      )).to_s)
                        ).to_s)

    _hamlout.buffer << ("\n</div>\n".freeze)
    ::Haml::Util.html_safe(_erbout)
  ensure
    @haml_buffer = @haml_buffer.upper if @haml_buffer
  end
```

I did some quick verifying, and this works because we rely on the return value of the template, and not `@output_buffer`, which is empty at render time for HAML templates. Maybe that eliminates the need for using `@output_buffer` at all, at least for view components?

jaredcwhite: @BlakeWilliams Sorry, I missed seeing a notification of this thread. I'd be cool with relocating the buffer instantiation to `render_in`. Is that something you'd like me to address in this PR? (BTW I didn't quite follow the point about how HAML's working.)

jaredcwhite:
@joelhawksley @BlakeWilliams Replied inline above.

lib/view_component/compiler.rb
jaredcwhite: Yeah let me revisit this…it's possible I may have already refactored out the need to change it.￼

joelhawksley:
@jaredcwhite do you want to continue this PR? If not, mind closing it? 🙉

joelhawksley:
Closing as stale. Please reopen if you'd like to continue your work here @jaredcwhite ❤️

Issue #517, opened 2020-11-05
React like Error Boundaries


francescobbo:
## Feature request

It would be really awesome to be able to capture errors in production and render fallback templates, instead of having a whole page break.

In React, one would create a component that wraps another, define a `componentDidCatch` method, and define what to render when something happens.

I thought that the React approach would be easily ported:

```ruby
class ErrorBoundaryComponent < ViewComponent::Base
  def call
    content

  rescue => exc
    # Report to some service
    Raven.capture_exception(exc)
    # possibly render a partial here instead of inlining
    "Something went wrong."
  end
end
```
and then using it like this:

```erb
<%= render(ErrorBoundaryComponent.new) do %>
  <%= render(InnerComponent.new(abc: 'test')) %>
<% end %>
```

The above however does not work, and I believe the reason is that inner components are rendered before parents, so `ErrorBoundaryComponent` `rescue` block is useless.

That however is not what I would like to have in a Rails app. Here's what I imagined:

```ruby
class FallibleComponent < ViewComponent::Base
  rescue_from SomeMoreSpecificError, with: :handle_specific_error
  rescue_from StandardError, with: :handle_errors

  def handle_errors
    # Report to some service
    Raven.capture_exception(exc)
    # possibly render a partial here instead of inlining
    "Something went wrong."
  end

  [...]
end
```

Then refactor that into a Mixin, and use everywhere.

If you believe this can be useful, I might give a shot at preparing a PR myself :)

### Motivation

Unpredictable production errors happen from time to time. If a component represents a negligible part of a page (say a widget in a sidebar, a footer), it can be really annoying for it to blow up (500) what might be a critical part of an application.

BlakeWilliams:
👋 This sounds like an interesting idea, and likely something that we'd want to support in view components. Would you mind opening up a PR for this behavior so we can continue the discussion?

joelhawksley:
Closing as stale.

ozzyaaron:
I know its closed but for anybody else looking for this type of thing the way we're doing it is just to override the `render_in` method on base component as we're finding we want different error boundary responses depending on the type of component 

e.g. 

```
class FormFieldWithErrorBoundary < ApplicationComponent
  ...
  render_in(*)
    super
  rescue => e
     ... handle error, send to logging, whatever ...

     # We use a component, but whatever 
     render(ErrorBannerComponent.new(message: ..)
  end
end
```

[...]

PR #522, opened 2020-11-10
Error handling

francescobbo:
### Summary

Implements #517. 

Uses `ActiveSupport::Rescuable` to let users catch and gracefully handle errors.

A `Rescuable` method had to be re-implemented to capture the return value of the error handler.


francescobbo:
@BlakeWilliams, I believe this is complete, tested and documented.

I've merged latest master (there was a conflict), and tests seem now to be having a random failure. Could you please give it a kick to retry the workflow?

docs/index.md
```diff
@@ -619,6 +619,24 @@ class UserComponent < ViewComponent::Base
...
 
+### Handling errors
+
+An exception occurring in a small fragment of the user interface can potentially break an entire page.
+
+To prevent this you can use `rescue_from`, just like in a controller:
+
+```ruby
+class FailingComponent < ViewComponent::Base
+  rescue_from User::NotAuthorized, with: :deny_access
```
BlakeWilliams, line 630: I wonder if this should be `with_component` instead of `with`. That way users can define, and re-use a common error component. Thoughts?

francescobbo: Not a bad idea, although I'm piggybacking `ActiveSupport::Rescuable` here. Implementing that keyword would mean a second override, and possibly a dynamic method definition:

`rescue_from Something, with_component: FallbackComponent` would create a method like this:

```ruby
def __rescue_something(exc)
  render FallbackComponent.new(exception: exc, source: self)
end
```

And then translate `with_component: FallbackComponent` to `with: :__rescue_something`

Am I overkilling or not seeing a better approach here?

Honestly, I can see both keywords being useful and maybe both should be defined. How about we start integrating this PR, and I'll try to sketch something on top of this for `with_component`?

BlakeWilliams: @joelhawksley and I have given this a bit more thought, and we think the solution to this problem should be consistent with the slot API we end up finalizing in #503 

We really like the direction this is going, but we think it's worth waiting a bit before moving forward so we end up with a cohesive API.

francescobbo: That makes sense. Let's wait for #503 and catch up later for a better API :)

alekhood317 (approved):

kylefox:
This API is really awesome, and feels a lot more Railsy than (somehow) using slots. Is there any further guidance/patterns on how to build components that _never_ completely break a page and, at worst, render an empty string?

BlakeWilliams:
Just now getting a chance to revisit this.

I want to compare React to view components, and I think that React needs this kind of exception handling more than view components. In React mutating state, props changing, side-effects that can raise like fetching JSON all contribute to the need for exception handling. View components avoids (or can avoid) a lot of that extra complexity which also means they can avoid some of the common causes of exceptions that React has. Given those difference, I think we should be careful about the React error handling patterns we implement in the library.

I'd be interested in seeing how view component error handling has been implemented in other applications, and cases where it has caught issues in production.

kylefox:
> I'd be interested in seeing how view component error handling has been implemented in other applications, and cases where it has caught issues in production.

We have some view components that should never, _ever_ cause Rails to throw an error (500) while rendering the component. 

For example, imagine a view component for a "Your cart" button on an e-commerce website. The component has a bug that, in some cases, raises a `NoMethodError`. **It's critical that this bug does not cause every page on the website to break.** We would rather render the page without the Cart button than render an error page.

One solution we've tried is to wrap rendering the component inside a helper which handles exceptions:

```ruby
# cart_helper.rb
def render_cart_button
  render CartButtonComponent.new(current_cart)
rescue StandardError => e
  "<!-- Error rendering cart button: #{e} -->".html_safe
end
```

```erb
<!-- Using the helper -->
<%= render_cart_button %>

<!-- Alternatively, rescue the view component inline (gross) -->
<%= render CartButtonComponent.new(current_cart) rescue '' %>
```

I think the API @francescobbo initially proposed is perfect because it exactly mimics how exception handling in controllers works. Personally I'd love to be able to do something like this, where the component output any truthy value returned by the exception handler:

```ruby
class CartButtonComponent < ApplicationComponent
  rescue_from StandardError, with: :log_component_error
  
  # ...snip...

  def log_component_error(error)
    message = "Error rendering cart button: #{error}"
    Rails.logger.error(message)
    "<!-- #{message} -->".html_safe
  end
end
```

francescobbo:
Hi @BlakeWilliams!

I've merged latest master and resolved some conflicts that arose in the last few months.

I get your point about React apps being more "fragile", having to handle both business and presentation logic, and that possibly includes randomly failing scenarios (network requests as you pointed out).

However, React, and this PR, do not propose Error Boundaries as a way to gracefully handle common errors. Try/catch (and begin/rescue) are still preferred.

What we are instead trying to do is preventing a small mistake, be it a rushed commit, an edge case, or whatnot, from bubbling up and bringing down an entire application.

Spone:
Thank you for your work on this @francescobbo, this will be super useful in several use-cases for the apps I'm working on!

docs/index.md
Spone: SlotsV2 are merged, maybe it's time to revisit this?

Most of the time, I would rather use a `with` keyword that points to a local method, than rendering another component using `with_component`.
I see the use case for a global error component though, so maybe the `with_component` keyword could be added in a separate PR as @francescobbo suggests?

BlakeWilliams: > Most of the time, I would rather use a `with` keyword that points to a local method, than rendering another component using `with_component`.

I ran into a case where I needed to rescue an error in a component (which should be rare, in my opinion). It was useful to have a method so I could add instrumentation and error logging. Given that, I think my proposal would be:

* Use the `rescue_from` API as shown above, where `with` can reference a method or a component (similar to how slots v2 works). 
* The handler method then behaves like a slot, as-in it will render the component or string that the method returns.

Thoughts?

Spone: Yes, exactly, I would probably add error logging in the handling method as well.

:+1: to your proposal.

ramontayag:
> What we are instead trying to do is preventing a small mistake, be it a rushed commit, an edge case, or whatnot, from bubbling up and bringing down an entire application.

I agree. I found myself in a situation where I want to rescue an exception, still ping by exception monitoring tool, but then render an "oops" in place of the component. No need to bring down the whole page.

If it's not in view component then writing a wrapper when to render components sounds like a solution.

francescobbo:
Hi @BlakeWilliams, long time no speak!

I just rebased this branch out of the latest changes on the main branch, adjusted with SlotsV2, changed all the tests and removed the deprecated stuff.

All the tests seem to be passing. Let me know if you still want this to go in!

Have a good weekend :)

docs/CHANGELOG.md
```diff
@@ -13,6 +13,10 @@ title: Changelog
 
     *Josh Clayton*
 
+* Use ActiveSupport::Rescuable to handle exceptions while rendering.
```
Spone, line 16:
```suggestion
* Use `ActiveSupport::Rescuable` to handle exceptions while rendering.
```

docs/guide/error-handling.md
```diff
@@ -0,0 +1,23 @@
...
+class FailingComponent < ViewComponent::Base
+  rescue_from User::NotAuthorized, with: :deny_access
+
+  def deny_access
+    'You cannot look at this.'
+  end
+end
+```
+
+The error handler can also render a fallback component.
```
Spone, line 23: Can you please add a code example for this?

BlakeWilliams:
@francescobbo 👋 Thanks for the ping!

In general, I like the idea of components and error handling, but I think this is something we'll need to see patterns and reports about how folks are implementing error handling in their components before we commit to something stable.

Internally we've been experimenting with a similar, but slightly different API (with a slight bit more focus on component composition) but have run into a few cases where the pattern falls short (happy to share that API/source code if you'd like).

Specifically, code like:

```erb
<% if some_condition_with_a_raisable_side_effect %>
  <%= render MyComponent.new %>
<% end %>
```

There's no good way of handling that outer `some_condition_with_a_raisable_side_effect` exception when error handling is encapsulated, which makes me question if it's the correct approach, bringing me back to the idea of composing components to rescue errors, like:

```erb
<% render RescueComponent.new(rescues: [StandardError]) do |c| %>
  <% c.try %>
    <% if some_condition_with_a_raisable_side_effect %>
      <%= render MyComponent.new %>
    <% end %>
  <% end %>
<% end %>
```

I don't think that even needs to be a component, but I have some WIP code that implements a pattern similar to ☝️ as a helper that's usable within components and partials.

Maybe a good next step could be starting a discussion, polling folks for how they implement error handling with components  (mentioning wrapper components, internal `rescue_from` like handling, or other approaches) so we can get a better understanding of how this type of behavior is used beyond our own perspectives and potentially determine a path forward?

joelhawksley:

PR #523, opened 2020-11-10
Revert "Always use current view context for helpers"

joelhawksley:
Reverts github/view_component#519

In testing v2.22.0 with `primer_view_components`, this change caused the entire test suite to fail due to captured content being rendered out of place. 

Let's revert it so we can ship a stable build while we investigate further.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -102,7 +102,11 @@ module ViewComponent
     # If trying to render a partial or template inside a component,
     # pass the render call to the parent view_context.
     def render(options = {}, args = {}, &block)
-      view_context.render(options, args, &block)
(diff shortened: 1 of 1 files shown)


BlakeWilliams (approved):

filippoliverani:
> Reverts #519
> 
> In testing v2.22.0 with `primer_view_components`, this change caused the entire test suite to fail due to captured content being rendered out of place.
> 
> Let's revert it so we can ship a stable build while we investigate further.

Thank you, that change broke a lot of components in my company's projects too. `ActionView::Helpers::TextHelper#concat` started to behave differently in inline rendering and some content blocks started to be rendered as escaped html  🙏

PR #534, opened 2020-11-25
Set content for components or slots without a block

jnraine:
### Summary

Prior to this pull request, providing content to a component could only be done by passing a block to `render`:

```erb
<%= render(TestComponent.new) do %>
  Hello, World!
<% end %>
```

This works great when you want to provide a snippet of HTML in an ERB file but can feel heavy weight when you only need to pass a piece of text, like above. It also makes it impossible to configure a component outside of a view, unless you use `instance_variable_set` or the like.

Now, you can set content using `#with_content`:

```erb
<%= render(TestComponent.new.with_content(“Hello, World!”)) %>
```

In additional to a string, you can also pass a component to `#with_content`, which will be rendered when `#content` is accessed:

```ruby
box = BoxComponent.new
box.with_content(FancyComponent.new(title: "My title"))
# ...later in the view:
render(box) # FancyComponent is is rendered along with BoxComponent
```

This makes it possible to build up a view from components.

### Slots

`with_content` has been added to slots too, making it possible to configure a component anywhere you can instantiate a class:

```ruby
box = BoxComponent.new
box.slot(:header).with_content("My title")

comments.each do |comment|
  box.slot(:row).with_content(comment.body)
end

box.slot(:footer).with_content("My footer")
```

However, slots don't support passing components to `with_content`:

```ruby
# NOT ALLOWED:
box = BoxComponent.new
box.slot(:header).with_content(FancyHeaderComponent.new)
```

This is because a slot doesn't know how to render things; it is just a plain old Ruby object. It would be useful to add this in the future.

### What makes this useful?

While working on a form builder, I found it difficult compose UIs without calling out to a template or creating a view component tailored to my single use case.

`#with_content` makes it possible to create beautifully rendered views prior to having a view context, making it possible to use ViewComponents in new and interesting ways.

As an added bonus, `#with_content` offers a more concise way to provide text while in ERB or other template languages.

### What's missing

This is something that could be added in a follow-up pull request.

### Other Information

This also removes `.ruby-version` to make it easier for developers to contribute. As per the gemspec, this library supports Ruby >2.3 and runs tests against [many Ruby versions](https://github.com/github/view_component/blob/master/.github/workflows/ci.yml#L38) so a strict Ruby version requirement while developing may not be necessary.

(I hit this while working on this pull request, where I had 2.7.0 instead of 2.7.1. Not having to install 2.7.1 made my setup a breeze.)


lib/view_component/base.rb
```diff
@@ -72,7 +72,13 @@ def render_in(view_context, &block)
       @current_template = self
 
       # Assign captured content passed to component as a block to @content
-      @content = view_context.capture(self, &block) if block_given?
+      if block_given?
+        if @content.nil?
```
BlakeWilliams, line 76: I think this may break existing cases where we render the same component instance twice.

e.g.

```ruby
my_component = MyComponent.new
render my_component { "foo" } # sets @content
render my_component { "bar" } # @content was set via the previous render call, but should now be overridden
```

What are your thoughts on setting an ivar like `@_content_set_by_with_content` or a similar approach?

BlakeWilliams (commented):
I had one thought about rendering the same component instance multiple times that might break, but otherwise LGTM

lib/view_component/base.rb
jnraine: Good point. I've adjusted to use a similar method name and added a test to  ensure that multiple renders of the same instance will use the latest block provided.

lib/view_component/base.rb
```diff
@@ -161,16 +167,47 @@ def with_variant(variant)
...
+
+      @content = render_or_return(@content_set_by_with_content.call(self))
+    end
+
     private
 
+    # Renders the given object and returns the result, if object can be rendered.
+    # Otherwise, returns object.
+    def render_or_return(object)
+      if object.respond_to?(:render_in)
```
BlakeWilliams, line 196: I'm not 100% sure, we may be able to avoid this check?

Each component implements `render` that *should* have the correct data available by the time the `with_content` block is eval'd. I wonder if the following would work and also solve the same use-cases?

```ruby
MyComponent.new.with_content do |c|
  c.render MyButtonComponent.new(text: "Submit")
end
```

jnraine: Yes, you're right: the block form allows you to call render yourself and was my initial approach.

After some use, I found myself trying to use components as content, just like strings. When the framework knows how to handle it, similar to how `ActionView` knows to escape a string or not, it is possible to build a view of components concisely.

I'm on mobile and can't provide a code example but would be happy to dive into this further tomorrow. M

jnraine:
I'm leaving on break starting now until the New Year. I will pick this up come January. 🍾🥂

manuelpuyol:
Closing in favor of https://github.com/github/view_component/pull/875

manuelpuyol:
Closing in favor of https://github.com/github/view_component/pull/875

PR #538, opened 2020-12-01
Default to original view_context for anything that is not a ViewComponent

caifara:
Rendering partials inside a component, with an object as argument, fails in specific cases:

`render some_object`

What happens is, we fall into the else case here because the argument is an object.

https://github.com/github/view_component/blob/bbe48fc73ff85c1565f92575be6460aef474269c/lib/view_component/base.rb#L104-L110

This tries to render the partial, but within the context of the component. This works, but not if the server already rendered that partial in normal context. In that case, the template is found, but the compiled template method is defined on the original context as seen here:

https://github.com/rails/rails/blob/6f9d4a000b4394bb13daacb30003bc4010a60d23/actionview/lib/action_view/template.rb#L151-L158
```ruby
    def render(view, locals, buffer = ActionView::OutputBuffer.new, add_to_stack: true, &block)
      instrument_render_template do
        compile!(view)
        view._run(method_name, self, locals, buffer, add_to_stack: add_to_stack, &block)
      end
    rescue => e
      handle_render_error(view, e)
    end
```

and here:  https://github.com/rails/rails/blob/6f9d4a000b4394bb13daacb30003bc4010a60d23/actionview/lib/action_view/template.rb#L280-L282

This results in an error like: 

`undefined method `_app_views_nodes__toc_html_haml__3805333848491569737_66100' for #CollectionsComponent:0x00007fe40e066720`

So the method is defined, but not within the current view_context (being the component). What I suppose is to make sure we use the original view context in all cases where we're not rendering other view_components.

(The problem occurs the other way round too: if rendered first within the component, it results in the above error if when trying to render within the normal view context.)

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -101,7 +101,7 @@ module ViewComponent
 
     # use original view_context if we're not rendering a component
     def render(options = {}, args = {}, &block)
-      if options.is_a? self.class
+      if options.is_a? ViewComponent::Base
         super
       else
         view_context.render(options, args, &block)
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -99,7 +99,11 @@ module ViewComponent
 
     def initialize(*); end
 
-    # use original view_context if we're not rendering a component
+    # Re-use original view_context if we're not rendering a component.
+    #
+    # This prevents an exception when rendering a partial inside of a component that has also been rendered outside
+    # of the component. This is due to the partials compiled template method existing in the parent `view_context`,
+    #  and not the component's `view_context`.
     def render(options = {}, args = {}, &block)
       if options.is_a? ViewComponent::Base
         super
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -99,13 +99,12 @@ module ViewComponent
 
     def initialize(*); end
 
-    # If trying to render a partial or template inside a component,
-    # pass the render call to the parent view_context.
+    # use original view_context if we're not rendering a component
     def render(options = {}, args = {}, &block)
-      if options.is_a?(String) || (options.is_a?(Hash) && options.has_key?(:partial))
-        view_context.render(options, args, &block)
-      else
+      if options.is_a? self.class
         super
+      else
+        view_context.render(options, args, &block)
       end
     end
 
@@ -291,7 +290,6 @@ module ViewComponent
       def provided_collection_parameter
         @provided_collection_parameter ||= nil
       end
-
     end
 
     ActiveSupport.run_load_hooks(:view_component, self)
--- a/test/app/components/partial_component.html.erb
+++ b/test/app/components/partial_component.html.erb
@@ -1,2 +1,3 @@
 <%= render partial: "test_partial" %>
 <%= render "test_partial" %>
+<%= render OpenStruct.new(to_partial_path: "test_partial") %>
--- a/test/view_component/view_component_test.rb
+++ b/test/view_component/view_component_test.rb
@@ -132,7 +132,7 @@ class ViewComponentTest < ViewComponent::TestCase
   def test_renders_partial_template
     render_inline(PartialComponent.new)
 
-    assert_text("hello,partial world!", count: 2)
+    assert_text("hello,partial world!", count: 3)
   end
 
   def test_renders_content_for_template
@@ -240,7 +240,6 @@ class ViewComponentTest < ViewComponent::TestCase
         "This is the footer"
       end
     end
-
 
     assert_selector(".card.mt-4")
 
@@ -389,7 +388,6 @@ class ViewComponentTest < ViewComponent::TestCase
   end
 
   def test_template_changes_are_not_reflected_if_cache_is_not_cleared
-
     render_inline(MyComponent.new)
 
     assert_text("hello,world!")


lib/view_component/base.rb
```diff
@@ -99,13 +99,12 @@ def render?
...
     def initialize(*); end
 
-    # If trying to render a partial or template inside a component,
-    # pass the render call to the parent view_context.
+    # use original view_context if we're not rendering a component
     def render(options = {}, args = {}, &block)
-      if options.is_a?(String) || (options.is_a?(Hash) && options.has_key?(:partial))
-        view_context.render(options, args, &block)
-      else
+      if options.is_a? self.class
```
BlakeWilliams, line 104:
```suggestion
      if options.is_a? ViewComponent::Base
```

I think we want `ViewComponent::Base` instead of `self.class`, otherwise I think we'll hit the `view_context.render` case unless we're rendering the same component recursively.

lib/view_component/base.rb
```diff
@@ -99,13 +99,12 @@ def render?
 
     def initialize(*); end
 
-    # If trying to render a partial or template inside a component,
-    # pass the render call to the parent view_context.
+    # use original view_context if we're not rendering a component
```
BlakeWilliams, line 102:
```suggestion
    # Re-use original view_context if we're not rendering a component.
    # 
    # This prevents an exception when rendering a partial inside of a component that has also been rendered outside
    # of the component. This is due to the partials compiled template method existing in the parent `view_context`,
    #  and not the component's `view_context`.
```

Thoughts on something like the above, explaining why this is necessary?

BlakeWilliams (commented):
This is looking good!

I had two small pieces of feedback, I think the `ViewComponent::Base` change may cause tests to pass since it should cause both branches of that conditional to be exercised instead of just the `else`.

lib/view_component/base.rb
caifara: This change tries not to interfere with the way Rails works, thus naturally should not exhibit unexpected results. In that sense it may not be necessary to explain in long form?

Is there a specific reason to render with the component as context? Implementing it like this passes all tests (and, in my Rails app it doesn't seem to lead to immediate problems):

``` ruby
    # Render using the original `view_context` Rails provides.
    #
    # This makes sure that compiled template methods live on that `view_context`
    # and not on the rendering component.
    def render(options = {}, args = {}, &block)
      view_context.render(options, args, &block)
    end
```

It makes sure we don't interfere with Rails expecting its methods to be defined within its `view_context`. On the other hand, the render monkey patch makes sure components can render as expected.

Not committing immediately as I may not see the reason why the `view_context` needs to change.

caifara: @BlakeWilliams oh, and thanks for reviewing! Really appreciate view_component and your fast responses!

BlakeWilliams: > Is there a specific reason to render with the component as context? Implementing it like this passes all tests (and, in my Rails app it doesn't seem to lead to immediate problems):

That's a good question. I'm not 100% sure, but I could see it being a performance reason.

Without diving into it, I think re-using the current `view_context` via `super` avoids instantiating a new view context instance, which is done when calling `view_context`.

There's a small benchmark script that can be run via `bundle exec rake benchmark`, we could verify that assumption by running that against `master` and this branch.

caifara: Well, ran it more than once, that's for sure :-).

I always get the message:

````
config.eager_load is set to nil. Please update your config/environments/*.rb files accordingly:

  * development - set it to false
  * test - set it to false (unless you use a tool that preloads your test environment)
  * production - set it to true
````

Always using `view_context` (twice):

````
          component:    16.155k i/100ms
            partial:     1.691k i/100ms
          component:    174.574k (± 5.6%) i/s -      1.745M in  10.029241s
            partial:     16.601k (± 6.8%) i/s -    165.718k in  10.034281s

Comparison:
          component::   174574.1 i/s
            partial::    16601.4 i/s - 10.52x  (± 0.00) slower


			
          component:    16.032k i/100ms
            partial:     1.614k i/100ms
          component:    177.213k (± 6.6%) i/s -      1.764M in  10.004032s
            partial:     16.234k (± 6.7%) i/s -    163.014k in  10.094963s

Comparison:
          component::   177213.0 i/s
            partial::    16234.1 i/s - 10.92x  (± 0.00) slower		
````
			

Master (changing context):

````

          component:    15.924k i/100ms
            partial:     1.712k i/100ms
          component:    174.103k (± 6.3%) i/s -      1.736M in  10.013919s
            partial:     16.490k (± 8.6%) i/s -    164.352k in  10.051274s

Comparison:
          component::   174103.2 i/s
            partial::    16489.8 i/s - 10.56x  (± 0.00) slower
			
			
          component:    14.450k i/100ms
            partial:     1.495k i/100ms
          component:    171.259k (± 9.0%) i/s -      1.705M in  10.051173s
            partial:     14.743k (± 9.9%) i/s -    146.510k in  10.055019s

Comparison:
          component::   171259.2 i/s
            partial::    14743.0 i/s - 11.62x  (± 0.00) slower			
````

Original version of this branch (changing context only for view components):

````

          component:    15.924k i/100ms
            partial:     1.712k i/100ms
          component:    174.103k (± 6.3%) i/s -      1.736M in  10.013919s
            partial:     16.490k (± 8.6%) i/s -    164.352k in  10.051274s

Comparison:
          component::   174103.2 i/s
            partial::    16489.8 i/s - 10.56x  (± 0.00) slower
			
			
          component:    14.450k i/100ms
            partial:     1.495k i/100ms
          component:    171.259k (± 9.0%) i/s -      1.705M in  10.051173s
            partial:     14.743k (± 9.9%) i/s -    146.510k in  10.055019s

Comparison:
          component::   171259.2 i/s
            partial::    14743.0 i/s - 11.62x  (± 0.00) slower			
````

BlakeWilliams: Thanks! It looks like we do have a small performance penalty when we don't share the `view_context` between component renders. Given that, I think we'll want to continue avoiding `view_context.render` for components.

caifara: Oh, right, I'll commit an explanation there. I don't really see the advantage of using a component specific context, it deviates from Rails default creating the origin of this issue and performs somewhat worse. Also note the lower standard deviation in the performance test. It's consistently lower than the one on master.
That said, using another context may fit better for reasons I don't know anything about so I'll commit an explanation making this ready for a merge.

BlakeWilliams (approved):

BlakeWilliams:
This is great, thank you!

Issue #234, opened 2020-02-28
Components should play nicely with Rails caching mechanisms


ozzyaaron:
### Feature request

Components should provide a digest to Rails views, or be able to use Rails explicit caching directive comments or be able to provide a cache_key. 

Essentially we should be able to invalidate a cache based on whether the Component class or template has changed. 

### Motivation

I'm just building on some previous reports primarily around where templates live and the filenames of the template/s. My concerns come down to integrating with Rails caching basics.  

The issue that we run into is basically summed up by 

```
## In any view/partial
cache[object, another_object] do
  render(MyComponent.new(attribute: value))
end
```

In this case if the template for `MyComponent` is changed the cache will not be invalidated. I tried to add `MyComponent` to the array used by `cache` and provide a `self.cache_key` method that used `ActionView::Digestor` but without having some further mechanism to invalidate the digestor's cache for the template/s when the template changes this didn't work. 

Rails allows for explicit directives to be used (https://guides.rubyonrails.org/caching_with_rails.html#explicit-dependencies) but by default this will only look inside `app/views`. You can add a view path `append_view_path("app/components")` but the digestor is expecting a partial and so a template filename would have to be prefixed with underscore. 

The way we are currently working around this is to have an `ApplicationComponent` that our components inherit from and where we override `ActionView::Component::Base.matching_views_in_source_location` so that templates are now found in `app/views/component_templates` and where the files are prefixed with underscore. 

```
  def self.matching_views_in_source_location
    return [] unless source_location

    *path, file_name = source_location.chomp(File.extname(source_location)).split("/")
    path.pop # remove components
    path.concat(["views", "component_templates"])
    path_to_template = path.push("_#{ file_name }").join("/")

    Dir["#{ path_to_template }.*{#{ActionView::Template.template_handler_extensions.join(',')}}"]
  end
```

Then in the above example where cache invalidation does not work via template digest changes we can do 

```
## In any view/partial
# Template Dependency: component_templates/my_component %>
cache[object, another_object] do
  render(MyComponent, attribute: value)
end
```

I think if AV::C is going to be core or integrate with Rails it would need a way to expose template digests nicely, or provide `cache_key` probably at the class level. If not then as a last resort we would need a way to use explicit directives in templates that use ActionView::Components internally. 

I'd also be all 👂 to people who have solved this already or if I've missed something about the features `ActionView::Component` provides!

pinzonjulian:
[...]

PR #542, opened 2020-12-03
provide cache digest for view

ermolaev:
fix if changed view, cache not expired

but, I think dynamicly modify `view_context.lookup_context.view_paths` bad for perfomance

https://github.com/github/view_component/issues/234

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Fix bug where `#with_request_url`, when used with query string, set the incorrect `request.path` and `request.fullpath`.
+
+    *Franz Liedke*
+
 * Add link to [ViewComponentAttributes](https://github.com/amba-Health/view_component_attributes) in Resources section of docs.
 
     *Romaric Pascal*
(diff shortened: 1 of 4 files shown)


lib/view_component/base.rb
```diff
@@ -52,7 +52,9 @@ def render_in(view_context, &block)
       self.class.compile(raise_errors: true)
 
       @view_context = view_context
-      @lookup_context ||= view_context.lookup_context
+      @lookup_context ||= view_context.lookup_context.tap do |c|
+        c.view_paths.unshift(Rails.root.join("app/components"))
```
BlakeWilliams, line 56: Is it possible to do this at the Railtie level instead of the component level?

I'm a little concerned about performance since this would run at least once per-component render.

ermolaev: I cached `lookup_context` in class attribute, bun it work only rails 6.0, 6.1, older versions is a hard work for support

joelhawksley:
@ermolaev thank you for the PR! 

I think we need to work to find a solution that works consistently across our supported Rails versions (>=5). I'm wary of introducing incomplete support for caching in the manner proposed by this PR. 

I recognize that this is probably not the answer you wanted to hear, and that taking on caching support for ViewComponent is no small effort, but I hope you agree that we need to take the time to do this right :heart:

stevehodgkiss:
This looks like it's almost there for Rails 6! It works for cache blocks with no dependencies, but with dependent view components the dependency tracker in ActionView fails find them. For example, with `<%= render Shared::Something.new %>` inside `cache "key" do` it logs "Couldn't find template for digesting: Shareds/Shared". To properly digest templates with dependencies, an extension or replacement of the [ERB tracker](https://github.com/rails/rails/blob/main/actionview/lib/action_view/dependency_tracker.rb#L32) in `ActionView::DependencyTracker` is probably needed, to return template paths for view component render calls.

PR #552, opened 2020-12-11
Sidecar translations

elia:
### Summary

Having sidecar translations is mentioned in multiple places in the project and the two apps using components I'm aware of are both doing it. Locally we solved with a less refined version of the attached code.

### Other Information

This builds on the work started in #74.

Expects a component to have a file with the same basename and the `.yml` extension, holding translations scoped to the component's path.

A component YML should look like this:

```yml
  en:
    hello: "Hello World!"
```

And inside the component will be possible to call

```erb
<%=  t(".hello") %>
```

Internally the scope will be expanded using the template's path.
As an example, the translation above from within `app/components/greeting/component.html.erb` will result in an internal `I18n` key like this:

```yml
  en:
    greeting:
      component:
        hello: "Hello World!"
```

---

_This PR is sponsored by @nebulab_

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1179,6 +1179,51 @@ app/components
 </div>
 ```
 
+##### Translations
+
+In order to support `I18n` translations scoped to the component you can add a dynamic translations loader:
+
+`config/locale/components.rb`
+
+```ruby
+require "view_component/i18n"
+ViewComponent::I18n.load
+```
+
+and setup reloading support for development:
+
+`config/application.rb`
+
+```ruby
+config.after_initialize do |app|
+  require "view_component/i18n"
+  ViewComponent::I18n.initialize_i18n(app)
+end
+
+config.before_eager_load do |app|
+  require "view_component/i18n"
+  ViewComponent::I18n.initialize_i18n(app)
+end
+```
+
+This allows for adding a "sidecar" YAML file that will define translations scoped to the component:
+
+`app/components/example_component.yml`
+
+```yml
+en:
+  # This key will be expanded to `en.example_component.hello`
+  hello: "Hello world!"
+```
+
+The translations will then be available with a leading "dot":
+
+`app/components/example_component.html.erb`
+
+```erb
+<p><%= t(".hello") %></p>
+```
+
 ## Known issues
 
 ### form_for compatibility
--- a/lib/view_component/i18n.rb
+++ b/lib/view_component/i18n.rb
@@ -1,0 +1,67 @@
+# frozen_string_literal: true
+
+require "yaml"
+require "pathname"
+
+module ViewComponent
+  module I18n
+    # Expects a component to have a file with the same basename and
+    # the .yml extension, holding translations scoped to the component's path.
+    #
+    # A component YML should look like this:
+    #
+    #   en:
+    #     hello: "Hello World!"
+    #
+    # And inside the component will be possible to call
+    #
+    #   t(".hello") # => "Hello World!"
+    #
+    # Internally the scope will be expanded using the template's path.
+    # As an example, the translation above from within
+    # `app/components/example/greeting_component.html.erb` will result in
+    # an internal I18n key like this:
+    #
+    #   en:
+    #     example:
+    #       greeting_component:
+    #         hello: "Hello World!"
+    #
+    #
+    def self.load(components_root: "#{Rails.root}/app/components", glob: "**/*component.{yml,yaml}")
+      Dir["#{components_root}/#{glob}"].each.with_object({}) do |path, translations|
+        relative_path = Pathname(path).relative_path_from(Pathname(components_root)).to_s
+        component_translations = YAML.load_file(path, fallback: {})
+        scopes = relative_path.sub(/\.ya?ml/, "").split("/")
+
+        component_translations.to_h.each do |locale, scoped_translations|
+          translations[locale] ||= {}
+          scopes.reduce(translations[locale]) do |nested_translations, scope|
+            nested_translations[scope] ||= {}
+          end
+          translations[locale].dig(*scopes).merge! scoped_translations
+        end
+      end
+    end
+
+    @i18n_initialized = false
+
+    # Install the sidecar translations reloader
+    def self.initialize_i18n(app = Rails.application, components_root: "#{Rails.root}/app/components")
+      return if @i18n_initialized
+
+      @i18n_initialized = true
+
+      reloader = app.config.file_watcher.new [], components_root => [".yml", ".yaml"] do
+        ::I18n.reload!
+      end
+
+      app.reloaders << reloader
+      app.reloader.to_run do
+        reloader.execute_if_updated { require_unload_lock! }
+      end
+
+      reloader.execute
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Add `ViewComponent::I18n` for sidecar translations
+
+    *Elia Schito*
+
 * Allow customization of the controller used in component tests.
 
     *Alex Robbin*
--- a/test/app/components/translations_component.html.erb
+++ b/test/app/components/translations_component.html.erb
@@ -1,4 +1,5 @@
 <div>
   <h1><%= t('.title') %></h1>
   <h2><%= t('translations_component.subtitle') %></h2>
+  <p><%= t('.body') %></p>
 </div>
--- a/test/app/components/translations_component.yml
+++ b/test/app/components/translations_component.yml
@@ -1,0 +1,2 @@
+en:
+  body: Even more lorem!
--- a/test/config/application.rb
+++ b/test/config/application.rb
@@ -14,6 +14,11 @@ require "slim"
 module Dummy
   class Application < Rails::Application
     config.action_controller.asset_host = "http://assets.example.com"
+
+    config.after_initialize do |app|
+      require "view_component/i18n"
+      ViewComponent::I18n.initialize_i18n(app)
+    end
   end
 end
 
--- a/test/config/locales/components.rb
+++ b/test/config/locales/components.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+require "view_component/i18n"
+
+ViewComponent::I18n.load
--- a/test/view_component/i18n_test.rb
+++ b/test/view_component/i18n_test.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+require "test_helper"
+require "fileutils"
+
+class ViewComponent::I18nTest < ViewComponent::TestCase
+  def setup
+    FileUtils.mkdir_p "#{components_root}/greeting"
+
+    File.write "#{components_root}/greeting/component.yml", <<~YAML
+      en:
+        hello: Hello!
+    YAML
+
+    File.write "#{components_root}/widget_component.yml", <<~YAML
+      en:
+        title: I'm a widget!
+        sub:
+          key: foobar
+    YAML
+  end
+
+  def test_multiple_translation_files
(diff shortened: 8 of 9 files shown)


elia:
For some reason Rails 5.0.0 on Ruby 2.7 was failing with a weird bundler error, updating to the latest patch seems to have fixed the problem.

joelhawksley:
@elia thank you for the PR! We should be able to review it early next week. 

In the meantime, I believe we can revert the dependency version changes. We see that build failure from time to time, but haven't been able to identify the cause. Rebuilding usually fixes it.

elia:
> @elia thank you for the PR! We should be able to review it early next week.

Awesome!

> In the meantime, I believe we can revert the dependency version changes. We see that build failure from time to time, but haven't been able to identify the cause. Rebuilding usually fixes it.

Done ✅ 

Do you want me to send [that change](https://github.com/elia/view_component/commit/e53b398) into another PR? Staying up-to-date seems a good way to avoid needless headaches.

lib/view_component/i18n.rb
```diff
@@ -0,0 +1,46 @@
...
+    # As an example, the translation above from within `app/components/greeting/component.html.erb`
+    # will result in an internal I18n key like this:
+    #
+    #   en:
+    #     greeting:
+    #       component:
+    #         hello: "Hello World!"
+    #
+    #
+    def self.load(components_root: "#{Rails.root}/app/components", glob: "**/*component.{yml,yaml}")
```
BlakeWilliams, line 30: I really love this change, but I think it'd be great if we could make this part of the compilation step.

I think that'd let us more tightly couple components to translation files, removing the need to configure translations. I *think* it might also help with template reloading, but that may need more investigation.

BlakeWilliams: Building on that thought a bit, it'd be great if we could use the component's file name to determine the translation file name instead of requiring the translation files to end in `component.*`.

e.g. If I have `modal.rb`, I'd expect the translation file to be called `modal.yml`,  or `modal/modal.yml` (or `.yaml` as an extension).

BlakeWilliams (commented):
Thanks for taking this on! This is a really great feature.

lib/view_component/i18n.rb
elia: The idea of using the compilation step to add paths to I18n is really neat, and `source_location` is perfect to do that. 

Also about reloading I missed an initializer during the extraction from our application, which would be integrated in `engine.rb` (it's modeled after the original I18n reloading code):

```rb
# Reload component translations whenever we change them.
Rails.application.configure do
  config.after_initialize do |app|
    app.config.file_watcher.new [], "#{Rails.root}/app/components" => ".yml" do
      I18n.reload!
    end.tap do |reloader|
      app.reloaders << reloader
      app.reloader.to_run do
        reloader.execute_if_updated { require_unload_lock! }
      end
    end
  end
end
```

With an hypothetical `modal.rb` component it should be already possible to adjust the glob, even to a generic `**/*.y{a,}ml`, to encompass any YAML in the components folder.

---

So, before proceeding with the implementation inside `compile`, I'd like to have your opinion on a couple of considerations:
- I foresee possible headaches for applications not being strict about only using the translations within the component
- We might still need the reloader to properly pickup changes (renames/deletions) 

The code would go more or less like this:

```rb
def compile
  unless I18n.paths.include? component_locale
    I18n.paths << component_locale
    I18n.reload!
  end
  # ...
end
```

lib/view_component/i18n.rb
```diff
@@ -0,0 +1,46 @@
...
+      Dir["#{components_root}/#{glob}"].each.with_object({}) do |path, translations|
+        relative_path = Pathname(path).relative_path_from(Pathname(components_root)).to_s
+        component_translations = YAML.load_file path
+        scopes = relative_path.sub(/\.ya?ml/, "").split("/")
+
+        component_translations.to_h.each do |locale, scoped_translations|
```
cj, line 36: I have done some testing, and to allow options like:

```yml
en:
  greeting:
    hello: "Hello World!"

# AND

en:
  greeting:
    component:
      hello: "Hello World!"
```

Doing:

```ruby
        component_translations.to_h.each do |locale, scoped_translations|
          translations[locale] ||= {}
          translations[locale].deep_merge!(scoped_translations)
        end
```

seems to work better.  Thoughts?

docs/index.md
```diff
@@ -1117,9 +1117,42 @@ app/components
...
+```yml
+en:
+  hello: "Hello world!"
+```
+
+This will be expanded internally to use the full path of the file as a scope:
+
+```yml
+en:
+  greeting:
```
jonspalmer, line 1141: should this be `greeter` given the file paths?

lib/view_component/i18n.rb
jonspalmer: > I foresee possible headaches for applications not being strict about only using the translations within the component

Can you share a bit more about that concern? What's an example of "not being strict"?

docs/index.md
```diff
@@ -1117,9 +1117,42 @@ app/components
...
+
+In order to support `I18n` translations scoped to the component you can add a dynamic translations loader in `config/locale/components.rb`.
+
+```ruby
+require "view_component/i18n"
+ViewComponent::I18n.load components_root: "#{Rails.root}/app/components", glob: "**/*component.yml"
+```
+
+This allows for adding a "sidecar" YAML file that will define translations
+scoped to the component, e.g. `app/components/greeter/component.yml`:
```
jonspalmer, line 1130:
```suggestion
scoped to the component, e.g. `app/components/greeting/component.yml`:
```

docs/index.md
```diff
@@ -976,7 +976,7 @@ function importAll(r) {
   r.keys().forEach(r)
 }
 
-importAll(require.context("../components", true, /_component\.js$/))
+importAll(require.context("../components", true, /[_\/]component.js$/))
```
Spone, line 979: I guess this unrelated change should be part of another PR.

Also, why removing the `\` in `component\.js`? It's useful to escape the dot that otherwise means "any character".

docs/index.md
Spone: What about using `example` as the component name to stay consistent with the rest of the docs?

lib/view_component/i18n.rb
```diff
@@ -0,0 +1,46 @@
...
+    #   t(".hello") # => "Hello World!"
+    #
+    # Internally the scope will be expanded using the template's path.
+    # As an example, the translation above from within `app/components/greeting/component.html.erb`
+    # will result in an internal I18n key like this:
+    #
+    #   en:
+    #     greeting:
+    #       component:
+    #         hello: "Hello World!"
```
Spone, line 27: I'm curious about the rationale behind this naming/namespacing.

By default the naming of the template file for an `ExampleComponent` with sidecar would be: `app/components/example_component/example_component.html.erb` so I think the docs should show this case instead.

What would the structure of the I18n key be in this case? `en.example_component.example_component.hello`, right?

lib/view_component/i18n.rb
elia: The idea is that a component would always use the relative notation, if that's the case the component's naming shouldn't matter, e.g. both these scenarios are expected to work:

---
`app/components/greeting.html.erb`
```erb
<%= t(".hello") %>
```
`app/components/greeting.yml`
```erb
en:
  hello: "Hello World!"
```
---
`app/components/greeting/component.html.erb`
```erb
<%= t(".hello") %>
```
`app/components/greeting/component.yml`
```erb
en:
  hello: "Hello World!"
```
---
Am I missing some context with regard to this?

lib/view_component/i18n.rb
Spone: :wave: @elia did you see this comment?

elia: _Sorry, I did read it but then somehow skipped it when replying_

I'll update the doc, the use of `example/component.*` was just coming from the way we organized them in the app I extracted this from.

elia:
The CI is having random failures with bundler (v1) and `Browser did not produce websocket url within 5 seconds (Ferrum::ProcessTimeoutError)` in the PVC job, I don't have the permission to restart the build, let me know if I should just keep trying using fake force-pushes.

Spone:
> The CI is having random failures with bundler (v1) and `Browser did not produce websocket url within 5 seconds (Ferrum::ProcessTimeoutError)` in the PVC job, I don't have the permission to restart the build, let me know if I should just keep trying using fake force-pushes.

I retried it a couple times, `CI / pvc (pull_request)` is failing every time. Not sure what to do @joelhawksley?

lib/view_component/i18n.rb
BlakeWilliams: > What I mean is that translations are expected to be there globally, but if we load them only when the component is compiled they might not be there in every environment/situation (think background jobs, rake tasks, partials, etc.) unless we ensure components are all eagerly compiled all the time.

Translations being available globally breaks the implicit and explicit encapsulation that view component assumes. For example, if we have two components with the same name they might have an i18n naming collision. I think it's important to handle that edge case.

lib/view_component/i18n.rb
Spone: *Don't worry about it ;)*

I find the `en.example_component.example_component.hello` I18n key not ideal, and I'm wondering if it shouldn't be something like `en.view_component.example_component.hello` or some other naming that:

- uses a `view_component` or `components` namespace to prevent collision with other parts of the app
- removes the component name duplication
- is consistent whether you have:
    - a *regular* component (with a template placed in `app/components/example_component.html.erb`)
    - a *sidecar* component (with a template placed in `app/components/example_component/example_component.html.erb`)
    - an *inline* component (so the `I18n.t` method is called from the component `call` method)

What do you think?

joelhawksley:
> > The CI is having random failures with bundler (v1) and `Browser did not produce websocket url within 5 seconds (Ferrum::ProcessTimeoutError)` in the PVC job, I don't have the permission to restart the build, let me know if I should just keep trying using fake force-pushes.
> 
> I retried it a couple times, `CI / pvc (pull_request)` is failing every time. Not sure what to do @joelhawksley?

We've marked that build as optional while we investigate ❤️

patchwork-her:
You finally did something. Can you admit to me please?

patchwork-her:
Illegal activity if you ask me because I didn't authorize none of this and
I've been hacked out of it.

lib/view_component/i18n.rb
elia: I like the idea of both the prefix and collapsing the sidecar components path with the duplicate name, the only thing I want to point out before trying to implement them is that we need to keep this configuration in synch with `virtual_path` in `ViewComponent::Base` because that's the one that will drive relative translations inside templates.

So, probably we should start moving some of this stuff to engine-level configurations, including maybe the components root and the glob for the YAML files. That would be ok?

Spone: I'd like to request @joelhawksley and @BlakeWilliams opinion about this, as I'm not sure what's the best move.

BlakeWilliams: Building on https://github.com/github/view_component/pull/552#discussion_r584988143, if we go down that path I think we could avoid coupling i18n key names to specific components because each component will have access to the global i81n, but also its sidecar i18n, but not the sidecar i18n of other components. 

> So, probably we should start moving some of this stuff to engine-level configurations, including maybe the components root and the glob for the YAML files. That would be ok?

We should avoid configuration wherever possible (or at least have some great defaults), so if we can make i18n work out of the box that would be preferable.

elia: @BlakeWilliams please checkout #660 which features a version of sidecar i18n as you intended, the global one is still available via `I18n.t` or `helpers.t`.

elia:
This has been replaced by #660 in which the I18n lookup will only include the sidecar translations instead of having them mixed with the global Rails I18n.

patchwork-her:
My "one" big thing I truly pay for is this. I need it

Issue #549, opened 2020-12-10
NoMethodError (private method `format' called)


xdmx:
### Steps to reproduce
I have a very simple component like
```ruby
class ActivityEntryComponent < ViewComponent::Base
  with_collection_parameter :event

  def initialize(event:)
    @event = event
  end
end
```
which is rendered in both views and in a controller action to load more activities. The rendering from the view is working, but after I've upgraded to rails 6.1 (and view component 2.23) the rendering from the controller is returning that it's calling the private format method.

The controller is also very simple
```ruby
class ActivitiesController < ApplicationController
  def index
    # fetch paginated events

    render ActivityEntryComponent.with_collection(@page.records)
  end
end
```

### Expected behavior
It should render the component without any error

### Actual behavior
```
NoMethodError (private method `format' called for #<ViewComponent::Collection:0x000055b3b4c5af68>):
  
app/controllers/activities_controller.rb:15:in `index'
```

### System configuration
**Rails version**: 6.1.0

**Ruby version**: 2.7.2

**Gem version**: 2.23.0

BlakeWilliams:
@xdmx Thanks for reporting this issue! Would it be possible to open a PR with a failing test for us to look at?

jcoyne:
Is this the same as #156?

xdmx:
I can't at the moment as I'm bit busy on the whole rails 6.1 upgrade and other work stuff. I believe it's just a matter of making sure that `render SomeComponent.with_collection` works from a controller in rails 6.1, as I don't have any specific complex setup (mainly standard rails and gem config).

Let me know if you could get the chance of debugging it out, otherwise I'll try to write a more complete example when I get some time.

xdmx:
Just did another quick try and `render ActivityEntryComponent.new(event: event)` from the controller works, but `render ActivityEntryComponent.with_collection(events)` doesn't (`format` error).. while `render inline: ActivityEntryComponent.with_collection(events)` renders `#`, so it seems that the problem is on the `render` itself, not the `with_collection` (or maybe both of them used together).

Just as a note, I also disabled the layout in the controller with `layout false`

Bramjetten:
Just hit this bug when upgrading to Rails 6.1. My current workaround is unfortunate, but working:
`render inline: view_context.render(Spina::MediaPicker::ImageComponent.with_collection(@images))`

joelhawksley:
Thanks for the report folks! I'm going to take a look at this one now ❤️

PR #554, opened 2020-12-15
Fix bug rendering component collections from controllers

joelhawksley:
Fixes https://github.com/github/view_component/issues/549.


BlakeWilliams (approved):
Looks great!

Issue #558, opened 2020-12-16
Do not capture block in ViewComponent::Base#render_in if render? is false


virolea:
## Feature request

Capturing the block content passed to a component you are rendering only if the component `render?` is true.

### Motivation

Let's take the following example: 

```ruby
# profile_component.rb 
class ProfileComponent < ViewComponent::Base 
  def initialize(user:) 
    @user = user 
    @name = user.name
  end 
end 

# index.html.erb
<%= render ConditionalComponent.new(should_render: current_user) do %>
  <%= render ProfileComponent.new(user: current_user) %>   
<% end %>
```

In the case when `current_user` is `nil`, this will break, as, in the current_implementation `render ProfileComponent.new(user: current_user)` will be called, and `user.name` as well: 

```ruby 
# lib/view_component_base L71 

def render_in(view_context, &block)
  # ...

  # Assign captured content passed to component as a block to @content
  @content = view_context.capture(self, &block) if block_given?  # ⬅️ block captured no matter what

  # ...
end
```

I could add a condition to `ProfileComponent` in the case when `user` is `nil` but since it is a compulsory argument, I want `ProfileComponent` to fail if I try to render it without a user. 

Plus, there's no need to capture the block if the parent component is not rendering.

PR #559, opened 2020-12-16
Do not capture block content of a component if component render is false

virolea:
### Summary

In #558 I detailed the motivations for a conditional block content capture when rendering a component. In the current implementation, we capture the block content passed to a component we are rendering, no matter what:

```ruby
# lib/view_component_base L71 

  # ...

  @content = view_context.capture(self, &block) if block_given?  # ⬅️ block captured no matter what

  # ...
end
```
The code in this PR conditions the capture of the block on the result of `render?`. 

== EDIT ==
After thinking about it, I don't see why we should not condition the call to `before_render` to `render?` as well !

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -68,12 +68,13 @@ module ViewComponent
       old_current_template = @current_template
       @current_template = self
 
-      # Assign captured content passed to component as a block to @content
-      @content = view_context.capture(self, &block) if block_given?
 
       before_render
 
       if render?
+        # Assign captured content passed to component as a block to @content
+        @content = view_context.capture(self, &block) if block_given?
+
         render_template_for(@variant)
       else
         ""
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -68,13 +68,14 @@ module ViewComponent
       old_current_template = @current_template
       @current_template = self
 
+      # Assign captured content passed to component as a block to @content
+      @content = view_context.capture(self, &block) if block_given?
 
       before_render
 
       if render?
-        # Assign captured content passed to component as a block to @content
-        @content = view_context.capture(self, &block) if block_given?
-
+        # Call the block from `render_content` and set it as @content if render? is true
+        @content = @block_from_render_content.call if @block_from_render
         render_template_for(@variant)
       else
         ""
@@ -93,6 +94,10 @@ module ViewComponent
 
     def render?
       true
+    end
+
+    def render_content(&block)
+      @block_from_render_content = block
     end
 
     def initialize(*); end
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -69,13 +69,11 @@ module ViewComponent
       @current_template = self
 
       # Assign captured content passed to component as a block to @content
-      @content = view_context.capture(self, &block) if block_given?
+      @_content = view_context.capture(self, &block) if block_given?
 
       before_render
 
       if render?
-        # Call the block from `render_content` and set it as @content if render? is true
-        @content = @block_from_render_content.call if @block_from_render
         render_template_for(@variant)
       else
         ""
@@ -96,8 +94,14 @@ module ViewComponent
       true
     end
 
-    def render_content(&block)
-      @block_from_render_content = block
+    def content(&block)
+      if block_given?
+        @block_from_content = block
+      else
+        @_content ||= if defined?(@block_from_content)
+          @block_from_content.call
+        end
+      end
     end
 
     def initialize(*); end
@@ -173,7 +177,7 @@ module ViewComponent
       @request ||= controller.request
     end
 
-    attr_reader :content, :view_context
+    attr_reader :view_context
 
     # The controller used for testing components.
     # Defaults to ApplicationController. This should be set early
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## master
 
+* Do not capture the block of a component if `render?` is false
+
+    *Vincent Rolea*
+
 ## 2.23.2
 
 * Fix bug where rendering a component `with_collection` from a controller raised an error.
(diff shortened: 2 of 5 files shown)


BlakeWilliams:
Thanks for the PR!

There are some downsides with using `before_render` to determine if content is captured, which is that `content` is no longer available in `before_render` if you need it. It also has other side-effects like slots not being instantiated/set which means they're no longer available in `before_render` either.

Maybe that points towards `content` needing to be handled more explicitly by components? or maybe ensuring that block content is captured when trying to access slots (there's a few dragons with that too)?

virolea:
Hey @BlakeWilliams, thanks for your insightful comment!

> There are some downsides with using before_render to determine if content is captured, which is that content is no longer available in before_render if you need it. It also has other side-effects like slots not being instantiated/set which means they're no longer available in before_render either.

Wouldn't moving the `before_render` call inside the `if render?` condition right after content capture do the trick? 

```ruby 
if render? 

   before_render

   render_template_for(@variant
else 
  "" 
end 
```

A more radical approach would be to render an empty string as soon as possible when calling `render_in`:

```ruby
  return "" unless render?
  
  self.class.compile(raise_errors: true)
  # [...] Normal render operations
end 
```

But here again there might be some issues in doing that. One I can see is that one might want to compile the template indifferently of `render?` to provide error feedback. 



If those solutions are not satisfactory, I'd happily dig deeper to find a more appropriate one, following your adivces!

BlakeWilliams:
> Wouldn't moving the before_render call inside the if render? condition right after content capture do the trick?

I think it would, but that also means that `render?` can't access `content`, which is desirable behavior.

I think we may be able to make this change if we change how we set and access the public `content` method.

Maybe something like:

```ruby
def content
  @_content ||= view_context.capture(self, &@_content_block) if @_content_block
end
```

That would also require us to change `render_in` so we set `@_content_block = block` before we call `render?` or `before_render?`. We should probably `raise` with a helpful message if `view_context` is `nil` similar to how the `controller` and `helpers` methods behave too.

Thoughts?

BlakeWilliams:
I've been thinking about this a bit more, and I think that would still break slots since we need to capture the block content in order to set slots, in most cases.

e.g.

```erb
<%= render MyComponent.new do |c| %>
  <%= c.my_slot(key: value) %>
<% end %>
```

To keep the existing API, we'd have to make sure that `@content` has been set before allowing the access of slots, which has a few other problems too, like when a slot is set on a component from outside of the content block.

virolea:
>  I think it would, but that also means that render? can't access content, which is desirable behavior.

Absolutely, your comment and the test case added in  #563 shed some light in the need for content to be available elsewhere than in the template. 

> I've been thinking about this a bit more, and I think that would still break slots since we need to capture the block content in order to set slots, in most cases.

It does you're right, it also breaks content area. After some experimentations and deep dive into slots, I am not sure that figuring out if a component has slots registered are not to conidtionnaly set `@content` is the way to go. 

Having put more thoughts into it, I do think however that having a more explicit way to manage `@content` as you first suggested, would be a more suitable answer to this problem. It would allow to manage slots and content independently. For instance: 

```ruby
<%= render MyComponent.new do |c| %>
  <%= c.my_slot do %>
    My slot
  <% end %> 

  <%= c.render_content do  %>
    # add all the content in that block  
  <% end %>
<% end %>
```

I've added a commit which demonstrates this behavior. One could you this method to render content only if `render?` is true, while keeping retrocompatibility for traditionnaly set content. I've chosen `render_content` since [`render` conflicts with `ActionView::Base#render`](https://apidock.com/rails/ActionView/Base/render). The content in `render_content` is captured and set as @content only if render? is true. The capture of @content rest unchanged, allowing for slots to be initialized and content to be accessible in `render?`

What do you think? 
And Merry christmas ! 🎄

BlakeWilliams:
ohhhhh, I like that idea! Alternatively, I wonder if we could change the eager evaluation into lazy evaluation? e.g. We only `capture` the content block if a user accesses `content` or something that depends on `content` being captured like a slot. I may spike that out in a separate PR so we can compare/contrast the two.

I've also opened a draft PR and the first commit is [adding a performance oriented test](https://github.com/github/view_component/pull/576) that I think could help highlight the importance of this change.

lib/view_component/base.rb
```diff
@@ -94,6 +96,10 @@ def render?
       true
     end
 
+    def render_content(&block)
+      @block_from_render_content = block
```
BlakeWilliams, line 100:
```suggestion
    def content(&block)
      if block_given?
        @block_from_render_content = block
      else
        @_content ||= block_from_render_content.call 
      end
```

Thoughts on renaming this to `content` and copying the slots API so this is both a getter and a setter?

virolea: I like it! Adds consistency with the slots API

virolea:
Thanks for the feedback, I added the lazy evaluation of the block and made it more a slot-like API. 

> I wonder if we could change the eager evaluation into lazy evaluation? e.g. We only capture the content block if a user accesses content or something that depends on content being captured like a slot.

This was my goal initially but it adds some overhead I was not happy with. For instance, we could track if the content is captured in the component, with an instance variable, like `@content_was_captured` initially set to false, and save the block for later use. For slots v2, we can check if the content is captured in `SlotableV2#get_slot`:

```ruby 
# slotable_v2.rb

def get_slot
  slot = self.class.registered_slots[slot_name]
  @_set_slots ||= {}
  
  capture_content unless @content_was_captured # ⬅️ ensure content capture here as soon as we get a slot

  if @_set_slots[slot_name]
    return @_set_slots[slot_name]
  end 

  if slot[:collection]
    []
  else
    nil
  end

end
```

Similar approach can be replicated for `content` but again I was not happy with it.

jonspalmer:
I'm late to the party and perhaps you've moved away from this idea anyway but IMO one of the valid uses of `before_render`  would be to perform some calculation or manipulation that `render?` depends on. As such they need to be independent.

PR #563, opened 2020-12-21
add test case for rendering component only if content is provided

joelhawksley:
In #558 / #559, @virolea is proposing changes to how capturing content and `render?` work together. 

After reading the proposed changes, I realized that we have a conflicting use case in the GitHub.com codebase, where we have components that check for the presence of `content` in `render?`.

While I'm open to the proposed changes, I felt the need to document our current usage of this API so as to not accidentally introduce a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,11 @@
 
 ## main
 
-## 2.23.3
+## 2.24.0
+
+* Add test case for checking presence of `content` in `#render?`.
+
+    *Joel Hawksley*
 
 * Rename `master` branch to `main`.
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,11 @@
 
 ## main
 
-## 2.23.3
+## 2.24.0
```
BlakeWilliams, line 5:
```suggestion
```

Feel free to ignore, but I thought we were keeping un-released changes under `## main`?

BlakeWilliams (approved):

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## main
 
-## 2.23.3
+
```
```suggestion
```

PR #565, opened 2020-12-29
Look up the correct path when slot-enabled components are nested

agrobbin:
Consider this component:

```ruby
module Nested
  class SlotsV2Component < ViewComponent::Base

    renders_many :items, 'MyHighlightComponent'

    class MyHighlightComponent < ViewComponent::Base
    end
  end
end
```

Before this change, the `MyHighlightComponent` template lookup would include this path:

```
app/components/nested/nested/slots_v2_component/my_highlight_component.*{...}
```

Notice the duplicated `nested/nested` directory structure.

This is due to `Compiler#matching_views_in_source_location` basing the nested component file lookup on the `deconstantize`d version of the component class name.

With this change, we now base it purely off of the component's filename, resulting in this path:

```
app/components/nested/slots_v2_component/slots_v2_component.*{...}
```


lib/view_component/compiler.rb
```diff
@@ -163,9 +163,8 @@ def matching_views_in_source_location
       # end
       #
       # Without this, `MyOtherComponent` will not look for `my_component/my_other_component.html.erb`
-      nested_component_files = if component_class.name.include?("::")
-        nested_component_path = component_class.name.deconstantize.underscore
-        Dir["#{directory}/#{nested_component_path}/#{component_name}.*{#{extensions}}"]
+      nested_component_files = if component_class.name.include?("::") && component_name != filename
+        Dir["#{directory}/#{filename}/#{component_name}.*{#{extensions}}"]
```
agrobbin, line 167: I'm definitely looking for another set of 👀 here, as this is not the simplest piece of code ever written.

lib/view_component/compiler.rb
```diff
@@ -163,9 +163,8 @@ def matching_views_in_source_location
       # end
       #
       # Without this, `MyOtherComponent` will not look for `my_component/my_other_component.html.erb`
-      nested_component_files = if component_class.name.include?("::")
-        nested_component_path = component_class.name.deconstantize.underscore
-        Dir["#{directory}/#{nested_component_path}/#{component_name}.*{#{extensions}}"]
+      nested_component_files = if component_class.name.include?("::") && component_name != filename
```
agrobbin, line 166: Adding this additional conditional seemed worth doing, to avoid doing this check when the names match up, in which case I don't believe we could possibly be dealing with a nested component.

BlakeWilliams: I think this seems fine since there shouldn't ever be a case where the component has a nested component with the same name.

lib/view_component/compiler.rb
BlakeWilliams: This will work for components that are nested one level deep, but I *think* it might break if they're nested any deeper than that since we're relying on `filename`.

e.g.

```ruby
# app/components/foo.rb
class Foo < ViewComponent::Base
  class Bar < ViewComponent::Base
    class Baz < ViewComponent::Base
    end
  end
end
```

I think that would look for `Baz`s template in `app/components/foo/baz.html.erb`, when I think we'd want `app/components/foo/bar/baz.html.erb`.

Can we add a test for this case?

BlakeWilliams (commented):
Thanks for fixing this!

Overall this looks good, but I had one question about deeply nested components.

lib/view_component/compiler.rb
agrobbin: Ahh yeah you're totally right @BlakeWilliams!

Part of me actually is inclined to have this behave exactly as it does with this change, essentially flattening the component hierarchy, but I can see how that could be surprising.

I'm curious if you have an idea for how we might cleanly handle the nested scenario, b/c I'm struggling to think of a good solution.

alekhood317:
How do I play the video ?

lib/view_component/compiler.rb
agrobbin: @BlakeWilliams the more I think about this, the more I'm inclined to have this be the "correct" behavior, even when a component is deeply nested. That keeps the file system layout identical for component files and template files, and happens to keep the template lookup relatively simple. I've updated the test case to at least cover the scenario.

If we *did* want to somehow look for deeply nested component template files differently, I think we'd have to somehow walk [`module_parents`](https://api.rubyonrails.org/v6.0.3.4/classes/Module.html#method-i-module_parents), determine if each parent is a `ViewComponent::Base`, compare each one's `source_location`, and then build up the path that way.

If others disagree, I can definitely try and tackle the deeply nested scenario differently.

BlakeWilliams: > That keeps the file system layout identical for component files and template files, and happens to keep the template lookup relatively simple. I've updated the test case to at least cover the scenario.

That makes sense to me. I think a good thing to follow here might be how controllers lookup their template paths. If we had a controller that was nested inside of another controller, would this be consistent with how its lookup paths behave?

agrobbin: Meaning something like this?

```ruby
class FooController < ApplicationController
  def index
  end

  class BarController < ApplicationController
    def index
    end
  end
end
```

When giving this a shot in an app locally, Rails isn't able to look up the controller successfully, let alone look up the template, so it might not be something we can mimic.

BlakeWilliams: Yeah, I was curious if Rails handled that situation. Thanks for trying that out though!

I've thought about this a bit and I think you're right that this is the way to go. 👍

BlakeWilliams (approved):

joelhawksley:
@agrobbin thanks for the PR! ❤️ 🙇🏻‍♂️

agrobbin:
Any time!

PR #568, opened 2021-01-04
Failing test for +link_to+ helper method

juanmanuelramallo:
### Summary

I think this is a bug. 

This PR adds a failing test to track and fix the bug.

## Steps to reproduce
1. Create a component
2. Use an AR instance in a +link_to+ method inside the component's template
```
link_to 'Something', Post.first
```
3. Render the component in a view
4. Visit the view 

## Actual
```
Error:
ActionView::Template::Error: undefined method `merge' for #<Post:0x00007fd155be2200 @id=1>
    app/components/link_component.html.erb:1:in `call'
    (eval):3:in `render_template_for'
```

## Expected
The link is displayed correctly

### DIFF
--- a/test/app/components/link_component.html.erb
+++ b/test/app/components/link_component.html.erb
@@ -1,0 +1,1 @@
+<%= link_to name, url_options, html_options %>
--- a/test/app/components/link_component.rb
+++ b/test/app/components/link_component.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class LinkComponent < ViewComponent::Base
+  attr_reader :name, :url_options, :html_options, :block
+
+  def initialize(name, url_options = nil, html_options = nil, &block)
+    @name = name
+    @url_options = url_options
+    @html_options = html_options
+    @block = block
+  end
+end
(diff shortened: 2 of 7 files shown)


joelhawksley:
@juanmanuelramallo thanks so much for the PR! It'd be great to get this working. Do you plan to take this on, or do you need a hand?

juanmanuelramallo:
@joelhawksley I'll need a hand

I'm not entirely familiar on how the internals work nor I have the time I'd like to dig into it, but I thought I could definitely write a failing test for it. 

I'm available for code reviews though

rewritten:
It is not a bug. The component under test overrides the `url_options` method, which is used by the routes to generate the url. If you change the name of the getter, or just keep it internal to create the url
```rb
def url
  url_for(@url_options)
end
```
it will work.

juanmanuelramallo:
Nice catch @rewritten I didn't realize about it! Thank you. 

I haven't tested it yet but it definitely makes sense, closing this.

rewritten:
The `url_options` method is called in https://github.com/rails/rails/blob/6-1-stable/actionpack/lib/action_dispatch/routing/route_set.rb#L199 on the same object that has the `url_for` and `post_path`.

PR #577, opened 2021-01-08
docs/index.md: Fix benchmarks link

jdennes:
### Summary

The benchmarks link on the GitHub Pages site at https://viewcomponent.org doesn't work as expected. It goes to https://viewcomponent.org/performance/benchmark.rb which 404s.

This branch makes the link go to the blob on the `main` branch instead: 

https://github.com/github/view_component/blob/main/performance/benchmark.rb

### Other Information

This seems like the simplest solution, but open to others if there's a preferred approach.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -48,7 +48,7 @@ ViewComponents use a standard Ruby initializer that clearly defines what is need
 
 #### Performance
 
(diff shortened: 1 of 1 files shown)


BlakeWilliams (approved):
Good catch! Thanks for fixing that.

PR #578, opened 2021-01-08
Remove no-cov from compiled_template call

maxbeizer:
### Summary

While pairing with @joelhawksley , he noticed that there was a `nocov` block and we decided to try to remove it and fix the coverage gap. Looks like CI passes without it so 🔥 .


joelhawksley (approved):

maxbeizer:
❤️🧡💛💚💙💜🖤

PR #585, opened 2021-01-15
Fix various warnings in test

maxbeizer:
### Summary

When running `bundle exec rake` Ruby emits over a dozen lines of warnings which is pretty distracting.

This PR aims to clean up some if not all of those warnings.


joelhawksley (approved):

maxbeizer:
❤️🧡💛💚💙💜🖤

PR #586, opened 2021-01-15
Only warn of deprecation if the Rails.env is not test

maxbeizer:
### Summary

related to #585  

### First pass

I split this one off from the other PR since it may involve more thought/discussion. Is there any reason to show these warnings in test? For development purposes, no. Would library consumers not see this warning with this change? I don't think they will and that may be a bad idea.

Still thinking about it.

### Second pass

I think this is a better approach. This uses an env var set up in `test_helper` to determine whether we are testing the library or a consumer is running tests against their app. If we are running VC tests, then swallow the warning.

In https://github.com/github/view_component/pull/586/commits/cddc367714294a6381c5f8826d303bdf415abd7e decided to introduce a new class to wrap the warning logic. One benefit here is that we could potentially capture those warnings and write tests against the, essentially asserting that various code paths will warn consumers. If you like this approach I can extend it a bit more and fix up the code coverage.

I thought I'd propose it and if you don't like it then we can go back to the simpler https://github.com/github/view_component/pull/586/commits/927c088cd3e5f7b8b5cfd500eca6c208e95e5438

cc @joelhawksley

### DIFF
--- a/lib/view_component/deprecation.rb
+++ b/lib/view_component/deprecation.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  # ViewComponent should warn consumers when they are using a deprecated method.
+  # In local development, however, those deprecation warnings are distracting.
+  # This class wraps the deprecation warning functionality in some basic logic
+  # so the deprecation warnings are only shown to consumers.
+  class Deprecation
+    def self.warn(message)
+      new.warn(message)
+    end
+
+    def initialize
+      @deprecation_class = if local_test_env?
+                             NoopDeprecation
+                           else
+                             ActiveSupport::Deprecation
+                           end
+    end
+
+    def warn(message)
+      @deprecation_class.warn(message)
+    end
+
+    private
+
+    def local_test_env?
+      ENV["VIEW_COMPONENT_ENV"] == "test"
+    end
+  end
+
+  class NoopDeprecation
+    def self.warn(_); end
+  end
+end
--- a/lib/view_component/deprecation.rb
+++ b/lib/view_component/deprecation.rb
@@ -12,10 +12,10 @@ module ViewComponent
 
     def initialize
       @deprecation_class = if local_test_env?
-                             NoopDeprecation
-                           else
-                             ActiveSupport::Deprecation
-                           end
+        NoopDeprecation
+      else
+        ActiveSupport::Deprecation
+      end
     end
 
     def warn(message)
--- a/lib/view_component/engine.rb
+++ b/lib/view_component/engine.rb
@@ -22,9 +22,9 @@ module ViewComponent
         )
 
         if options.preview_path.present?
-          ActiveSupport::Deprecation.warn(
+          ViewComponent::Deprecation.warn(
             "`preview_path` will be removed in v3.0.0. Use `preview_paths` instead."
-          ) unless ENV["VIEW_COMPONENT_ENV"] == "test"
+          )
           options.preview_paths << options.preview_path
         end
       end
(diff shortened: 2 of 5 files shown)


joelhawksley:
@maxbeizer IMO these warnings should be shown in tests. They are a primary form of feedback in a lot of folks' development lifecycle.

maxbeizer:
@joelhawksley yeah that's kind of where I landed on this one. I was looking into whether I could omit then when running the library's tests but ran out of time. That's what I'll be investigating more when I carve out more time to work on this

lib/view_component/engine.rb
```diff
@@ -22,7 +22,7 @@ class Engine < Rails::Engine # :nodoc:
         )
 
         if options.preview_path.present?
-          ActiveSupport::Deprecation.warn(
+          ViewComponent::Deprecation.warn(
             "`preview_path` will be removed in v3.0.0. Use `preview_paths` instead."
           )
```
joelhawksley, line 27: 👋🏻 what do you think about inlining this check instead of making a new Deprecation class?

```suggestion
          ActiveSupport::Deprecation.warn(
            "`preview_path` will be removed in v3.0.0. Use `preview_paths` instead."
          ) unless ENV["VIEW_COMPONENT_ENV"] == "test"
```

maxbeizer: I'd say the only argument against it is the following:

We could look at asserting that deprecation warnings invoked. A separate class could-- instead of noop-ing-- retain the warning and allow us to write tests that it would warn outside of local test. So you could write a test that says something like, 
```ruby
# a bit pseudo codey
assert_deprecation_warning "`preview_path`...", preview_path`
```

I certainly get the inclination to avoid the indirection here though. So we could just go back to https://github.com/github/view_component/pull/586/commits/927c088cd3e5f7b8b5cfd500eca6c208e95e5438 if you prefer.

maxbeizer: The more I think about it, I think `ViewComponent::Deprecation` is actually a good case for a singleton class too 🤔

maxbeizer: All that said, if you prefer in-lining it then I'm 👍  on that too

lib/view_component/deprecation.rb
```diff
@@ -0,0 +1,35 @@
...
+    def self.warn(message)
+      new.warn(message)
+    end
+
+    def initialize
+      @deprecation_class = if local_test_env?
+                             NoopDeprecation
+                           else
+                             ActiveSupport::Deprecation
+                           end
```
BlakeWilliams, line 18:
```suggestion
      @deprecation_class = if local_test_env?
        NoopDeprecation
      else
        ActiveSupport::Deprecation
      end
```

Thoughts on keeping the indentation consistent for readability?

test/test_helper.rb
```diff
@@ -17,6 +17,7 @@
 
 # Configure Rails Envinronment
 ENV["RAILS_ENV"] = "test"
+ENV["VIEW_COMPONENT_ENV"] = "test"
```
BlakeWilliams, line 20: As an alternative to setting an environment variable, what are your thoughts on adding a method to `ViewComponent::Deprecation` like `.silence_deprecation_warnings!` that we can call early on? That might also give us something to hook into if we wanted to re-enable deprecation warnings for a test later on.

lib/view_component/deprecation.rb
maxbeizer: TBH I could go either way. The original, typed indentation was as you suggest but vim's `V}=` moved it to this so ¯\_(ツ)_/¯. Happy to defer to @joelhawksley on that one

test/test_helper.rb
maxbeizer: Interesting idea. I'm still not sure we're going to go down the `ViewComponent::Deprecation` route but I'll think about it some more 💖

maxbeizer:
I got pulled into other things but I'm hoping to pick this back up soon. Sorry 🙇

PR #590, opened 2021-01-21
Fix broken link to slots v1 docs

DanceParty:
### Summary

Link to the original slots v1 api docs was a 404. Hopefully this fixes it 🍔


joelhawksley (approved):
That URL worked for me: https://viewcomponent.org/slots_v1.

Thanks for the fix! ❤️

Issue #580, opened 2021-01-12
Add validation that raises if a component's initializer accepts an argument named `content` 


BlakeWilliams:
Currently components can be defined with an initializer that accepts a `content` argument. In this case, we should raise to help prevent overwriting of the `content` accessor that provides access to the passed in content.

### Steps to reproduce
```ruby
class MyComponent < ViewComponent::Base
  def initialize(content:)
    @content = content
  end
end
```

### Expected behavior
It should raise to help prevent overwriting the `content` accessor

### Actual behavior
No exception is raised.

### System configuration

**Gem version**: 2.24.0

BlakeWilliams:
I've marked this as `help wanted`, so if anyone tackles this issue but needs assistance feel free to reach out.

PR #592, opened 2021-01-21
Validate component arguments

manuelpuyol:
### Summary

Closes https://github.com/github/view_component/issues/580

Adds a validation on compile time that checks if any blacklisted arguments are used on the component `initialize`. At the moment, the only invalid argument name is `content` but it can be expanded

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
   *Bob Maerten*
 
-* Add argument validation to avoid `content` override
+* Add argument validation to avoid `content` override.
 
     *Manuel Puyol*
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
   *Bob Maerten*
 
+* Add argument validation to avoid `content` override
+
+    *Manuel Puyol*
+
 ## 2.24.0
 
 * Add `--inline` option to the erb generator. Prevents default erb template from being created and creates a component with a call method.
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -304,9 +304,7 @@ module ViewComponent
           "public ViewComponent method."
         end
 
-        raise ArgumentError.new(
-
-        )
+        raise ArgumentError.new message
       end
 
       private
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -15,6 +15,8 @@ module ViewComponent
 
     ViewContextCalledBeforeRenderError = Class.new(StandardError)
 
+    BLACKLISTED_PARAMETERS = %i[content].freeze
+
     # For CSRF authenticity tokens in forms
     delegate :form_authenticity_token, :protect_against_forgery?, :config, to: :helpers
 
@@ -264,7 +266,7 @@ module ViewComponent
         parameter = validate_default ? collection_parameter : provided_collection_parameter
 
         return unless parameter
-        return if initialize_parameters.map(&:last).include?(parameter)
+        return if initialize_parameter_names.include?(parameter)
 
         # If Ruby cannot parse the component class, then the initalize
         # parameters will be empty and ViewComponent will not be able to render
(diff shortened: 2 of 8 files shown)


manuelpuyol:
Hey @BlakeWilliams, would love some opinions here!
Also, I'm not sure how to test this since the change only affects the compiler

lib/view_component/base.rb
```diff
@@ -281,8 +283,26 @@ def validate_collection_parameter!(validate_default: false)
...
+      # invalid parameters that could override the framework's
+      # methods.
+      def validate_initialization_parameters!
+        invalid_parameters = initialize_parameter_names & BLACKLISTED_PARAMETERS
+
+        return unless invalid_parameters.any?
+
+        raise ArgumentError.new(
+          "#{self} initializer must not have " \
+          "`#{invalid_parameters.join(", ")}` as parameters."
```
BlakeWilliams, line 296: Thoughts on elaborating here? Maybe something similar to `initializer cannot contain #{invalid_parameters.join(", ")} since they will override public ViewComponent  methods`?

BlakeWilliams (commented):
> Also, I'm not sure how to test this since the change only affects the compiler

Could we define a component like?

```ruby
class MyContentComponent < ViewComponent::Base
  def initialize(content)
  end
end
```

and have a test roughly like?

```ruby
error = assert_raises ArgumentError do
  MyContentComponent.compile(raise_errors: true)
end

assert_equal "the error message text", error.message
```

I think that *may* work. I believe we should also have at least 1 existing component defined in the test suite that should need to be updated to not use a `content` initializer.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
   *Bob Maerten*
 
+* Add argument validation to avoid `content` override
```
joelhawksley, line 9:
```suggestion
* Add argument validation to avoid `content` override.
```

lib/view_component/base.rb
```diff
@@ -15,6 +15,8 @@ class Base < ActionView::Base
 
     ViewContextCalledBeforeRenderError = Class.new(StandardError)
 
+    BLACKLISTED_PARAMETERS = %i[content].freeze
```
joelhawksley, line 18:
```suggestion
    DENYLISTED_PARAMETERS = %i[content].freeze
```

lib/view_component/base.rb
```diff
@@ -281,8 +283,27 @@ def validate_collection_parameter!(validate_default: false)
         )
       end
 
+      # Ensure the component initializer does not define
+      # invalid parameters that could override the framework's
+      # methods.
+      def validate_initialization_parameters!
+        invalid_parameters = initialize_parameter_names & BLACKLISTED_PARAMETERS
```
joelhawksley, line 290:
```suggestion
        invalid_parameters = initialize_parameter_names & DENYLISTED_PARAMETERS
```

test/app/components/invalid_parameters_component.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class InvalidParametersComponent < ViewComponent::Base
+  def initialize(content); end
```
joelhawksley, line 4: Do we want to test a named parameter as well?

lib/view_component/base.rb
```diff
@@ -281,8 +283,27 @@ def validate_collection_parameter!(validate_default: false)
...
+      # invalid parameters that could override the framework's
+      # methods.
+      def validate_initialization_parameters!
+        invalid_parameters = initialize_parameter_names & DENYLISTED_PARAMETERS
+
+        return unless invalid_parameters.any?
+
+        raise ArgumentError.new(
+          "#{self} initializer cannot contain " \
+          "`#{invalid_parameters.join(", ")}` since it will override " \
```
joelhawksley, line 296: Might we pluralize the `it` here if there are multiple parameters?

lib/view_component/base.rb
```diff
@@ -281,8 +283,27 @@ def validate_collection_parameter!(validate_default: false)
...
+          "public ViewComponent methods."
```
joelhawksley, line 297: Same re: `methods`

BlakeWilliams (approved):

Issue #571, opened 2021-01-04
render_many doesn't work right with Slim


dkniffin:
### Steps to reproduce

```
# app/components/foo_component.rb
class FooComponent < ViewComponent::Base
  include ViewComponent::SlotableV2

  renders_many :bars, BarComponent
end
```
```
# app/components/foo_component.html.slim
- bars.each do |bar|
    = bar
```
```
# app/components/foo_component.html.slim
class BarComponent < ViewComponent::Base
end
```
```
# app/components/foo_component.html.slim
= content
```
```
# app/views/whatever.html.slim
= render FooComponent.new do |c|
  - c.bar do
    p Something with HTML formatting
```

### Expected behavior
When the HTML is rendered, the text "Something with HTML formatting" should be formatted correctly.

### Actual behavior
The HTML is escaped and rendered in the page so the end-user sees the HTML tags instead of the formatting.

### System configuration
**Rails version**:
6.0.3.4

**Ruby version**:
2.7.2p137

**Gem version**:
2.24.0

joelhawksley:
@dkniffin thank you for the bug report! Would you be willing to open a PR with a failing test?

dkniffin:
@joelhawksley Yeah, I can try and do that tomorrow

dkniffin:
@manuelpuyol I think I tried that and it didn't work, but it's been a few weeks now. I started on a failing test but haven't gotten around to finishing it.

manuelpuyol:
@dkniffin @joelhawksley I think I've found the issue!

Looking at the render stacktrace here:

```
 "/Users/manuelpuyol/github/view_component/lib/view_component/slot_v2.rb:26:in `to_s'",
 "/Library/Ruby/Gems/2.6.0/gems/temple-0.8.2/lib/temple/utils.rb:38:in `escape_html'",
 "/Library/Ruby/Gems/2.6.0/gems/temple-0.8.2/lib/temple/utils.rb:21:in `escape_html_safe'",
 "/Users/manuelpuyol/github/view_component/test/app/components/slim_renders_many_component.html.slim:3:in `block in call'",
 "/Users/manuelpuyol/github/view_component/test/app/components/slim_renders_many_component.html.slim:1:in `each'",
 "/Users/manuelpuyol/github/view_component/test/app/components/slim_renders_many_component.html.slim:1:in `call'",
 "(eval):3:in `render_template_for'",
[...]

PR #594, opened 2021-01-22
Slim SlotV2 components are rendered as escaped HTML

manuelpuyol:
closes https://github.com/github/view_component/issues/571

### Summary

This PR adds a way to delegate html safety to the component's content instead of always being `false`. This will fix an issue where `temple` was escaping the component HTML since it was flagged as unsafe

### Benchmarks

No noticeable changes here

| Before | After |
|--------|--------|
| image  | image  | 
| image  |  image  |


lib/view_component/slot_v2.rb
```diff
@@ -22,8 +22,10 @@ def initialize(parent)
     # If there is no slot renderable, we evaluate the block passed to
     # the slot and return it.
     def to_s
+      return @content if defined?(@content)
```
BlakeWilliams, line 25: I think this makes this problem more complex, but I don't think we want to memoize here. I don't believe components memoize their output, so I'm a little hesitant about memoizing slots output too.

manuelpuyol: the reason I did this is because `html_safe?` has to call `to_s` and later on temple also calls `to_s` so I wanted to avoid the second "render"

BlakeWilliams: Yep, makes sense.

I've been thinking on it a bit this morning, and it's probably fine. I'm not sure we'll run into the use-case of 'rendering a slot once, changing some properties, then rendering it again' very often. So this is probably an acceptable trade-off. 👍

lib/view_component/slot_v2.rb
```diff
@@ -58,6 +62,10 @@ def method_missing(symbol, *args, &block)
       @_component_instance.public_send(symbol, *args, &block)
     end
 
+    def html_safe?
+      to_s.html_safe
```
BlakeWilliams, line 66:
```suggestion
      to_s.html_safe?
```

test/app/components/slim_renders_many_component.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class SlimRendersManyComponent < ViewComponent::Base
+  include ViewComponent::SlotableV2
+
+  renders_many :components, SlimComponent
```
BlakeWilliams, line 6:
```suggestion
  renders_many :slim_components, SlimComponent
```

Thoughts on this naming? It might help make it more explicit that we're rendering slim components as slots.

lib/view_component/slot_v2.rb
manuelpuyol: at some point I have to assume the HTML is safe, otherwise `html_safe?` will always be `false`.
What I can do is change the `to_s` capture to:

```rb
@content = view_context.capture do
        if defined?(@_component_instance)
          # render_in is faster than `parent.render`
          @_component_instance.render_in(view_context, &@_content_block)
        elsif defined?(@_content)
          @_content
          @_content_block.call
        end
      end.html_safe
```

and here use `to_s.html_safe?`, but this is basically the same as defining 

```rb
def html_safe?
  true
```

since we are always assuming safety :/

BlakeWilliams: > at some point I have to assume the HTML is safe, otherwise `html_safe?` will always be `false`.

I thought that slim/erb/haml would tell us if it's `html_safe` or not by us calling `to_s.html_safe?`? I don't think we'd have to mark it as safe anywhere explicitly in the library.

Would it be worth adding a test that outputs content we'd want to ensure is escaped?

 e.g. something roughly like

```ruby
class UserBioComponent < ViewComponent::Base
  renders_one :header, -> (user:) do
    user.bio # This is unsafe, since this could be something like `<script>alert('hello')</script>`
  end
end

# erb <%= header %>

# test
def test_slot_escapes_unsafe_content
  user = OpenStruct.new(bio: '<script>alert('hello')</script>')
  render_inline UserBioComponent.new { |component| component.header(user: user) }

  assert_text '&lt;script&gt;alert('hello')&lt;/script&gt;'
end
```

manuelpuyol: from what I see here, Rails always sets strings as unsafe no matter what. e.g:
```rb
"foo".html_safe? -> false
"<div>hello</div>".html_safe? -> false
```

If we want to escape unsafe parts we'd need something like [sanitize](https://github.com/rgrove/sanitize).

manuelpuyol: What I actually don't know is if the content safety is indeed a responsibility of the gem or of the user.
I see the `content` use something like React's [dangerouslySetInnerHTML](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) which infers that the user knows what is being rendered

BlakeWilliams: > from what I see here, Rails always sets strings as unsafe no matter what. e.g:

That is true, but in this case `@content` should be the result of the slim template rendering, so I'd expect it to be marked as `html_safe?` appropriately by slim itself.

> What I actually don't know is if the content safety is indeed a responsibility of the gem or of the user.

In the example above, I'd say it's on the framework (us) to ensure the consumer of the API can't easily create a vulnerability.

Similar to how Rails works, I'd expect a user to have to opt-in to rendering the user-input as-is with the `raw` helper method.

BlakeWilliams (approved):
LGTM!

patchwork-her (commented):
No, that did not fix it. What should be done is to leave end user completely out of the string all together! I say NO! i will go forwards to legal reasons if need be. This ends now!

halo:
And here I was adding `to_s` for several months without wondering 😆 Should have updated more often :)

Thank you very much for fixing this! 😍

PR #602, opened 2021-01-29
Docs: replace "test" and "my" with "example"

Spone:
### Summary

There are quite a few occurrences of `TestComponent` and `MyComponent` in the docs.

To avoid confusion with the word "test" in the context of testing the components (which is mentioned in the docs as well), I suggest we always use `ExampleComponent` instead.

It also makes some of the paths (such as `test/components/previews/test_component_preview.rb`) easier to understand when using "example": `test/components/previews/example_component_preview.rb`

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -100,17 +100,17 @@ bin/rails generate component Example title content --preview
 
 A ViewComponent is a Ruby file and corresponding template file with the same base name:
 
-`app/components/test_component.rb`:
+`app/components/example_component.rb`:
 
 ```ruby
-class TestComponent < ViewComponent::Base
+class ExampleComponent < ViewComponent::Base
   def initialize(title:)
     @title = title
   end
 end
 ```
 
-`app/components/test_component.html.erb`:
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -544,10 +544,10 @@ _To assert that a component has not been rendered, use `refute_component_rendere
 
 Components can define a `before_render` method to be called before a component is rendered, when `helpers` is able to be used:
 
-`app/components/confirm_email_component.rb`
```
Spone, line 547: Here the file name was inconsistent with the class name below. I changed both to `example`.

docs/index.md
```diff
@@ -698,9 +698,9 @@ Capybara matchers are available if the gem is installed:
 ```ruby
 require "view_component/test_case"
 
-class MyComponentTest < ViewComponent::TestCase
```
Spone, line 701: Here the class name was inconsistent with the component name below. I changed both to `example`.

joelhawksley (approved):
Thank you for the thoughtful PR! I like it.

PR #603, opened 2021-01-29
Call `.after_compile` class method after a component is compiled.

joelhawksley:
### Summary

This PR introduces the `after_compile` hook, allowing for components to define logic that should be executed when the component is compiled.

### Other Information

The compilation step is a critical part of the ViewComponent architecture that has previously been more or less invisible to the developer. I'm hoping that this hook is the first step towards allowing developers to leverage this powerful feature in new and exciting ways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## main
 
-* Call `.after_compile` class method after a component is compiled.
+* Experimental: call `._after_compile` class method after a component is compiled.
 
     *Joel Hawksley*
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Call `.after_compile` class method after a component is compiled.
+
+    *Joel Hawksley*
+
 * Fix bug where SlotV2 was rendered as an HTML string when using Slim.
 
     *Manuel Puyol*
--- a/docs/index.md
+++ b/docs/index.md
@@ -554,6 +554,12 @@ class ExampleComponent < ViewComponent::Base
 end
 ```
 
+### `.after_compile`
+
+ViewComponents can define an `after_compile` class method that will be called after the component is compiled.
+
+`.after_compile` is called at request time if `Rails.application.config.eager_load` is `false`, or at application boot if it set to `true`.
+
 ### Rendering collections
 
 Use `with_collection` to render a ViewComponent with a collection:
--- a/docs/index.md
+++ b/docs/index.md
@@ -540,9 +540,9 @@ end
 
 _To assert that a component has not been rendered, use `refute_component_rendered` from `ViewComponent::TestHelpers`._
 
-### `before_render`
+### `#before_render`
 
-Components can define a `before_render` method to be called before a component is rendered, when `helpers` is able to be used:
+ViewComponents can define a `before_render` method to be called before a component is rendered, when `helpers` is able to be used:
 
 `app/components/example_component.rb`
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -554,12 +554,6 @@ class ExampleComponent < ViewComponent::Base
 end
 ```
 
-### `.after_compile`
-
-ViewComponents can define an `after_compile` class method that will be called after the component is compiled.
-
-`.after_compile` is called at request time if `Rails.application.config.eager_load` is `false`, or at application boot if set to `true`.
-
 ### Rendering collections
 
 Use `with_collection` to render a ViewComponent with a collection:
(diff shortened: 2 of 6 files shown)


manuelpuyol (approved):

lib/view_component/base.rb
```diff
@@ -23,6 +23,13 @@ class Base < ActionView::Base
     class_attribute :content_areas
     self.content_areas = [] # class_attribute:default doesn't work until Rails 5.2
 
+    # Hook for allowing components to do work as part of the compilation process.
+    #
+    # For example, one might compile component-specific assets at this point.
+    def self.after_compile
```
BlakeWilliams, line 29: I could also see this being nice as a DSL (similar to `after_filter`, `after_save`, etc), maybe something roughly like:

```ruby
  after_compile :load_i18n
  
  private

  def load_i18n
    # find i18n files
  end
end
```

To avoid having to remember to call `super` each time you add more behavior, and the potential ability to opt-out of specific hooks like how `skip_before_action` works.

BlakeWilliams (approved):
I had one non-blocking thoughts, but LGTM! I'm excited to see what all we can do with compiler hooks ✨

lib/view_component/base.rb
BlakeWilliams, line 29:
```suggestion
    # EXPERIMENTAL: This API is experimental and may be removed at any time.
    # Hook for allowing components to do work as part of the compilation process.
    #
    # For example, one might compile component-specific assets at this point.
    def self._after_compile
```

docs/index.md
```diff
@@ -554,6 +554,12 @@ class ExampleComponent < ViewComponent::Base
 end
 ```
 
+### `.after_compile`
+
+ViewComponents can define an `after_compile` class method that will be called after the component is compiled.
+
+`.after_compile` is called at request time if `Rails.application.config.eager_load` is `false`, or at application boot if set to `true`.
+
```
BlakeWilliams, line 562:
```suggestion
```

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* Call `.after_compile` class method after a component is compiled.
```
```suggestion
* Experimental: call `._after_compile` class method after a component is compiled.
```

lib/view_component/compiler.rb
```diff
@@ -67,6 +67,8 @@ def #{method_name}
 
       define_render_template_for
 
+      component_class.after_compile
```
joelhawksley, line 70:
```suggestion
      component_class._after_compile
```

test/app/components/after_compile_component.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class AfterCompileComponent < ViewComponent::Base
+  @@compiled_value = ""
+
+  def self.after_compile
```
joelhawksley, line 6:
```suggestion
  def self._after_compile
```

joelmoss:
Would be good to get a little more detail on what the compilation actually does and why. Would help us understand its capabilities and scope.

Also, any way we could expose the view context to this hook? One way I want to use it is to sidecar JS/CSS by simply calling rails `javascript_include_tag`, etc. helpers. But seems that is not currently possible?

joelhawksley:
@joelmoss I'm not sure I follow. This callback is called as part of the compilation process, so it does not have access to the view context. 

Can you be more specific about what you're trying to accomplish with `javascript_include_tag`?

joelmoss:
Hey Joel (great name by the way 😜),

I simply want to side load any CSS or JS file alongside the component when it is rendered, by simply adding a `javascript_include_tag` or `stylesheet_link_tag` into a `content_for` block, that is then yielded into my layout.

But as a component could be used multiple times on the same page, it should only include its assets the once. The `before_render` hook gets called each time a component renders, so that is no good really. I figured compilation only happens the once, so the new hook seems like a better option.

However...

I have since been able to make it work with the `before_render` hook, thanks to Rails de-duping the sources it receives in the above helpers.

However... again... 😏

Using `before_render` gets a little messy, as I define that in my `ApplicationComponent` so that all components can sideload their assets if any. But that means that if a component needs to use the `before_render` hook, I have to be sure to call `super`.

Thinking that implementing the hook using `ActiveSupport::Callback` would be better for that reason.

Just thinking out loud now, but thx 👍

joelhawksley:
@joelmoss thanks for providing more context! We're looking into similar issues in https://github.com/github/view_component/pull/677.

I could see us using callbacks for this problem, but I'd want to make sure we benchmark any changes we make in that direction as the throughput of rendering ViewComponents is significant compared to other existing use cases of callbacks.

PR #604, opened 2021-01-30
Docs: escape dot in regexp + use camelCase in JS

Spone:
### Summary

Dot should be escaped in regular expressions, else it matches any character.

Also, variables are usually in camelCase.


jonspalmer (approved):
I like it.

PR #609, opened 2021-02-04
add note about assert_selector visibility

joelhawksley:
This PR adds a small note about using `assert_selector` with hidden elements, as its behavior has bit a few developers enough to warrant a warning here.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -709,6 +709,8 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+_Note: `assert_selector` only matches on visible elements by default. To match on hidden elements, add `visible: false`._
+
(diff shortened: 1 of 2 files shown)


jonspalmer (approved):
LGTM

docs/index.md
```diff
@@ -709,6 +709,8 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+_Note: `assert_selector` only matches on visible elements by default. To match on hidden elements, add `visible: false`._
```
BlakeWilliams, line 712: Would it be worth linking to the capybara docs as well?

BlakeWilliams (approved):
One small question, but LGTM

PR #611, opened 2021-02-08
Add docs for how to add validations

dkniffin:
I noticed that ViewComponent no longer has validation support built-in, but [it's easily added](https://github.com/github/view_component/issues/206#issuecomment-602715364). I didn't see this in the documentation, so I added it.

Also, side-note: The docs are very difficult to read currently. I would recommend using something like [Gitbook](https://www.gitbook.com/) or similar to make them easier to navigate

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -421,6 +421,7 @@ class ExampleComponent < ViewComponent::Base
   end
 end
 ```
+
 _Note: Using validations in this manner can lead to runtime exceptions. Use them wisely._
 
 ### Template Inheritance
--- a/docs/index.md
+++ b/docs/index.md
@@ -405,6 +405,24 @@ And render them `with_variant`:
 
 _**Note**: `call_*` methods must be public._
 
+### Validations
+
+ViewComponent does not include built-in validation capabilities. However, this functionality
+can be easily added by using `ActiveModel::Validations`:
+
+```ruby
+class ExampleComponent < ViewComponent::Base
+  include ActiveModel::Validations
+
+  # Requires that a content block be passed to the component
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -405,6 +405,24 @@ And render them `with_variant`:
 
 _**Note**: `call_*` methods must be public._
 
+### Validations
+
+ViewComponent does not include built-in validation capabilities. However, this functionality
+can be easily added by using `ActiveModel::Validations`:
```
joelhawksley, line 411:
```suggestion
ViewComponent does not include support for validations. However, it can be added by using `ActiveModel::Validations`:
```

docs/index.md
```diff
@@ -405,6 +405,24 @@ And render them `with_variant`:
...
+
+  # Requires that a content block be passed to the component
+  validate :content, presence: true
+
+  def before_render
+    validate!
+  end
+end
+```
+
```
joelhawksley, line 425:
```suggestion
_Note: Using validations in this manner can lead to runtime exceptions. Use them wisely._

```

joelhawksley (approved):

docs/index.md
```diff
@@ -405,6 +405,24 @@ And render them `with_variant`:
...
+  # Requires that a content block be passed to the component
+  validate :content, presence: true
+
+  def before_render
+    validate!
+  end
+end
+```
+_Note: Using validations in this manner can lead to runtime exceptions. Use them wisely._
```
joelhawksley, line 424:
```suggestion

_Note: Using validations in this manner can lead to runtime exceptions. Use them wisely._
```

PR #624, opened 2021-02-15
Fix sample code for Delegate slots

### LINKED ISSUE #622: SlotableV2 docs example for Delegate Slots throws errors
The example code given in https://viewcomponent.org/#delegate-slots appears not to be working out of the box and raises two different errors:

- `missing keyword: :title`
- `Could not find a template file or inline render method for BlogComponent::HeaderComponent`

### Steps to reproduce

Here’s a simplified version of the example code focusing only on `HeaderComponent`:

`# blog_component.rb`

```ruby

  # Since `HeaderComponent` is nested inside of this component, we have to
  # reference it as a string instead of a class name.
  renders_one :header, "HeaderComponent"

  class HeaderComponent < ViewComponent::Base
    attr_reader :title

    def initialize(title:)
      @title = title
    end
  end
end
```

`# blog_component.html.erb`

```ruby
<div>
  <h1><%= header %></h1> <!-- render the header component -->
</div>
```

`# index.html.erb `

```ruby
<%= render BlogComponent.new do |c| %>
  <% c.header do %>
    <%= link_to "My Site", root_path %>
  <% end %>
<% end %>
```

### Expected behavior

The example code in the docs should not raise errors.

### Actual behavior

Running the above snippets first raises a `missing keyword: :title` title, which can be solved with the following change:

```diff
<%= render BlogComponent.new do |c| %>
-  <% c.header do %>
+  <% c.header(title: "My Title") do %>
    <%= link_to "My Site", root_path %>
  <% end %>
<% end %>
```

First error is gone, but we now get to another one: `Could not find a template file or inline render method for BlogComponent::HeaderComponent`

I was able to work around _that one_ with the following change:

```diff
class BlogComponent < ViewComponent::Base
  include ViewComponent::SlotableV2

  # Since `HeaderComponent` is nested inside of this component, we have to
  # reference it as a string instead of a class name.
  renders_one :header, "HeaderComponent"

  class HeaderComponent < ViewComponent::Base
    attr_reader :title

    def initialize(title:)
      @title = title
    end
+ 
+  def call
+    content
+  end
  end
end
```

### System configuration
**Rails version**: `6.0.3.4`

**Ruby version**: `2.7.2`

**Gem version**: `2.25.1`

### Suggestions

For the first error, I’m happy to create a docs PR to update that example to remove that error and actually use the parameter we pass to the component, e.g.:

```ruby

  # Since `HeaderComponent` is nested inside of this component, we have to
  # reference it as a string instead of a class name.
  renders_one :header, "HeaderComponent"

  class HeaderComponent < ViewComponent::Base
    attr_reader :classes

    def initialize(classes:)
      @classes = classes
    end
  end
end
```

```ruby
<div>
  <h1 "<%= @classes %>"><%= header %></h1> <!-- render the header component -->
</div>
```

For the second error, I’m not sure if the addition of

```ruby
def call
  content
end
```

would be considered good practice or an antipattern. Would appreciate your opinions on this, and I’d be happy to add that to the PR.

phacks:
### Summary

Fixes #622 (the code samples for that section of the docs raised errors)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -220,7 +220,7 @@ class BlogComponent < ViewComponent::Base
     end
     
     def call
-      content_tag :h1, content, {class: classes}
+      content_tag :h1, content, { class: classes }
     end
   end
 end
@@ -269,7 +269,7 @@ class BlogComponent < ViewComponent::Base
   # Renders the returned string
   renders_one :header, -> (classes:) do
     content_tag :h1 do
-      link_to title, root_path, {class: classes}
+      link_to title, root_path, { class: classes }
     end
   end
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -213,10 +213,14 @@ class BlogComponent < ViewComponent::Base
   renders_many :posts, PostComponent
 
   class HeaderComponent < ViewComponent::Base
-    attr_reader :title
+    attr_reader :classes
 
-    def initialize(title:)
-      @title = title
+    def initialize(classes:)
+      @classes = classes
+    end
+    
+    def call
+      content_tag :h1, content, {class: classes}
     end
   end
 end
@@ -226,7 +230,7 @@ end
 
 ```erb
 <div>
-  <h1><%= header %></h1> <!-- render the header component -->
+  <%= header %> <!-- render the header component -->
 
   <% posts.each do |post| %>
     <div class="blog-post-wrapper">
@@ -240,7 +244,7 @@ end
 
 ```erb
 <%= render BlogComponent.new do |c| %>
-  <% c.header do %>
+  <% c.header(classes: "") do %>
     <%= link_to "My Site", root_path %>
   <% end %>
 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -213,10 +213,14 @@ class BlogComponent < ViewComponent::Base
...
+    attr_reader :classes
 
-    def initialize(title:)
-      @title = title
+    def initialize(classes:)
+      @classes = classes
+    end
+    
+    def call
+      content_tag :h1, content, {class: classes}
```
BlakeWilliams, line 223:
```suggestion
      content_tag :h1, content, { class: classes }
```

docs/index.md
```diff
@@ -263,9 +267,9 @@ class BlogComponent < ViewComponent::Base
   include ViewComponent::SlotableV2
 
   # Renders the returned string
-  renders_one :header, -> (title:) do
+  renders_one :header, -> (classes:) do
     content_tag :h1 do
-      link_to title, root_path
+      link_to title, root_path, {class: classes}
```
BlakeWilliams, line 272:
```suggestion
      link_to title, root_path, { class: classes }
```

BlakeWilliams (approved):

BlakeWilliams:

PR #628, opened 2021-02-15
Fix destination problem when generating an ERB component with `--inline` flag

### LINKED ISSUE #626: Destination problem when generating an ERB component with `--inline` flag
### Steps to reproduce

```
bin/rails g component Example --template-engine erb --inline
Running via Spring preloader in process 15365
      create  app/components/example_component.rb
      invoke  test_unit
      create    test/components/example_component_test.rb
      invoke  erb
      create    component.html.erb
```

Related to #579 

### Expected behavior

No `.erb` file should be created, as the inline component do not need one.

### Actual behavior

A `component.html.erb` file is generated at the root of the project.

### System configuration
**Rails version**: 5.2

**Ruby version**: 2.7.2

**Gem version**: 2.25.1

Spone:
Close #626

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Fix template generated at the project root when generating an ERB component with the `--inline` flag
+
+    *Hans Lemuet*
+
 ## 2.25.1
 
 * Experimental: call `._after_compile` class method after a component is compiled.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## main
 
-* Fix template generated at the project root when generating an ERB component with the `--inline` flag
+* Do not generate template when using `--inline` flag.
 
     *Hans Lemuet*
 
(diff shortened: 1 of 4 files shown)


test/generators/component_generator_test.rb
```diff
@@ -57,6 +57,7 @@ def test_component_with_inline
     end
 
     assert_no_file "app/components/user_component.html.erb"
+    assert_no_file "component.html.erb"
```
Spone, line 60: I added this test to check that the bug was fixed, but I'm not sure it makes sense to keep it in the codebase...

test/generators/erb_generator_test.rb
```diff
@@ -40,4 +40,10 @@ def test_component_with_namespace_and_sidecar
 
     assert_file "app/components/admins/user_component/user_component.html.erb"
   end
+
+  def test_component_with_inline
```
Spone, line 44: I think it more logical to have this test here instead of inside `ComponentGeneratorTest`.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* Fix template generated at the project root when generating an ERB component with the `--inline` flag
```
```suggestion
* Do not generate template when using `--inline` flag.
```

joelhawksley (approved):

CHANGELOG.md
Spone: Just so you know, I explained it like this in the changelog because it's actually a fix of the `--inline` flag feature introduced in 1.24.0, not a new feature.

PR #633, opened 2021-02-18
Update index.md

compostbrain:
Why:

* the generators as written adds a 'content' parameter to the intialize method for the class which causes the instantiation of the class to fail.

How:

* Remove 'content' from the generator command examples


joelhawksley (approved):

joelhawksley:
Thanks for the PR!

PR #635, opened 2021-02-19
docs(adr): add 001 Record Architecture Decisions

keithamus:
### Summary

This adds a basic document that describes how we might make architectural decisions, and record such decisions for future reference.

### DIFF
--- a/adr/0001-record-architecture-decisions.md
+++ b/adr/0001-record-architecture-decisions.md
@@ -1,0 +1,19 @@
+# 1. Record architecture decisions
+
+Date: 2021-02-19
+
+## Status
+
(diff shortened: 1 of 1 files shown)


joelhawksley (approved):
Fantastic! Thank you for the PR 🙇🏻

Spone:
I'm late to the party, but may I suggest to put this in `docs/architecture-decisions` instead of `adr` at the root, to keep all documentation in the same place?

joelhawksley:
@Spone good idea! I think it might be worth thinking about how we might incorporate ADRs into the documentation, perhaps once we have one beyond this establishing document?

PR #636, opened 2021-02-19
Add breaking test with default args slot call

manuelpuyol:
Related to https://github.com/github/view_component/issues/637

### Summary

This PR adds a failing test for SlotsV2 calls without providing any args (expected to use defaults). It also adds a test showing the slot being rendered if arguments are provided

### DIFF
--- a/test/view_component/slotable_v2_test.rb
+++ b/test/view_component/slotable_v2_test.rb
@@ -231,4 +231,14 @@ class SlotsV2sTest < ViewComponent::TestCase
     assert_selector(".greeting", text: "Hello, John Doe")
     assert_selector(".greeting", text: "Hello, Jane Doe")
   end
+
+  def test_renders_slots_with_default_args
+    render_inline(SlotsV2WithDefaultArgsComponent.new) do |component|
+      component.icon
+    end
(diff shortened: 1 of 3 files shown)


test/view_component/slotable_v2_test.rb
```diff
@@ -231,4 +231,24 @@ def test_renders_slots_with_before_render_hook
     assert_selector(".greeting", text: "Hello, John Doe")
     assert_selector(".greeting", text: "Hello, Jane Doe")
   end
+
+  def test_renders_slots_with_default_args
+    render_inline(SlotsV2WithDefaultArgsComponent.new) do |component|
+      component.icon
```
BlakeWilliams, line 237: I think this is failing because this isn't setting the value of `component.icon`, but accessing it.

To set a slot, you have to pass an argument or a block.

joelhawksley:
@manuelpuyol this is an interesting proposed API. I'm a little cautious of having a macro that calls itself simplicity if it hasn't been called explicitly. I can imagine this being an unexpected behavior. Might we consider a way to opt-in to this instead of it being the default?

PR #638, opened 2021-02-19
Don't call compile in collections

BlakeWilliams:
This removes a call to `ViewComponent::Base#compile` from
`ViewComponent::Collection` since components should already be compiled
when their `render_in` method is called.

This saves us a decent amount of time when rendering a large number of
components.


BlakeWilliams:
In the example below, ~11k components were rendered.
image

joelhawksley (approved):
👍🏻 looks good to me! If it turns out this is the incorrect change, we need more tests 😄

PR #650, opened 2021-02-26
Make the components.js example work with Sidecar directories

elia:
### Summary

The original code didn't work well with sidecar components.


joelhawksley (approved):
👋🏻 thanks for the PR! I sent you an email the other day, not sure if you got it. Mind emailing me at joelhawksley@github.com?

PR #654, opened 2021-03-02
add docs note about nested URL helpers

joelhawksley:
### Summary

This PR adds a note about why nested URL helpers need their options to be provided explicitly, a source of confusion internally ❤️

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -713,6 +713,18 @@ class UserComponent < ViewComponent::Base
 end
 ```
 
+#### Using nested URL helpers
+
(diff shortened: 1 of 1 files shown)


BlakeWilliams (approved):
Thoughts on defining the relevant `config/routes.rb` line as well?

```ruby
resources :users, only: :edit
```

joelhawksley:
> Thoughts on defining the relevant `config/routes.rb` line as well?
> 
> ```ruby
> resources :users, only: :edit
> ```

I was hoping to avoid making this example more complex, as the `bad` case only works when on `users/show` 😬 (or another path with a `user_id` segment)

PR #659, opened 2021-03-08
update components to match zeitwerk expectation

manuelpuyol:
Updated the component names according to zeitwerk errors. Checked it locally using `zeitwerk:check`
This was breaking builds running against rails main


BlakeWilliams (approved):
:shipit: !

PR #660, opened 2021-03-08
Isolated I18n

elia:
### Summary

A proof of concept for https://github.com/github/view_component/pull/552#discussion_r586829617 with I18n translations that are only accessible from within the component. 

There's still some work to do, but if the general idea is ok this PR can replace #552. 
One thing to note is that this code has not been tried on any real-world production app.

### Other Information

```
              global     6.194k i/100ms
      global_missing     2.233k i/100ms
    sidecar_absolute     8.213k i/100ms
             sidecar     7.650k i/100ms
     sidecar_missing     1.459k i/100ms
    sidecar_fallback     2.572k i/100ms
              global     62.470k (± 1.7%) i/s -    315.894k in   5.058230s
      global_missing     22.641k (± 2.2%) i/s -    113.883k in   5.032216s
    sidecar_absolute     82.636k (± 2.0%) i/s -    418.863k in   5.070853s
             sidecar     76.282k (± 2.6%) i/s -    382.500k in   5.017815s
     sidecar_missing     14.882k (± 2.2%) i/s -     74.409k in   5.002379s
    sidecar_fallback     27.377k (± 3.2%) i/s -    138.888k in   5.078596s

Comparison:
    sidecar_absolute:    82636.3 i/s
             sidecar:    76281.6 i/s - 1.08x  (± 0.00) slower
              global:    62469.6 i/s - 1.32x  (± 0.00) slower
    sidecar_fallback:    27377.3 i/s - 3.02x  (± 0.00) slower
      global_missing:    22641.5 i/s - 3.65x  (± 0.00) slower
     sidecar_missing:    14882.2 i/s - 5.55x  (± 0.00) slower
```


lib/view_component/isolated_i18n.rb
```diff
@@ -0,0 +1,44 @@
...
+    result = catch(:exception) do
+      if key.is_a?(Array)
+        key.map { |k| backend.translate(locale, k, options) }
+      else
+        backend.translate(locale, key, options)
+      end
+    end
+
+    if result.is_a?(I18n::MissingTranslation)
+      handle_exception(false, result, locale, key, options)
```
BlakeWilliams, line 25: I wonder if we could fall-back to global translations for this case so you don't have to keep track of "global" vs "local" translations? What are your thoughts?

lib/view_component/isolated_i18n.rb
```diff
@@ -0,0 +1,44 @@
...
+    else
+      result
+    end
+  end
+  alias :t :translate
+
+  module ClassMethods
+    def compile(*,**)
```
BlakeWilliams, line 33: It's currently private and experimental, but I think `self._after_compile` is another potential place to hook into the compile process. https://github.com/github/view_component/blob/main/lib/view_component/compiler.rb#L48

BlakeWilliams (commented):
I had a few comments/questions but so far this is looking great!

Thank you for adding a benchmark too!

lib/view_component/isolated_i18n.rb
elia: We could even go with the [Chain](https://github.com/ruby-i18n/i18n/blob/master/lib/i18n/backend/chain.rb) backend and avoid the throw/catch game altogether, but I thought that there's some value in strictly using the sidecar translations.

No strong feelings though, with regards to this I'm also open to falling back when a missing-translation is thrown, or the chain backend, or even a "switch" parameter like `fallback_to_global: true` etc.

lib/view_component/isolated_i18n.rb
```diff
@@ -0,0 +1,44 @@
...
+      else
+        backend.translate(locale, key, options)
+      end
+    end
+
+    if result.is_a?(I18n::MissingTranslation)
+      handle_exception(false, result, locale, key, options)
+    else
+      result
+    end
```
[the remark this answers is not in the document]
elia: Not really, I18n for some reason (could it be performance?) uses `throw` and `catch` instead of `raise` and `rescue`. The code in this method is copied almost verbatim from I18n, I just removed a things that could be safely cut.

lib/view_component/isolated_i18n.rb
elia: I saw that (also the experimental warning) but finally decided that using it would put the burden of calling super on anyone that wanted to use it in their components. Also being `IsolatedI18n` part of ViewComponent seemed ok to hook into a core method.

The assumption is of course that this is meant to be used in the final app, probably inside the `ApplicationComponent`, if that is not the case I'm happy to move there the implementation.

elia (commented):
Just a few comments to address concerns and questions (still didn't had the time to finish up the PR, fix specs, remove debug statements etc.)

docs/index.md
```diff
@@ -1179,6 +1179,47 @@ app/components
 </div>
 ```
 
+##### Translations
```
BlakeWilliams, line 1182: Can we mark this as experimental similar to how we've marked V2 as experimental?

The goal is to include this for V3 but don't want to commit to a specific API/behavior until then.

lib/view_component/base.rb
```diff
@@ -205,6 +205,44 @@ def content_evaluated?
     class << self
       attr_accessor :source_location, :virtual_path
 
+      # Find sidecar files for the given extensions.
+      #
+      # The provided array of extensions is expected to contain
+      # strings starting without the "dot", example: `["erb", "haml"]`
+      def sidecar_files(extensions)
```
BlakeWilliams, line 212: I'm not sure that we want to make this part of the public API just yet. Thoughts on prefixing this with an `_` and making it private, or moving it back to the compiler?

lib/view_component/compiler.rb
```diff
@@ -125,39 +125,8 @@ def templates
...
-        []
-      end
-
-      # view files in the same directory as the component
-      sidecar_files = Dir["#{directory}/#{component_name}.*{#{extensions}}"]
-
-      sidecar_directory_files = Dir["#{directory}/#{component_name}/#{filename}.*{#{extensions}}"]
-
-      (sidecar_files - [source_location] + sidecar_directory_files + nested_component_files)
+    def template_paths
```
BlakeWilliams, line 128: Thoughts on inlining this call in `#templates` since it looks like that's the only place it's used?

lib/view_component/sidecar_i18n.rb
```diff
@@ -0,0 +1,48 @@
...
+          key.map { |k| i18n_backend.translate(locale, k, options) }
+        else
+          i18n_backend.translate(locale, key, options)
+        end
+      end
+    end
+    alias :t :translate
+
+    module ClassMethods
+      def compile(*, **)
```
BlakeWilliams, line 39: Can we migrate this to use `_after_compile`?

I think it's fine to require the `super` call on it for now since there will be at least a few consumers of that API now.

test/app/components/sidecar_i18n_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<p class="isolated"><%= t("isolated_translation", default: "This won't be used") %></p>
+<p class="missing"><%= t("translations_component.title", default: "Not the original value") %></p>
+<p class="helpers"><%= helpers.t("translations_component.title", default: "Not the original value") %></p>
+<p class="global"><%= ::I18n.t("translations_component.title", default: "Not the original value") %></p>
```
BlakeWilliams, line 4: This was a little confusing since I didn't know we had an existing translations file that had a component namespace. Thoughts on using some other namespace that's more general?

BlakeWilliams (commented):
I had a few small questions/comments, but looking good!

docs/index.md
```diff
@@ -1179,6 +1179,47 @@ app/components
...
+
+In order to support `I18n` translations scoped to the component you can include `ViewComponent::SidecarI18n`
+into your component:
```
joelhawksley, line 1185:
```suggestion
To use experimental support `I18n` translations, include `ViewComponent::SidecarI18n`:
```

docs/index.md
```diff
@@ -1179,6 +1179,47 @@ app/components
...
+
+`app/components/example_component.rb`
+
+```ruby
+module ExampleComponent < ApplicationComponent
+  include ViewComponent::SidecarI18n
+end
+```
+
+This allows for adding a "sidecar" YAML file that will define translations for the component:
```
joelhawksley, line 1195:
```suggestion
Then add a sidecar YAML file with translations for the component:
```

docs/index.md
```diff
@@ -1179,6 +1179,47 @@ app/components
...
+The translations will then be available as usual:
+
+`app/components/example_component.html.erb`
+
+```erb
+<p><%= t("hello") %></p>
+```
+
+It's important to notice that if the component needs to access global translations
+that can be done explicitly mentioning `helper` or `I18n`:
```
joelhawksley, line 1214:
```suggestion
To access global translations, explicitly mention `helper` or `I18n`:
```

That being said, I don't know if this is ideal. I feel like we shouldn't have to require folks to use two different APIs here.

docs/index.md
```diff
@@ -1179,6 +1179,47 @@ app/components
...
+end
+```
+
+This allows for adding a "sidecar" YAML file that will define translations for the component:
+
+`app/components/example_component.yml`
+
+```yml
+en:
+  # This key will be expanded to `en.example_component.hello`
```
joelhawksley, line 1201: What happens if there are multiple `example_component`s? Perhaps we need to hash this identifier?

joelhawksley (commented):
Some docs feedback. Thanks again for your hard work on this feature 😄

test/app/components/sidecar_i18n_component.html.erb
elia: I cleaned it up a bit, hope is more understandable now.

elia (commented):
I set both APIs as experimental, I think `_sidecar_files` might come useful for other plugins as well, as it encapsulates what is considered a sidecar file by ViewComponent.

docs/index.md
```diff
@@ -1128,6 +1128,44 @@ app/components
 </div>
 ```
 
+##### Translations (experimental)
+
+To use experimental support `I18n` translations, include `ViewComponent::SidecarI18n`:
```
joelhawksley, line 1133:
```suggestion
To use experimental support for `I18n` translations, include `ViewComponent::SidecarI18n`:
```

docs/index.md
```diff
@@ -1128,6 +1128,44 @@ app/components
...
+
+`app/components/example_component.rb`
+
+```ruby
+module ExampleComponent < ApplicationComponent
+  include ViewComponent::SidecarI18n
+end
+```
+
+Then add a sidecar YAML file with translations for the component:
```
joelhawksley, line 1143:
```suggestion
Add a sidecar YAML file with translations for the component:
```

docs/index.md
```diff
@@ -1128,6 +1128,44 @@ app/components
...
+
+`app/components/example_component.yml`
+
+```yml
+en:
+  hello: "Hello world!"
+```
+
+The translations will then be available as usual:
```
joelhawksley, line 1152:
```suggestion
Access translations as usual:
```

docs/index.md
```diff
@@ -1128,6 +1128,44 @@ app/components
...
+
+`app/components/example_component.html.erb`
+
+```erb
+<p><%= t("hello") %></p>
+```
+
+To access global translations, explicitly mention `helper` or `I18n`:
```
joelhawksley, line 1160: Would it be possible to have `t` also access global translations? I think it's a little awkward to re-define `t` with a local scope and then expect consumers to know when to use the `helpers` prefix.

elia (commented):
@joelhawksley the fallback to the global I18n has been added, I also updated the benchmark to reflect the different scenarios, and a note in the docs to explain how to reach translation keys shadowed by the sidecar file.

Spone:
This is looking great @elia! :raised_hands: 

I'm a bit late to the party, but what about using a dot as a way to disambiguate between local and global I18n?

`app/components/example_component.yml`

```yml
en:
  hello: "Hello world!"
```

To access the local translation:

`app/components/example_component.html.erb`

```erb
<p><%= t(".hello") %></p>
```

To access a global translation

```erb
<%= t("my.global.translation") %>
```

So if a global translation is shadowed by the sidecar file you just have to drop the dot:

`app/components/example_component.html.erb`

```erb
<%= t("hello") %>
```

Seems very Railsy to me but maybe I'm missing something?

docs/index.md
```diff
@@ -1128,6 +1128,51 @@ app/components
...
+```
+
+Global Rails translations are available as well:
+
+```erb
+<%= t("my.global.translation") %>
+```
+
+If you have Rails translations shadowed by the sidecar file you can access
+them using `helpers` or `I18n`:
```
joelhawksley, line 1167:
```suggestion
Global translations shadowed by sidecar translations can be accessed with `helpers` or `I18n`:
```

joelhawksley (commented):
Just leaving a small docs tweak here. I think @spone might be on to something with the local syntax proposal 🤔

elia:
@Spone @joelhawksley that's surely feasible, although I have some concerns about that:

1. introducing more complexity, either by copying some code from the Rails own [`TranslationHelper`](https://github.com/rails/rails/blob/5aaaa1630ae9a71b3c3ecc4dc46074d678c08d67/actionview/lib/action_view/helpers/translation_helper.rb#L132-L144), the current implementation is super simple, if we want to have it to mimic Rails maybe we should just use Rails and move back to something more like #552
2. `t(".my-key")` already has a meaning in Rails, I'm not too concerned, since I don't expect it to be a very popular feature, but we should probably add instructions on how to get the original behavior
3. how likely it is to have a key shadowed by the component's translations? _my purely speculative guess: not very often_

I don't necessarily oppose doing that, but I lean more toward keeping things simple, adding features later, if required.
What are your thoughts?

---

_One of the first things I tried was to include the original `TranslationHelper` and have it point to the components I18n. That is actually easier said than done, the helper is full of `I18n` references and the I18n library itself was built to be global._

Spone:
> I don't necessarily oppose doing that, but I lean more toward keeping things simple, adding features later, if required.

I totally agree. Let's just keep in mind that changing how locale lookup works in a future version of `view_component` could be a massive breaking change, and migration would be painful in some cases...

For the sake of the argument, I'm taking the point of view of a developer using `view_component` here, regardless of the internal implementation (whose complexity has obviously to be considered).

### `t` helper lookup behavior

As stated in the docs, let's follow the principle of least surprise. If I read the code of a project using `view_component`, and I see this anywhere:

```rb
t("hello")
```

knowing Rails conventions on [translations lookup](https://guides.rubyonrails.org/i18n.html#looking-up-translations), I would expect it to reference the `hello` key at the root of the global locale file. It would be surprising that this code looks up at the root when in a controller or view, but in the component sidecar locale file if it's in a component class or template.

If I see this:

```rb
t(".hello")
```

I know that a ["lazy" lookup](https://guides.rubyonrails.org/i18n.html#lazy-lookup) is used, so I'll look a the file structure to figure out what the path to the translation key is. Lazy lookup, that can be used in controllers and views, could be likened to a "local lookup".
In the current version of `view_component`, calling `t(".hello")` from `ExampleComponent` will look up `en.example_component.hello`, so it can actually also be used in components already (but without strict isolation).

### Switch between global and local translations

In a component, seeing both `helpers.t`/`I18n.t` (for global translations) and `t` (for local translations) could be surprising, especially because you usually pick the former only when you don't have access to `t` in the current context (eg. in a decorator, in a service object...). Without reading `view_component` documentation, I would consider this a lack of consistency in coding style, and not necessarily infer that it triggers a different lookup behavior.

Compare it to `t("hello")` vs. `t(".hello")`: it shows a clear intent.

### Side note : where to put translations

In some projects, I prefer to put my translations in multiple small files, matching the structure of the project (it's easier to find the key you're looking for). Alternatively, especially when I'll need to request translations from other people, it makes more sense to have them all in `config/locales/en.yml`.

`view_component` could enable both solutions with `t(".hello")` by:
- looking up the key `hello` in the sidecar locale file first,
- then, if not found, look up the namespaced key globally at `example_component.hello` (current behavior)

But is it still real isolation? 🤔

elia:
> Is it still real isolation?

_Probably doesn't matter, the worst that can happen is that we change the title of this PR 😄_ 

@Spone I've been sitting on your comment for 3 days, pondering how could we possibly do that in a way that makes sense for the consumer of this API and the code that needs to implement it. Below I tried to summarize what we're trying to accomplish and the problems we need to overcome.

### Desired features

It seems like what we want is:

1. having the _lazy lookup_ as the preferred way to get the sidecar translations
2. have it fall-back on the lazy lookup in Rails
3. have the sidecar translations unreachable from anywhere else in the application

### Current issues

I would add the following as issues, but @BlakeWilliams and @joelhawksley please check if these are things we don't want to worry too much about:

- try to avoid duplicating existing Rails functionality
  _Would be quite easy to copy and paste the TranslationHelper from Rails along with its tests, do a few tweaks and make it work. I'm not how well that would work with:_
  - _supporting different Rails versions and potentially diverging implementations_
  - _maintaining all that code and keeping up with changes in Rails_ 

-  acknowledge the fact that the I18n library is global in nature
  _I had to find out this about the `I18n` library at my own expense while working on this PR, in multiple occasions I had to fight the fact that I18n really likes to rely on the `::I18n` constant for many things. That's why I ended up using just a backend instead of an isolated I18n instance… because there's no such thing! 😅_ 

### Open questions

- in which direction we want to proceed?
- on which features we're ready to compromise?
- how much this can become it's own thing (vs. relying on Rails)?

Especially if we want to use the "lazy lookup" I'd say we need to rely on Rails, which in turn means that we need to keep the sidecar translations global either respecting the `virtual_path` of the templates or adding our own `view_components` scope.

I'm interested in your thoughts on this, but also on alternative approaches I might have missed.

elia:
@Spone @BlakeWilliams I just pushed an additional commit with an attempt at isolated lazy path translations for components.

I also dropped support for multiple translations using the `t/translate` helper directly in the component, that can be added back, either now, or down the road. In this case I think we're on a good long term path for improvements (unlike changing from lazy to non-lazy and vice-versa, as pointed out by @Spone), and we're free to ship a minimalistic version initially.

What else can change to have this ready for a merge? / What other criteria should it meet?

docs/index.md
```diff
@@ -1128,6 +1128,50 @@ app/components
 </div>
 ```
 
+##### Translations (experimental)
+
+To use experimental support for `I18n` translations, include `ViewComponent::SidecarI18n`:
```
joelhawksley, line 1133: What do you think about renaming this to `Translatable` be more idiomatic Ruby?

```suggestion
To use experimental support for `I18n` translations, include `ViewComponent::Translatable`:
```

joelhawksley (commented):
👋🏻 leaving a naming comment here. I'm going to defer to @BlakeWilliams for the technical review.

docs/index.md
elia: That's a great suggestion, it fits well with existing modules 👍

joelhawksley:
@elia I have a small favor to ask you ❤️ 

Might you be willing to open a separate PR introducing `_sidecar_files`? I'd love to use it on a PR I'm working on 😄

elia:
@joelhawksley sure, will take just a minute 👍

elia:
@joelhawksley there you go https://github.com/github/view_component/pull/676

OT is there a reason for sticking with bundler v1?

joelhawksley:
> OT is there a reason for sticking with bundler v1?

I think it's due to how many Ruby/Rails versions we support in CI?

docs/index.md
```diff
@@ -1128,6 +1128,50 @@ app/components
...
+en:
+  hello: "Hello world!"
+```
+
+Access translations as usual:
+
+`app/components/example_component.html.erb`
+
+```erb
+<p><%= t("hello") %></p>
```
BlakeWilliams, line 1157: Is this still accurate? Do we need to update the docs to use the "lazy" format? e.g. `t(".hello")`?

BlakeWilliams (approved):
I had two small questions, but otherwise this looks great!

joelhawksley:
Thanks for your work on this PR @elia!

elia:
Thanks for making the final changes and merging!

PR #662, opened 2021-03-11
Prepend preview routes instead of appending

joelhawksley:
### Problem

Given an application with a catchall route, previews are inaccessible, as they are _appended_ to the application's routes file.

### Solution

_Prepend_ them instead.

### Notes

I'd be curious if folks think this is worth testing. Doing so could be a bit of a hassle, but I'm up for the challenge 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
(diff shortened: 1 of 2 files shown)


BlakeWilliams (approved):
Makes sense to me 👍

PR #664, opened 2021-03-15
Ensure blocks passed to slots are captured

BlakeWilliams:
Currently blocks given to a lambda slot aren't always captured which can
cause them to be rendered incorrectly in certain rendering situations.

This fix resolves the issue by ensuring the block passed to the lambda
slot is always captured.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
     *Joel Hawksley*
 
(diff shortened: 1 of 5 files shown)


joelhawksley (approved):
How about a CHANGELOG? 😄

BlakeWilliams:
Added 👍

BlakeWilliams:
Ah, those PVC failures are legit. Looking 👀

joelhawksley (approved):

PR #667, opened 2021-03-17
504 previews with sample source code

johannesengl:
### Summary

This PR introduces a source code example snippet on view component previews. 

Screenshot 2021-06-02 at 20 12 00

In order to add syntax highlighting we rely on [prism.js](https://prismjs.com/). To use the library we need to add one js and one css asset file. Those files are served as static assets from the gems `assets/vendor` directory and we therefore don't require the host app to use a specific setup for handling assets like webpacker or the assets pipeline. 

The prism javascript file includes syntax highlighting for the following languages: ruby, erb, haml. 

To enable the source previews users can set `config.view_component.show_previews_source = true`. The config option defaults to `false`. A view helper method `preview_source` is added to render the preview source. 
The helper method is conditionally called based on the config option at the end of the `preview.html.erb` view file. 
Users can also use the helper in their views for more flexibility of where to render the preview source.


jonspalmer:
A few thoughts:

1. Can we come up with a better mechanism for including the prism css and js?
  * Are they necessary?
  * What are the alternatives?
  * How can we avoid snapshotting a particular version which makes upgrades and maintenance a PITA. is there an npm type solution?
2. Can we add test coverage of this new behavior?

johannesengl:
@jonspalmer thanks for your input on this! ❤️

I had a call with @joelhawksley before getting started on this and we decided on using prism and starting simple by just adding the assets via the assets pipeline. The benefit of using the assets pipeline is that we don't force users to use a specific tool for managing assets which keeps the deployment setup easy - no webpacker or npm required for using view component. 

> 1. Can we come up with a better mechanism for including the prism css and js?
> 
> * Are they necessary?

If we want to give users syntax highlighting I guess the most adequate approach would be to rely on a library like prism.js for this which requires us to add assets.
> * What are the alternatives?

Do you mean alternative ways to include the assets or alternative libraries for syntax highlighting? 

> * How can we avoid snapshotting a particular version which makes upgrades and maintenance a PITA. is there an npm type solution?
> 

Interesting. Yes, it would be nice if we could update the prim assets in an easier matter. Wondering how frequent we need to update the prim version and if it justifies making the asset setup of the gem more complex. 

> 1. Can we add test coverage of this new behavior?
Sure. This is still WIP and I will add tests for this shortly.

jonspalmer:
@johannesengl Thanks for the feedback. Sounds like you're way ahead of most of my questions. 

I'm curious about use of the asset pipeline usage. How does that play with apps that are using pure Webpacker and none of the traditional asset pipeline? Have we tested that combination?

johannesengl:
> @johannesengl Thanks for the feedback. Sounds like you're way ahead of most of my questions.
> 
> I'm curious about use of the asset pipeline usage. How does that play with apps that are using pure Webpacker and none of the traditional asset pipeline? Have we tested that combination?

I just spend some more time testing out different setups as you suggested and indeed when removing sprockets in the host application the following raises an error: 

```
    initializer "view_component.assets.precompile" do |app|
      app.config.assets.precompile << "view_component/prism.css"
      app.config.assets.precompile << "view_component/prism.min.js"
    end
```
https://github.com/github/view_component/pull/667/files#diff-fe26dc052312860cc076b99a7f3c0913bf834fc73fe246def378024ffc038164R89

I did some research of how other rails engines include assets in the host application and found the following discussion: 
https://discuss.rubyonrails.org/t/how-to-pack-js-from-a-different-gem-engine-in-rails-6-webpacker-it-seems-frustrating/75364/10

> My resolution is that if I need to share js between projects I am better of keeping the js in npm, like rails-ujs, like turbolinks for example. They don’t come as engines, but as npms. Much cleaner.

**Sidekiq Web UI**
Looking at sidekiq web UI they register their own rack app and rely on `Rack::Static` to reference the static assets. I don't think this is a valid solution for us:
https://github.com/mperham/sidekiq/blob/f0a20b1a5e6b06853083774d35b79d39821085b7/lib/sidekiq/web.rb#L146
https://github.com/mperham/sidekiq/blob/f0a20b1a5e6b06853083774d35b79d39821085b7/web/views/layout.erb#L21 

**Active Admin**
I also had a look at the gem [active admin ](https://github.com/activeadmin/activeadmin). They provide two different generators depending if your using webpacker or not and copy the assets files into the right location in the host application. 

https://github.com/activeadmin/activeadmin/blob/master/lib/generators/active_admin/webpacker/webpacker_generator.rb
 
I don't really like the idea of adding an extra step to setup view component and requiring users to run a generator for installation. 

### Resolution 
Given that rails 6 comes by default with webpack for including javascript and sprocket for including other assets I feel like view component should support both webpack and sprocket for including the assets.

I am wondering how about we just don't add assets at all for a specific library like prism.js and let this step be up to the users of the gem?

@jonspalmer @joelhawksley What is your opinion on the above?

joelhawksley:
@johannesengl this might not be a great long-term solution, but since this is only meant for diagnostic use, could we start by vendoring the JS and CSS assets _directly_ into the gem and serve them inline on the page? It's not ideal, but could unblock our work here.

johannesengl:
> @johannesengl this might not be a great long-term solution, but since this is only meant for diagnostic use, could we start by vendoring the JS and CSS assets _directly_ into the gem and serve them inline on the page? It's not ideal, but could unblock our work here.

@joelhawksley Updated the way we load the assets based on your comment in https://github.com/github/view_component/pull/667/commits/1103f43a9c8ffb7a3898745f10f3290b48f550f1. 

I have a question regarding the test setup. For some combination of ruby / rails versions the test suit fails now. Is there a way to see which specific test fails or run the test suit locally for specific version?

johannesengl:
> > @johannesengl this might not be a great long-term solution, but since this is only meant for diagnostic use, could we start by vendoring the JS and CSS assets _directly_ into the gem and serve them inline on the page? It's not ideal, but could unblock our work here.
> 
> @joelhawksley Updated the way we load the assets based on your comment in [1103f43](https://github.com/github/view_component/commit/1103f43a9c8ffb7a3898745f10f3290b48f550f1).
> 
> I have a question regarding the test setup. For some combination of ruby / rails versions the test suit fails now. Is there a way to see which specific test fails or run the test suit locally for specific version?

@joelhawksley @jonspalmer Any update on this?

Spone:
Hi @johannesengl it seems that the test suite is now passing. I see "All checks have passed - 17 successful checks".

johannesengl:
> Hi @johannesengl it seems that the test suite is now passing. I see "All checks have passed - 17 successful checks".

Amazing :) I guess there was an issue with CI or the main branch had failing tests. 

@joelhawksley or @jonspalmer Would appreciated a review on this :)

joelhawksley:
@jonspalmer mind having another look?

johannesengl:
> @jonspalmer mind having another look?

Any update on this?

joelhawksley:
@jonspalmer pinging you one more time before I take on reviewing this ❤️

test/app/views/layouts/application.html.erb
```diff
@@ -6,5 +6,6 @@
 </head>
 <body>
   <%= yield %>
+  <%= content_for?(:view_component_preview_source) ? yield(:view_component_preview_source) : yield %>
```
jonspalmer, line 9: why do we render `view_component_preview_source` in the application template?

jonspalmer: Also isn't this code double yielding if there is no `view_component_preview_source`

test/app/views/layouts/preview.html.erb
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html>
+
+<head>
+  <title>ViewComponent - Preview - Test</title>
+</head>
+
+<body>
+  <%= yield %>
+  <%= content_for?(:view_component_preview_source) ? yield(:view_component_preview_source) : yield %>
```
jonspalmer, line 10: Isn't this code double yielding if there is no `view_component_preview_source`

lib/view_component/preview.rb
```diff
@@ -86,6 +86,12 @@ def preview_example_template_path(example)
                 .sub(/\..*$/, "")
       end
 
+      # Returns the method body for the example from the preview file.
+      def preview_example_source(example)
+        source = self.instance_method(example.to_sym).source.split("\n")
+        source[1...(source.size - 1)].join("\n")
```
jonspalmer, line 92: What's going on with the split and join here? What is being removed?

app/views/layouts/previews.html.erb
```diff
@@ -0,0 +1,26 @@
...
+        </code>
+      <% end %>
+    </pre>
+  </div>
+  <script type="text/javascript" src="<%= asset_path('prism.min.js', skip_pipeline: true) %>"></script>
+<% end %>
+
+<% if @host_application_layout == false %>
+  <%= yield %>
+  <%= yield(:view_component_preview_source) %>
```
jonspalmer, line 23: Can you share what's happening here?

app/controllers/view_components_controller.rb
```diff
@@ -32,7 +32,8 @@ def previews
       template = @render_args[:template]
       locals = @render_args[:locals]
       opts = {}
-      opts[:layout] = layout if layout.present? || layout == false
+      opts[:layout] = "previews"
+      @host_application_layout = layout if layout.present? || layout == false
```
jonspalmer, line 36: It's not clear to me why we need to mess around with layouts to achieve what we need here. Can you share the reasons for this approach?

johannesengl: We want to append some HTML for the source code example below every preview. My thinking was that this would be easiest archived if all previews are being rendered in the same layout where `yield` renders the preview HTML and below we render the HTML for the source code example. 
I initially intended this to work without the user having to add anything in the layout the preview gets rendered in. This didn't work and users now have to add `yield(:view_component_preview_source_code_example)` in their layout which adds the source code example. 

Given that a user anyways has to explicitly render the preview source code example in a layout in the host application we can also solve this using a render call of a partial. That way we don't have to touch any logic regarding layouts in the view components controller. Does this sound like a better approach or do you have any suggestion of how to solve this without a user having to change anything in their layout?

app/views/layouts/previews.html.erb
johannesengl: `yield` on line 22 renders the view component, `yield(:view_component_preview_source)` on line 23 renders source code example.

lib/view_component/preview.rb
johannesengl: Given the following method: 

```
  def default
    render(
      ExampleComponent.new(
        title: "Example Component"
      )
    )
  end
```

`self.instance_method(example.to_sym)` returns the method above as a string. The split and join returns the string above without the first and last line so we remove the method signature and the closing `end`. The result of the method is being displayed with syntax highlighting to the user.

app/controllers/view_components_controller.rb
jonspalmer: Can we just add to `app/views/view_components/preview.html.erb` to append the source code after the preview? Maybe I'm missing something.

johannesengl: This does not work for [preview templates](https://viewcomponent.org/guide/previews.html#preview-templates) for which the view files are stored in the host application.

test/app/views/layouts/preview.html.erb
johannesengl: Fixed in 6b9fccd62f888cca8bbcf3889e1f56734897cd37

test/app/views/layouts/application.html.erb
johannesengl: Fixed the double yielding in 6b9fccd62f888cca8bbcf3889e1f56734897cd37. 

> why do we render view_component_preview_source in the application template?
By default, previews are rendered in the application layout. We need to render `view_component_preview_source ` in order to append the example source code after the preview. `content_for?(:view_component_preview_source)` only returns true if we render a preview.

lib/view_component/preview.rb
jonspalmer: I think I'd expect to include the method signature too

app/views/layouts/previews.html.erb
jonspalmer: Wouldn't you expect to guard this with `if content_for?(:view_component_preview_source) %>` ?

jonspalmer (commented):
To core code for generating the source code makes sense to me. One question about stripping out the method signature or not - I'd say not.

My question around what the API to enable this should be? At the moment this feature is turned on by adding a line to the application template. While this works I have a few concerns with that approach:

1. It introduces 'test'/'debug' code into the application template - If I typo this I break my production app
2. It's not a simple config setting that you set to true
3. There is a bunch of code in the PR to handle the various layout combos and has a unusual pattern of creating a preview layout that's really just composing the existing template. I worry this is brittle, non-obvious and not particularly flexible
4. I've I'm overriding Preview Templates I don't have a lot of configuration options. The source content just gets yielded after the regular content

IMO it would be simpler to organize this a little differently:
1. Add a new configuration setting like `show_previews_source` that defaults to false
2. Introduce a helper that renders the preview source if `show_previews_source == true`
3. Call the new helper in `app/views/view_components/preview.html.erb`
4. If consumers are using their own Preview Templates then they're responsible for calling the new helper where they want the source to be rendered. 

This would give less code in this PR, the simplest config in common use cases and also gives maximum flexibility for consumers that are customizing Preview Templates.

johannesengl:
I agree with your concerns and like the approach you suggested! 

We can also suggest users to include their own layout for previews if they don't want to call the helper method below every preview template. In this case they can set the config option to false to prevent having the code preview twice on non preview templates. 

Thanks for your input. ❤️

Regarding the question of including the method signature or not. Why would you include it? 
Isn't the purpose of the source code preview to explain with an example how the view component can be used? If so the method signature is in my opinion not relevant for that.

johannesengl:
> One question about stripping out the method signature or not - I'd say not.
Updated in b8db0ca469d13cf4583bd40c23fd8551299b5567

johannesengl:
@jonspalmer Mind having another look? I updated the PR based on your suggestions.

lib/view_component/preview.rb
Spone: I find it confusing to keep the `default` method definition around the `render`. The sample source code is usually meant to be ready to copy-paste, and having to remove the first and last lines manually makes it harder.

See for instance: https://primer.style/view-components/components/button

johannesengl:
> @jonspalmer Mind having another look? I updated the PR based on your suggestions.

Would love seeing this being used at some point :) @joelhawksley or @jonspalmer mind having another look?

Spone:
Just to reiterate on https://github.com/github/view_component/pull/667#discussion_r646996768, I really think the method signature (`def render`) should be stripped, I see no point in displaying it.

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # CHANGELOG
 
+## 2.33.0
+
+* Add view helper `preview_source` for rendering a source code preview below previews.
+* Add config option `config.view_component.show_previews_source` for enabling the soure preview.
+
+    *Johannes Engl*
+
 ## main
```
Spone, line 11: The changelog item should be below `main`, as it's not released yet.

```suggestion
## main

* Add view helper `preview_source` for rendering a source code preview below previews.
* Add config option `config.view_component.show_previews_source` for enabling the soure preview.

    *Johannes Engl*

```

CHANGELOG.md
```diff
@@ -1 +0,0 @@
-docs/changelog.md
```
johannesengl, line 1: How does the symbolic link for the change log work?

docs/CHANGELOG.md
```diff
@@ -19,6 +19,13 @@ title: Changelog
 
     *Joel Hawksley*
 
+## 2.35.0
```
Spone, line 22: Please remove this heading, it will be added on the next release

CHANGELOG.md
Spone: I think you have to keep the file as it is and make your changes in `docs/changelog.md`

@joelhawksley could it be the other way around (put the symlink in the docs folder and keep the real changelog at the root)? It's confusing when you want to access the changelog from GitHub web UI.

lib/view_component/base.rb
```diff
@@ -241,6 +241,9 @@ def content_evaluated?
     #
     mattr_accessor :render_monkey_patch_enabled, instance_writer: false, default: true
 
+    # Configure if previews should include example source or not.
+    mattr_accessor :show_previews_source, instance_writer: false, default: false
```
Spone, line 245: Now that we have yard documentation (see #968) I guess we should add a yard-style documentation block here as well.

Can you confirm @joelhawksley?

Spone (changes requested):

CHANGELOG.md
joelhawksley: @johannesengl we need to update the file in `docs`, as @Spone mentioned. 

@Spone I tried doing the symlink in the other direction, but Jekyll was unhappy with that setup due to Github Pages running Jekyll in `safe` mode.

lib/view_component/base.rb
joelhawksley: Yes, please do ☝🏻, and be sure to run `rake docs:build` after you do 😄

Spone: :+1: 

@johannesengl let me know if you want me to wrap this up for you!

johannesengl: @Spone Sure feel free to wrap this up. Thanks!

Spone (approved):

Spone:
Here you go! Thanks again for your contribution @johannesengl :rocket:

joelhawksley:
@johannesengl would you be willing to add slim highlighting support to this feature?

PR #671, opened 2021-03-18
Fix handling of jbuilder templates

nielsslot:
### Summary

I had the crazy idea of trying to use ViewComponent with a Jbuilder template today and noticed fairly quickly that it doesn't work, but not for any reason I would've thought.

Rails ships builder as an included template handler and with jbuilder installed that means that both `.builder` and `.jbuilder` are valid template extensions. ViewComponent will look for template files with either extension but will find any jbuilder template twice, because it matches both `*builder` and `*jbuilder`. Because it finds the template twice it raises an error and doesn't render.

In this pull request I add a call to uniq to ensure that ViewComponent continues with a list of template paths that doesn't have any duplicates. I use jbuilder as an example, but any template handler whose extension is a sub string of another would show the same problem.

### Other Information

This is the same issue that was mentioned in #492. I solve it in a slightly different way, mostly because the code was changed since that pull request. And I use jbuilder as an example instead of relying on a custom template handler definition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## main
 
-* Fix handling of Jbuilder and other template engines with certain names.
+* Fix bug where a single `jbuilder` template matched multiple template handlers.
 
     *Niels Slot*
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Fix handling of Jbuilder and other template engines with certain names.
+
+    *Niels Slot*
+
 ## 2.28.0
 
 * Include SlotableV2 by default in Base. **Note:** It's no longer necessary to include `ViewComponent::SlotableV2` to use Slots.
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -129,6 +129,8 @@ GEM
     i18n (1.8.7)
       concurrent-ruby (~> 1.0)
     jaro_winkler (1.5.4)
+    jbuilder (2.11.2)
+      activesupport (>= 5.0.0)
     loofah (2.9.0)
       crass (~> 1.0.2)
       nokogiri (>= 1.5.9)
@@ -220,6 +222,7 @@ DEPENDENCIES
   bundler (~> 1.14)
   capybara (~> 3)
   haml (~> 5)
+  jbuilder (~> 2)
   minitest (= 5.6.0)
   pry (~> 0.13)
   rails!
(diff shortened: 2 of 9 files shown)


joelhawksley (approved):
👋🏻 thanks for the patch! Mind adding a CHANGELOG entry?

patchwork-her:
My My did someone not think ahead and say to themselves as they do this
hacking and stealing of  my household utilities, fraud, impersonation,
theft of services, driving without insurance, theft bye taking, strong
armed robbery and wow....i can name all night and their would be more to
add. Coercion, try me up. I'm ready

nielsslot:
Sure, no problem! I just added an entry to the CHANGELOG as described in the CONTRIBUTING file.

I noticed that the FAQ mentions testing on Haml and Slim explicitly. Should Jbuilder be added to this list? Or do we consider it an example of the support for "most Rails template handlers"?

Spone:
@nielsslot I'm curious about your use cases for building component with Jbuilder templates. Can you give us a few examples about what you use that for? :)

nielsslot:
> @nielsslot I'm curious about your use cases for building component with Jbuilder templates. Can you give us a few examples about what you use that for? :)

I thinking "use" is a big word at this moment :) I'm not even sure yet if view_component would work or be reasonable in my case.

I was looking for a convenient way to compose non-trivial Elasticsearch queries. I want a Ruby object which represents the parameters of a specific query. Then use this object as a model with Jbuilder to actually create the query. So in code this looks something like:

```ruby
class MyQuery
  attr_accessor :term, :size

  def downcase_term
    term.downcase
  end
end
```

combined with a template like:

```ruby
json.query do
  json.match do
    json.body do
      json.query @query.downcase_term
    end
  end
end
json.size @query.size
```

Which outputs JSON that can be send to Elasticsearch. My queries are typically a lot more complicated, but I hope you get the idea. (I'm using Jbuilder similar to the [elasticsearch-dsl gem](https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-dsl), which doesn't add enough over Jbuilder for me, but the idea is nice.)

It occurred to me that view_component is also kind of like this. Having some domain or model logic in a Ruby object with a template. So i thought to try it out and then came across this bug.

Besides my crazy out-of-the-box idea, I guess there could be other valid use cases for this as well. Maybe embedding a bunch of JSON data inside your HTML using a `<script>` tag with `application/json` as type? Or why not use view_component to build the Jbuilder views of a Rails API back-end application? If your API has a certain JSON structure that's used in multiple places, view_component could be better then partials and helpers.

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* Fix handling of Jbuilder and other template engines with certain names.
```
```suggestion
* Fix bug where a single `jbuilder` template matched multiple template handlers, resulting in error.
```

joelhawksley (approved):

joelhawksley:
Thanks for the PR, Niels! Come back any time ❤️

nielsslot:
Thanks for merging! Happy to help out, will be back when I can :-)

PR #675, opened 2021-03-24
SlotableV2 using shared state by leveraging procs (issue caused by #664)

xtr3me:
### Summary

After updating to version 2.28.0 our view components gem started to fail rendering. This was caused by #664 
`undefined method 'cell' for nil:NilClass`

In this PR is simplified extraction of the code which shows the need for state that is passed on to the children that ran without issues before #664. 

### Other Information

We pass on the state of the parent component so our developers only need to set a boolean in a single place rather than to provide it on each row they add.

How can we still pass on certain state and still have the changes in #664?
@BlakeWilliams


BlakeWilliams:
@xtr3me Thanks for the report and the PR!

Would it be possible to update the components in the test to be a minimal reproduction of the issue?

xtr3me:
Yes, what would you like to have removed? I could remove the rows and leave the header + cell. Is that enough?

xtr3me:
What we could do is pass down the parent to it's children, for example:

```ruby
class ViewComponent::Base
  attr_accessor :parent_component

  #...logic to set the parent_component after initializing the child component...
end

  attr_accessor :selectable

  renders_one :header, HeaderComponent

  def initialize(selectable: selectable)
    @selectable = selectable
  end
end

class HeaderComponent < ViewComponent::Base
  delegate :selectable, to: :parent_component, allow_nil: true
end
```

@BlakeWilliams how would you feel about this potential solution? What are the drawbacks about this approach?

BlakeWilliams:
@xtr3me I'd like to avoid having to inject another parameter if possible.

Thank you for simplifying the test case! I took a peek this morning and I got the tests passing with this change:

```diff
diff --git a/test/app/components/nested_shared_state/table_component.rb b/test/app/components/nested_shared_state/table_component.rb
index c13e096..94b124d 100644
--- a/test/app/components/nested_shared_state/table_component.rb
+++ b/test/app/components/nested_shared_state/table_component.rb
@@ -9,9 +9,8 @@ module NestedSharedState
       header_system_arguments[:selectable] = @selectable

       header_component = NestedSharedState::HeaderComponent.new(**header_system_arguments)
-      render(header_component) do
-        block.call(header_component)
-      end
+
+      header_component
     end

     # @param selectable [Boolean] When enabled it allows the user to select rows using checkboxes
```

The two cases look similar but the difference is that `render` will return the string representation of the component instead of the component instance itself. That being said, both examples *should* work.

I got both working by making sure we pass arguments through to the block in the `capture` call:

```diff
diff --git a/lib/view_component/slotable_v2.rb b/lib/view_component/slotable_v2.rb
index e69772a..08f382c 100644
--- a/lib/view_component/slotable_v2.rb
+++ b/lib/view_component/slotable_v2.rb
@@ -227,7 +227,9 @@ module ViewComponent
         # current component. This is necessary to allow the lambda to access helper
         # methods like `content_tag` as well as parent component state.
         renderable_value = if block_given?
-          slot_definition[:renderable_function].bind(self).call(*args, **kwargs) { view_context.capture(&block) }
+          slot_definition[:renderable_function].bind(self).call(*args, **kwargs) do |*args|
+            view_context.capture { block.call(*args) }
+          end
         else
           slot_definition[:renderable_function].bind(self).call(*args, **kwargs)
         end
```

I think it would be worth testing that it works well with keyword args since it may need to be updated for that, but otherwise should make both cases work. If you have any interest incorporating that diff into your PR and testing out the keyword arg case that would be awesome, otherwise I can dedicate more time to this later.

xtr3me:
@BlakeWilliams i'm super happy that you were able to take a peek and got it to work. I will update my PR tomorrow with your changes and add some tests to make sure arguments and keyword arguments work as expected.

I will also add an entry to the `CHANGELOG.md` so it will be easy to merge in.

xtr3me:
@BlakeWilliams I have integrated your patch and added an additional fixes when supplying keyword arguments.
To be able to check these cases I added additional asserts in the tests.

Also the `CHANGELOG.md` is updated and I have added additional documentation about sharing data from the parent component whilst still being able to chain upon the returned component.

Let me know if there is anything else you would like to see changed. 
Have a great weekend!

xtr3me:
Hi @BlakeWilliams while i was implementing a component based on the result of this PR I came across the following weirdness:

The following bits are not captured in the cell `content` method:
```
<% row.cell(classes: "col-xs-1") do %>
   <%= 'Hi Blake' %>
<% end %>

# <div class="col-xs-1"></div>
```

But wrapping it in a `capture` block within the view of my program it is captured into the content:
```
<% row.cell(classes: "col-xs-1") do %>
  <% capture do %>
    <%= 'Hi Blake' %>
  <% end %>
<% end %>

# <div class="col-xs-1">Hi Blake</div>
```

Without the ERB tag `<%= %>` it works fine:
```
<% row.cell(classes: "col-xs-1") do %>
  Hi Blake
<% end %>
# <div class="col-xs-1">Hi Blake</div>
```

Would you be so kind to assist me with this and/or point me in the right direction so I can also squash bug as well?

BlakeWilliams:
@xtr3me thanks for implementing that patch and getting things working!

Since that issue seems to be related, can you create a failing test for it?

xtr3me:
@BlakeWilliams I'm working on that, but it seems that Rails does some kind of hack to be able to use `<%=` on a block which isn't supported by ERB out of the box.

Currently I have:
```
template = ERB.new <<-EOF
  <%= controller.view_context.render(NestedSharedState::TableComponent.new(selectable: false)) do |table_card| %>
    <% table_card.header do |header| %>
      <% header.cell { 'Cell1' } %>
      <% header.cell(class_names: '-has-sort') do %>
        <%= 'Cell2' %>
      <% end %>
    <% end %>
  <% end %>
EOF
@rendered_component = template.result(binding)

assert_selector("div.table div.table__header div.table__cell", text: "Cell1")
assert_selector("div.table div.table__header div.table__cell.-has-sort", text: "Cell2")

# Check shared data through Proc
refute_selector("div.table div.table__header span", text: "Selectable")
```

But that crashes with:
```
SyntaxError: (erb):1: syntax error, unexpected ')'
...able: false)) do |table_card| ).to_s); _erbout.<< "\n       
```

As it seems Rails does a workaround where it rewrites the ERB input to something that is valid for ERB to process.
Is there a way to render this for the test without the ERB workaround that i'm trying?

BlakeWilliams:
@xtr3me I think you could create a component with that erb as a template instead of trying to implement the `ERB.new` call in the test. That may help you work around it?

xtr3me:
@BlakeWilliams that does work, but it doesn't trigger my issue. Since we also have a wrapper to lookup the component and render it I will investigate if that combination is the culprit:

```
def render_our_component(name, *args, **kwargs, &block)
  render(ComponentResolver.resolve(:internal, name).new(*args, **kwargs), &block)
end
```

lib/view_component/slotable_v2.rb
```diff
@@ -227,7 +227,9 @@ def set_slot(slot_name, *args, **kwargs, &block)
         # current component. This is necessary to allow the lambda to access helper
         # methods like `content_tag` as well as parent component state.
         renderable_value = if block_given?
-          slot_definition[:renderable_function].bind(self).call(*args, **kwargs) { view_context.capture(&block) }
+          slot_definition[:renderable_function].bind(self).call(*args, **kwargs) do |*args, **kwargs|
+            view_context.capture { block.call(*args, **kwargs) }
```
BlakeWilliams, line 231:
```suggestion
            view_context.capture(*args, **kwargs, &block)
```

Related to https://github.com/github/view_component/pull/683, I think this will need to be changed to prevent breaking HAML templates.

BlakeWilliams:
@xtr3me Any updates on the other issue you are seeing?

xtr3me:
@BlakeWilliams I didn't manage to create a minimal test case within the gem and didn't find enough time to make a sample app that I could share with you. Is it okay to come back on this early next week?

(we did experience the issue in different code bases where we use our gem that is based on ViewComponents.)

BlakeWilliams:
@xtr3me Totally, no rush!

Do you feel like this change is good to merge, or do you feel like you need to validate the other issue isn't related first?

xtr3me:
@BlakeWilliams It would be super nice to merge this already, I will make sure a new issue is opened with reference to a repo that displays the additional issue.

test/app/components/nested_shared_state/header_component.html.erb
```diff
@@ -0,0 +1,4 @@
+<%= content_tag(:div, **@tag_arguments) do %>
+  <% if @selectable %><span>Selectable</span><% end %>
+  <% cells.each do |cell| %><%= cell %><% end %>
+<% end %>
```
joelhawksley, line 4:
```suggestion
<% end %>

```

test/app/components/nested_shared_state/table_component.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= content_tag(:div, class: 'table') do %>
+  <%= header %>
+<% end %>
```
joelhawksley, line 3:
```suggestion
<% end %>

```

test/app/components/nested_shared_state/table_component.rb
```diff
@@ -0,0 +1,21 @@
...
+      header_system_arguments = system_arguments
+      header_system_arguments[:selectable] = @selectable
+
+      header_system_arguments[:data] ||= {}
+      header_system_arguments[:data][:argument] = arg if arg.present?
+
+      NestedSharedState::HeaderComponent.new(**header_system_arguments)
+    end
+
+    # @param selectable [Boolean] When enabled it allows the user to select rows using checkboxes
```
joelhawksley, line 16:
```suggestion
```

test/app/components/nested_shared_state/table_component.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module NestedSharedState
+  class TableComponent < ViewComponent::Base
+
```
joelhawksley, line 5:
```suggestion
```

joelhawksley (approved):

joelhawksley (approved):
Nicely done! Thanks @xtr3me!

PR #676, opened 2021-03-24
Add Base.sidecar_files to fetch sidecar paths

elia:
### Summary

This has been extracted from #660 but can actually be useful for supporting other types of sidecar assets.

### Other Information

The original method in the compiler was only referencing the component class, which is a good hint this method should rather live there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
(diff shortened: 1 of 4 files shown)


joelhawksley (approved):

joelhawksley:
🙇🏻 ❤️ thank you!

PR #677, opened 2021-03-25
CSS encapsulation

joelhawksley:
_Please read this pull request as a **proposal** that can and **should** change._

This pull request introduces a prototype implementation of CSS encapsulation. We've been running this code in production for about a month without issue. 

Wading into this area is not something we take lightly, and we recognize the significant amount of prior art and context to take into account.

I've included extensive documentation in the form of an Architectural Decision Record. I'd love to see us use that document as the grounds of debate for this proposal. 

I'd love to hear your thoughts at all abstraction levels, from the technical details to the broad concepts.


adr/0677-css-encapsulation.md
```diff
@@ -0,0 +1,67 @@
...
+
+The proposed design avoids Webpacker and other build tools. Instead, it uses a Ruby implementation of CSS Modules built by @rmosolgo, enabling greater portability and a convention-over-configuration experience that Rails developers expect.
+
+### No bundles
+
+At least for now, the proposed design eschews the concept of shipping CSS in bundles.
+
+### No global syntax
+
+In keeping with the existing ViewComponent encapsulation mental modal, all styles are locally scoped by default, with no global escape hatch.
```
elia, line 43: nit: `mental modal` → `mental model`

elia (commented):
Just had a quick read, super interesting, but this is just to notify about a couple of nits.
One of the two is that the first commit is missing a `:` after `Co-authored-by`.

adr/0677-css-encapsulation.md
```diff
@@ -0,0 +1,65 @@
...
```
joelhawksley, line 43:
```suggestion
In keeping with the existing ViewComponent encapsulation mental model, all styles are locally scoped by default, with no global escape hatch.
```

jaredcwhite:
This seems like a very interesting approach to solving for the use cases described. My comment is more a philosophical one: should this be part of ViewComponent at all? I can think of some reasons why VC could simply provide a few hooks in the base class(es) so that this particular CSS solution would function through a separate gem. That way (a) it's optional, and (b) others could build different solutions in a similar fashion. To be perfectly honest, I never use CSS modules (aka fingerprinting class names) nor HTML-embedded stylesheets, and prefer custom properties, custom elements, shadow DOM, etc. for encapsulation techniques…and so this isn't a solution I personally would use.

Spone:
This is great! I'll take some time to try it next week, but here are some preliminary thoughts after reading the PR:

- **Generator support**: will the component generator be updated to generate the sidecar CSS file? (with a `--css` flag or global setting)
- **Naming conventions**: I see that you went with a `.Css_0343d_foo` style selector for the `foo` class in `StylableComponent`. Most of the time, my personal choice would be `.stylable-foo` (I'm always making sure that my component naming is unique).  I guess there are a lot of strong personal preferences on this topic (also, CSS naming conventions such as BEM, SmaCSS...). How can we make sure this is not an obstacle to adoption?
- **Third-party CSS**: let's say I build a ViewComponent that depends on a 3rd-party library such as [Flickity](https://github.com/metafizzy/flickity), so I need to include the library CSS to my component. How would you approach this?
- **Slim templates**: I'm a bit concerned about the readability of having `div(class="#{styles['foo']} another-class")` instead of `.stylable-foo.another-class` in my Slim templates, but there's probably a way to improve this

lib/view_component/css_module.rb
```diff
@@ -0,0 +1,119 @@
...
+        mappings: mappings,
+        css: @css_root.render
+      }
+    end
+
+    private
+
+    # Generate a short, random-ish token to prevent CSS selector collisions.
+    def compute_hash
+      Digest::MD5.hexdigest(@module_name).first(5) # rubocop:disable GitHub/InsecureHashAlgorithm
```
Spone, line 36: What about using the first 5 characters of the Base64 representation of the hexdigest? This way there's more "specificity" in 5 chars, and even less risk of collision.

- inspired by `css-loader`'s `localIdentName: "[path][name]__[local]--[hash:base64:5]"` ([source](https://github.com/webpack-contrib/css-loader/blob/f1e0f00472fdd8fc1971c0602665b357b339e3b8/README.md#object))
- see https://stackoverflow.com/a/9987466/1789900 for an example of implementation

keithamus, line 36: `MD5` is pretty good at generating (mostly) collision free hashes, but a couple of points:

 - It's counted as encryption and federal compliance regulations have specific rubrics around encryption standards. Using this for "enterprise" software involves some bureaucratic complexity. It might be easier to migrate from this just to avoid having the conversation.
 - JS does not have MD5 in its standard library. It's unlikely to get it as it's now an obsolete standard. It's also a non-trivial implementation; it requires quite a lot of code to implement or pull in, so if we need to mirror these on the client side we'd be looking at adding 1-2kb of JS code just to do this.

Perhaps a better option is to use a simpler digest mechanism, such as `base64`? is probably more ubiquitous than MD5, and where it isn't it can be implemented in a dozen or so LOC. It does have a much higher potential for collisions; but given it's used as _part_ of a class selector here, the risk of collisions is likely the same as using MD5.

Spone: The problem with simple truncated `base64` is that you'll have the same value if components have the same suffix -- it's only encoding the string with a different radix, not hashing it:

```rb
Base64.encode64("HelloWorld").first(5)
=> "SGVsb"
Base64.encode64("HelloThere").first(5)
=> "SGVsb"
```

I'm not sure if there is some other hashing function that does not carry the same compliance regulations you mention?

keithamus: I don't think this is necessarily a problem as the class names would still be different as `@module-name` is used both undigested and digested parts of the selector:  a full class would be `.HelloThere_SGVsb_foo` or `.HelloWorld_SGVsb_foo`. The intent of the `compute_hash` function seems to be to provide enough entropy in a class name so as not to collide with human authored classes; `.HelloThere_foo` is more likely to already exist in existing CSS.

adr/0677-css-encapsulation.md
```diff
@@ -0,0 +1,65 @@
...
+
+We will implement significant portions of the long-standing [CSS Modules spec](https://github.com/css-modules/css-modules) natively in ViewComponent for [encapsulated styles](https://medium.com/seek-blog/the-end-of-global-css-90d2a4a06284):
+
+### No Webpacker
+
+The proposed design avoids Webpacker and other build tools. Instead, it uses a Ruby implementation of CSS Modules built by @rmosolgo, enabling greater portability and a convention-over-configuration experience that Rails developers expect.
+
+### No bundles
+
+At least for now, the proposed design eschews the concept of shipping CSS in bundles.
```
srt32, line 39: Could we add a "why" explanation to this statement?

adr/0677-css-encapsulation.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Alternatives we considered
+
+### CSS-in-Ruby
+
+Much like our issue with [inlining ERB inside ViewComponent Ruby files](https://github.com/github/view_component/commit/5d3806a2a9ec0187574ac6c64a0ae90e655691ae#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R173), writing CSS inline (like Styled Components and others do) raised similar concerns over compatibility with Ruby 3 type systems.
+
+### Ruby-in-CSS
+
+We considered allowing for evaluation of Ruby inside CSS files (at compile, runtime, or both) but instead decided to lean on [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) for injecting dynamic values.
```
srt32, line 57: If true, is it worth adding something here like, "The recent dotcom overhaul related to dark mode heavily relies on this pattern"?

adr/0677-css-encapsulation.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Bundle splitting
+
+We attempted to split our CSS into multiple smaller bundles without significant success. We found it hard to be confident that such changes weren't breaking our UI in subtle ways. Ensuring heavily reused templates always rendered with their corresponding styles proved to be practically impossible.
+
+## Consequences
+
+ViewComponent authors will write encapsulated CSS, avoiding several common errors and resulting in more resilient user interfaces.
```
srt32, line 65: I'm not sure where the right place is but I am left wondering how a project like this one that relies on Primer CSS will work. Will component authors have to pull in (duplicate) every class their component depends or is there some set of global primer css classes we get "for free"?

joelhawksley: @srt32 good question. I didn't mean for this statement to be so far-reaching. I've updated it to _ViewComponent authors will be able to write encapsulated CSS_

elia (commented):
So happy to finally see this released! :clap:
Just few considerations/needs I would have as a potential user of `ViewComponent::Stylable`.

How can we support porting **existing SCSS code** and running tools like **postCSS and CSSLint**?

**Using `params` to store the list of compiled CSS** might not play well with caching. Using a *current attribute* with a `Set` in it would probably look better, but I'm not sure it really solves the problem.

I'm under the impression that for some application would be as fine to **just ship the collected CSS of all existing components**. That has the advantage of leveraging the browser cache, or not even parsing the bundle more than once by using Turbo. This approach retains most of the properties/advantages as the current solution but simplifies a lot the delivery.

I'd like to see **a configurable CSS scoping prefix**, ideally I should be able to replicate popular solutions like turning `Foo::BarBaz` into `.foo--bar-baz`. As long as the components are stored in global constants any string derived from the class name should work.

**That's all I've got for now**, hope this feedback will help in shaping a solution that can be effective for apps of all sizes! :sparkles:

---

PS.

**My experience building two e-commerce stores with CSS and components** has been great, we implemented a poor man's version of styled components by just using unique CSS class names based on the Ruby class name. In one case it was built from scratch and we only included a reset and a micro CSS framework. In the other there's plenty of legacy CSS, so far the extraction has been quite painless. The only headaches we had were around components accepting blocks which potentially have conflicting CSS class names and clashes with names coming from utility classes.

BlakeWilliams (commented):
Overall this looks good to me!

docs/index.md
```diff
@@ -967,122 +966,85 @@ With the monkey patch disabled, use `render_component` (or  `render_component_to
...
-  def commenter
-    @comment.user
-  end
+class StyleableComponent < ViewComponent::Base
+  include ViewComponent::Styleable
 
-  def commenter_name
-    commenter.name
+  def call
+    content_tag(:div, "Hello, World!", class: styles['foo'])
```
BlakeWilliams, line 1004:
```suggestion
    content_tag(:div, "Hello, World!", class: styles.foo)
```

Thoughts on using methods instead of hashes? I think API wise, methods are a little easier to use plus it gives us the ability to later "enhance" the style methods with additional functionality. e.g. Finding unused CSS in the test environment

lib/view_component/stylable.rb
```diff
@@ -0,0 +1,81 @@
...
+require "active_support/concern"
+require "view_component/css_module"
+
+module ViewComponent
+  # EXPERIMENTAL - This module is subject to change
+  # at any time without warning. Use at your own peril!
+  #
+  # Adds support for sidecar CSS modules.
+  # See docs/index.md for directions for use.
+  module Stylable
```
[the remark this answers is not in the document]
joelhawksley: Done!

adr/0677-css-encapsulation.md
```diff
@@ -0,0 +1,65 @@
...
+
+### CSS-in-Ruby
+
+Much like our issue with [inlining ERB inside ViewComponent Ruby files](https://github.com/github/view_component/commit/5d3806a2a9ec0187574ac6c64a0ae90e655691ae#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R173), writing CSS inline (like Styled Components and others do) raised similar concerns over compatibility with Ruby 3 type systems.
+
+### Ruby-in-CSS
+
+We considered allowing for evaluation of Ruby inside CSS files (at compile, runtime, or both) but instead decided to lean on [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) for injecting dynamic values.
+
+### Bundle splitting
```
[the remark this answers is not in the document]
koddsson: I think we should eventually support "sidecar" components as well as "svelte" encapsulating components. Having both options would allow you to write little encapsulating components and large sidecar ones.

For simplicity's sake, we should only support sidecars in the beginning.

adr/0677-css-encapsulation.md
```diff
@@ -0,0 +1,65 @@
...
+
+We considered allowing for evaluation of Ruby inside CSS files (at compile, runtime, or both) but instead decided to lean on [CSS custom properties](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties) for injecting dynamic values. The recent Dark Mode feature heavily depended on CSS variables.
+
+### Bundle splitting
+
+We attempted to split our CSS into multiple smaller bundles without significant success. We found it hard to be confident that such changes weren't breaking our UI in subtle ways. Ensuring heavily reused templates always rendered with their corresponding styles proved to be practically impossible.
+
+## Consequences
+
+o, avoiding several common errors and resulting in more resilient user interfaces.
```
koddsson, line 65: Seems like there's text cut off here @joelhawksley?

adr/0677-css-encapsulation.md
```diff
@@ -0,0 +1,65 @@
...
+
+At least for now, the proposed design eschews the concept of shipping CSS in bundles.
+
+### No global syntax
+
+In keeping with the existing ViewComponent encapsulation mental model, all styles are locally scoped by default, with no global escape hatch.
+
+### No SCSS
+
+We're not very sure about this one, but we're playing with the idea of just using plain CSS, at least to start.
```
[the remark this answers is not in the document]
Spone: Yes it seems. I was also a bit surprised when I saw the added `sass` dependency.

joelhawksley:
@koddsson @BlakeWilliams @jaredcwhite @Spone @elia @keithamus @srt32 thanks for taking the time to review!

### Should this be a part of ViewComponent at all?

> I can think of some reasons why VC could provide a few hooks in the base class(es) so that this particular CSS solution would function through a separate gem. That way (a) it's optional, and (b) others could build different solutions in a similar fashion. - @jaredcwhite

I think this is a fair point, Jared. As proposed here, one would need to include the `Stylable` module, but I'd like to see us offer _some_ sort of built-in solution eventually. We should make it easy to opt out of this behavior.

### Class naming conventions

There is a _lot_ of subtle complexity here, and the proposed approach is ultimately quite naive. For example, it does not consider the case where a ViewComponent's CSS is updated, but a stale copy of the ViewComponent exists in a view cache. For that reason and others, I expect we'll eventually want to use a hash of the ViewComponent _and_ its sidecar files to generate the fingerprint.

### API for Slim templates

@Spone, thanks for bringing this up! We use ERB exclusively at GitHub, so it's easy to miss these kinds of things. Given the desire to have fingerprinted CSS selectors, could you suggest how we might implement them in a more Slim-friendly fashion?

### Third-party CSS

@Spone, I don't think this proposed architecture would work in that case. I'd expect a consumer of a library like Flickity to write ViewComponents that wrap its behavior. But perhaps you're suggesting that there might be a way to fingerprint third-party CSS?

### Caching

> Using params to store the list of compiled CSS might not play well with caching. Using a current attribute with a Set in it would probably look better, but I'm not sure it solves the problem. - @elia

:facepalm: I had a feeling that the `params` approach was a bit naive. Let's remove it for now, then.

### How can we support porting existing SCSS code and running tools like postCSS and CSSLint?

I think that `autoprefixer-rails` might be our ticket here. It already uses PostCSS under the hood, and I'd love to see us come up with a convention-over-configuration solution to this problem. Got any ideas here @palkan?

### Ship the collected CSS of all existing components

@elia, you're right. For a lot of applications, this is probably the correct approach.

### configurable CSS scoping prefix

@elia, I think this depends on what our goals are for CSS scoping. If it's to prevent collisions, I think using the Ruby constant name is excellent, as being globally unique. However, I _feel_ like we'll also want it to be a cache key effectively.

palkan:
Hey everyone!

> I think that `autoprefixer-rails` might be our ticket here.

I would say "I don't think so". Here is how autoprefixer-rails works: it ships with the prebuilt JS script containing both PostCSS core and Autoprefixer (PostCSS plugin). During assets compilation, it just calls JS runtime (via ExecJS) to run this script against a CSS text.

Packing every single PostCSS plugin this way doesn't sound like a good plan.

----

Now, let me share my take on other questions.

First of all, I love the idea of using CSS Modules as well as inlining `<style>` along with the template. That's brilliant.

> Should this be a part of ViewComponent at all?

I'm a bit worried about additional runtime dependencies (currently, `sass`). If we can make this dependency optional (similarly, for example, like Rails does for `redis` in [Action Cable](https://github.com/rails/rails/blob/f7e19dee026a6cec4d50c28f84da88b0f1397c40/actioncable/lib/action_cable/subscription_adapter/redis.rb#L5-L6)), that would be a good compromise, in my opinion.

However, abstractizing `Stylable` and `#styles` also makes sense. The current implementation could be one of the possible adapters. And I already have another one [here][vc-postcss-modules] 🙂.

> Class naming conventions

Maybe, there is already the answer to the question in the thread but: why can't we rely on the component name only? E.g., `<configurable-namespace>-<component_name.parameterize>-<class>`.

> API for Slim templates

I've been thinking of using shortcuts for that, smth like:

```slim
p.regular-class$container // expands into <p class="regular-class #{styles['container']}"></p>
```

Not sure that this is possible but I would like to experiment with this.

----

And the last but not least.

> No Webpacker

I'm a big fan of the classic Rails way and HTML-first approach. But ignoring the fact that frontend ecosystem has greatly progressed since the _Golden Age of Asset Pipeline_ (like, about 10 years ago). PostCSS and others changed the way we write CSS. Tailwind is getting more and more popular in the Rails community (and not everyone hates `@apply`). 

I think the statement "enabling greater portability" is over-estimated. A lot of projects won't be able to use Stylable due to the dependency on PostCSS, Autoprefixer or whatever.

That brings us back to the question whether this particular implementation should be the default one or not. Maybe, we can provide a common API to work with both Ruby CSS Modules and postcss-modules? That would allow us to satisfy everyone.

[vc-postcss-modules]: https://github.com/palkan/view_component-contrib#isolating-css-with-postcss-modules.

Spone:
> > API for Slim templates
> 
> I've been thinking of using shortcuts for that, smth like:
> 
> ```
> p.regular-class$container // expands into <p class="regular-class #{styles['container']}"></p>
> ```
> 
> Not sure that this is possible but I would like to experiment with this.

Quick try: https://github.com/Spone/slim/pull/1 (requires some minor changes to the `slim` parser, see the diff for details)

```slim
~text= @text
/ <div class="Styled_6c266_text">Lorem ipsum</div>

p~text.is-big LOREM IPSUM
/ <p class="Styled_6c266_text is-big">LOREM IPSUM</p>
```

Spone:
To follow up on **Third-party CSS**, here is an example to make it a bit more concrete. Let's consider a `GalleryComponent` that is a wrapper around Flickity, so the component requires both the JS and CSS from Flickity.

Currently, I see 4 possible approaches:

- `#1` Using the provided [CDN](https://github.com/metafizzy/flickity#cdn) obviously works out of the box if you add the `<script>` and `<link>` to your layout (but it's up to you to find a way to include them only if the component is used on the page)
- `#2` Install the library through [NPM](https://github.com/metafizzy/flickity#package-managers), then require it with the Asset Pipeline:

```js
//= require flickity
```
- `#3` Same as above but with Webpacker, you would then import the assets in the component JS:

```js
// app/components/gallery_component/gallery_component.js
import Flickity from "flickity";
import "flickity/dist/flickity.css";
```

- `#4` Finally, you can vendor the assets in your component sidecar folder, but then you have a duplication problem when multiple components depend on the library (and you can't update the library easily).

In approaches `#2` and `#3`, you still need a bundle of JS and CSS alongside your inlined components' CSS (and that's very probably what most projects will have anyway, if they have legacy CSS, or use a CSS reset, default values, variables, utility CSS...).

To sum up, I'm not sure we need to tackle third-party CSS encapsulation (ie. alter the selectors), at least not in the initial implementation.

(@joelhawksley you mentioned fingerprinting, in the case of NPM packages, maybe the `integrity` attribute in `package-lock.json` or `yarn.lock` could be of some use?)

lib/view_component/styleable.rb
```diff
@@ -0,0 +1,81 @@
...
+    # the rendering of the style tag more than once for the
+    # same component in a single request.
+    def _style_tag
+      rendered_components = request.params["___rendered_view_components"] ||= Set.new
+      rendered_components_identifier = "#{Rails.application.class.name}::#{self.class.name}"
+
+      if !rendered_components.include?(rendered_components_identifier)
+        request.params["___rendered_view_components"] << rendered_components_identifier
+
+        content_tag(:style, self.class._css.html_safe)
```
halo, line 72: I [saw](https://securityheaders.com/?q=github.com&followRedirects=on) that github.com uses the permissive content security policy `style-src 'unsafe-inline' github.githubassets.com`.

I suppose `unsafe-inline` is then required for anyone using these styleable view components because of the rendering of `<style>...</style>` directly into HTML.

tomasc:
I have been running CSS modules with ViewComponents for about a year now — based on this pattern: https://daniel.heath.cc/2018/07/02/css-modules-with-webpacker

Based on my experience an important part of the CSS modules idea is the `composes` directive — which requires a global index of class names and their encoded forms, so that they can be easily referenced.

The above mentioned implementation using JSON files (similarly to the original CSS modules) does work and performs well, based on my experience.

Recently I switched back to Sprockets and simply output reference to each components stylesheet tag along with the component. So far works great, reaping the benefits of HTTP/2.

At the same time I had to implement my own CSS modules Sprockets postprocessor which I would be happy to share, along with my update of the above referenced CSS modules implementation, if anyone is interested.

The postprocessor simply outputs a JSON file next to components sidecar file, which is in turn read (and in production cached) and used as global index.

It would be fairly easy to extract both as gems, which would be totally independent of the view component gem.

jaredcwhite:
I have to admit a large amount of trepidation around VC recommending a "CSS Modules" pattern, which by-and-large I consider an anti-pattern. I think it's reasonable to say VC shouldn't promote any particular flavor of how to write CSS￼. *But* if it were to do so, I would advocate for a design system based on CSS Custom Properties combined with custom element names for individual components to scope styles, plus Shadow DOM where true encapsulation is required. A far better—and more authentically "modern"—pattern￼, IMHO, and it requires little-to-none additional tooling.

It's fine if you want to disagree with me, but that's actually my point…I don't see why either flavor of CSS should somehow get "blessed" by the project. My 2c FWIW.

existentialmutt:
Really cool to see this happening! Scoped CSS is a very compelling feature. I work on a few long-lived apps that have been phasing in ViewComponent, and I would love to start replacing our mess of BEM/SMCSS stylesheet partials with something like this.

Reading over the proposal, here's a few points of feedback I can offer:

- It would be really nice to have an option for SCSS that could integrate global variables and mixins.
- How much performance overhead does it add to generate styles in ruby at request time compared to precompiling everything?
- Maybe I'm missing something, but won't inlining a style tag every time the component is rendered result in redundant CSS for components that are called multiple times on a page?

Our shop only recently transitioned to using webpacker for building css.  Long term I'm intrigued by the idea of moving it back to a ruby-based tool.  Short term my team is going to prefer a solution that hooks into our webpack setup, with an eye towards transitioning to a ruby-based solution as it matures.

Spone:
@existentialmutt good points!

> Maybe I'm missing something, but won't inlining a style tag every time the component is rendered result in redundant CSS for components that are called multiple times on a page?

Actually, the clever thing is that it's only inlining the style tag after the first render of the component on the page, see: https://github.com/github/view_component/blob/e0c1e7b4c092eebabea8531227a8c0dd0cf5c294/lib/view_component/styleable.rb#L62-L64

tomasc:
@Spone This has its major pitfalls though: Imagine a list of items, style is inlined in the first one, the first one gets removed (for example user the deletes it via UI), and … oops.

Spone:
> @Spone This has its major pitfalls though: Imagine a list of items, style is inlined in the first one, the first one gets removed (for example the user deletes it via UI), and … oops.

The style tag is not *in* the component, but just *after* it, so if you remove the component root element, the style will stick around.

tomasc:
@Spone I see, thanks!

joelhawksley:
👋🏻 Hi folks! Thank you for all the feedback so far. We are continuing to experiment in this space internally, and I hope to have more to share as we try more approaches.

The main concern on my mind at the moment is trying to come with an approach that _just works_ and feels conventional in the way that most of Rails does. I recognize that this is quite a challenge when it comes to assets as there seems to be less consensus on on conventions in that area in the Rails ecosystem.

test/view_component/css_module_test.rb
```diff
@@ -0,0 +1,82 @@
...
+    before_css = "#container .title, #container .subtitle {\n  background: rbga(255, 255, 255, 0.8);\n}"
+    after_css = "#item_447b7_container .item_447b7_title, #item_447b7_container .item_447b7_subtitle {\n  background: rbga(255, 255, 255, 0.8); }\n"
+    assert_rewrite("item", before_css, after_css)
+  end
+
+
+  def test_it_doesnt_rewrite_bare_element_selectors
+    before_css = "span {\n  background: rbga(255, 255, 255, 0.8);\n}"
+    after_css = "span {\n  background: rbga(255, 255, 255, 0.8); }\n"
+    assert_rewrite("item", before_css, after_css)
```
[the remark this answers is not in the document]
hyphenized: Would it be possible to add some sort of special selector to target global styles like other frameworks do? e.g. `:global(span)`

collimarco:
I find really interesting the possibility of using CSS Modules with ViewComponents. This avoid possible conflicts with names in the global namespace.

Are there any updates? Or have you found better solutions?

tomasc:
@collimarco https://github.com/rails/tailwindcss-rails

Slightly different approach, but works well with components.

jaredcwhite:
@collimarco My 2c: the vanilla CSS way to avoid conflicts in the global namespace is to use custom elements and selectors based off of those (`my-app-card > header` and that sort of thing)—or for even more control, you can use Declarative Shadow DOM and write truly encapsulated HTML & CSS. I've used both of those techniques along with ViewComponent and also similarly-handled components in the Bridgetown framework. It's rad. ☺️

I also would recommend avoiding Tailwind. I've had nothing but problems with it across multiple projects, and it's particularly ill-suited for use in a stack-agnostic design system build on top of modern web standards. Much better to reach for tools like [Open Props](https://open-props.style) (a fantastic set of CSS variables you can use to build your own components) or full-on web component libraries like [Shoelace](https://shoelace.style)—either way future-proofing your efforts for decades to come.

tomasc:
@jaredcwhite out of curiosity -- can you please be more specific about what makes Tailwind ill-suited? I have recently switched to Tailwind on a fairly large codebase and so far I am having fun. Curious to hear what issues might lie ahead. Thanks!

Spone:
Please, let's not hijack this issue with a Tailwind pros and cons argument 🙂

PR #679, opened 2021-03-26
Update Rails dependencies with mimemagic fix

joelhawksley:
I also took the liberty of removing 5.0.x from CI, as it did not receive the mimemagic fix.


BlakeWilliams (approved):
:shipit:

PR #683, opened 2021-04-01
Rails 6.1 + HAML + passthrough slots are rendering escaped HTML

agrobbin:
As requested in the New Issue template, I've written a failing test for a bug with Rails 6.1, HAML, and passthrough slots. It is double-rendering, first the non-escaped content (as expected), and then escaped content.

The test is influenced by the Slim bug fix in #594.


BlakeWilliams:
Thanks for opening this!

I'm not sure why HAML would behave differently than other template engines, but I can take a peek and see if anything stands out.

agrobbin:
@BlakeWilliams let me know if I can be of any assistance! I was very surprised to see this behavior as well.

BlakeWilliams:
Interesting enough, this technically fixes the problem (but isn't shippable)

```ruby
diff --git a/lib/view_component/slot_v2.rb b/lib/view_component/slot_v2.rb
index d40d2c6..cb88457 100644
--- a/lib/view_component/slot_v2.rb
+++ b/lib/view_component/slot_v2.rb
@@ -36,7 +36,9 @@ module ViewComponent
         elsif defined?(@_content)
           @_content
-          @_content_block.call
+          view_context.capture_haml do
+            @_content_block.call
+          end
         end
       end
```

I'm still looking at it, but I wonder if the problem is that we're wrapping the entire block in `view_context.capture`, which would return `false` if the `block_is_haml?` method is called on it.

BlakeWilliams:
It looks like the `block_is_haml?` issue was the problem.

I pushed a change in https://github.com/github/view_component/pull/683/commits/2ab9d09aa87d13fb3a0c48a1799566e7c62e4e06 that removes the wrapping `view_context.capture` call in favor of capturing each branch explicitly. This includes the `@_content_block` call, which is passed directly to `view_context.capture` so HAML can correctly determine if it should override the capture logic.

joelhawksley (approved):
Might we add a changelog entry?

agrobbin:
@BlakeWilliams thanks so much for the quick work here! I just applied this patch to our Rails 6.1 branch and it works as expected.

Looking forward to this getting merged + released!

BlakeWilliams:
Added a changelog entry here: https://github.com/github/view_component/pull/683/commits/734c0293b6421d3a87740fe1b581f305a7981d14

Issue #505, opened 2020-10-15
[TRACKING] ViewComponent documentation site


joelhawksley:
## Summary

With Rails 6.1 on the horizon, the ViewComponent project will be under increased attention from the public. 

This is a great opportunity for us to make the best first impression possible to the Rails community. 

Let's level up our README into a documentation website we'll be proud of!

## TODO

- [x] Register domain name (`viewcomponent.org`)
- [x] Pick out / design Jekyll template
- [x] Move docs from README to site
- [x] Publish site
- [x] Review site content and plan for any additions
- [x] Add quick start to README (https://github.com/github/view_component/pull/512#pullrequestreview-517923634)

## Open questions

- Should we have a logo? #649 
- What are some documentation sites we might be inspired by?

jaredcwhite:
A few Ruby/Rails-aligned documentation sites I'll mention:

* [StimulusReflex](https://docs.stimulusreflex.com)
* [Jumpstart](https://jumpstartrails.com/docs/)
* [Bridgetown](https://www.bridgetownrb.com) (full disclosure — I work on this one)

joelhawksley:
@jaredcwhite those are all fantastic examples!

I'm really impressed with the Bridgetown site. Would you be up for helping out with this project?

jaredcwhite:
@joelhawksley Well, if you don't mind using Bridgetown rather than Jekyll, I think it'd be pretty easy to clone the BT site and clear out the content and tweak the design for use here. You should be able to use GH actions to run the build process and deploy on pages￼￼ here… slightly more inconvenient than using Jekyll directly, but not by a ton. (Bonus: you get Webpack for front-end assets as well that way…) Anyway, I could help out with all that if need be.

g13ydson:
> A few Ruby/Rails-aligned documentation sites I'll mention:
> 
> * [StimulusReflex](https://docs.stimulusreflex.com)
> * [Jumpstart](https://jumpstartrails.com/docs/)
> * [Bridgetown](https://www.bridgetownrb.com) (full disclosure — I work on this one)

I really like how the Stimulus Reflex documentation works. Very clean, but it's a personal taste 😁

g13ydson:
@joelhawksley I really like how things are organized using the gitbook + github integration. I created a documentation reorganizing some subjects by sessions. So, if it looks good, I can open a PR with the gitbook files.
https://g13ydson.gitbook.io/view-component-docs

pinzonjulian:
[...]

PR #684, opened 2021-04-01
MVP documentation site

joelhawksley:
### Summary

This PR implements an MVP of a new documentation site using Just The Docs, splitting up our _gigantic_ single-page site into sub-pages 😄 

I plan to follow up this PR with more changes. I tried to limit the scope as much as possible to keep this manageable.

### Other Information

Tracking issue: https://github.com/github/view_component/issues/505


docs/contributing.md
```diff
@@ -1,5 +1,12 @@
+---
```
BlakeWilliams, line 1: I think it could be valuable to keep a `CONTRIBUTING.md` file in addition to `docs/contributing.md` since it will show up in the GitHub UI, pointing folks towards the contribution guidelines.

docs/index.md
```diff
@@ -1,1241 +1,21 @@
...
+---
+layout: default
+title: Overview
+nav_order: 1
+---
+
+# ViewComponent
 
-[View on GitHub →](https://github.com/github/view_component)
+A framework for building reusable, testable & encapsulated view components in Ruby on Rails.
```
BlakeWilliams, line 9: As-is this page is a little light on content but seems like the best place to "sell" the idea of view components to visitors.

I don't think it's blocking, but it could be cool to show off some of the features of view components. Maybe a good example could be comparing helpers/partials to a component that implements slots and some other functionality?

docs/guide/validations.md
```diff
@@ -0,0 +1,24 @@
+---
+layout: default
+title: Validations
+parent: Building ViewComponents
+---
+
+# Validations
```
BlakeWilliams, line 7: Somewhat related to my top-level comment, but this seems more like an FAQ entry or a recipe, thoughts on removing it for now in favor of bringing it back in some other, less official form?

docs/guide/lifecycle.md
```diff
@@ -0,0 +1,21 @@
+---
+layout: default
+title: Lifecycle
+parent: Building ViewComponents
+---
+
+# Lifecycle
```
BlakeWilliams, line 7: I think a diagram would be awesome to have on this page. Thoughts on opening an issue for one? Maybe something similar to https://twitter.com/dan_abramov/status/981712092611989509/photo/1 would be nice.

docs/changelog.md
```diff
@@ -1,4 +1,9 @@
-# CHANGELOG
+---
```
BlakeWilliams, line 1: Similar to the `CONTRIBUTING.md` file comment, would it be possible to import `CHANGELOG.md` into `docs/changelog.md` instead of moving it?

Thoughts being, seeing a `CHANGELOG` file in the directory root is a relatively common convention.

BlakeWilliams (commented):
I checked out the deployment, overall this is looking really good! I do have some high-level thoughts:

* The index page doesn't have a lot of content but is likely the best place to show off the value of view components. Not blocking, but I think we could put a really convincing example there showing off some of the features of view components as well as the value they provide.
* We list all of the functionality of view components, but don't really have anywhere that explains _when_ you would want to use that functionality. I wonder if there's a way to fill that gap. Maybe a recipes section? (also non-blocking).
* The light gray in the syntax highlighting is very low contrast, should we increase it?

docs/guide/collections.md
```diff
@@ -0,0 +1,98 @@
+---
```
manuelpuyol, line 1: tbh, I read this section multiple times and I'm still not sure what `with_collection` does

docs/guide/lifecycle.md
manuelpuyol: I'd love to see `_after_compile` here too :)

docs/guide/previews.md
```diff
@@ -0,0 +1,161 @@
...
+parent: Building ViewComponents
+---
+
+# Previews
+
+`ViewComponent::Preview`, like `ActionMailer::Preview`, provides a quick way to preview components in isolation.
+
+_For a more interactive experience, consider using [ViewComponent::Storybook](https://github.com/jonspalmer/view_component_storybook)._
+
+`ViewComponent::Preview`s are defined as:
```
manuelpuyol, line 13:
```suggestion
`ViewComponent::Previews` are defined as:
```

It was looking weird
image

docs/guide/collections.md
BlakeWilliams: Kind of meta, but I wonder how often collections are used with view components. Are they valuable? Do we use them? etc.

docs/guide/validations.md
joelhawksley: Moved to FAQs 👍🏻

docs/guide/collections.md
manuelpuyol: I've never seen `with_collection` before reading this doc 😅

joelhawksley: https://github.com/github/view_component/pull/684/commits/0bc8acd195798f39b652bbb647773fbbbc00432d

docs/guide/lifecycle.md
joelhawksley: Filed in https://github.com/github/view_component/issues/505#issuecomment-812577891

joelhawksley:
@BlakeWilliams @manuelpuyol thanks for the review! I've filed a couple TODOs here: https://github.com/github/view_component/issues/505#issuecomment-812577891.

In the meantime, I moved the contributors images to the homepage so it's not so empty:

Screen Shot 2021-04-02 at 9 44 23 AM

docs/guide/collections.md
joelhawksley: We have over a dozen uses internally. Not a ton, but it's definitely used 😄

BlakeWilliams (approved):
LGTM!

PR #787, opened 2021-04-05
include ViewComponent::SlotableV2 module

otorain:
When I am not include the `ViewComponent::SlotableV2` module，It raises the renders_many/renders_one method not found exception.


BlakeWilliams (approved):
This is great, thanks for clearing that up!

PR #788, opened 2021-04-05
Add note about when including slotable is necessary

BlakeWilliams:
Since 2.28.0 slots are included by default, so this removes the `include
ViewComponent::SlotableV2` call in the docs and adds a note about
versions that will need to include the module.

### DIFF
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -18,7 +18,7 @@ _To view documentation for content_areas (deprecated) and the original implement
 
 ## Defining slots
 
-**Note**: Versions of ViewComponent `< 2.28.0` need to `include ViewComponent::SlotableV2` in each component. Versions `>= 2.28.0` do not have to include the slotable module as it is now included by default.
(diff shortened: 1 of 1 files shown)


docs/guide/slots.md
```diff
@@ -18,6 +18,8 @@ _To view documentation for content_areas (deprecated) and the original implement
 
 ## Defining slots
 
+**Note**: Versions of ViewComponent `< 2.28.0` need to `include ViewComponent::SlotableV2` in each component. Versions `>= 2.28.0` do not have to include the slotable module as it is now included by default.
```
joelhawksley, line 21:
```suggestion
**Note**: In versions `< 2.28.0`, `include ViewComponent::SlotableV2` to use slots.
```

joelhawksley (approved):

joelhawksley:
@otorain FYI, we've updated the documentation to communicate when you need to include `SlotableV2` 😄

otorain:
> @otorain FYI, we've updated the documentation to communicate when you need to include `SlotableV2`

thanks :smile:

PR #805, opened 2021-04-06
fix link to "Slots" in content_areas.md

lxxxvi:
### Summary

Hey there 👋 , I just discovered that the documentation for `content_areas.md` had a broken link.

Screenshot 2021-04-06 at 07 58 20

PR #806, opened 2021-04-06
Update "Slots v2" broken link in docs

Spone:
### Summary

After reviewing #805 I found another similar broken link.


joelhawksley (approved):
🤔 I wonder if there's a linter that would catch these?

PR #849, opened 2021-04-06
Added test for rendering integers when using slots

xtr3me:
### Summary

Hi @BlakeWilliams, the bug I was mentioning turned out an issue in our own codebase which I was able to resolve by simply following the document which I wrote myself in the previous PR 🤣 

There is an additional issue I wanted to show and with this PR you will find a test case to demonstrate the issue. In the testcase `'2'` will render, where as `1` won't render.

We ran into when displaying `.count` from for example.

### DIFF
--- a/test/view_component/slotable_v2_test.rb
+++ b/test/view_component/slotable_v2_test.rb
@@ -304,7 +304,7 @@ class SlotsV2sTest < ViewComponent::TestCase
   def test_renders_integers
     render_inline(SlotsV2Component.new) do |component|
       component.title { 1 }
-      component.subtitle { '2' }
+      component.subtitle { "2" }
     end
 
     assert_selector(".title", text: "1")
--- a/test/view_component/slotable_v2_test.rb
+++ b/test/view_component/slotable_v2_test.rb
@@ -300,4 +300,14 @@ class SlotsV2sTest < ViewComponent::TestCase
     # Check shared data through Proc
     refute_selector("div.table div.table__header span", text: "Selectable")
   end
+
+  def test_renders_integers
+    render_inline(SlotsV2Component.new) do |component|
+      component.title { 1 }
+      component.subtitle { '2' }
+    end
+
+    assert_selector(".title", text: "1")
+    assert_selector(".subtitle", text: "2")
+  end
 end
--- a/lib/view_component/slot_v2.rb
+++ b/lib/view_component/slot_v2.rb
@@ -40,7 +40,7 @@ module ViewComponent
       elsif defined?(@_content)
         @_content
       elsif defined?(@_content_block)
-        view_context.capture(&@_content_block)
+        view_context.capture { @_content_block.call.to_s }
       end
 
       @content


test/view_component/slotable_v2_test.rb
```diff
@@ -300,4 +300,14 @@ def test_slot_with_nested_blocks_content_selectable_false
     # Check shared data through Proc
     refute_selector("div.table div.table__header span", text: "Selectable")
   end
+
+  def test_renders_integers
+    render_inline(SlotsV2Component.new) do |component|
+      component.title { 1 }
+      component.subtitle { '2' }
```
Spone, line 307: To make Rubocop happy:

```suggestion
      component.subtitle { "2" }
```

BlakeWilliams:
👋 thanks for opening this!

I wonder if we could get away with calling `to_s` when rendering slot content that isn't a renderable (something that responds to `render_in`)?

Maybe:

```diff
diff --git a/lib/view_component/slot_v2.rb b/lib/view_component/slot_v2.rb
index 8f7a507..8f40934 100644
--- a/lib/view_component/slot_v2.rb
+++ b/lib/view_component/slot_v2.rb
@@ -38,9 +38,9 @@ module ViewComponent
           end
         end
       elsif defined?(@_content)
-        @_content
+        @_content.to_s
       elsif defined?(@_content_block)
-        view_context.capture(&@_content_block)
+        view_context.capture(&@_content_block).to_s
       end
```

xtr3me:
@BlakeWilliams I altered your suggestion a bit and that solves the issue for the integers not being rendered. Unfortunately it causes a different test to fail: `ViewComponentTest#test_renders_haml_with_html_formatted_slot`

I also checked out the capture method: https://github.com/rails/rails/blob/914caca2d31bd753f47f9168f2a375921d9e91cc/actionview/lib/action_view/helpers/capture_helper.rb#L43

But didn't see an obvious solution, especially because under normal Rails circumstances it works perfectly fine

BlakeWilliams:
@xtr3me Sorry for the delay!

I took a look and I'm not sure I have any good solutions at the moment except for maybe overriding the default behavior of `capture`. The issue seems to be that Rails will filter out non-string values, https://github.com/rails/rails/blob/914caca2d31bd753f47f9168f2a375921d9e91cc/actionview/lib/action_view/helpers/capture_helper.rb#L46

I'm still thinking on the best way to proceed, but in the meantime calling `to_s` is likely a good work around.

xtr3me:
@BlakeWilliams no problem! We are not stopped by this little inconvenience and frankly i'm reporting/trying to fix this mostly to just improve the VC gem.

I'm not really a fan of overriding the `capture` method itself as it would probably impose problems we cannot foresee with the limited scope we have between our projects. Maybe we can take this up with someone from Rails and ask if there was a specific reason to filter out non-string values, based on that information we could determine some possible ways forward.

Maybe @joelhawksley and his network within the Rails team can answer this question?

PR #861, opened 2021-04-16
component with `link_to` with `url_for` throws `ActionController::UrlGenerationError`

lxxxvi:
Hello dear ViewComponent people 👋

I think this is a bug, or I may use `link_to` the wrong way. Please let me know.

### Steps to reproduce

(See failing test)

Use `url_for` in a Rails `link_to` method:

```ruby
<%= link_to "Home", url_for(key: "value") %>
```

### Expected behavior

Should render something like this:

```html
<a href="/?key=value">Home</a>
```

### Actual behavior

It throws an error:

```
ViewComponentTest#test_render_link_to_with_url_for:
ActionController::UrlGenerationError: No route matches {:key=>"value"}
```

### System configuration
**Rails version**: 6.1.3.1

**Ruby version**: 3.0.1

**Gem version**: 2.30.0 (resp. master)

### DIFF
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -120,32 +120,21 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
-## Setting request.path_parameters
+## Setting `request.path_parameters`
 
-Some Rails helpers won't work unless `request.path_parameters` are set correctly. In this case, you may see this error:
+Some Rails helpers won't work unless `request.path_parameters` are set correctly, resulting in an `ActionController::UrlGenerationError`.
 
-```
-ActionController::UrlGenerationError: No route matches {...}
-```
-
-To set `request.path_parameters` for a test case, you can use `with_request_url` from `ViewComponent::TestHelpers`.
+To set `request.path_parameters` for a test case, use `with_request_url` from `ViewComponent::TestHelpers`:
 
 ```ruby
 class ExampleComponentTest < ViewComponent::TestCase
   def test_with_request_url
-    with_request_url "/" do
-      render_inline ExampleComponent.new # contains i.e. `link_to "French", url_for(locale: "fr")`
-      assert_link "French", href: "/?locale=fr"
-    end
-
     with_request_url "/products/42" do
       render_inline ExampleComponent.new # contains i.e. `link_to "French", url_for(locale: "fr")`
       assert_link "French", href: "/products/42?locale=fr"
     end
   end
 end
-```
-
 ## RSpec configuration
 
 To use RSpec, add the following:
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -120,6 +120,32 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+## Setting request.path_parameters
+
+Some Rails helpers won't work unless `request.path_parameters` are set correctly. In this case, you may see this error:
+
+```
+ActionController::UrlGenerationError: No route matches {...}
+```
+
+To set `request.path_parameters` for a test case, you can use `with_request_url` from `ViewComponent::TestHelpers`.
+
+```ruby
+class ExampleComponentTest < ViewComponent::TestCase
+  def test_with_request_url
+    with_request_url "/" do
+      render_inline ExampleComponent.new # contains i.e. `link_to "French", url_for(locale: "fr")`
+      assert_link "French", href: "/?locale=fr"
+    end
+
+    with_request_url "/products/42" do
+      render_inline ExampleComponent.new # contains i.e. `link_to "French", url_for(locale: "fr")`
+      assert_link "French", href: "/products/42?locale=fr"
+    end
+  end
+end
+```
+
 ## RSpec configuration
 
 To use RSpec, add the following:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Add `with_request_url` test helper.
+
+    *Mario Schüttel*
+
 * Improve feature parity with Rails translations
   * Don't create a translation backend if the component has no translation file
   * Mark translation keys ending with `html` as HTML-safe
--- a/lib/view_component/test_helpers.rb
+++ b/lib/view_component/test_helpers.rb
@@ -60,6 +60,17 @@ module ViewComponent
       @controller = old_controller
     end
 
+    def with_request_url(path)
+      old_request_path_parameters = request.path_parameters
+      old_controller = defined?(@controller) && @controller
+
+      request.path_parameters = Rails.application.routes.recognize_path(path)
+      yield
+    ensure
+      request.path_parameters = old_request_path_parameters
+      @controller = old_controller
+    end
+
     def build_controller(klass)
       klass.new.tap { |c| c.request = request }.extend(Rails.application.routes.url_helpers)
     end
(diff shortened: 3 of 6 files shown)


BlakeWilliams:
I rarely use `url_for`, but I'm not sure that's the intended way to use that helper. Can you verify that this same behavior doesn't also exist in partials?

I think `root_path(key: "value")` may achieve the output you want though.

Spone:
> I think `root_path(key: "value")` may achieve the output you want though.

It's a different behavior: `url_for(key: "value")` will add the `key` parameter to the current path.
Maybe there is a problem accessing the current path from within a component?

BlakeWilliams:
> It's a different behavior: `url_for(key: "value")` will add the `key` parameter to the current path.

Ahhh, that would help explain the behavior some. Maybe it's trying to read the request's `controller` and `action` properties, which when rendered via `render_inline` might not have a route associated with them? At least that would be the case in view component tests.

BlakeWilliams:
It looks like that's the issue, when using `render_inline` we don't set any path parameters.

I was able to get `url_for` to work by doing the following:

```ruby
# config/routes.rb
get "/:login", to: "Users#show"
```

```ruby
test "my component do
  request.path_parameters[:controller] = "users"
  request.path_parameters[:action] = "show"
  request.path_parameters[:login] = "fake"

  render_inline MyComponent.new # calls `url_for` in the template and does not raise
end
```

BlakeWilliams:
It looks like that's the issue, when using `render_inline` we don't set any path parameters.

I was able to get `url_for` to work by doing the following:

```ruby
# config/routes.rb
get "/:login", to: "Users#show"
```

```ruby
test "my component do
  request.path_parameters[:controller] = "users"
  request.path_parameters[:action] = "show"
  request.path_parameters[:login] = "fake"

  render_inline MyComponent.new # calls `url_for` in the template and does not raise
end
```

BlakeWilliams:
@lxxxvi Thanks for opening a PR with a failing test, do you have any interest in implementing the `with_request_url` helper?

lxxxvi:
> do you have any interest in implementing the with_request_url helper?

@BlakeWilliams Sure, let me try it! 👍 

Thanks for your inputs so far.

lxxxvi:
Hello @BlakeWilliams

Thanks for pointing me into the direction how to implement the proposed helper `with_request_url`.

Using the information from your and @Spone's comments, I learned that we need to set two values in `request.path_parameters`: `:controller` and `:action`. The rest seems not relevant for `url_for`.

Eventually, I realized that the required logic _could_ be extended in the existing helper method `with_controller_class` by adding one argument to the method and then set the `request.path_parameters` accordingly.
So, instead of implementing a new method `with_request_url`, I tried it with the existing helper and **it works**!

Here is the draft I have so far:

```ruby
# Note: I simply copied the logic from `with_controller_class` for the moment.
def with_controller_class_copy(klass, action: "index")

  @controller = build_controller(klass)
  @controller.request.path_parameters[:controller] = klass.controller_name
  @controller.request.path_parameters[:action] = action
  yield
ensure
  @controller = old_controller
end
```

At this state I'd love to have your feedback, just to see if I go into the right direction.

How do you like the idea of extending the existing helper method?
Also, I am very much uncertain about defaulting the new `action:` argument and I fear it _could_ be a "breaking change" in certain cases, what do you think?

Thank you very much for your feedback.

BlakeWilliams:
@lxxxvi Thanks for trying that out!

I'm not positive that we'll want to build on top of the existing helper. Even though the two are somewhat related, I think it makes sense to introduce this as a separate helper. I could see this possibly replacing the `with_controller` helper at some point though.

As for the code, I think that will work in basic cases but not for cases where paths require other parameters. e.g. `/posts/:id` since we're not setting the `:id` path parameter. 

Maybe it would make sense to use something like `Rails.application.routes.recognize_path url` to return the correct path parameters that we could then merge into `@request.path_parameters`?

lxxxvi:
@BlakeWilliams 

I agree with your arguments, I thought I'd ask before I continue or try something else.
I like the idea with `Rails.application.routes.recognize_path url`, so I am going to try this next.
I'll let you know once I have something useful 😄

lxxxvi:
Hello @BlakeWilliams

I followed your suggestion to implement a separate helper `with_request_url` using `Rails.application.routes.recognize_path`.

I'd like to know if that looks good for you?

```ruby
def with_request_url(path)
  old_request_path_parameters = request.path_parameters

  request.path_parameters = Rails.application.routes.recognize_path(path)
  yield
ensure
  request.path_parameters = old_request_path_parameters
end
```

I added two tests, on for `/` and one for `/products`, is this sufficient?

Side question:
At first I tried to test _both routes_ in the the same test method, like this


```ruby
def test_with_request_url
  with_request_url "/" do
    render_inline UrlForComponent.new
    assert_text "/?key=value"
  end

  with_request_url "/products" do
    render_inline UrlForComponent.new
    assert_text "/products?key=value"
  end
end
```

But this does not work, as the response does not change after the second `render_inline` call. Is this intended (may be due to a caching mechanism)?

Thank you for your feedback.

BlakeWilliams:
@lxxxvi Code looks good!

I might spend some time diving into why your example doesn't work though. Maybe there's some deeply nested memoization somewhere causing that issue?

lxxxvi:
@BlakeWilliams

I think I found a workaround/solution for the mentioned problem.

In saw that in `with_controller_class` the instance variable for `@controller` is reset in the `ensure` block at the end of the method.
It works if we do the same in `with_request_url`, see latest commit.

But honestly, I don't understand why exactly this is needed.

I tried to debug the problem in the template, by changing `request.path_parameters` on the fly and then call `url_for`.
These tests always returned correct results. So, I think that's weird (for me), but there's probably a good reason or explanation for that behaviour. 😄 

---

Anyway, would it make sense to add a little description to the documentation for the new helper? Do we need other tests?

Thank you for your time.

joelhawksley (commented):
👋🏻 I'll let @BlakeWilliams provide the final review, but might you include a CHANGELOG entry and updates to the `testing` docs?

lxxxvi:
I added a new chapter to the testing docs, and the entry in the CHANGELOG.
Let me know if I can do anything else. 

Eventually I could rebase and squash all my commits into one (or I can split it as you prefer it)?

docs/guide/testing.md
```diff
@@ -120,6 +120,32 @@ class ExampleComponentTest < ViewComponent::TestCase
...
+    end
+
+    with_request_url "/products/42" do
+      render_inline ExampleComponent.new # contains i.e. `link_to "French", url_for(locale: "fr")`
+      assert_link "French", href: "/products/42?locale=fr"
+    end
+  end
+end
+```
+
```
joelhawksley, line 148:
```suggestion
## Setting `request.path_parameters`

Some Rails helpers won't work unless `request.path_parameters` are set correctly, resulting in an `ActionController::UrlGenerationError`.

To set `request.path_parameters` for a test case, use `with_request_url` from `ViewComponent::TestHelpers`:

```ruby
class ExampleComponentTest < ViewComponent::TestCase
  def test_with_request_url
    with_request_url "/products/42" do
      render_inline ExampleComponent.new # contains i.e. `link_to "French", url_for(locale: "fr")`
      assert_link "French", href: "/products/42?locale=fr"
    end
  end
end
```

```

joelhawksley:
> Eventually I could rebase and squash all my commits into one (or I can split it as you prefer it)?

No need to change your commit history. We don't worry too much about commit hygiene on this project ❤️

lxxxvi:
@joelhawksley Thanks for fixing the documentation 🙈

BlakeWilliams (approved):
This looks great, thank you for adding this helper!

PR #862, opened 2021-04-16
Increase feature-parity between Translatable and ActionView's TranslationHelper

elia:
### Summary

* Improve feature parity with Rails translations
  * Don't create a translation backend if the component has no translation file
  * Mark translation keys ending with `html` as HTML-safe
  * Always convert keys to String
  * Support multiple keys

### Other Information

While migrating from shared translations to `Translatable` in an application we run into problems related to the missing support for translation keys ending in `_html`.

Also we're including `Translatable` inside `ApplicationComponent`, doing this shouldn't affect memory after this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## main
 
+* Improve feature parity with Rails translations
+  * Don't create a translation backend if the component has no translation file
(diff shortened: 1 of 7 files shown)


BlakeWilliams (approved):

BlakeWilliams:
This is really great, thank you for the updates!

PR #875, opened 2021-04-22
Set content for components or slots without a block

manuelpuyol:
Successor of https://github.com/github/view_component/pull/534

### Summary

Prior to this pull request, providing content to a component could only be done by passing a block to `render`:

```erb
<%= render(TestComponent.new) do %>
  Hello, World!
<% end %>
```

This works great when you want to provide a snippet of HTML in an ERB file but can feel heavy weight when you only need to pass a piece of text, like above. It also makes it impossible to configure a component outside of a view, unless you use `instance_variable_set` or the like.

Now, you can set content using `#with_content`:

```erb
<%= render(TestComponent.new.with_content(“Hello, World!”)) %>
```

In additional to a string, you can also pass a component to `#with_content`, which will be rendered when `#content` is accessed:

```ruby
box = BoxComponent.new
box.with_content(FancyComponent.new(title: "My title"))
# ...later in the view:
render(box) # FancyComponent is is rendered along with BoxComponent
```

This makes it possible to build up a view from components.

### Slots

`with_content` has been added to slots too, making it possible to configure a component anywhere you can instantiate a class:

```ruby
box = BoxComponent.new
box.header(args).with_content("My title")

comments.each do |comment|
  box.row(args).with_content(comment.body)
end

box.footer(args).with_content("My footer")
```

### What makes this useful?

While working on a form builder, I found it difficult compose UIs without calling out to a template or creating a view component tailored to my single use case.

`#with_content` makes it possible to create beautifully rendered views prior to having a view context, making it possible to use ViewComponents in new and interesting ways.

As an added bonus, `#with_content` offers a more concise way to provide text while in ERB or other template languages.

### Other Information

This also removes `.ruby-version` to make it easier for developers to contribute. As per the gemspec, this library supports Ruby >2.3 and runs tests against [many Ruby versions](https://github.com/github/view_component/blob/master/.github/workflows/ci.yml#L38) so a strict Ruby version requirement while developing may not be necessary.

(I hit this while working on this pull request, where I had 2.7.0 instead of 2.7.1. Not having to install 2.7.1 made my setup a breeze.)


lib/view_component/with_content_helper.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module WithContentHelper
+    def with_content(value = nil, &block)
```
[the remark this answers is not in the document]
manuelpuyol: `content=` gives me the impression that the content will already be evaluated, which is not the case with `with_content`.

I think the main intention of `with_content` is telling the component: "hold this content and evaluate it when you have access to a view_context".

That's why we can do `MyComponent.new.with_content(OtherComponent.new)` without a `view_context` available, since `OtherComponent` isn't rendered when `with_content` is called, only when we render the `MyComponent` instance.

As for some other points:

> Allow us to avoid the complexity of supporting blocks

Actually, we don't need to support blocks afaik 🤔 

> The one downside is that we lose some amount of composability

I think composability is a great part of the API, since it can allow us to do:

```rb
component = MyComponent.new.content(
  OtherComponent.new.with_content(
    AnotherComponent.new...
```

lib/view_component/with_content_helper.rb
```diff
@@ -0,0 +1,29 @@
...
+      else
+        @_content_set_by_with_content = -> (_component) { value }
+      end
+
+      self
+    end
+
+    # Renders the given object and returns the result, if object can be rendered.
+    # Otherwise, returns object.
+    def render_or_return(object)
```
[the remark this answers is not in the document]
manuelpuyol: I'd be ok starting only with string support to test the waters

manuelpuyol:
> For example, when talking about the view_context dependence we assume the library will take ownership of the initialize method. If we control the initialize method, I could imagine us adding support for passing blocks to it as a potential alternative to with_content.

I think the blocks as an alternative could be great! But since we do not control the initialize (for now), `with_content` could pave the path for it. I don't think the `with_content -> block` migration would be troublesome

docs/guide/content.md
```diff
@@ -0,0 +1,47 @@
...
+Content passed to a ViewComponent as a block is captured and assigned to the `content` accessor.
+
+```erb
+<%= render(MyComponent.new) do %>
+  <!-- This will be accessible via `content` -->
+  <div>This is my content</div>
+<% end %>
+```
+
+## `#with_content` (Experimental)
```
joelhawksley, line 18: Do we want this feature to be experimental? I'm pretty confident in the API.

```suggestion
## `#with_content`
```

docs/guide/content.md
```diff
@@ -0,0 +1,47 @@
...
+
+Content can also be passed to a ViewComponent by calling `#with_content`. This is especially useful to build UI when you don't have a `view_context` available and render it at a later moment.
```
joelhawksley, line 20:
```suggestion
Content can also be passed to a ViewComponent by calling `#with_content`, which is especially useful when rendering components outside of views (such as in a controller, background job, script, etc):
```

docs/guide/content.md
```diff
@@ -0,0 +1,47 @@
...
+
+```rb
+component = MyComponent.new.with_content("This is my content")
+...
+render(component)
```
joelhawksley, line 25:
```suggestion
class MyController < ApplicationController
  def show
    render(MyComponent.new.with_content("This is my content"))
  end
end
```

docs/guide/content.md
```diff
@@ -0,0 +1,47 @@
...
+
+`#with_content` also accepts passing another component as content:
+
+```rb
+component = MyComponent.new.with_content(SomeComponent.new)
+...
+render(component)
+```
+
+### Slots
```
joelhawksley, line 36: I almost wonder if this document should just be part of the existing docs on passing content / slots documentation. I'd be afraid of folks missing this context when reading the Slots docs page.

docs/guide/content.md
```diff
@@ -0,0 +1,47 @@
...
+
+`#with_content` is also available for slots:
+
+```rb
+component = MyComponent.new
```
joelhawksley, line 41: While it _is_ valid Ruby, I'd prefer we keep our documentation free of assigning component instances to variables ❤️, instead inlining this into the `render` call.

lib/view_component/compiler.rb
```diff
@@ -13,17 +13,23 @@ def compiled?
...
 
-      if component_class.instance_methods(false).include?(:before_render_check)
+      if subclass_instance_methods.include?(:before_render_check)
         ActiveSupport::Deprecation.warn(
           "`before_render_check` will be removed in v3.0.0. Use `before_render` instead."
         )
       end
 
+      if subclass_instance_methods.include?(:with_content) && raise_errors
+        raise ViewComponent::ComponentError.new("#{component_class} implements a reserved method, `with_content`.")
```
joelhawksley, line 25: We currently use `ArgumentError` in `Base#validate_initialization_parameters!`. If we're going to introduce a custom error class, let's use it there too. I'm fine with either approach (ArgumentError or ComponentError)

lib/view_component/with_content_helper.rb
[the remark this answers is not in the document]
joelhawksley: I agree with @manuelpuyol here. That being said, we _could_ start with just the string API and add support for passing in components later? I'm fine with either approach.

docs/guide/content.md
manuelpuyol: I **want** this to be in a variable to showcase that it will be rendered later, even in a different context

lib/view_component/compiler.rb
manuelpuyol: I'd argue they are different kind of errors. `validate_initialization_parameters` checks for **arguments** in the initializer method, so `ArgumentError` makes sense.
Here, I'm checking if a **method** exists. Using `ArgumentError` could be confusing

lib/view_component/with_content_helper.rb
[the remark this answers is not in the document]
BlakeWilliams: > I think composability is a great part of the API, since it can allow us to do:
> 
> ```ruby
> component = MyComponent.new.content(
>   OtherComponent.new.with_content(
>     AnotherComponent.new...
> ```

I agree, but at the same time I see it as us re-implementing some part of Rails view `render` API (at least when we support renderables).

I could also see something like the following if we wanted to pre-render content for components.

```ruby
# alternative:  some kind of block wrapper
component = ViewComponent.render(MyComponent.new) do |c|
  render OtherComponent.new do
    AnotherComponent.new
  end
end
```

I think that, combined with `with_content` (imagining that it only supports strings) could be a powerful combination.

```ruby
# with some kind of block wrapper
component = ViewComponent.render(MyComponent.new) do |c| # provides a controllerless view_context to render
  render OtherComponent.new do
    AnotherComponent.new.with_content(content_tag(:h1, "hello!")) # with_content is helpful for injecting pre-rendered or Ruby generated content
  end
end
```

lib/view_component/compiler.rb
joelhawksley: By that logic, I think that `ArgumentError` is incorrect, as I _believe_ it's meant to be used when an argument passed to a method is incorrect, not that the method's argument declaration itself is incorrect. Maybe we just move to `ComponentError` instead for both?

lib/view_component/with_content_helper.rb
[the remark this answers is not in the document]
manuelpuyol: I think that pivoting to string only is a great idea, we can mature the API and have further discussions (like this one) as we see more use-cases in production!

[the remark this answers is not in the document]
manuelpuyol: removed the Component implementation in https://github.com/github/view_component/pull/875/commits/963dd774375a0cadba8eda033f6d96981d0a00bd

lib/view_component/slot_v2.rb
```diff
@@ -27,9 +31,15 @@ def to_s
       view_context = @parent.send(:view_context)
 
       @content = if defined?(@_component_instance)
-        # render_in is faster than `parent.render`
-        if defined?(@_content_block)
+        if defined?(@_content_set_by_with_content)
```
BlakeWilliams, line 34: Will we raise if there's an `@_content_block` and an `@_content_set_by_with_content` value passed?

lib/view_component/slotable_v2.rb
```diff
@@ -255,7 +255,7 @@ def set_slot(slot_name, *args, **kwargs, &block)
         @_set_slots[slot_name] = slot
       end
 
-      nil
+      slot
```
BlakeWilliams, line 258: I'm a fan of this change 👍

lib/view_component/with_content_helper.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module WithContentHelper
+    def with_content(value = nil, &block)
```
BlakeWilliams, line 5: Since we only support strings, how do you feel about removing block support to focus `with_content` even more for v1?

BlakeWilliams (approved):
LGTM!

I had one question about raising, and one more question about block support for `with_content`, but nothing blocking,

lib/view_component/with_content_helper.rb
joelhawksley: I agree ☝🏻

lib/view_component/slot_v2.rb
manuelpuyol: hmm we could to keep it more consistent, do you think it is worth it?

lib/view_component/with_content_helper.rb
manuelpuyol: done in https://github.com/github/view_component/pull/875/commits/8833c4b776820115d112d615377a2f80e36db9cd

lib/view_component/slot_v2.rb
BlakeWilliams: If someone tries to pass a block to a slot that has already had `with_content` called on it I think we should raise instead of silently overriding one, since it could save some amount of debugging time.

manuelpuyol: added in https://github.com/github/view_component/pull/875/commits/87db42624b8eaf5b539ec5d26650af9e8567c526

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## main
 
+* Add #with_content to allow setting content without a block.
```
```suggestion
* Add `#with_content` to allow setting content without a block.
```

lib/view_component/with_content_helper.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module WithContentHelper
+    def with_content(value = nil)
+      if value.nil?
+        raise ArgumentError.new("No content provided.")
+      else
+        @_content_set_by_with_content = -> (_component) { value }
```
jonspalmer, line 9: Why are we converting this to a proc vs just storing the value?

jonspalmer:
This is very similar to some of the options for passing content in the very first version of content areas. We anticipated uses cases like this but never 100% had them to use in anger. Glad to see this back with real world examples of it application.

lib/view_component/with_content_helper.rb
manuelpuyol: good catch! This was done because we used to accept `blocks` here, so I wanted to keep the interface instead of always checking if `@_content_set_by_with_content` is callable or not

lib/view_component/with_content_helper.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module WithContentHelper
+    def with_content(value = nil)
```
jonspalmer, line 5: Should we remove the default `nil` value if it's invalid?

manuelpuyol: done!

Issue #886, opened 2021-04-27
Deprecation warning for `before_render_check`


davekaro:
### Steps to reproduce

I just upgraded to the 2.31.0 release and now am getting:

```
DEPRECATION WARNING: `before_render_check` will be removed in v3.0.0. Use `before_render` instead. (called from <main> at app/components/application_component.rb:1)
```

My `ApplicationComponent` inherits `ViewComponent::Base`, and the only place I could find the `before_render_check` method being called is https://github.com/github/view_component/blob/d485d9ebb6e185b53a258e67fd62a183d3029f6a/lib/view_component/base.rb#L101. Should this method call be removed here?

### Expected behavior
No deprecation warning

### Actual behavior
Deprecation warning is displayed

### System configuration
**Rails version**: 6.1.3.1

**Ruby version**: 3.0.1

**Gem version**: 2.31.0

leighhalliday:
I've been digging into this as well, and I think it occurs because `ViewComponent::Base` has an `inherited` method (https://github.com/github/view_component/blob/ba90dcc29256c6b840da4efdbd6312b461ec3bc6/lib/view_component/base.rb#L269) that calls `compile` (deprecation warning is given in compiler), but it is compiling `ViewComponent::Base` which does implement this method, and I believe should continue to until this method is officially removed.

Guessing at solutions...

- The deprecation warning could check if it is compiling `ViewComponent::Base`, which is allowed to implement this method for now
- Don't call `compile` when `ViewComponent::Base` is inherited from

davekaro:
Ah yes, I see now we'd have to keep calling the method until it's officially removed.

If it wasn't clear, this is happening in the new release because this method https://github.com/github/view_component/pull/875/files#diff-fa28fa6e2d5f267384e7793b855281451a46b8437431867871a298fc52fe4940L16-L19  was moved below the deprecation warning, and before it was returning false because errors were present - `Could not find a template file or inline render method for ViewComponent::Base`. So, I think your first option makes sense, as that was effectively how it was behaving before. Or, can we move the check for errors back up to the top?

PR #887, opened 2021-04-27
move the deprecation warning below the template_errors check

davekaro:
### Summary

this was changed in https://github.com/github/view_component/pull/875/commits/6799a29af7daff334c34ccb0c6b26d2afd9d0e13,
but that causes deprecations when compiling ViewComponent::Base, which is
still allowed to call the before_render_check method

Fixes https://github.com/github/view_component/issues/886

### Other Information

Let me know if this seems good and needs a CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Fix `DEPRECATION WARNING: before_render_check` when compiling `ViewComponent::Base`
+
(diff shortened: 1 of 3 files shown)


joelhawksley (approved):
Thanks for the fix! Mind adding a CHANGELOG?

davekaro:
All set, thanks.

PR #889, opened 2021-04-27
Add basic API docs for instance methods, powered by YARD

joelhawksley:
### Summary

We don't have API-focused documentation. This PR aims to give us a basic system for generating it from YARD code comments, inspired by a similar approach we've taken in Primer ViewComponents.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -34,4 +34,3 @@ A proxy through which to access helpers. Use sparingly as doing so introduces co
 ### #request → [ActionDispatch::Request]
 
 The current request. Use sparingly as doing so introduces coupling that inhibits encapsulation & reuse, often making testing difficult.
-
--- a/docs/api.md
+++ b/docs/api.md
@@ -9,17 +9,29 @@ title: API
 
 ## Instance methods
 
-### before_render → [void]
+### #before_render → [void]
 
 Called before rendering the component.
 
-### before_render_check → [void] (Deprecated)
+### #before_render_check → [void] (Deprecated)
 
 Called after rendering the component.
 
 _Use `before_render` instead. Will be removed in v3.0.0._
 
-### render? → [Boolean]
+### #render? → [Boolean]
 
 Whether the ViewComponent should render.
+
+### #controller → [ActionController::Base]
+
+The current controller. Use sparingly as doing so introduces coupling that inhibits encapsulation & reuse, often making testing difficult.
+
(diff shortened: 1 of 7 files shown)


docs/api.md
```diff
@@ -0,0 +1,36 @@
+---
+layout: default
+title: API
+---
+
+<!-- Warning: AUTO-GENERATED file, do not edit. Add code comments to your Ruby instead <3 -->
+
+# API
```
BlakeWilliams, line 8: I like this, but we do miss out on a lot of functionality that the YARD generated documentation pages provide like viewing the source inline, links to related code (like exceptions), search, etc.

What if we added a top-level documentation link to https://rubydoc.info/gems/view_component?

joelhawksley: @BlakeWilliams I hear you. It'd be easy to add inline source, and we already have search built into the docs site.

More generally, I see some downsides to using RubyDoc:

- Users would have to navigate between two sites.
- Anecdotally, RubyDoc is very slow to load for me.
- We'd be giving up control over how our API docs are organized. For example, the Collections feature exists in a module, but it's included automatically. Should the consumer of the docs have to navigate this complexity, knowing to look in `ViewComponent::Collection`? I personally would rather have a single page for all of the public methods that are available by default, regardless of where they come from.

lib/view_component/base.rb
```diff
@@ -97,25 +99,37 @@ def render_in(view_context, &block)
       @current_template = old_current_template
     end
 
+    # Called before rendering the component.
```
BlakeWilliams, line 102:
```suggestion
    # Overridable lifecycle that is called before rendering the component.
```
Thoughts on mentioning that these are overridable lifecycle methods?

lib/view_component/base.rb
```diff
@@ -97,25 +99,37 @@ def render_in(view_context, &block)
...
 
+    # Called after rendering the component.
+    #
+    # @deprecated Use `before_render` instead. Will be removed in v3.0.0.
+    # @return [void]
     def before_render_check
       # noop
     end
 
+    # Whether the ViewComponent should render.
```
BlakeWilliams, line 117:
```suggestion
    # Overridable lifecycle method that determines if the component should render.
```

Similar to the above comment, thoughts on mentioning it's an overridable lifecycle method?

BlakeWilliams (approved):
LGTM

PR #891, opened 2021-04-28
fix slot examples using incorrect syntax 

### LINKED ISSUE #890: Duplicated content started to appear when using `<%=` to capture slots in v2.31
[This change](https://github.com/github/view_component/commit/358fbe2ef60b1d2e2d7de308a58f718bade6dd5f#diff-f4d01a0600f7ce443a17504bbf4275827066c6a3430e92e0a440b3d344be8ab5L258-R258), introduced in v2.31, broke some of our components that were wrongly using `<%=` instead of `<%` to capture slot blocks.

Although it's not strictly a bug I think the documentation is not consistent/explicit in discouraging that, and a changelog entry would probably be useful as well. Or, if the change is not functional to any other feature, we can revert the change and reintroduce it in v2.32.

joelhawksley:
cc @elia 

[Closes #890]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ## 2.31.0
 
+_Note: This release includes an underlying change to Slots that may affect incorrect usage of the API, where Slots were set on a line prefixed by `<%=`. The result of setting a Slot should not be returned. (`<%`)_
+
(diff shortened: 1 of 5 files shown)


BlakeWilliams (approved):

elia (commented):
🙏 thanks!

PR #910, opened 2021-05-05
Add instrumentation

### LINKED ISSUE #893: Instrumentation (ActiveSupport notifications)
## Feature request

Add ActiveSupport notifications support (`render.view_component`?).

### Motivation

Instrumentation could be useful for different reasons:
- Measuring rendering performance (APM-like stuff)
- Performing runtime analysis: e.g., detect unused components, build component dependency trees (could be used for caching).

Since ViewComponent aims to be as performant as possible, this feature could be opt-in (e.g., `ActiveSupport.on_load(:view_component) { include ViewComponent::Instrumentation }` or whatever).

skryukov:
### Summary

Closes #893

This PR adds a new configuration option `config.view_component.instrumentation_enabled`. When set true, ActiveSupport notification event `render.view_component` is enabled.

### DIFF
--- a/docs/guide/instrumentation.md
+++ b/docs/guide/instrumentation.md
@@ -1,0 +1,25 @@
+---
+layout: default
+title: Instrumentation
+parent: Building ViewComponents
+---
+
+# Instrumentation
+
+To enable ActiveSupport notifications, use the `instrumentation_enabled` option:
+
+```ruby
+# config/application.rb
+# Enable ActiveSupport notifications for all ViewComponents
+config.view_component.instrumentation_enabled = true
+```
+
+Then subscribe to the `!render.view_component` event:
+
+```ruby
+ActiveSupport::Notifications.subscribe("!render.view_component") do |*args|
+  event = ActiveSupport::Notifications::Event.new(*args)
+  event.name    # => "!render.view_component"
+  event.payload # => { name: "MyComponent", identifier: "/Users/mona/project/app/components/my_component.rb" }
+end
+```
--- a/lib/view_component.rb
+++ b/lib/view_component.rb
@@ -8,6 +8,7 @@ module ViewComponent
   autoload :Base
   autoload :Compiler
   autoload :ComponentError
+  autoload :Instrumentation
   autoload :Preview
   autoload :PreviewTemplateError
   autoload :TestHelpers
--- a/lib/view_component/engine.rb
+++ b/lib/view_component/engine.rb
@@ -13,6 +13,7 @@ module ViewComponent
 
       options.render_monkey_patch_enabled = true if options.render_monkey_patch_enabled.nil?
       options.show_previews = Rails.env.development? || Rails.env.test? if options.show_previews.nil?
+      options.instrumentation_enabled = false if options.instrumentation_enabled.nil?
       options.preview_route ||= ViewComponent::Base.preview_route
       options.preview_controller ||= ViewComponent::Base.preview_controller
 
@@ -30,7 +31,17 @@ module ViewComponent
       end
 
       ActiveSupport.on_load(:view_component) do
-        options.each { |k, v| send("#{k}=", v) }
+        options.each { |k, v| send("#{k}=", v) if respond_to?("#{k}=") }
+      end
+    end
+
+    initializer "view_component.enable_instrumentation" do |app|
+      ActiveSupport.on_load(:view_component) do
+        if app.config.view_component.instrumentation_enabled.present?
+          # :nocov:
+          ViewComponent::Base.prepend(ViewComponent::Instrumentation)
+          # :nocov:
+        end
       end
     end
 
--- a/lib/view_component/instrumentation.rb
+++ b/lib/view_component/instrumentation.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require "active_support/notifications"
+
+module ViewComponent # :nodoc:
+  module Instrumentation
+    def self.included(mod)
+      mod.prepend(self)
+    end
+
+    def render_in(view_context, &block)
+      ActiveSupport::Notifications.instrument("!render.view_component", name: self.class.name, identifier: self.class.identifier) do
+        super(view_context, &block)
+      end
+    end
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Add the ability to enable ActiveSupport notifications (`!render.view_component` event) with `config.view_component.instrumentation_enabled`.
+
+  *Svyatoslav Kryukov*
+
 ## 2.32.0
 
 * Enable previews by default in test environment.
--- a/docs/index.md
+++ b/docs/index.md
@@ -70,6 +70,7 @@ ViewComponent is built by:
 <img src="https://avatars.githubusercontent.com/rmacklin?s=64" alt="rmacklin" width="32" />
 <img src="https://avatars.githubusercontent.com/seanpdoyle?s=64" alt="seanpdoyle" width="32" />
 <img src="https://avatars.githubusercontent.com/simonrand?s=64" alt="simonrand" width="32" />
+<img src="https://avatars.githubusercontent.com/skryukov?s=64" alt="skryukov" width="32" />
 <img src="https://avatars.githubusercontent.com/smashwilson?s=64" alt="smashwilson" width="32" />
 <img src="https://avatars.githubusercontent.com/spdawson?s=64" alt="spdawson" width="32" />
 <img src="https://avatars.githubusercontent.com/swanson?s=64" alt="swanson" width="32" />
--- a/test/app/components/instrumentation_component.html.erb
+++ b/test/app/components/instrumentation_component.html.erb
@@ -1,0 +1,1 @@
+<div>hello,world!</div>
--- a/test/app/components/instrumentation_component.rb
+++ b/test/app/components/instrumentation_component.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class InstrumentationComponent < ViewComponent::Base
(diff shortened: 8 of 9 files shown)


lib/view_component.rb
```diff
@@ -14,4 +14,5 @@ module ViewComponent
   autoload :TestCase
   autoload :TemplateError
   autoload :Translatable
+  autoload :Instrumentation
```
Spone, line 17: Maybe keep the `autoload`s in alphabetic order?

skryukov: Good catch!

lib/view_component/engine.rb
```diff
@@ -29,6 +30,12 @@ class Engine < Rails::Engine # :nodoc:
         end
       end
 
+      app.reloader.to_prepare do
```
palkan, line 33: Can't we move this into `ActiveSupport.on_load(:view_component)` block below? As far as I remember, it's also called every time the code is reloaded.

lib/view_component/instrumentation.rb
```diff
@@ -0,0 +1,25 @@
...
+        mod.prepend(self)
+      end
+
+      def enabled?
+        Rails.application.config.view_component.instrumentation_enabled
+      end
+    end
+
+    def render_in(view_context, &block)
+      return super(view_context, &block) unless Instrumentation.enabled?
```
palkan, line 18: Is this necessary? As I see, we use this check in tests.

Maybe, we can instead manually include the module in the view component class under test?

palkan (commented):
Thanks! Overall looks good.

Left some questions.

lib/view_component/instrumentation.rb
skryukov: Sure we can, but this also helps with hot reloading during development: prepended module remains after reload, so I decided to leave this line. But yeah, it's kinda ugly.

lib/view_component/engine.rb
skryukov: `ActiveSupport.on_load(:view_component)` does not run block during `reload!`, at least in a test suite. ¯\\_(ツ)_/¯

lib/view_component/instrumentation.rb
palkan: > hot reloading during development

Actually, I think, generally, any updates to `config/*.rb` are not handled by `reload!`; so, we don't need to care about it. We only extend the core functionality once on boot.

docs/guide/instrumentation.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: default
+title: Instrumentation
+parent: Building ViewComponents
+---
+
+# Instrumentation
+
+You can enable ActiveSupport notifications for all ViewComponents with `instrumentation_enabled` option:
```
joelhawksley, line 9:
```suggestion
To enable ActiveSupport notifications, use the `instrumentation_enabled` option:
```

docs/guide/instrumentation.md
```diff
@@ -0,0 +1,25 @@
...
+
+```ruby
+# config/application.rb
+# Enable ActiveSupport notifications for all ViewComponents
+config.view_component.instrumentation_enabled = true
+```
+
+Now you can subscribe to `render.view_component` event:
```
joelhawksley, line 17:
```suggestion
Then subscribe to the `render.view_component` event:
```

lib/view_component/instrumentation.rb
```diff
@@ -0,0 +1,25 @@
...
+
+      ActiveSupport::Notifications.instrument("render.view_component", identifier: self.class.short_identifier) do
```
joelhawksley, line 20: What do you think about using the class name here instead of the path?

```suggestion
      ActiveSupport::Notifications.instrument("render.view_component", identifier: self.class.name) do
```

joelhawksley (commented):
This looks great! Thank you for the lovely contribution ❤️

lib/view_component/base.rb
```diff
@@ -223,6 +223,9 @@ def content_evaluated?
       @_content_evaluated
     end
 
+    # Configure if ActiveSupport notifications are enabled for all ViewComponents.
+    mattr_accessor :instrumentation_enabled, instance_writer: false, default: false
```
BlakeWilliams, line 227: Should this be included as part of the `Instrumentation` module instead of being defined directly on base?

Since we're utilizing Rails config I think we can use something like `Rails.application.config.view_component.instrumentation_enabled` when needed?

lib/view_component/instrumentation.rb
BlakeWilliams: I agree, I think we could remove this check. 👍

lib/view_component/instrumentation.rb
BlakeWilliams: Should we follow the Rails pattern here?

https://github.com/rails/rails/blob/722303b4a77307c85b988793a7aa299b1f478aca/actionview/lib/action_view/template.rb#L357-L359

```ruby
{ virtual_path: @virtual_path, identifier: @identifier }
```

BlakeWilliams, line 20:
```suggestion
      ActiveSupport::Notifications.instrument("!render.view_component", identifier: self.class.short_identifier) do
```

I think we should follow the Rails convention and prefix this with a `!` to help signal that this shouldn't be used in production.

lib/view_component/engine.rb
skryukov: Used `ActiveSupport.on_load(:view_component)` since we drop `reload!` compatibility

lib/view_component/instrumentation.rb
skryukov: I decided to add both `name` and full `identifier`

lib/view_component/engine.rb
```diff
@@ -30,7 +31,16 @@ class Engine < Rails::Engine # :nodoc:
       end
 
       ActiveSupport.on_load(:view_component) do
-        options.each { |k, v| send("#{k}=", v) }
+        virtual_options = [:instrumentation_enabled]
+        options.select { |k, _| virtual_options.exclude?(k) }.each { |k, v| send("#{k}=", v) }
```
BlakeWilliams, line 35: I read this a few times and found it difficult to follow. Is there a way to make what this is trying to do more explicit, or is it possibly unnecessary?

lib/view_component/instrumentation.rb
```diff
@@ -0,0 +1,20 @@
...
+require "active_support/notifications"
+
+module ViewComponent # :nodoc:
+  module Instrumentation
+    def self.included(mod)
+      mod.prepend(self)
+    end
+
+    # Configure if ActiveSupport notifications are enabled for all ViewComponents.
+    mattr_accessor :instrumentation_enabled, instance_writer: false, default: false
```
BlakeWilliams, line 12: Is this still being used? It looks like we're checking `Rails.application.config` elsewhere.

lib/view_component/engine.rb
skryukov: Checked with `respond_to?("#{k}=")`, an explicit variant is a bit verbose for my taste.

BlakeWilliams (approved):

BlakeWilliams:
Thanks for this addition!

It looks like there's one failure due to code coverage, otherwise I think this is good to go!

lib/view_component/engine.rb
```diff
@@ -30,7 +31,15 @@ class Engine < Rails::Engine # :nodoc:
...
       ActiveSupport.on_load(:view_component) do
-        options.each { |k, v| send("#{k}=", v) }
+        options.each { |k, v| send("#{k}=", v) if respond_to?("#{k}=") }
+      end
+    end
+
+    initializer "view_component.enable_instrumentation" do |app|
+      ActiveSupport.on_load(:view_component) do
+        if app.config.view_component.instrumentation_enabled.present?
+          ViewComponent::Base.prepend(ViewComponent::Instrumentation)
```
skryukov, line 41: I don't get how to test this line without setting `instrumentation_enabled` for all tests by default or additional code, which we deleted earlier ¯\\_(ツ)_/¯

skryukov: We can make `ViewComponent::Base` configurable and substitute it for tests. @BlakeWilliams wdyt?

BlakeWilliams: Sorry for the delay here. I think I'd be okay skipping coverage here for now via `# :nocov:` comments if you'd like to make that change to get tests ✅

lib/view_component/instrumentation.rb
```diff
@@ -0,0 +1,17 @@
...
+require "active_support/notifications"
+
+module ViewComponent # :nodoc:
+  module Instrumentation
+    def self.included(mod)
+      mod.prepend(self)
+    end
+
+    def render_in(view_context, &block)
+      ActiveSupport::Notifications.instrument("!render.view_component", name: self.class.name, identifier: self.class.identifier) do
```
Spone, line 12: The convention seems to be that the leading `!` means "*that the event is private of the framework and should not be used by applications and libraries*".

see https://twitter.com/rafaelfranca/status/1289978377550442496

I'm not sure it's the case here?

BlakeWilliams: Oh interesting, I had a different interpretation of the `!` prefix. I had made the suggestion to prefix it with a `!` for similar reasons that [this comment outlines](https://github.com/rails/rails/blob/91f0db1353e1b24e2d26ebbbadce510735f66848/actionview/lib/action_view/template.rb#L148-L150).

```
    # This method is instrumented as "!render_template.action_view". Notice that
    # we use a bang in this instrumentation because you don't want to
    # consume this in production. This is only slow if it's being listened to.
```

Instrumentation can become surprisingly slow when there are listeners, so I thought it made sense to try and discourage it where possible.

joelhawksley (approved):

Issue #913, opened 2021-05-06
Final newline is rendered


boardfish:
Reduced test case: https://github.com/boardfish/view_component-bug

### Steps to reproduce

Render a component that uses a HTML element with `display: inline` by default, e.g. `span` or `a`, inline.

### Expected behavior

There should be no trailing whitespace. i.e. I should be able to do this:

```html.erb
<%= render TestComponent.new %>,
```

and receive:

```html
<a href="https://simon.fish">Add Test template here</a>,
```

as opposed to:

```html
<a href="https://simon.fish">Add Test template here</a>
,
```

### Actual behavior

It seems as though the final newline of the template file is rendered, even though this usually indicates the end of a file. [This is interpreted by browsers to mean whitespace.](https://stackoverflow.com/a/588386)

![image](https://user-images.githubusercontent.com/20680337/117316759-ca6b2b00-ae80-11eb-89cc-51492ebaecc1.png)

![image](https://user-images.githubusercontent.com/20680337/117317136-25048700-ae81-11eb-80fb-17cdc18c03b4.png)

The newline preceding the comma is rendered from the component template. The following whitespace is part of the view template.

### System configuration
**Rails version**:  6.1.3.2

**Ruby version**: 2.7.3

**Gem version**: 2.31.1

BlakeWilliams:
Thanks for opening this! We should definitely support this case.

Any interest in opening a pull request to fix this issue? Even if the PR only includes a failing test that would help us resolve the issue sooner.

boardfish:
Yeah, I'd be happy to do that. I'll get on it right away :sparkles:

boardfish:
Alright, done :+1:

boardfish:
That PR got merged, so I'll raise another one here to change those tests back to their original form. If the change to the ERB renderer fixed the issue here, then there shouldn't be a need for further work. Otherwise, I'll consider adding a config option to ViewComponent for this.

boardfish:
`ActionView::Base::Template::Handlers::ERB.strip_trailing_newlines = true` is the fix for this issue. Add it to your `config/application.rb` and it should work.

boardfish:
I'm updating the tests in #958 - can I get some advice on how best to do this (if at all just yet) given that the functionality is only in edge Rails?

boardfish:
Just encountered this again - if `render_in` is used, the final newline will still be rendered in spite of the setting. I imagine all that needs to be done is (pseudocode):

```rb
def render_in(view_context)
[...]

PR #914, opened 2021-05-06
Add test for rendering component from template without newline

boardfish:
Test case for #913 

### Summary

Adds a failing test that renders the same component twice in sequence to ensure they're not separated by a newline.

### DIFF
--- a/test/app/components/display_inline_component.html.erb
+++ b/test/app/components/display_inline_component.html.erb
@@ -1,0 +1,1 @@
+<span>Hello, world!</span>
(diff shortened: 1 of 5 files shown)


boardfish:
The test renders this (I've added newlines for easier parsing, but the rendered newlines are shown as \n):

```html
<!-- BEGIN app/components/display_inline_component.rb -->
<span>Hello, world!</span>\n
<!-- END app/components/display_inline_component.rb -->
<!-- BEGIN app/components/display_inline_component.rb -->
<span>Hello, world!</span>\n
<!-- END app/components/display_inline_component.rb -->\n
```

Will it be matching on the comments in the output? If so, it's not currently failing correctly.

boardfish:
Thanks Joel, I'd love to! Where would be best to discuss this?

PR #919, opened 2021-05-09
Add preview command line, format change, fix typo

mikepmunroe:
### Summary

I thought it would be nice to present the command line argument to generate a preview again before the first examples for localhost are listed in case someone new comes directly to Previews before reading https://viewcomponent.org/guide/#quick-start.
I put the list of localhost URLs in a list
I fixed a typo where `the` should come after `preview`

### DIFF
--- a/docs/guide/previews.md
+++ b/docs/guide/previews.md
@@ -33,9 +33,15 @@ class ExampleComponentPreview < ViewComponent::Preview
 end
 ```
 
-Which generates <http://localhost:3000/rails/view_components/example_component/with_default_title>,
-<http://localhost:3000/rails/view_components/example_component/with_long_title>,
-and <http://localhost:3000/rails/view_components/example_component/with_content_block>.
+Generate the preview using:
+```console
+bin/rails generate component Example title --preview
+```
+
+Which generates: 
+* <http://localhost:3000/rails/view_components/example_component/with_default_title>
+* <http://localhost:3000/rails/view_components/example_component/with_long_title>
+* <http://localhost:3000/rails/view_components/example_component/with_content_block>
 
 It's also possible to set dynamic values from the params by setting them as arguments:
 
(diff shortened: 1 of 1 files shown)


docs/guide/previews.md
```diff
@@ -33,9 +33,15 @@ class ExampleComponentPreview < ViewComponent::Preview
 end
 ```
 
-Which generates <http://localhost:3000/rails/view_components/example_component/with_default_title>,
-<http://localhost:3000/rails/view_components/example_component/with_long_title>,
-and <http://localhost:3000/rails/view_components/example_component/with_content_block>.
+Generate the preview using:
```
joelhawksley, line 36: I think `generate` was the wrong word to use in the existing docs. It should instead say something like `which can then be accessed at:`

docs/guide/previews.md
```diff
@@ -65,7 +71,7 @@ class ExampleComponentPreview < ViewComponent::Preview
 end
 ```
 
-You can also set a custom layout to be used by default for previews as well as the preview index pages via the `default_preview_layout` configuration option:
```
joelhawksley, line 68: I _believe_ this line is correct as it stands, but could be written more clearly.

docs/guide/previews.md
```diff
@@ -65,7 +71,7 @@ class ExampleComponentPreview < ViewComponent::Preview
...
+You can also set a custom layout to be used by default for previews as well as preview the index pages via the `default_preview_layout` configuration option:
```
joelhawksley, line 74:
```suggestion
To set a custom layout for previews and the previews index page, set: `default_preview_layout`:
```
```suggestion
You can also set a custom layout to be used by default for previews as well as preview the index pages via the `default_preview_layout` configuration option:
```

joelhawksley (commented):
👋🏻 thanks for pointing out the lack of clarity in these docs ❤️

PR #920, opened 2021-05-10
Add test for newline behavior

joelhawksley:
@boardfish thanks for the failing test!

@BlakeWilliams and I had a look into it this morning and believe we have pinpointed the issue: `test/app/components/display_inline_component.html.erb` had a trailing newline! 

We've removed the newline and added the ability to turn off template annotations in tests, and the test passes. 

Does this feel correct to you?

(Original PR: https://github.com/github/view_component/pull/914)


Spone:
My 2 cents: I think the issue also appears "in the wild" because some code editors automatically add a trailing newline when you save a file.

Also, related: the issue is probably specific to ERB templates. Slim squishes the whitespace by default.

joelhawksley:
> My 2 cents: I think the issue also appears "in the wild" because some code editors automatically add a trailing newline when you save a file.
> 
> Also, related: the issue is probably specific to ERB templates. Slim squishes the whitespace by default.

Right. I wonder if we might want to have a `strip_newlines` option that could be set per-component or globally?

boardfish:
Hey @joelhawksley, thanks for looking into this. I'm afraid this solution doesn't tackle my original issue.

> It seems as though the final newline of the template file is rendered, even though this usually indicates the end of a file.

As @Spone mentioned, trailing newlines are something of a standard ([here's one stance on why, if you're curious](https://unix.stackexchange.com/a/18789)). It's definitely something we lint for over @raisedevs and something [RuboCop](https://www.rubydoc.info/gems/rubocop/0.20.1/Rubocop/Cop/Style/FinalNewline) checks out.

A per-component or global option would be a great idea. :sparkles:

boardfish:
(Semi-related, but thank you to you and your fellow contributors for ViewComponent. I'm really glad it's here to make the view situation in Rails much nicer!)

boardfish:
The ability to turn off template annotations in tests is probably still of value to some, though, so it might be good to merge this for that. But the test should pass without the original file having been modified.

joelhawksley:
@boardfish I'm seeing the same error when rendering the template via a partial:

```erb
<%= render("inline") %><%= render("inline") %>
```

```
Failure:
IntegrationTest#test_does_not_render_additional_newline [/Users/joelhawksley/github/view_component/test/view_component/integration_test.rb:442]:
Expected "<!DOCTYPE html>\n<html>\n<head>\n  <title>ViewComponent - Test</title>\n  <link rel=\"stylesheet\" href=\"http://assets.example.com/assets/application-a431f1c297dd1ba0d918190bcb9fa2cce7956fdd4e66d8832e109d7fb298171c.css\" media=\"screen\" />\n</head>\n<body>\n  <span>Hello, world!</span>\n<span>Hello, world!</span>\n\n\n</body>\n</html>\n" to include "<span>Hello, world!</span><span>Hello, world!</span>".
```

Which makes me think that it's not _really_ ViewComponent's problem to solve.

joelhawksley:
That being said, I think we should land this PR so we have a clear example of what it _does_ currently take to avoid newlines in rendered output, even if it's not the ideal solution.

PR #921, opened 2021-05-11
Add experimental `_output_postamble` method

joelhawksley:
This PR adds an experiment `_after_render` method, as proposed in https://github.com/github/view_component/pull/677/files, for the sake of internal testing.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -91,12 +91,19 @@ module ViewComponent
       before_render
 
       if render?
-        render_template_for(@variant)
+        render_template_for(@variant) + _after_render
       else
         ""
       end
     ensure
       @current_template = old_current_template
+    end
+
+    # EXPERIMENTAL: Optional content to be returned after the rendered template.
+    #
+    # @return [String]
+    def _after_render
+      ""
     end
 
     # Called before rendering the component. Override to perform operations that depend on having access to the view context, such as helpers.
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -91,7 +91,7 @@ module ViewComponent
       before_render
 
       if render?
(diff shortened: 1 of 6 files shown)


lib/view_component/base.rb
```diff
@@ -91,14 +91,21 @@ def render_in(view_context, &block)
...
         ""
       end
     ensure
       @current_template = old_current_template
     end
 
+    # EXPERIMENTAL: Optional content to be returned after the rendered template.
+    #
+    # @return [String]
+    def _after_render
```
BlakeWilliams, line 105: Thinking a bit more broadly about this callback, I'd love to start chatting about/looking into compiler level hooks instead of relying on component level hooks.

Ignoring that and focusing on naming, thoughts on something like the following:

```suggestion
    def _output_postamble
```

`_after_render` sounds like a callback since we have `before_render` and not something that will impact the output of the component.

BlakeWilliams (approved):

lib/view_component/base.rb
```diff
@@ -91,14 +91,21 @@ def render_in(view_context, &block)
       before_render
 
       if render?
-        render_template_for(@variant)
+        render_template_for(@variant) + _output_postamble
```
cbeer, line 94: It looks like with this change means `call` must return a string (or something that works with `+` at least)?

In our usage, we've overridden `call` a couple places (using the [inline rendering](https://viewcomponent.org/guide/templates.html#inline)) and happened to return some other things like `nil` (or the contents of a slot, an instance of ViewComponent::SlotV2). I guess something later in the pipeline string-ified them but now breaks with `undefined method `+' for nil:NilClass`.

We might be in the wrong, but would it be useful to clarify in the documentation that `call` must return a string at least?

PR #922, opened 2021-05-11
Fix invalid slots with_content example

BlakeWilliams:
The example in the documentation won't work because no arguments and no
block are being passed to the slot, so it is attempting to get the slot
instead of setting it. This example will raise when run.

This updates the example to pass an argument to the slot, setting it
instead of fetching it.


Spone (approved):
:+1:

PR #925, opened 2021-05-18
Ensure session is set on request

Tonkpils:
### Summary

Rails 7.0 deprecated writing to a session when the session is disabled
and defaults a session to a disabled session. https://github.com/rails/rails/pull/42231

In the past, ActionDispatch::TestRequest.create would set the
TestSession explicitly but this is not the case anymore and the default
session is used. With the PR mentioned above, this raises an exception
when testing view components that implement sessions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Fix test helper compatibility with Rails 7.0, TestRequest, and TestSession.
+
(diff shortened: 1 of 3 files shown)


BlakeWilliams (approved):
Thanks for fixing this!

Would you mind adding a changelog entry for this change?

Tonkpils:
Done! @BlakeWilliams

PR #926, opened 2021-05-18
Fix bug with helpers not respecting latest assigned value for a member variable

cheshire137:
### Summary

This addresses a bug I experienced where a method in my view component that called a helper method via `helpers` was getting an out-of-date return value. I was able to fix the bug locally, in my Rails app that uses the view_component gem, by changing my component method from using `helpers.some_helper` to `view_context.some_helper`. This PR makes the same change to the gem so that workaround won't be necessary in my app. We can return `view_context` directly from ViewComponent::Base instead of going through `controller` to get its `view_context`.

### Other Information

Two tests failed after I made my fix. I was able to make them pass again by changing them to be integration tests. My thinking is this is acceptable, that we're not reducing our test coverage and that the spirit of the original tests remains intact with them as integration tests.


lib/view_component/base.rb
```diff
@@ -158,7 +158,7 @@ def controller
     # @return [ActionView::Base]
     def helpers
       raise ViewContextCalledBeforeRenderError, "`helpers` can only be called at render time." if view_context.nil?
-      @helpers ||= controller.view_context
+      @helpers ||= view_context
```
cheshire137, line 161: Making this change resulted in a test error:

```
Error:
ViewComponentTest#test_renders_helper_method_within_nested_component:
NoMethodError: undefined method `message' for #<ContainerComponent:0x00007fd6ac718d08>
    /Users/cheshire137/code/github/view_component/test/app/components/helpers_proxy_component.html.erb:1:in `call'
```

I made the test pass by turning it into an integration test in test/view_component/integration_test.rb instead.

cheshire137:
Hmm, all tests are passing for me locally when I run `bundle exec rake`, but on CI there's a failure:

```
Failure:
ViewComponentTest#test_renders_button_to_component [/home/runner/work/view_component/view_component/test/view_component/view_component_test.rb:155]:
expected to find css "input[type='hidden'][name='authenticity_token']" but there were no matches
```

I wonder if this is another one that would pass if I made it an integration test.

cheshire137:
Another failing CI build but this one looks unrelated:

```
Error:
CssCoverageTest#test_docs_do_not_generate_primer_css_classes_that_do_not_exist:
Errno::ENOENT: No such file or directory @ rb_sysopen - /home/runner/work/view_component/view_component/primer_view_components/node_modules/@primer/css/dist/stats/primer.json
    /home/runner/work/view_component/view_component/primer_view_components/test/lib/css_coverage_test.rb:41:in `read'
    /home/runner/work/view_component/view_component/primer_view_components/test/lib/css_coverage_test.rb:41:in `setup'
```

I'll try rebuilding.

cheshire137:
The rebuild didn't make it pass. I consulted with @BlakeWilliams though who thought the tests might be flaky, since the majority of the CI builds are passing. Going to open this up for review!

CONTRIBUTING.md
```diff
@@ -15,6 +15,8 @@ Please note that this project is released with a [Contributor Code of Conduct](C
 1. [Fork](https://github.com/github/view_component/fork) and clone the repository.
 1. Configure and install the dependencies: `bundle`.
 1. Make sure the tests pass on your machine: `bundle exec rake`.
+    - Use `bundle exec ruby -I test test/view_component/some_test_file.rb` to run all tests in a single file.
+    - Use `bundle exec ruby -I test test/view_component/some_test_file.rb -n test_some_particular_test` to run a single test.
```
BlakeWilliams, line 19: I love that this is now documented!

test/app/views/integration_examples/member_var_in_controller.html.erb
```diff
@@ -0,0 +1 @@
+<%= render HelpersProxyComponent.new %>
```
BlakeWilliams, line 1: I'm curious if we could reproduce the test exception we saw above.

My hunch is that changing this to render the `ContainerComponent` would cause the same exception since `view_context` will sometimes be component instances instead of the original `view_context` that has helpers included, instance variables defined, etc.

BlakeWilliams (commented):
This looks great, but I'd like to verify that rendering the `ContainerComponent` from the integration test won't cause the same exception as we saw in the unit tests before merging.

test/app/views/integration_examples/member_var_in_controller.html.erb
cheshire137: I'll try adding a new app/views/integration_examples view that renders `ContainerComponent` and corresponding controller endpoint to test this!

cheshire137: Yep, it definitely fails! Pushed up the test in https://github.com/github/view_component/pull/926/commits/02428f3573bad20d5fc366e7ab16140136f4e188.

```
Error:
IntegrationTest#test_renders_container_component_that_uses_helpers:
ActionView::Template::Error: undefined method `message' for #<ContainerComponent:0x00007f984a18e680>
    /Users/cheshire137/code/github/view_component/test/app/components/helpers_proxy_component.html.erb:1:in `call'
    /Users/cheshire137/code/github/view_component/lib/view_component/compiler.rb:73:in `render_template_for'
    /Users/cheshire137/code/github/view_component/lib/view_component/base.rb:94:in `render_in'
```

cheshire137: > view_context will sometimes be component instances instead of the original view_context

So we need a way to get at that original view context. 🤔 It feels strange to me that this `view_context` method would sometimes have a component instance in it, why does that happen?

BlakeWilliams: `ViewComponent::Base` inherits from `ActionView::Base` which implements `render` like: https://github.com/rails/rails/blob/91f0db1353e1b24e2d26ebbbadce510735f66848/actionview/lib/action_view/helpers/rendering_helper.rb#L43

When we call `render` in a component, it's passing `self` as an argument and not the original `view_context`. We could try to thread the original `view_context` all the way down, but I think that might introduce some new challenges (that are probably worth fixing anyways).

cheshire137: Adding this line in ViewComponent::Base#helpers was informative:

```ruby
puts "has helpers? #{view_context.class <= ActionController::Base::HelperMethods ? 'yes' : 'no'}"
```

When I `bundle exec rake`, I see a lot of 'yes' lines and one 'no' right before the test error:

```
has helpers? yes
.....................has helpers? yes
has helpers? yes
.has helpers? no
E

Error:
IntegrationTest#test_renders_container_component_that_uses_helpers:
ActionView::Template::Error: undefined method `message' for #<ContainerComponent:0x00007fe712a673d0>
```

cheshire137:
I'm going to close this as stale since I haven't looked at it in a while.

PR #927, opened 2021-05-19
Install node modules for PVC test runs

BlakeWilliams:
PVC hasn't passed consistently in a while which I think may be due to
not having the correct node modules installed.

This installs node in the pvc workflow, caches the `node_modules`
directory, and installs node modules to hopefully make the PVC tests
pass more consistently.


cheshire137:
Possibly related to some failures I hit in https://github.com/github/view_component/pull/926 that I couldn't explain, thanks!

manuelpuyol (approved):

PR #929, opened 2021-05-19
Define component helpers automatically

### LINKED ISSUE #930: Define component helpers
### Summary

Define an helper for each component to make them easily usable from the view/controller:

```ruby
class ExampleComponent < ViewComponent::Base
  def initialize(argument, keyword:)
    @argument = argument
    @keyword = keyword
  end

  def call
    tag.div do
      concat(tag.span(@argument))
      concat(tag.span(@keyword))
      concat(content)
    end
  end
end
```

This will *automatically* define the helper `example_component`:

```ruby
# from the view
example_component("Hello", keyword: "World") do
  tag.span("!")
end

# <div>
#   <span>Hello</span>
#   <span>World</span>
#   <span>!</span>
# </div>

# This is equivalent to:
render ExampleComponent.new("Hello", keyword: "World") do
  tag.span("!")
end
```

The main idea is to be closer to other Rails helpers, so a frontend developer uses all components the same way:

```ruby
# view

# a rails helper
form_with user do |f|
  # ...
end

# another rails helper
image_tag(user.avatar_url)

# a view component helper
badge_component(user)
``` 

What do you think? I've started a PR to show what it could look like: https://github.com/github/view_component/pull/929

BigBigDoudou:
> // suggestion // work in progress

### Summary

Define an helper for each component to make them easily usable from the view/controller:

```ruby
class ExampleComponent < ViewComponent::Base
  def initialize(argument, keyword:)
    @argument = argument
    @keyword = keyword
  end

  def call
    tag.div do
      concat(tag.span(@argument))
      concat(tag.span(@keyword))
      concat(content)
    end
  end
end
```

This will *automatically* define the helper `example_component`:

```ruby
helpers.example_component("Hello", keyword: "World") do
  tag.span("!")
end

# <div>
#   <span>Hello</span>
#   <span>World</span>
#   <span>!</span>
# </div>
```

The main idea is to be closer to other Rails helpers, so a frontend developer uses all components the same way:

```ruby
# view

# a rails helper
form_with user do |f|
  # ...
end

# another rails helper
image_tag(user.avatar_url)

# a view component helper
badge_component(user)
``` 

### Other Information

- [ ] Ensure components are loaded before helpers are defined

resolve #930


app/helpers/view_component_helper.rb
```diff
@@ -0,0 +1,10 @@
+module ViewComponentHelper
+  # For each component, define an helper that renders the component
+  # Forward arguments, keywords and block
+  ViewComponent::Base.subclasses.each do |component|
```
BigBigDoudou, line 4: ⚠️ Are we sure components are loaded here?

app/helpers/view_component_helper.rb
```diff
@@ -0,0 +1,12 @@
+module ViewComponentHelper
+  # For each component, define an helper that renders the component
+  # Forward arguments, keywords and block
+  Dir[Rails.root.join("app/components/*.rb")].each { |file| require file }
```
BigBigDoudou, line 4: ⚠️ How to ensure components are loaded here? Without these line, it works with ruby 2.7 + rails 7 but not with ruby 2.6.7 + rails 6.1.3.2... With these line it works but it does not seem to be the best way to do that...

collimarco:
It's sad that this has been closed, I think it's a really good idea.

Why did you close it? Have you found a better solution?

PR #933, opened 2021-05-20
Bug test/block becomes empty hash

### LINKED ISSUE #931: Wrong number of arguments in slot initialize
### Steps to reproduce

Create this component:

```ruby
class ListComponent < ViewComponent::Base
  renders_many :items, "ListItemComponent"

  def call
    tag.ul do
      items.each { |item| concat(item) }
    end
  end

  class ListItemComponent < ViewComponent::Base
    def initialize(value: nil)
      @value = value
    end

    # we want the same behaviour than Rails `tag`
    # i.e. be able to call `tag.div("Hello World")` and `tag.div { "Hello World" }`
    def call
      tag.li(@value || content)
    end
  end
end
```

Then use it with:

```ruby
render(AbstractComponent::ListComponent.new) do |list|
  list.item(value: "Hello World")
  list.item do
    tag.span("Hello World")
  end
end
```

Ensure it works.

Now, replace the keyword argument `value` by a simple argument:

```ruby
def initialize(value = nil)
  @value = value
end
```

and

```ruby
list.item("Hello World")
```

You should have an error:

```
ActionView::Template::Error (wrong number of arguments (given 2, expected 1))
    1: <% if ViewComponent::Base.render_monkey_patch_enabled || Rails.version.to_f >= 6.1 %>
    2:   <%= render(@render_args[:component], @render_args[:args], &@render_args[:block])%>
    3: <% else %>
    4:   <%= render_component(@render_args[:component], &@render_args[:block])%>
    5: <% end %>
```

This second argument is `{}` (due to `&@render_args[:block]`?). So the workaround is:

```ruby
def initialize(value = nil, _)
  @value = value
end
```

### Expected behavior

We should be able to use arguments in slots.

### Actual behavior

Raises error `ActionView::Template::Error (wrong number of arguments (given 2, expected 1))`.

### System configuration
**Rails version**: 6.1.3.1

**Ruby version**: ruby 2.6.6p146

**Gem version**: 2.31.1

BigBigDoudou:
# Summary

When mixing argument and block in a slot component, block "becomes" an empty hash argument, leading to an error: `ActionView::Template::Error (wrong number of arguments (given 2, expected 1))`.

See https://github.com/github/view_component/issues/931 for full description and steps to reproduce.

This PR contains the failing tests.

# Other Information

⚠️ Test pass with Ruby 2.7.0 and Rails 7.0.0. When downgrading to Ruby 2.6.6 and Rails 6.1.3.1 (I had the errors with these versions) tests fail.

resolve #931


Gemfile
```diff
@@ -6,4 +6,4 @@ gemspec
 rails_version = "#{ENV['RAILS_VERSION'] || 'main'}"
 
 gem "capybara", "~> 3"
-gem "rails", rails_version == "main" ? { git: "https://github.com/rails/rails", ref: "main" } : rails_version
+gem "rails", "6.1.3.2"
```
BigBigDoudou, line 9: We do not want version `7.0.0.alpha` that needs ruby version `2.7.0` in which the bug does not appear anymore... Downgrade Rails so we can validate that test fail before the fix with a previous version.

⚠️ b4bbbaee19dc6012ac543b367243363f673ba4c2 should be revert before merge!

lib/view_component/slotable_v2.rb
```diff
@@ -221,10 +221,18 @@ def set_slot(slot_name, *args, **kwargs, &block)
 
       # If class
       if slot_definition[:renderable]
-        slot._component_instance = slot_definition[:renderable].new(*args, **kwargs)
+        slot._component_instance = if kwargs.present?
+                                     slot_definition[:renderable].new(*args, **kwargs)
+                                   else
+                                     slot_definition[:renderable].new(*args)
+                                   end
```
BigBigDoudou, line 228: If there is no keyword arguments then `**{}` becomes `{}`:

```ruby
def foo; end

foo(**{}) # equivalent to `foo({})`
ArgumentError: wrong number of arguments (given 1, expected 0)
```

BlakeWilliams: Thanks for opening this!

I'm just now getting a chance to take a peek. I think the issue is how we're delegating args + kwargs and the (unfortunate) differences between Ruby versions.

In order to get things working 100% I think we can:

1. Call `ruby2_keywords(METHOD_NAME_AS_SYMBOL) if respond_to?(:ruby2_keywords, true)` on all methods that currently accept `**kwargs` (including methods defined via `define_method`)
2. Remove all uses of `**kwargs` and forward only `*args` since the `ruby2_keywords` method should ensure that kwargs work for 2.7+

If you'd like to give that a try, that would be awesome. No pressure though, I'm happy to make that change and push it up to this branch if you'd like!

BigBigDoudou: Ok so it seems to work like a charm! 👍 Tested with ruby 2.6.7 and 2.7.

I've reverted rails downgrade. Do you want me to clean the branche history?

BigBigDoudou: It's failing with Ruby 3.0 and I don't find how to fix it. Any clue?

BlakeWilliams: Sorry for the delayed response, I've been very busy this week.

> I've reverted rails downgrade. Do you want me to clean the branche history?
No need, we're not strict about commit history in this project, but thank you for offering!

>It's failing with Ruby 3.0 and I don't find how to fix it. Any clue?

Not totally sure yet, I'll try to find some time this week to take a peek and see if anything stands out.

jonspalmer:
I understand the need for this fix but boy does it feel ugly to be going in this direction. I hope we can revert this soon once 2.6 is EOL.

PR #941, opened 2021-06-02
Coerce rendered component to a string even if nothing is rendered

### LINKED ISSUE #937: Empty components/slots now raising whiny nil error in 2.32.0
### Steps to reproduce

If a component renders nothing, for whatever reason, the changes made in #921 result in an `ActionView::Template::Error` (well, technically a `NoMethodError`):

```
undefined method `+' for nil:NilClass
```

It looks like something similar was raised post-merge in https://github.com/github/view_component/pull/921#discussion_r638121759, but in our case, it's not just b/c we've overridden `call` to return `nil` (though we have in a couple instances!), but we also have a component with a slot that looks something like this:

```haml
= component.foo do
  - if something
    = something.interesting
```

### Expected behavior

This doesn't raise an error.

### Actual behavior

It raises an error starting in v2.32.0.

### System configuration
**Rails version**: 6.1.3.2

**Ruby version**: 2.7.3

**Gem version**: 2.32.0

agrobbin:
### Summary

If a component or slot is empty and the rendering of the component returns `nil`, we would attempt to add the postamble to `nil`, which throws an exception:

```
undefined method `+' for nil:NilClass
```

Now we are explicitly coercing the result of the render to a string, to ensure it can be appended to.

### Other Information

Closes #937.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Don't raise an error when rendering empty components.
+
+    *Alex Robbin*
+
 ## 2.32.0
 
 * Enable previews by default in test environment.
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -91,7 +91,7 @@ module ViewComponent
       before_render
 
       if render?
-        render_template_for(@variant) + _output_postamble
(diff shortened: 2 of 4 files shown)


joelhawksley (approved):
🙇🏻 🙏🏻

PR #943, opened 2021-06-02
Add `_iteration` parameter support to get access to collection iteration

### LINKED ISSUE #939: Pass iterator to collection instead of count
## Feature request
Currently when you render a component using the `with_collection` method, you can define a named parameter in the initializer for `#{collection_parameter}_counter` and receive the position counter of the component being rendered. This is tracked using `ActionView::PartialIterator` which provides a few other useful methods, like `first?` (which is knowable from the position count `count == 1`) and `last?` (which is _not_ knowable from the position count `count == ?`)

### Motivation

It can be helpful to modify how the template renders if it is the first or last element in the collection, but it's currently not possible to do that for the last element of a collection because it is not able to know that it is last. Since we already are using a `PartialIterator` to keep track of this, why not just pass the iterator to the components instead of the position count?

This would mimic the current capabilities of rendering a partial as a collection, which exposes the `PartialIterator` with the `#{partial_name}_iteration` local variable.

willcosgrove:
Closes #939

### Summary

This adds support for a `{collection_name}_iteration` named parameter for the component initializer. It will automatically be populated with the current collection iteration object, which responds to methods: `first?`, `last?`, `index`, and `size`.

This keeps things more in line with Rails' own partial collection rendering, which exposes both a `{collection_name}_counter` and a `{collection_name}_iteration` local variable to the partials being rendered.

### Other Information

I added docs and a changelog entry in separate commits in case they weren't wanted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## main
 
+* Add support for `_iteration` parameter when rendering in a collection
+
+    *Will Cosgrove*
+
 * Don't raise an error when rendering empty components.
 
     *Alex Robbin*
--- a/docs/guide/collections.md
+++ b/docs/guide/collections.md
@@ -88,3 +88,24 @@ end
   <%= @counter %> <%= @product.name %>
 </li>
 ```
+
+## Collection iteration
+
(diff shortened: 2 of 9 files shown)


joelhawksley (approved):
😍 thank you for the clean PR, merging!

PR #945, opened 2021-06-03
tests: check if within is supported.

KonnorRogers:
### Summary

A quick test to see if `within` works with minitest.

### Other Information

Speaking with @joelhawksley it appears there some uncertainty around supporting `within`. it works with RSpec, but unsure with minitest.

EDIT: 

`within` is supported under RSpec since it gets aliased under ProxyMatchers. 

https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/RSpecMatcherProxies#within-instance_method

There is no equivalent for minitest, and the base `within` provided by Capybara relies on the session which as far as I can tell is only set in tests which use a controller and is scoped under `request`.

`request.session.within()`

Anyways, Capybara docs recommend just using chaining instead of `within`

https://rubydoc.info/github/teamcapybara/capybara/master/Capybara/Session#within-instance_method

It is worth noting that Capybara scopes `page` to the `within` so using an `expect(page).to have_text("blah")` has different effects when used within a `within` block as opposed to outside of a `within` block. All that to say, there are workarounds and I dont think this is a show stopper.

### DIFF
--- a/test/view_component/slotable_test.rb
+++ b/test/view_component/slotable_test.rb
@@ -37,18 +37,20 @@ class SlotableTest < ViewComponent::TestCase
 
     assert_selector(".card.mt-4")
 
-    assert_selector(".title", text: "This is my title!")
+    within(".card.mt-4") do
+      assert_selector(".title", text: "This is my title!")
 
-    assert_selector(".subtitle", text: "This is my subtitle!")
+      assert_selector(".subtitle", text: "This is my subtitle!")
 
-    assert_selector(".tab", text: "Tab A")
-    assert_selector(".tab", text: "Tab B")
+      assert_selector(".tab", text: "Tab A")
+      assert_selector(".tab", text: "Tab B")
 
-    assert_selector(".item", count: 3)
-    assert_selector(".item.highlighted", count: 1)
-    assert_selector(".item.normal", count: 2)
+      assert_selector(".item", count: 3)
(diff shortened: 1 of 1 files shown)


joelhawksley:
@ParamagicDev thanks for digging into this! I'm wondering if nested `assert` blocks like this https://github.com/primer/view_components/pull/613/files#diff-16cb864a5d937ab896d96307338b86ad788ad440b349c5ad5209024333a8e41eR95 are functionally equivalent to `within`. Do you know?

javierm:
> it appears there some uncertainty around supporting within. it works with RSpec

@ParamagicDev Correct me if I'm wrong :pray:. There seems to be a trap I keep falling into (several times per month :sweat_smile:). Using `within` in a component spec (with RSpec) causes the expectations inside the block to always pass, because the `within` method is actually an alias for the `be_within` matcher in RSpec.

Including the `Capybara::DSL` module in component specs correctly loads the `within` method, as it includes the ProxyMatchers you mentioned. However, when doing so the test crashes:

```
undefined method `within_element' for class `Capybara::Node::Simple'
```

This is pretty much the same error we get in the test provided in this pull request if we replace `within` with `page.within` or `page.within_element`. After all, Capybara's `within` method is an alias to `page.within` AFAIK.

joelhawksley:

PR #947, opened 2021-06-03
add docs, test for #with_variant

joelhawksley:
This PR adds documentation and a (missing 😱) test for `#with_variant`.

I also took the liberty of reconfiguring the generator specs to not create a `tmp` directory in the `test` folder, as it resulted in `bundle exec rake` running the generated tests on my machine.

### DIFF
--- a/test/generators/component_generator_test.rb
+++ b/test/generators/component_generator_test.rb
@@ -8,7 +8,7 @@ Rails.application.load_generators
 
 class ComponentGeneratorTest < Rails::Generators::TestCase
   tests Rails::Generators::ComponentGenerator
-  destination File.expand_path("../../tmp", __dir__)
+  destination Dir.mktmpdir
   setup :prepare_destination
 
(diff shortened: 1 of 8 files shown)


test/generators/component_generator_test.rb
```diff
@@ -8,7 +8,7 @@
 
 class ComponentGeneratorTest < Rails::Generators::TestCase
   tests Rails::Generators::ComponentGenerator
-  destination File.expand_path("../tmp", __dir__)
+  destination File.expand_path("../../tmp", __dir__)
```
BlakeWilliams, line 11: Could it make sense to use the OS's tempdir? https://ruby-doc.org/stdlib-2.5.3/libdoc/tmpdir/rdoc/Dir.html#method-c-tmpdir

BlakeWilliams (approved):
LGTM, thanks for fixing the tmp dir issue!

Issue #670, opened 2021-03-18
Remove the Component suffix from components


BlakeWilliams:
As we've written and used components the `Component` suffix has been questioned a few times. Looking at current uses of components, it looks like the namespaces used are unique even when the `Component` suffix is removed. I think this is due to following the naming convention of views, which is to use the plural paths in the namespace. e.g. `users/index.html.erb` -> `Users::IndexComponent`/`Users::Index`.

Lately I've been thinking about components as "models that return markup". Models don't have a suffix (and neither do views, although they are usually templates), so maybe components shouldn't either?

Here's the pro's and con's I can think of at the moment:

* **Pro**: Less boilerplate and better readability. No more `render Primer::LinkComponent.new`, we can now call `render Primer::Link.new`. Once you learn that components are objects that can be rendered, the `Component` suffix loses a lot of its value.
* **Pro**: Component names now completely align with the `views` naming conventions. This will make it easier to put components in the `views` directory if we go down that path.
* **Con**: It's no longer immediately clear that your class/object is a component.

keithamus:
I'm 100% _for_ this idea. I think the `Con` is intrinsic to all software development: it's hard to know what type a "thing" is. The existing `Component` suffix is a bit like [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation), I feel as there is a somewhat low enthusiasm for Hungarian Notation in general.

jaredcwhite:
So instead of a `Post` model and a `PostComponent` view, what would be the recommended namespaced solution? `Posts::Item`? `Components::Post`? I think I'm on board with this idea but I'd want to name everything this way and not go piecemeal (aka some with `Component` suffix and some not).

BlakeWilliams:
> So instead of a `Post` model and a `PostComponent` view, what would be the recommended namespaced solution? `Posts::Item`? `Components::Post`? I think I'm on board with this idea but I'd want to name everything this way and not go piecemeal (aka some with `Component` suffix and some not).

Great question! I agree we wouldn't want to have some components with the suffix, and some without.

If it's a component you're rendering from a controller, I really like `Posts::Index` or `Posts:Show`. If it's a generic post that you want to render on both the index and show pages, I like`Posts::Single`, but you could also follow Rails via `Posts::Post` (since that aligns with `app/views/post/_post.html.erb`). The latter would mean that you would need to access the `Post` model via `::Post`, but I think that's an acceptable trade-off.

boardfish:
[...]

PR #951, opened 2021-06-07
Can set custom templates explicitly now.

vikks:
### Summary

When working with many components, having an option to name custom templates and set them in 
the components makes editing and fixing code much easier rather than having to double check every time
a template file is opened for editing.

```ruby
# app/components/image/component.rb

class Image::Component < ViewComponent::Base
  # Uses app/components/image/image_template.html.erb
  template "image_template"
end
```
```ruby
# app/components/image/component.rb

class Button::Component < ViewComponent::Base
  # Uses the default app/components/button/component.html.erb
end
```
```
components/
   button/
     component.html.erb
     component.rb
   image/
     image_template.html.erb
     component.rb  
     controller.js
```

### Other Information

Added test files for CustomTemplate component and its corresponding html template.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## 2.33.0
 
+* Add support for custom `template` option
+
+    *Vikas Raveendran*
+
 * Add support for `_iteration` parameter when rendering in a collection
 
     *Will Cosgrove*
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 ## main
 
-## 2.33.0
-
 * Add support for custom `template` option
 
     *Vikas Raveendran*
+
+## 2.33.0
 
 * Add support for `_iteration` parameter when rendering in a collection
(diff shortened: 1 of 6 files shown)


Sergei2208 (commented):
test/view_component/view_component_test.rb

Sergei2208:
Good

Sergei2208:
Good

joelhawksley:
@vikks thank you for taking the time to open a PR with this idea!

Per https://github.com/github/view_component/issues/670, we are planning to remove the recommendation for the `Component` suffix naming convention. Given that context, I don't think we should add this API to the project. 

I would be open to making it possible to change configure this option via a gem, in a similar manner to @boardfish's approach in https://github.com/github/view_component/issues/670#issuecomment-822349700.

I'm going to go ahead and close this PR for housekeeping purposes, but I am happy to continue discussing it. 

Please do continue to share ideas you have for ViewComponent- the project is built through the work of folks from the community just like you ❤️

PR #958, opened 2021-06-10
Update tests for trailing newlines

boardfish:
### Summary

The tests introduced in #920 were made to pass by removing the trailing newline. These were supposed to check that trailing newlines in template files would not become part of the output. Now that the source issue has reached a conclusion thanks to an update to `ActionView::Template::Handlers::ERB`, the implementation for the test can be updated to reflect this. It also makes clear that we're dependent on that class for this change.


joelhawksley:
@boardfish we'll likely need to `bundle update rails` here. I tried to do it for you, but I wasn't able to push to your branch ❤️

boardfish:
The test matrix runs against many versions of Rails, and most won't have that option on the ERB template handler. Is there something that can be done here? My instinct is to skip that test.

The reason that the build is failing seems to be unrelated to that, though.

joelhawksley:
> My instinct is to skip that test unless we're running against main

Yes, let's only have this test run if the Rails::VERSION::MAJOR >= 7.

joelhawksley:
> The reason that the build is failing seems to be unrelated to that, though.

😬 I've seen this CI issue before. Rebuilding _usually_ helps. I think it's due to our usage of mixed versions of Bundler, which I've been meaning to look into. I'll take this opportunity to do just that now ❤️

joelhawksley:
@boardfish the `bundler` fix is in `main` if you wanted to rebase 😄

test/view_component/integration_test.rb
```diff
@@ -437,10 +437,14 @@ def test_renders_the_inline_component_preview_examples_with_default_behaviour_an
...
 
   def test_does_not_render_additional_newline
+    skip unless Rails::VERSION::MAJOR >= 7
     without_template_annotations do
+      ActionView::Template::Handlers::ERB.strip_trailing_newlines = true
       get "/rails/view_components/display_inline_component/with_newline"
       assert_includes response.body, "<span>Hello, world!</span><span>Hello, world!</span>"
     end
+  ensure
+    ActionView::Template::Handlers::ERB.strip_trailing_newlines = false  unless Rails::VERSION::MAJOR >= 7
```
BlakeWilliams, line 447:
```suggestion
    ActionView::Template::Handlers::ERB.strip_trailing_newlines = false if Rails::VERSION::MAJOR >= 7
```

boardfish:
Just needed a rebase after that.

joelhawksley (approved):

PR #960, opened 2021-06-11
use Bundler v2

joelhawksley:
I think it's time for us to move to Bundler v2. Having both
v1 and v2 in our CI workflow has been a source of headaches
and friction for some time. Seeing it trip up @boardfish
in https://github.com/github/view_component/pull/958 was
the last straw for me.

There are also now security concerns around Bundler v1
that to my knowledge are unresolved.

Removing Bundler v1 means that we can't easily test
Ruby 2.4 in CI. I'm OK with this, as we've never had a 2.4
compatibility issue caught by running against it.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,14 @@
 
 _This project is intended to be a safe, welcoming space for collaboration. Contributors are expected to adhere to the Contributor Covenant code of conduct._
 
+ViewComponent would only be a shadow of what it is today if it weren't for the work of people like you. Thank you for your interest in contributing and we look forward to
+
+How you can help
+
+open a PR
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -2,6 +2,14 @@
 
 _This project is intended to be a safe, welcoming space for collaboration. Contributors are expected to adhere to the Contributor Covenant code of conduct._
 
+ViewComponent would only be a shadow of what it is today if it weren't for the work of people like you. Thank you for your interest in contributing and we look forward to
```
BlakeWilliams, line 5: Is this sentence complete?

BlakeWilliams (approved):
one question, but LGTM

Issue #639, opened 2021-02-19
Parameter DSL


dkniffin:
## Feature request

Add a built-in parameter DSL to ViewComponent:

```rb
class Button < ViewComponent::Base
  property :size, 
    type: String, 
    default: "lg", 
    validations: { 
      inclusion: { in: ["sm", "gt", "md"] }
    }

  # lack of `default` value means it's required. 
  # However, if `fallback` is given, it will return that in production environments, so as to avoid exceptions
  property :text, fallback: "Lorem ipsum" 
end
```

### Motivation

@rosendi proposed an idea [here](https://github.com/github/view_component/issues/206#issuecomment-617588434) and [here](https://github.com/github/view_component/issues/206#issuecomment-617588434) for a DSL for defining the attributes for a ViewComponent. I agree with his reason for it: there's a lot of boilerplate that goes into creating a view component. 

In most cases, there's a few things I want to do in `initialize`:
1. Define the parameters coming, whether they are required, and what the default value is
2. Set an instance variable for the value. Most of the time this has the same name as the parameter. A lot of the time I also want to set this in attr_reader (though sometimes I would rather override the value coming in with some behavior to make it smarter)
3. I want to validate that the value is allowed.

The way to do this right now requires repeating the name of the parameter several times (arg in `initialize()`, instance variable, attr_reader). 

Additionally, using ActiveModel::Validations is less than ideal because has potential to raise exceptions in production. I think there's potential for a built-in DSL to leverage ActiveModel::Validation, while still making it safe, by allowing it to raise in a development environment, and in a production environment it can use the fallback values.

This would also avoid the issue of `def initialize` needing to sometimes take many many parameters, which makes the line very very long.

dzunk:
Interested in the outcome of this discussion. FWIW, we use [dry-initializer](https://dry-rb.org/gems/dry-initializer/3.0/) in all our view components for this exact purpose.

joelhawksley:
@dkniffin thanks for raising this issue!

I like the API you've proposed! However, I agree with @dzunk's approach to use `dry-initializer` for this functionality, as I'm wary of having ViewComponent be the arbiter of such a DSL. 

Perhaps this might be worth an update to the documentation? I'd be fine with recommending `dry-initializer`.

pinzonjulian:
Rails already has a couple of ways to enhance the experience of initializing objects that could be used in ViewComponent. 

## ActiveModel attribute assignment

[...]

PR #963, opened 2021-06-14
Add basic attributes API (experimental)

BlakeWilliams:
This adds a basic and experimental attributes API that is used to define the
properties that a component accepts when initialized.

The idea behind this API is to give the library more control over the
`initializer`, attempt to close API gaps like `helpers` sometimes not being
callable, and reduce boilerplate when assigning attributes.

As-is, the API introduces two new methods:

1. `requires :name` - defines a required attribute accessible via the `name` property.. If this value is not provided when initializing the component an `ArgumentError` is raised.
1. `accepts :name` - defines an optional attribute accessible via the `name` property. Optionally, a default value can be provided by passing a `default` kwarg. e.g. `accepts :name, default: "NullUser"`.

This API is not added to all components by default, but can be opted-into by
adding `include ViewComponent::Attributes` to a component.

Feedback/thoughts welcome!

### DIFF
--- a/test/view_component/attributes_test.rb
+++ b/test/view_component/attributes_test.rb
@@ -22,8 +22,13 @@ class AttributesTest < ViewComponent::TestCase
     end
   end
 
+  class MyInheritedAttributeComponent < MyAttributeComponent
+  end
+
   def test_basic_attributes
     render_inline MyAttributeComponent.new(title: "foo")
+
+    assert_selector('h1', text: 'foo')
   end
 
   def test_required_attribute_raises_if_missing
@@ -47,5 +52,11 @@ class AttributesTest < ViewComponent::TestCase
   def test_explicit_nil_overrides_default_values
     component = MyAttributeComponent.new(title: "foo", posted_at: nil)
     assert_nil component.posted_at
+  end
+
+  def test_inheritance_works
+    render_inline MyInheritedAttributeComponent.new(title: "foo")
+
+    assert_selector('h1', text: 'foo')
   end
 end
--- a/test/view_component/attributes_test.rb
+++ b/test/view_component/attributes_test.rb
@@ -31,6 +31,30 @@ class AttributesTest < ViewComponent::TestCase
     assert_selector("h1", text: "foo")
   end
 
+  def test_exposes_attributes
+    component = MyAttributeComponent.new(title: "foo", body: "hello world!")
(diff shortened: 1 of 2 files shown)


BlakeWilliams:
related to https://github.com/github/view_component/issues/639

test/view_component/attributes_test.rb
```diff
@@ -0,0 +1,98 @@
...
+    requires :title
+    accepts :body
+    accepts :posted_at, default: Date.today
+
+    def call
+      content_tag :div do
+        safe_join([
+          content_tag(:h1, @title),
+          content_tag(:p, @body),
+          content_tag(:date, @posted_at),
```
BlakeWilliams, line 19:
```suggestion
          content_tag(:h1, title),
          content_tag(:p, body),
          content_tag(:date, posted_at),
```

BlakeWilliams:
This API does somewhat deviate from Rails, but I think if Rails does decide to adopt ViewComponent (or parts of it) we can easily omit this extra module.

There is *some* overlap with [`ActiveModel::AttributeAssignment`](https://api.rubyonrails.org/classes/ActiveModel/AttributeAssignment.html). Maybe one potential path forward here is trying to build a similar API on top of it (maybe by introducing `attr_required`?)

pinzonjulian:
I believe Rails already has a powerful attribute assignment API: `ActiveModel::Attributes`. The problem is that it still needs a bit of work before it can become a public API in Rails. 

See this thread where Rafael França talks a little bit about it. 
https://twitter.com/rafaelfranca/status/1361444817578971139?s=20

The `AttributeAssignment` module is more of a convenient but simple way to initialise allowed parameters. 

How would you feel about checking in with Rails Core about the status of `ActiveModel::Attributes` and maybe help make it public to use it in VC?

jonspalmer:
I like something along the lines of this API. We've been just using ActiveModel::Attributes and we'll deal with it if it breaks down the line. I'm also experimenting with dry-initializer. 

What would be really terrific would be to have some common API that ViewComponent accepts and that we could write a number of "Adapters" that would translate dry-initializer -> VS attributes etc. We could cover common cases and consumers could write or contribute their own for more esoteric patterns. You could imagine auto detecting common cases too.

joelhawksley:
Closing as stale.

PR #967, opened 2021-06-16
rename @variant to @__vc_variant to prevent conflicts

joelhawksley:
In our pairing today, @boardfish and I realized that this instance variable is named such that it might be unintentionally conflicted with by consumers.

I ran with this idea and renamed a few other private instance variables.

Co-authored-by: Simon Fish <si@mon.fish>


Rakefile
```diff
@@ -7,7 +7,7 @@ require "yard"
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
   t.libs << "lib"
-  t.test_files = FileList["test/**/*_test.rb"]
+  t.test_files = FileList[ENV["TESTS"] || "test/**/*_test.rb"]
```
manuelpuyol, line 10: I'm pretty sure Minitest has this built in using the `TEST` env var
```sh
TEST=test/lib/**/*_test.rb bundle exec rake
```

joelhawksley: Ah, cool! I guess we can remove this from PVC then too, as I copied this line from there 😄 cc @khiga8

CONTRIBUTING.md
```diff
@@ -40,7 +40,7 @@ The maintainers team is @blakewilliams, @joelhawksley, @jonrohan, and @manuelpuy
 
 1. [Fork](https://github.com/github/view_component/fork) and clone the repository.
 1. Configure and install the dependencies: `bundle`.
-1. Make sure the tests pass on your machine: `bundle exec rake`.
+1. Make sure the tests pass on your machine: `bundle exec rake`. (Run a subset of tests by supplying a file glob to the test command: `TESTS="test/components/YOUR_COMPONENT_test.rb" bundle exec rake`)
```
manuelpuyol, line 43:
```suggestion
1. Make sure the tests pass on your machine: `bundle exec rake`. (Run a subset of tests by supplying a file glob to the test command: `TEST="test/components/YOUR_COMPONENT_test.rb" bundle exec rake`)
```

manuelpuyol (approved):

PR #968, opened 2021-06-16
Dynamically generate configuration docs using YARD

### LINKED ISSUE #953: Add documentation for configuration options
Per https://github.com/github/view_component/pull/952#discussion_r647465286, let's add a documentation section for configuration options to the API page, preferably generated via YARD.

cc @Spone

joelhawksley:
This PR adds documentation for configuration options, generated using YARD.

I've reused some code @khiga8 wrote for Primer ViewComponents to extract `mattr_accessor`s.

I also took the liberty of sorting the existing instance methods alphabetically.

[Closes #953]

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,7 @@
 require "bundler/gem_tasks"
 require "rake/testtask"
 require "yard"
+require "yard/mattr_accessor_handler"
 
 Rake::TestTask.new(:test) do |t|
   t.libs << "test"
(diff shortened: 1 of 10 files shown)


Rakefile
```diff
@@ -90,6 +91,35 @@ namespace :docs do
...
+        sort_by { |method| method[:name] }.
+        each do |method|
+
+        suffix =
+          if method.tag(:deprecated)
+            " (Deprecated)"
+          end
+
+        f.puts
+        f.puts("### #{method.sep}#{method.name}#{suffix}")
```
Spone, line 111:
```suggestion
        f.puts("### `#{method.sep}#{method.name}`#{suffix}")
```

lib/view_component/previewable.rb
```diff
@@ -2,51 +2,49 @@
...
-      #
-      mattr_accessor :show_previews, instance_writer: false
-
-      # Set the entry route for component previews through app configuration:
+      # Set the entry route for component previews:
       #
       #     config.view_component.preview_route = "/previews"
       #
-      # Defaults to +/rails/view_components+ when `show_previews' is enabled
+      # Defaults to `/rails/view_components` when `show_previews' is enabled.
```
Spone, line 37:
```suggestion
      # Defaults to `/rails/view_components` when `show_previews` is enabled.
```

Spone:
:+1: 

Do you think it could be useful to mention the `config.generators.template_engine` and `config.generators.test_framework` settings on this page as well? They are not specific to ViewComponent but I guess some people may look for this kind of setting on this page.

Also, what do you think about having a dedicated "Configuration" item in the website navigation? (maybe as a child of "API")

joelhawksley:
> They are not specific to ViewComponent but I guess some people may look for this kind of setting on this page.

@Spone that's a tricky one. I could go either way. I'll give it some thought, this PR isn't going to be done today 😄

joelhawksley:
@Spone I think I've gotten this to a good point for now. I ended up removing the backticks as they made visually scanning the page difficult to my eyes.

I think we should aim to flesh out the documentation in further PRs and then revisit the idea of splitting it up into pages.

manuelpuyol (approved):

PR #971, opened 2021-06-17
Validate collection parameter with Active Record attribute names

### LINKED ISSUE #970: `with_collection_parameter` validation fails when used with ActiveModel
### Steps to reproduce

1. Include Active Model on your component:

```rb
  include ActiveModel::Model
  include ActiveModel::Attributes
```

2. Define an attribute, which you'd hope to use as the `collection_parameter`:

```
        attribute :message, ActiveModel::Type::Value.new
        with_collection_parameter :message
```

3. Attempt to render the component via `with_collection`, passing in a valid `message`

### Expected behavior

No validation error should be raised and the collection should be rendered.

### Actual behavior

I receive [this error](https://github.com/github/view_component/blob/e443c1f80afc16b65b10eae9aff046e48b8e1be6/lib/view_component/base.rb#L368-L371): "initializer must accept `message` collection parameter.

### System configuration
**Rails version**: 6.1.3.2

**Ruby version**: 2.7.2

**Gem version**: 2.33.0

boardfish:
### Summary

If the component responds to `#attribute_names`, use those to validate that a change to `with_collection_parameter` is correct.

### Other Information

There are a couple of ways we could go about this that I can see:

- checking that the component will `respond_to? :attribute_names`, which is what I've done here
- checking that the component has specifically included `ActiveModel::Attributes`

I've elected to go with the former, since we might need to check that `ActiveModel::Attributes` is even defined otherwise. We could also choose not to check that and assume that the Rails installation includes it.

Closes #970


lib/view_component/base.rb
```diff
@@ -411,6 +411,8 @@ def iteration_argument_present?
       private
 
       def initialize_parameter_names
```
joelhawksley, line 413: Might we rename this to `parameter_names` as this is not just for the initializer now?

test/app/components/collection_parameter_with_active_model_component.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+class CollectionParameterWithActiveModelComponent < ViewComponent::Base
+  include ActiveModel::Model
+  include ActiveModel::Attributes
```
joelhawksley, line 5: Do we need to include both of these?

test/app/components/collection_parameter_with_active_model_component.rb
```diff
@@ -0,0 +1,13 @@
...
+
+  attribute :name, :string
+
+  with_collection_parameter :name
+
+  def call
+  end
```
joelhawksley, line 12:
```suggestion
  def call; end
```

test/app/components/missing_collection_parameter_with_active_model_component.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class MissingCollectionParameterWithActiveModelComponent < ViewComponent::Base
+  include ActiveModel::Model
+  include ActiveModel::Attributes
+
+  with_collection_parameter :name
+
+  def call
+  end
```
joelhawksley, line 10:
```suggestion
  def call; end
```

joelhawksley (commented):
Looking good! Mind adding a CHANGELOG entry describing the bug this fixes?

lib/view_component/base.rb
boardfish: I suppose there's a case for renaming it, but `parameter_names` is unqualified, so it's not as descriptive. The return value of this method returns the valid parameters for the method - even though something that includes Active Model takes a params hash instead of explicit keyword arguments, these are the only ones we expect to be valid. So perhaps `valid_initialize_parameter_names` is the right name for this now?

test/app/components/collection_parameter_with_active_model_component.rb
boardfish: Most likely not. I'll play Jenga with this.

lib/view_component/base.rb
joelhawksley: Let's just leave it as is 😄

BlakeWilliams:
👋 This looks good! Can you add a changelog entry when you get a chance?

boardfish:
Looks like something changed about `CHANGELOG.md` upstream. Wasn't quite sure how to fix that in the rebase.

pinzonjulian:
FYI: Even though ActiveModel::Attributes is widely used is a private API of Rails and it could in traduce breaking changes without warning. 

See this thread where Rafael França talks a little bit about it. 
https://twitter.com/rafaelfranca/status/1361444817578971139?s=20

Just wanted to make sure you were aware of that

boardfish:
[Thankfully, it ought to become public soon](https://twitter.com/rafaelfranca/status/1361470266782662656), as can be seen a little further down that thread.

boardfish:
Alright, I finally got around to rebasing and fixing the business with the changelog. This should be good to go now.

Spone (approved):

boardfish:
I'm struggling to make this compatible with Rails 5. I was pretty confident a class using ActiveModel would have its `attribute_types` available on an accessor. @pinzonjulian, would you happen to know any more about how this might work?

lib/view_component/base.rb
```diff
@@ -507,6 +507,10 @@ def iteration_argument_present?
...
+        return attribute_names.map(&:to_sym) if respond_to?(:attribute_names)
+
+        return self.class.attribute_types.keys.map(&:to_sym) if Rails::VERSION::MAJOR <= 5 and self.class.respond_to?(:attribute_types)
```
Spone, line 512:
```suggestion
        return attribute_types.keys.map(&:to_sym) if Rails::VERSION::MAJOR <= 5 and respond_to?(:attribute_types)
```

boardfish:
Ah, looks like I got close with that. Thanks for stepping in @Spone! Tests are passing and this one's good to merge.

Spone (approved):
:ok_hand:

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Validate collection parameter with Active Model attribute names
```
```suggestion
* Validate collection parameter with Active Model attribute names.
```

joelhawksley (approved):

PR #972, opened 2021-06-17
Move root docs into site, add symlinks

joelhawksley:
### Summary

This PR moves the code of conduct, changelog, and contributing documents into `docs`, preserving their presence in the root directory with symlinks.


manuelpuyol:
since you changed `CHANGELOG` you should update `script/release` with the new path

joelhawksley:
@manuelpuyol good call! Fixed.

manuelpuyol (approved):

Issue #894, opened 2021-05-04
Localized I18n templates / views


23tux:
## Feature request
When using Rails' `#render` method you can use localized views (see https://guides.rubyonrails.org/i18n.html#localized-views)

```
<%= render("my_component") %>

# _my_component.html.erb
my template

# _my_component.de.html.erb
my template in german
```

Now when a template is rendered, and `I18n.locale == :de` it renders `my template in german`.

When I try to define a ViewComponent template this way, an error is thrown:

```
<%= render(MyComponent.new) %>

ERROR: More than one template found for MyComponent. There can only be one default template file per component.
```

As there is an explicit error, I guess it is something that is not supported yet.

I'm happy to try for a PR if someone could give me some hints why this error is thrown and where to start.

### Motivation
We have a multi language platform, and sometimes there are requirements that apply to a certain language only and can't be met with pure CSS and translations. It would be great if we could use ViewComponents the same way we use this feature with standard Rails templates and partials.

joelhawksley:
@23tux I'm on board. Would you like to build this?

23tux:
@joelhawksley Sorry for the delay, it's been busy at work :)

Yes, I'd love to contribute! How shall we start?

lfalcao:
Any progress here?

joelhawksley:
@lfalcao no. Would you like to help?

lfalcao:
@joelhawksley sure, I'll take a look :+1:

lfalcao:
hey, sorry for the delay. I can continue the PR #977 if there's no one looking...

joelhawksley:
Closing as stale.

PR #977, opened 2021-06-23
Adds support to localized components files

lfalcao:
Related: #894 

I just combined the "localized" information from component files (my_component.**es**.html.erb) into variant attribute to avoid adding new parameters or functions and worked well :smile: 

It's not done yet, needs some work, more tests, benchmarks... but I wanted to see some opinions about this approach, if there's something else I'm not seeing.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -96,9 +96,8 @@ module ViewComponent
 
       if render?
         if defined?(I18n) && !I18n.locale.eql?(I18n.default_locale)
-          attr = (I18n.locale || I18n.default_locale).to_s.downcase.tr("-", "_")
-
-          custom_variant = [@__vc_variant, attr].compact.reduce { |s, x| "#{s}_#{x}" }
+          attr = (I18n.locale || I18n.default_locale).to_s.underscore
+          custom_variant = [@__vc_variant, attr].compact.join('_')
         end
 
         render_template_for(custom_variant || @__vc_variant).to_s + _output_postamble
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -95,9 +95,10 @@ module ViewComponent
       before_render
 
       if render?
-        if defined?(I18n.locale) && I18n.locale != :en
-          custom_variant = [@__vc_variant, I18n.locale.to_s.downcase.tr("-", "_")]
-          custom_variant = custom_variant.compact.reduce { |s, x| s + "_" + x }.to_sym
+        if defined?(I18n) && !I18n.locale.eql?(I18n.default_locale)
+          attr = (I18n.locale || I18n.default_locale).to_s.downcase.tr("-", "_")
+
+          custom_variant = [@__vc_variant, attr].compact.reduce { |s, x| "#{s}_#{x}" }
         end
 
         render_template_for(custom_variant || @__vc_variant).to_s + _output_postamble
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -95,12 +95,12 @@ module ViewComponent
       before_render
 
       if render?
-        custom_variant = [@__vc_variant]
         if defined?(I18n.locale) && I18n.locale != :en
-          custom_variant << I18n.locale.to_s.downcase.tr("-", "_")
+          custom_variant = [@__vc_variant, I18n.locale.to_s.downcase.tr("-", "_")]
+          custom_variant = custom_variant.compact.reduce { |s, x| s + "_" + x }.to_sym
         end
 
-        render_template_for(custom_variant.join("_").to_sym).to_s + _output_postamble
+        render_template_for(custom_variant || @__vc_variant).to_s + _output_postamble
       else
         ""
       end
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -95,7 +95,12 @@ module ViewComponent
       before_render
 
       if render?
-        render_template_for(@__vc_variant).to_s + _output_postamble
+        custom_variant = [@__vc_variant]
+        if defined?(I18n.locale) && I18n.locale != :en
+          custom_variant << I18n.locale.to_s.downcase.tr("-", "_")
+        end
+
+        render_template_for(custom_variant.join("_").to_sym).to_s + _output_postamble
       else
         ""
       end
--- a/lib/view_component/compiler.rb
+++ b/lib/view_component/compiler.rb
@@ -38,11 +38,9 @@ module ViewComponent
       templates.each do |template|
         # Remove existing compiled template methods,
         # as Ruby warns when redefining a method.
-        method_name = call_method_name(template[:variant]).to_sym
+        method_name = call_method_name(template[:variant])
 
-        if component_class.instance_methods.include?(method_name)
-          component_class.send(:undef_method, method_name)
-        end
+        component_class.send(:undef_method, method_name.to_sym) if component_class.instance_methods.include?(method_name.to_sym)
 
         component_class.class_eval <<-RUBY, template[:path], -1
           def #{method_name}
@@ -64,9 +62,7 @@ module ViewComponent
     attr_reader :component_class
 
     def define_render_template_for
-      if component_class.instance_methods.include?(:render_template_for)
-        component_class.send(:undef_method, :render_template_for)
-      end
+      component_class.send(:undef_method, :render_template_for) if component_class.instance_methods.include?(:render_template_for)
 
       variant_elsifs = variants.compact.uniq.map do |variant|
         "elsif variant.to_sym == :#{variant}\n    #{call_method_name(variant)}"
@@ -132,12 +128,12 @@ module ViewComponent
           pieces = File.basename(path).split(".")
 
           # if template is a localized file (ie file.country_code.html.erb)
-          # append into locale inside variant and fix regional dash (pt-BR, en-GB...)
+          # append locale inside variant and fix regional dash (pt-BR, en-GB...)
           if pieces.length >= 4
             variant = [
-              pieces.third.split("+").second&.to_sym,
-              pieces.second.downcase.tr("-", "_")
-            ].compact.reduce { |s, x| "#{s}_#{x}" }
+              pieces.third.split("+").second,
+              pieces.second.underscore
+            ].compact.join('_')
           else
             variant = pieces.second.split("+").second&.to_sym
           end
--- a/lib/view_component/compiler.rb
+++ b/lib/view_component/compiler.rb
@@ -38,8 +38,11 @@ module ViewComponent
       templates.each do |template|
         # Remove existing compiled template methods,
         # as Ruby warns when redefining a method.
-        method_name = call_method_name(template[:variant])
-        component_class.send(:undef_method, method_name.to_sym) if component_class.instance_methods.include?(method_name.to_sym)
+        method_name = call_method_name(template[:variant]).to_sym
+
+        if component_class.instance_methods.include?(method_name)
+          component_class.send(:undef_method, method_name)
+        end
 
         component_class.class_eval <<-RUBY, template[:path], -1
           def #{method_name}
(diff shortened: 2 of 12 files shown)


lib/view_component/base.rb
```diff
@@ -95,7 +95,12 @@ def render_in(view_context, &block)
       before_render
 
       if render?
-        render_template_for(@__vc_variant).to_s + _output_postamble
+        custom_variant = [@__vc_variant]
+        if defined?(I18n.locale) && I18n.locale != :en
```
Spone, line 99: I'm not sure about the `!= :en` part. What is the goal? What about using `I18n.default_locale` instead?

lib/view_component/compiler.rb
```diff
@@ -121,13 +126,23 @@ def template_errors
 
     def templates
       @templates ||= begin
-        extensions = ActionView::Template.template_handler_extensions
+        extensions = %w[builder erb haml html jbuilder raw ruby slim]
```
Spone, line 129: Is there a way to achieve this without hardcoding these values?

lib/view_component/compiler.rb
```diff
@@ -121,13 +126,23 @@ def template_errors
...
 
         component_class._sidecar_files(extensions).each_with_object([]) do |path, memo|
           pieces = File.basename(path).split(".")
+          variant = pieces.second.split("+").second&.to_sym
+
+          # grab localized templates and fix regional dash (pt-BR, en-GB...)
+          localized_template = pieces.second.downcase.tr("-", "_")
+
+          # if template is a localized (ie file.country_code.html.erb) append as variant
+          if !localized_template.match(/builder|erb|html|jbuilder|raw|ruby|slim|tablet|iphone|json|html|\+/).present?
```
Spone, line 139: Same here, can we avoid hardcoding (or at least reuse the values from above)?

Spone (commented):
Thanks for tackling this! A few quick comments/questions.

lib/view_component/compiler.rb
lfalcao: This is just a cosmetic change, the value of `ActionView::Template.template_handler_extensions` is exactly the array added. I think we should try to minimize Rails dependency, so the library could be used from other frameworks... but that's not the objective of this merge, so I removed it.

lib/view_component/compiler.rb
lfalcao: I changed how identify it, take a look in the next commit.

lib/view_component/base.rb
lfalcao: I'm still thinking about this.

The problem is that files without localized filename (ie "mycomponent.html.erb") generates a `call` function. If I consider `I18n.default_locale` or `I18n.locale` it will create a function `call_{locale}` that will break other things...  :thinking:

lfalcao:
Benchmark: 

localized components:
> 
> Warming up --------------------------------------
>           component:    13.862k i/100ms
>             partial:     2.296k i/100ms
> Calculating -------------------------------------
>           component:    139.812k (± 0.9%) i/s -      1.400M in  10.014746s
>             partial:     22.876k (± 3.2%) i/s -    229.600k in  10.047667s

main branch:

> 
> Warming up --------------------------------------
>           component:    13.007k i/100ms
>             partial:     2.440k i/100ms
> Calculating -------------------------------------
>           component:    129.883k (± 1.9%) i/s -      1.301M in  10.018044s
>             partial:     24.145k (± 3.6%) i/s -    241.560k in  10.021579s

Spone:
I'm doing some real world tests in a dummy app, with the following configuration:

```rb
I18n.available_locales = [:en, :es, :fr]
I18n.default_locale = :en
```

I'm adding a `FooComponent` with 3 partials:

```
app/components/foo_component.en.html.erb
# no partial for `es`
app/components/foo_component.fr.html.erb
app/components/foo_component.html.erb
```

- If I set `I18n.locale = :en` in my request, I get the `foo_component.html.erb` :arrow_right: I would expect `foo_component.en.html.erb`
- If I set `I18n.locale = :es` in my request, I get the `foo_component.html.erb` :heavy_check_mark: 
- If I set `I18n.locale = :fr` in my request, I get the `foo_component.fr.html.erb` :heavy_check_mark: 
- If I set no locale in my request, I get the `foo_component.html.erb` :arrow_right: I would expect the `I18n.default_locale` to be used, so `foo_component.en.html.erb`

Now If I'm removing this partial :

```diff
app/components/foo_component.en.html.erb
app/components/foo_component.fr.html.erb
- app/components/foo_component.html.erb
```

I get an error in all cases: `No such file or directory @ rb_sysopen - /home/lfalcao/review/app/components/foo_component.html.erb` 

I guess it should work without a "default" partial, what do you think?

lfalcao:
Thanks for reviewing it @Spone 

I think every case is expected as I used the `I18n.default_locale` as the default without locale information in the filename. I mean, if you set `I18n.default_locale` there's no need to have all your files with '.en'.... :thinking: 

There's a ([test](https://github.com/github/view_component/pull/977/files#diff-cd50ec609aed15d820305c1ab53ceebd6fa83e56f2bfe657236e5653ab810a9fR38)) for this. 

> If I set I18n.locale = :es in my request, I get an error undefined local variable or method 'call' for #<FooComponent:0x00007f9720c56610>

This happened because there's no `foo_component.es.html.erb`. You think we should to fallback to the `foo_component.html.erb` ?

> If I set no locale in my request, I get an error undefined local variable or method 'call' for #<FooComponent:0x00007f9720c56610>

Also expected, as the file `foo_component.html.erb` was deleted.

> If I set no locale in my request, I get the show.en.html.erb
> Actually, I never get the show.html.erb and I can remove it without getting errors

Without locale, but with `I18n.default_locale = :en` right?

Spone:
> This happened because there's no `foo_component.es.html.erb`. You think we should to fallback to the `foo_component.html.erb` ?

In this case, Rails localized views fall back to the default locale (so `foo_component.en.html.erb` in this case).

> Also expected, as the file `foo_component.html.erb` was deleted.

Yes but in this case Rails localized views fall back to the default locale as well.

> Without locale, but with `I18n.default_locale = :en` right?

Yes. Actually you can also remove it from the config since `:en` is the default value.

To sum up, I think the fall back behavior should be the same as Rails standard behavior with [localized views](https://guides.rubyonrails.org/i18n.html#localized-views), as described in the **EDIT** of my previous comment. What do you think?

lib/view_component/compiler.rb
```diff
@@ -125,9 +130,21 @@ def templates
...
         component_class._sidecar_files(extensions).each_with_object([]) do |path, memo|
           pieces = File.basename(path).split(".")
+
+          # if template is a localized file (ie file.country_code.html.erb)
+          # append into locale inside variant and fix regional dash (pt-BR, en-GB...)
+          if pieces.length >= 4
+            variant = [
+              pieces.third.split("+").second&.to_sym,
+              pieces.second.downcase.tr("-", "_")
+            ].compact.reduce { |s, x| "#{s}_#{x}" }
```
jonspalmer, line 140: Likely simpler as just
```suggestion
            variant = [
              pieces.third.split("+").second,
              pieces.second.underscore
            ].compact.join("_")
```

Does the result need to be a symbol?

test/test_helper.rb
```diff
@@ -33,6 +33,31 @@ def with_preview_paths(new_value)
...
+
+  apply_config = ->(c) { c.each { |ee| eval(ee) } }
+
+  begin
+    apply_config.call(configs)
+    yield
+  ensure
+    apply_config.call(old_config)
+  end
+end
```
jonspalmer, line 59: I understand the motivation here but the use of eval is scary and the assumption that each config is only set with an `=` seems like a bit of a stretch. This would be just as simple and more direct to just have a helper like 


```suggestion
def with_locale(locale, default_locale = I18n.default_locale)
  old_locale = I18n.locale
  old_default_locale = I18n.default_locale
  
  begin
    I18n.locale = locale
    I18n.default_locale = default_locale
    yield
  ensure
     I18n.locale = old_locale
     I18n.default_locale = old_default_locale
  end
end
```

lib/view_component/base.rb
```diff
@@ -95,7 +95,13 @@ def render_in(view_context, &block)
       before_render
 
       if render?
-        render_template_for(@__vc_variant).to_s + _output_postamble
+        if defined?(I18n) && !I18n.locale.eql?(I18n.default_locale)
+          attr = (I18n.locale || I18n.default_locale).to_s.downcase.tr("-", "_")
+
+          custom_variant = [@__vc_variant, attr].compact.reduce { |s, x| "#{s}_#{x}" }
```
jonspalmer, line 101:
```suggestion
          attr = (I18n.locale || I18n.default_locale).to_s.underscore

          custom_variant = [@__vc_variant, attr].compact.join("_")
```

jonspalmer (commented):
Looks powerful. A couple of suggested simplifications and cleanups

lfalcao:
Thanks @Spone and @jonspalmer for the comments. I'll have a look into it carefully.

test/test_helper.rb
lfalcao: Yep, I know it's scary... but still, it's just for tests. My idea was to use this function to replace all those other functions in test_helper.rb (with_preview_route, with_preview_controller, with_custom_component_path, with_default_preview_layout, with_render_monkey_patch_config...).

I also think its helps in readability (as the configuration is close to the test itself). But I can remove it, no problem.

lib/view_component/compiler.rb
```diff
@@ -38,8 +38,11 @@ def compile(raise_errors: false)
       templates.each do |template|
         # Remove existing compiled template methods,
         # as Ruby warns when redefining a method.
-        method_name = call_method_name(template[:variant])
-        component_class.send(:undef_method, method_name.to_sym) if component_class.instance_methods.include?(method_name.to_sym)
+        method_name = call_method_name(template[:variant]).to_sym
+
+        if component_class.instance_methods.include?(method_name)
+          component_class.send(:undef_method, method_name)
+        end
```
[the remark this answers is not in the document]
lfalcao: There's a little change where I moved two '.to_sym' to `method_name`.

lib/view_component/compiler.rb
```diff
@@ -61,7 +64,9 @@ def #{method_name}
     attr_reader :component_class
 
     def define_render_template_for
-      component_class.send(:undef_method, :render_template_for) if component_class.instance_methods.include?(:render_template_for)
+      if component_class.instance_methods.include?(:render_template_for)
+        component_class.send(:undef_method, :render_template_for)
+      end
```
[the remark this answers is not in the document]
lfalcao: Cosmetic only, will remove it.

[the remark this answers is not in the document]
jonspalmer: I'm a little surprised rubocop doesn't force the choice

test/test_helper.rb
jonspalmer: Sure but those other test helpers have other things they need like reloading. Also it's somewhat drier to use the explicit helper methods. IMO its much cleaner and declarative to see:

```ruby
with_preview_controller("MyPreviewController") do
```

than
```ruby
with_custom_config("Rails.application.config.view_component.preview_controller = 'MyPreviewController'")
```

It's less for me to read which aids in the comprehension of the code under test. 

Lastly its much more maintainable. If the setting for `preview_controller` changed for any reason with the explicit test_helper we'd only need to update one location. The `with_custom_config` approach requires you to change it everywhere its used.

lfalcao: I had done this change to work with reloading and changed all tests locally, it would be the next PR... :sweat_smile: 

One last comment, it also would help to join multiple configurations too...

``ruby
with_custom_config(
  "Rails.application.config.view_component.preview_controller = 'MyPreviewController'",
  "I18n.locale = ':en'"
)
``

But no problem, I'll remove it.

23tux:
@lfalcao is there any news on this PR?

Spone:
I can take some time to revisit this @lfalcao if you're still interested. Please let me know!

night91:
I think it's worth to get this. Sometimes this is better than the current translations support

joelhawksley:
Closing as stale. Please reopen if/when you pick this up, @Spone!

PR #980, opened 2021-06-24
Fragment Caching POC

Matt-Yorkley:
### Summary

This is a rough draft POC to explore adding compatibility for Fragment Caching in ViewComponent. 

It makes some modifications to `ActionView::Digestor` and `ActionView::ERBTracker` in order to get **digests** working correctly with components, which means (for example) if you cache a rails view that renders a component inside it and the contents of the component class file change, the digest used in the cache key will be updated and the cached fragment will be correctly invalidated.

Initial notes in issue https://github.com/github/view_component/pull/476#issuecomment-843063274

#### Current state:

- It more or less works :partying_face: (see point two)
- It needs a lot of tests
- It was thrown together with the aim of getting quick results, but the way these ActionView classes are being overridden probably needs a rethink and iterating on before it'll start to look like something acceptable :sweat_smile: 
- Currently works with ERB, but other template engines register dependency trackers in different ways, for example here in the [haml-rails gem](https://github.com/haml/haml-rails/blob/fc6d0ace6e1e24079a6c879129e37a3396b558fd/lib/haml-rails.rb)

### DIFF
--- a/lib/view_component/caching/component_tracker.rb
+++ b/lib/view_component/caching/component_tracker.rb
@@ -1,0 +1,39 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module Caching
+    class ComponentTracker < ERBTracker
+      def initialize(name, template, view_paths = nil)
+        super
+        @component_class = name.safe_constantize
+      end
+
+      def dependencies
+        return [] unless view_component?
+
+        super + templates + relevant_ancestors
+      end
+
+      private
+
+      attr_reader :component_class
+
+      def view_component?
+        component_class&.ancestors&.include? ViewComponent::Base
+      end
+
+      # Returns relevant templates for the component
+      def templates
+        component_class.
+          _sidecar_files(ActionView::Template.template_handler_extensions).
+          map { |path| path.gsub(%r{(.*app/components/)|(\..*)}, "") }
+      end
+
+      # Returns relevant ancestors the component inherits from (if any)
+      def relevant_ancestors
+        ((component_class.ancestors & ViewComponent::Base.descendants) - [component_class]).
+          map(&:to_s)
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


lfalcao:
Related: #476

jonspalmer:
This feature set is super exciting. Curious about the monkey patch - Are these changes that could reasonably be a part of Rails itself? Could we eliminate the monkey patch entirely and make this just about registering trackers?

Matt-Yorkley:
> Are these changes that could reasonably be a part of Rails itself?

Joel said the exact same thing :smile: It needs some thought/discussion.

> Could we eliminate the monkey patch entirely and make this just about registering trackers?

I think it won't work without the changes to `ActionView::Digestor`, but I'm open to any ideas :+1:

Matt-Yorkley:
I started working on this in a Rails 6.0 app, but there's a coupe of things that are slightly broken in 6.1+. I'll push some fixes shortly and add some initial tests.

Matt-Yorkley:
If anyone's interested what the `Digestor`'s dependency tree object actually looks like, I've made a Gist [here](https://gist.github.com/Matt-Yorkley/9799f3ead134711693bfc429f611ac54) with some sample debugging output.

Matt-Yorkley:
I had a go at reformulating this as a PR on ActionView and it looks a lot better. Preview [here](https://github.com/rails/rails/compare/main...Matt-Yorkley:component-digests).

From the ActionView perspective it's introducing support for a feature set that's ultimately external to ActionView, which feels a bit weird... I managed to introduce it without any explicit coupling to ViewComponent though, and it looks fairly reasonable and unobtrusive :man_shrugging:

dkniffin:
@Matt-Yorkley Any progress on this? Are you planning on making a PR to ActionView for that change?

Matt-Yorkley:
> Any progress on this? Are you planning on making a PR to ActionView for that change?

Yes and no. I did make that PR, but it was rejected: https://github.com/rails/rails/pull/42850

If anyone else wants to take up the baton on this one, be my guest.

joelhawksley:
Closing as stale ❤️

PR #981, opened 2021-06-24
Allow accessing slot arguments via new args method

dkniffin:
### Summary

When using "generic" slots (aka, slots that are not classes or lambdas), it doesn't seem currently possible to access arguments passed to the slot. In this PR, I've added that ability.

For example, this is now possible, where it wasn't before:
```rb
# Usage:
render_inline(SlotsV2WithArgsComponent.new) do |component|
  component.slot_with_arg("Arg 0") do
    "Arg slot content"
  end
  component.slot_with_kwarg(a: "Kwarg a") do
    "Kwarg slot content"
  end
end
```

```rb
# app/components/slots_v2_with_args_component.rb
class SlotsV2WithArgsComponent < ViewComponent::Base
  renders_one :slot_with_arg
  renders_one :slot_with_kwarg
end
```

```erb
<%#  app/components/slots_v2_with_args_component.html.erb %>
<div class="slot-with-arg">
  <div class="arg-0"><%= slot_with_arg.args[0] %></div>
  <div class="content"><%= slot_with_arg %></div>
</div>

<div class="slot-with-kwarg">
  <div class="kwarg-a"><%= slot_with_kwarg.args[:a] %></div>
  <div class="content"><%= slot_with_kwarg %></div>
</div>
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add support for slot arguments
+
+    *dkniffin*
+
 * Add @boardfish to Triage.
 
     *Joel Hawksley*
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -29,10 +29,10 @@ To render a `renders_many` slot, iterate over the name of the slot:
 
 ```erb
 <%# blog_component.html.erb %>
-<h1><%= header %></h1>
+<h1 class="<%= header.args[:classes] %>"><%= header %></h1>
 
 <% posts.each do |post| %>
-  <%= post %>
+  <%= post.args.post %>
 <% end %>
 ```
 
--- a/lib/view_component/slot_v2.rb
+++ b/lib/view_component/slot_v2.rb
@@ -6,7 +6,7 @@ module ViewComponent
   class SlotV2
     include ViewComponent::WithContentHelper
 
-    attr_writer :__vc_component_instance, :__vc_content_block, :__vc_content
+    attr_writer :__vc_component_instance, :__vc_content_block, :__vc_content, :__vc_args, :__vc_kwargs
 
     def initialize(parent)
       @parent = parent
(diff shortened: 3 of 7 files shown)


dkniffin:
@joelhawksley What do you think of this?

BlakeWilliams:
👋 Thanks for taking the time to write and open this PR!

I have to admit that I'm somewhat hesitant about introducing functionality that allow slots to store arbitrary values.  We've seen a lot of value from backing slots with components or lambdas since it makes the API and intended use of the slot more clear and easier to understand.

Can you share any specific use-cases that inspired this solution? I'd love to discuss it and identify any potential gaps in the API. 🙂

dkniffin:
🤔 I suppose there's no reason not to share the code for the component I'm working on, and it provides a good example for this.

I've created a gist here: https://gist.github.com/dkniffin/b16e44ff1519b01782124878ff1d2082

There's a few different things going on, but if you look at the bottom of `some_view.html.slim` there, you'll see how this is being used. I would be fine using a separate component for the `columns` there, but the problem I ran into with that is I don't have any way of knowing where it shows up in order of columns. I created #982 as a way to address that, but I ended up going with this version of things, mostly because it was the path I went down first 😅

jonspalmer:
I'm with @BlakeWilliams on understanding the need for this use case. If we go this route I'd recommend that we keep args and kwargs separate:

```ruby
<%#  app/components/slots_v2_with_args_component.html.erb %>
<div class="slot-with-arg">
  <div class="arg-0"><%= slot_with_arg.args[0] %></div>
  <div class="content"><%= slot_with_arg %></div>
</div>

<div class="slot-with-kwarg">
  <div class="kwarg-a"><%= slot_with_kwarg.kwargs[:a] %></div>
  <div class="content"><%= slot_with_kwarg %></div>
</div>
```

This also has the advantage that we could treat `args` as a true array with access to `first` and `last` methods etc. that wouldn't work as expected with the current implementation of a Hash backing args.

dkniffin:
I'm going to go ahead and close this. I no longer need this functionality. I really only needed it for the reason documented in #982, and I've since found a workaround (which I documented there as well)

PR #983, opened 2021-06-24
Changelog requirement

joelhawksley:
This PR introduces a build that ensures folks include CHANGELOG entries in their PRs so that reviewers don't have to ask for them. I also took the liberty of changing our CI configuration to only run on PRs, not all pushes.


Spone:
@joelhawksley is there a way to skip this check for tiny changes (such as https://github.com/github/view_component/pull/984)?

joelhawksley:
@Spone It's probably possible! One intent here was to force us to recognize all contributions, no matter how small. Perhaps I should update the contribution guide to mention that?

BlakeWilliams (approved):
I'm a fan of this change, 👍 to trying this out and seeing how it goes.

joelhawksley:
@Spone I added a note to that effect: https://github.com/github/view_component/pull/989

PR #986, opened 2021-06-26
Add slot_index method

### LINKED ISSUE #982: Allow for accessing index number when using renders_many
## Feature request

I would like to be able to do something like this:
```slim
/ Usage (in slim)
= render ListComponent.new do |c|
  = c.list_item
    span Item A
  = c.list_item
    span Item B
  = c.list_item
    span Item C
```

```rb
class ListComponent < BaseComponent
  renders_many :list_items, ListItemComponent
end
```

```rb
class ListItemComponent < BaseComponent
  def item_number
    slot_index + 1
  end
end
```

`slot_index` is the new method I'm requesting, and it would be automatically set based on the position of that slot item in it's collection of slots. For example, calling that from within "Item A" in my example would return `0`, whereas calling it from within "Item B" would return `1`, etc.

### Motivation

The reason I'd like this feature is that I'm considering creating a component that creates a page layout, using Bootstrap's grid columns. I want to have `renders_many :columns`, but in order for that to work properly, the slot needs to know where it appears in the list of columns.

dkniffin:
Resolves #982 

I've taken a first stab at this. @joelhawksley let me know what you think of this. I'd definitely want to add tests and docs, but I want to make sure I'm headed in the right direction first.


joelhawksley:
@dkniffin it looks like you're making changes in the right places!

FWIW, I generally prefer to see failing tests before anything else, so we can chat about proposed APIs.

dkniffin:
@joelhawksley Alright, I pushed up a commit w/ a failing test (and rebased so it's in the right order). I'm noticing now that the implementation is definitely a bit wrong because there's some test failures. Thoughts?

lib/view_component/slotable_v2.rb
```diff
@@ -250,6 +256,10 @@ def set_slot(slot_name, *args, **kwargs, &block)
 
       if slot_definition[:collection]
         @__vc_set_slots[slot_name] ||= []
+
+        # set the index to the next index in the array (aka, the current length)
+        slot.slot_index = @__vc_set_slots[slot_name].length
```
BlakeWilliams, line 261: What happens if we sort the collection before rendering?

dkniffin: :/ Yeah, that wouldn't work with the current implementation. I'm not sure if that'd be a common use-case or not though. It's certainly not for me.

joelhawksley: From my experience, slot introspection is not out of the ordinary. With that in mind, do you think we should still pursue this feature?

dkniffin: I definitely would still find value in this feature. Since creating this PR, I've run into a couple other spots where it would be useful. I just haven't found the time to make the improvements to it yet 😓

Spone: I would find it useful as well. I have a few components where it would help.

primer-css:

PR #987, opened 2021-06-26
Add view_component_path option to ViewComponent::Base

Matt-Yorkley:
### Summary

Another little refactor proposal that came up while working on #980...

Exposes the (recently added) `view_component_path` option as an attribute on `ViewComponent::Base`. This clarifies the config option as the single source of truth for the components path and makes it more readily accessible from other parts of the engine (it's currently only used in generators, but would be useful elsewhere). It also makes the value slightly easier to modify in tests, because it's not tied to initialization.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,9 +7,9 @@ title: Changelog
 
 ## main
 
-* Make view_component_path config option available on ViewComponent::Base
+* Make view_component_path config option available on ViewComponent::Base.
 
-  *Matt-Yorkley*
+    *Matt-Yorkley*
 
 * Add @boardfish to Triage.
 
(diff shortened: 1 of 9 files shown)


lfalcao:
Thanks @Matt-Yorkley !

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Make view_component_path config option available on ViewComponent::Base
+
+  *Matt-Yorkley*
+
```
```suggestion
* Make view_component_path config option available on ViewComponent::Base.

    *Matt-Yorkley*

```

lib/view_component/base.rb
```diff
@@ -249,6 +249,13 @@ def content_evaluated?
     #
     mattr_accessor :show_previews_source, instance_writer: false, default: false
 
+    # Path for component files
```
joelhawksley, line 252: 👋🏻 mind running `rake docs:build` locally? Doing so will parse this file and update the API docs ❤️

joelhawksley (approved):

Issue #445, opened 2020-08-13
Can't render ViewComponent with rich_text_area


helphop:
### Steps to reproduce

I have the following View Component:

```
<%= form_with(model: [post, element]) do |form| %>
  <%= form.rich_text_area :content %>
  <div class="mt-3 d-flex justify-content-between">
    <div>
      <%= form.submit "Save", class: 'btn btn-primary' %>
      <a href="#" class="cancel btn btn-secondary">Cancel</a>
    </div>
      <%= link_to 'Delete', post_element_path(post, element), method: :delete, class: 'btn btn-danger', data: { confirm: 'Are you sure?'} %>
  </div>
<% end %>
```
### Expected behavior
This should render the Trix Action Text Editor

### Actual behavior
ActionView::Template::Error (undefined method `rich_text_area_tag' for #<ParagraphFormComponent:0x00007f9734219b48>)

### System configuration
**Rails version**:
Rails 6.0.3.2
**Ruby version**:
ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin19]
**Gem version**:
Using view_component 2.18.1

helphop:
Can someone please look into this I need to be able to use rich_text_area in my view components.

helphop:
I was able to figure out how to get this working. I hope this helps others with a similar problem.

In your  View Component Class add the following:

```
require 'action_text'
class ParagraphFormComponent < ViewComponent::Base
  include ActionText::Engine.helpers

   def main_app
     Rails.application.class.routes.url_helpers
  end

```
 In config environment development (or other environment file):

`Rails.application.routes.default_url_options[:host] = "localhost"`

It might be helpful to require the action_text in the ViewComponent::Base so it's available to all components as well as the ActionText::Engine.helpers for future releases.

joelhawksley:
@helphop sorry for the delay in getting to this issue, we've been pretty swamped here at work 😅 

Would you be willing to open a PR with a failing test for us to look at? I'd also be happy to pair with you on fixing this issue. Send me an email: joelhawksley@github.com

helphop:
I have forked the application and added the following code to the view_component/base.rb class:

require "action_view"
require "active_support/configurable"
require "view_component/collection"
require "view_component/compile_cache"
require "view_component/previewable"
require "view_component/slotable"
require "action_text/engine"

```
module ViewComponent
  class Base < ActionView::Base
    include ActiveSupport::Configurable
    include ViewComponent::Previewable
[...]

Issue #351, opened 2020-05-20
[Possible bug] Incompatibility with ActionText?


coderberry:
### Steps to reproduce

I am trying to convert one of our components to use the rich_text_area form helper in place of the regular text area. For some reason, it will not recognize the form method `rich_text_area`. The error message makes me think it may have something to do with ViewComponent.

### Expected behavior

Here's the old code:

```erb
<%= simple_form_for(:comment, url: comments_path, method: :post, wrapper: :front_form) do |f| %>
  <div class="row">
    <%= f.hidden_field :sgid, value: commentable.to_sgid %>
    <%= tag.div f.input(:body, as: :text, label: false, required: true, input_html: {rows: 2, class: "form-control"}), class: "col-12 mb-3" %>
  </div>
  <div class="d-flex justify-content-end">
    <%= f.submit "Add Comment", class: "btn btn-primary" %>
  </div>
<% end %>
```

and here's the change I'm trying to implement:

```erb
<%= form_with(model: Comment.new) do |f| %>
  <div class="row">
    <%= f.hidden_field :sgid, value: commentable.to_sgid %>
    <div clas="col-12 mb-3">
      <%= f.rich_text_area :content %>
    </div>
  </div>
  <div class="d-flex justify-content-end">
    <%= f.submit "Add Comment", class: "btn btn-primary" %>
  </div>
<% end %>
```

### Actual behavior
Here's the error that is resulting:

```
ActionView::Template::Error (undefined method `rich_text_area_tag' for #<Comments::RichTextFormComponent:0x000055a8a1b12cc8>                             │2020-05-20T22:53:12.503Z pid=44313 tid=gtiqx4nv1 class=DownloadAndExtractMax
Did you mean?  rich_text_area):                                                                                                                          │mindFileJob jid=5ce88f291615856fb6f38320 elapsed=0.188 INFO: done
     9:     <%= render(CommentsComponent.new(commentable: @organization, comments: @comments)) %>                                                        │2020-05-20T23:03:02.141Z pid=44313 tid=gtiqx4jlx class=DownloadAndExtractMax
    10:   <% end %>                                                                                                                                      │mindFileJob jid=b415c53f7c54684dfbe71ef0 INFO: start
    11: <% end %>                                                                                                                                        │2020-05-20T23:03:02.413Z pid=44313 tid=gtiqx4jlx class=DownloadAndExtractMax
[...]

PR #990, opened 2021-06-30
Add compatibility notes for using ActionText with ViewComponent

jaredplanter:
### Summary

Using ActionText with ViewComponent is not supported out of the box. It results in an error.

This updates the compatibility documents with a workaround until it is supported.

This has been previously reported in these issues:

https://github.com/github/view_component/issues/445
https://github.com/github/view_component/issues/351

Both proposed solutions. The latter being the simplest in requiring the least configuration.

Thank you @coderberry and @andrewmcodes: https://github.com/github/view_component/issues/351#issuecomment-631796883


docs/compatibility.md
```diff
@@ -24,3 +24,18 @@ With the monkey patch disabled, use `render_component` (or  `render_component_to
...
+undefined method `rich_text_area_tag'
+```
+
+This is because ViewComponent does not have access to the helpers it needs via ActionText. To avoid this, add the following to your component (or base component):
+
+```
+delegate :rich_text_area_tag, to: :helpers
+```
+
+*_This is the simplest route. Adding the helpers directly via `include ActionText::Engine.helpers` leads to several more errors and requires additional configuration that is not necessary._
```
andrewmcodes, line 41: Looks great but I'm wondering if it's worth mentioning the other method without saying what it is or linking to it? Imo I believe this to be the "right" way to fix this and it doesn't just pertain to ActionText!

[Here's a great example](https://www.bridgetownrb.com/docs/components/ruby#helpers-1)

jaredplanter: Good call! I went ahead and removed it.

docs/compatibility.md
```diff
@@ -24,3 +24,16 @@ With the monkey patch disabled, use `render_component` (or  `render_component_to
 ## Bridgetown (Static Site Generator)
 
 [Bridgetown](https://www.bridgetownrb.com/) supports ViewComponent via an experimental shim provided by the [bridgetown-view-component gem](https://github.com/bridgetownrb/bridgetown-view-component). More information available [here](https://www.bridgetownrb.com/docs/components/ruby#need-compatibility-with-rails-try-viewcomponent-experimental).
+
+## ActionText Support
```
joelhawksley, line 28:
```suggestion
## ActionText
```

docs/compatibility.md
```diff
@@ -24,3 +24,16 @@ With the monkey patch disabled, use `render_component` (or  `render_component_to
...
+
+Trying to use `rich_text_area` via ActionText out of the box with ViewComponent will result in this error:
```
joelhawksley, line 30:
```suggestion
Using `rich_text_area` from ActionText in a ViewComponent will result in this error:
```

docs/compatibility.md
```diff
@@ -24,3 +24,16 @@ With the monkey patch disabled, use `render_component` (or  `render_component_to
...
+```
+undefined method `rich_text_area_tag'
+```
```
joelhawksley, line 33:
```suggestion

```
undefined method `rich_text_area_tag'
```
```

docs/compatibility.md
```diff
@@ -24,3 +24,16 @@ With the monkey patch disabled, use `render_component` (or  `render_component_to
...
+
+This is because ViewComponent does not have access to the helpers it needs via ActionText. To avoid this, add the following to your component (or base component):
```
joelhawksley, line 35:
```suggestion
This is due to ViewComponent not having access to the helpers it needs via ActionText. As a workaround, add the following to your component (or base component):
```

joelhawksley (approved):
👋🏻 mind adding a CHANGELOG?

docs/compatibility.md
rmacklin: Hey @joelhawksley, I don't think this suggestion applied correctly due to using ``` inside the suggestion block

Spone (approved):

docs/CHANGELOG.md
```diff
@@ -56,6 +56,10 @@ title: Changelog
 
     *Johannes Engl*
 
+* Add documentation for compatibility with ActionText
```
joelhawksley, line 59:
```suggestion
* Add documentation for compatibility with ActionText.
```

PR #996, opened 2021-07-01
Fix references to documentation files that were moved into `docs/`

rmacklin:
### Summary

We moved these files into `docs/` in https://github.com/github/view_component/pull/972 but there were a few lingering references that still assumed they were in the repository root, including the `files` list in the gemspec which would've bundled a broken symlink instead of the actual CHANGELOG with the gem.


joelhawksley (approved):

joelhawksley:
Thanks Richard! Glad to see you around these parts again ❤️

PR #997, opened 2021-07-01
Ensure consistent indentation with Rubocop

joelhawksley:


Rakefile
```diff
@@ -84,7 +84,7 @@ namespace :docs do
           end
 
         types = if method.tag(:return)&.types
-          " → [#{method.tag(:return).types.join(',')}]"
+                  " → [#{method.tag(:return).types.join(',')}]"
```
BlakeWilliams, line 87: This seems like it's indented too far. I generally have a preference for consistent indentation instead of trying to align to the right of the `if` statement above.

Thoughts?

BlakeWilliams: I think this may be the relevant cop for this behavior? https://docs.rubocop.org/rubocop/cops_layout.html#layoutmultilineassignmentlayout

BlakeWilliams (approved):
LGTM

Issue #976, opened 2021-06-22
Which words are reserved under ViewComponent?


boardfish:
I've just encountered an issue whereby having a slot like `renders_many :tags` results in a conflict between the `tag` slot and the `tag` view helper. It'd be good to keep track of which other words, if any, are likely to cause issues with ViewComponent.

joelhawksley:
@boardfish I agree. Our work on reducing the number of potential instance variable clashes was a good step in this direction. How do you think we should keep track of these words?

boardfish:
I wonder if tests would be a good way to do this. In the same way that the test for the trailing newline issue uses something peripheral to ViewComponent to document the solution to the issue, we could do something similar for namespace clashes. 

For example, in the case I had above, we could have passing tests to suggest alternative names like `topics` or `categories` for the slot. `attributes` is another namespace clash, now that I think about it, if you're using Active Model on top of your base component class.

joelhawksley:
@boardfish hmm. I wonder if we could enforce this at the framework level. For example, we could check slot names against a list of reserved words at compile time and raise an error.

boardfish:
That could be really good for addressing this in a way that's more developer-friendly.

I wonder if reserved words could affect the experience in other ways, perhaps such as:
- defining a method that's the same as a slot name
- doing something similar with some of ViewComponent's convenience features like collection parameters
- accidentally overriding private methods
I'm interested in experimenting with these to see what kinds of situations can arise.

I suppose the logic we'd use to raise errors in case of reserved words would be common to all these cases, but the words themselves may differ between each case.

joelhawksley:
`accidentally overriding private methods`

Oh, that is an interesting one. I hadn't thought of namespacing our private methods.

Would you be up for taking on the slot naming issue and seeing what you can come up with?

boardfish:
Yeah, I'll see what I can whip up there.

boardfish:
Turns out there's a validation in place for if `renders_one :content` is called, but there isn't for `renders_many :contents`, so I'll try to address that in isolation first.

I'll see how it interfaces with other things too, but this is one well-defined area we can tackle for sure.

boardfish:
[...]

PR #1001, opened 2021-07-03
Raise error on renders_many :contents

boardfish:
### Summary

If `renders_many :contents` is called, you'd have to use the slot like this:

```html.erb
<%= render Component.new do |c| %>
  <% c.content %>
<% end %>
```

Calling `content` here calls the private method `content` that already exists on the component.

Related to #976

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Raise an error if the slot name for renders_many is :contents
+
+    *Simon Fish*
+
 ## 2.35.0
 
 * Only load assets for Preview source highlighting if previews are enabled.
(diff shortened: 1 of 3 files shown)


lib/view_component/slotable_v2.rb
```diff
@@ -174,11 +174,23 @@ def register_slot(slot_name, collection:, callable:)
         self.registered_slots[slot_name] = slot
       end
 
-      def validate_slot_name(slot_name)
+      def validate_plural_slot_name(slot_name)
```
boardfish, line 177: We could use inflections here to make sure the plural and singular versions of the slot name don't clash with things in future, but until we're confident of that, I thought I'd keep things simple for now.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Raise an error if the slot name for renders_many is :contents
```
```suggestion
* Raise an error if the slot name for `renders_many` is `:contents`.
```

joelhawksley (approved):
Nice! I like this incremental approach to start 👍🏻

PR #1004, opened 2021-07-13
Update error messaging to be more descriptive and consistent

joelhawksley:
### Summary

This PR updates our error messages to follow a more consistent format and to provide suggested resolutions when possible.


jonspalmer:
This looks great 💯 
Are there plans to push these error message strings to i18n bundles so they could one day be translated?

joelhawksley:
@jonspalmer I hadn't thought about i18n, but that would be great! One day 😄

Issue #623, opened 2021-02-15
Project-wide config for generators


Spone:
## Feature request

Currently, you have to pass the flags every time you run a generator. Being able to override the default values in `config/application.rb` would be useful.

Here is the suggested syntax, that would follow [Rails convention](https://guides.rubyonrails.org/generators.html#customizing-your-workflow):

```rb
config.generators do |g|
  g.view_component preview: true, sidecar: true # this would always create a preview and a sidecar directory
end
```

Another example:

```rb
config.generators do |g|
  g.view_component inline: true # this would always create inline components
end
```

There already is a way to configure the template engine (see below) so we wouldn't change that.

Such global config option could obviously be overridden when running the generator, by using flags as usual.

### Motivation

In some cases, it makes sense to use generator flags every time you generate components (such as `--sidecar`), but you can easily forget.

Related:
- https://github.com/github/view_component/pull/579#issuecomment-758189083
- previous work on Komponent: https://github.com/komposable/komponent#configuration

### Existing generator flags

```rb
# Override the default template engine
bin/rails generate component Example title content --template-engine slim

# Generate a preview
bin/rails generate component Example title content --preview

# Place the view and assets in a sidecar directory
bin/rails generate component Example title content --sidecar

# Use inline rendering (no template file)
bin/rails generate component Example title content --inline
```
By the way, the `--inline` flag appears to be only taken into account by the [ERB generator](https://github.com/github/view_component/blob/4db9e5896b06aec7f4a18b89de61691f45f490af/lib/rails/generators/erb/component_generator.rb#L10), it should be added to Slim and Haml as well. I will open a separate issue for that.

Also, what about a `--stimulus` flag to generate the Stimulus controller? I will also open a separate issue.

### Existing configuration settings

Here is a quick recap of the current configuration settings in ViewComponent.

The `generators.template_engine` option from Rails is taken into account.

Apart from that, most of the ViewComponent-specific settings are for now related to previews.

```rb
# config/application.rb

# Use your template engine of choice (reusing the existing Rails option)
config.generators.template_engine = :slim

# Set the default layout for previews
config.view_component.default_preview_layout = "component_preview"

# Set the path to previews
config.view_component.preview_paths << "#{Rails.root}/lib/component_previews"

# Set the route for previews
config.view_component.preview_route = "/previews"

# Set the controller to use for previews
config.view_component.preview_controller = "MyPreviewController"

# Set the controller to use for tests
config.view_component.test_controller = "BaseController"

[...]

Issue #618, opened 2021-02-11
Component and View Colocation


BlakeWilliams:
Currently, components live in a dedicated `app/components` directory. This has worked well so far, but we'd like to explore a more unified approach.

This isn't set in stone, but the goals would be to:

* Allow components to be defined in `app/views`
* Enable `render "users/show", user: current_user` to render a component instead of explicitly referring to the component class. e.g. This would be equivalent to `render Users::Show.new(user: current_user)`
* Remove the `Component` suffix from components so their names have a 1-to-1 mapping to view directories. e.g. `app/views/users/user.rb` would map to `class Users::User < ViewComponent::Base`

A more complete example may look like:

```ruby
# app/controllers/users_controller.rb
class UsersController < ApplicationController
  def show
    user = User.find(params[:id])
    render :show, user: user
  end
end
```

```ruby
# app/views/users/show.rb
class Users::Show < ViewComponent::Base
  def initialize(user:)
    @user = user
  end
end
```

```erb
<%# app/views/users/show.html.erb %>
<div>
  <h1><%= @user.name %>'s page!</h1>
   <%= render "users/profile_content", user: user # will render the app/views/users/profile_content.rb component %>
</div>
```

part of https://github.com/github/view_component/issues/617

MatheusRich:
Related: https://github.com/github/view_component/discussions/588

BlakeWilliams:
I've spiked this out a bit here: https://github.com/BlakeWilliams/view_component-rails-test (PR's showing off other ideas or approaches are welcome)

Some take-aways:
* `_` prefixed file names for use as partials doesn't feel like the correct approach for components.
* Using strings, e.g. `render "users/profile, user: user"` instead of `render Users::Profile.new(user)` has some limitations, like requiring that component arguments are keyword arguments.
  * How would `render User::Profile.new(user)` translate to `render "user/profile`? The component would be required to change the initializer from `def initialize(user)` to `def initialize(user:)`.
* Implicit controller renders may be tricky or impractical.

I think the flexibility of view components (e.g. user defined initializers) may make it more challenging to make them work with Rails as seamlessly as we'd like, but I'll keep experimenting with it.

choilive:
@BlakeWilliams I actually prefer the explicit object instantiation over using strings. Makes it easier to refactor and also makes it clearer that it is a "component" and not a partial. Would be OK with being able to use both ways though as a matter of user preference.

[...]

PR #1005, opened 2021-07-13
[ADR] Remove suffix from naming conventions

### LINKED ISSUE #670: Remove the Component suffix from components

joelhawksley:
[closes #670]

_Please read this pull request as a **proposal** that can and **should** change._

### Summary

I think we should move away from naming ViewComponents with the `-Component` suffix.

I've included documentation in the form of an Architectural Decision Record. I'd love to see us use that document as the grounds of debate for this proposal. 

Assuming we can agree on this change, I will take on the work of updating the project to follow these guidelines.

### DIFF
--- a/adr/0002-naming-conventions-for-view-components.md
+++ b/adr/0002-naming-conventions-for-view-components.md
@@ -1,0 +1,26 @@
+# 2. Naming conventions for ViewComponents
+
+Date: 2021-07-13
+
+## Status
+
+Proposed
+
+## Context
+
+The `Component` suffix has been questioned a few times. Based on our ViewComponents at GitHub and those shared on the ViewComponent repo, it looks like the namespaces used are unique even when the `Component` suffix is removed. This is due to following the naming convention of views, which is to use the plural paths in the namespace. e.g. `users/index.html.erb` -> `Users::IndexComponent`/`Users::Index`.
+
+## Decision
+
+We will recommend that ViewComponents are named without the `-Component` suffix, for the User Interface elements they render, namespaced like views and controllers.
+
+## Alternatives considered
+
+We've used the `-Component` suffix for some time, so it is a viable alternative.
+
+## Consequences
+
+* **Pro**: Less boilerplate and better readability. No more `render Primer::LinkComponent.new`, we can now call `render Primer::Link.new`. Once you learn that components are objects that can be rendered, the `Component` suffix loses a lot of its value.
+* **Pro**: Component names now completely align with the `views` naming conventions. This will make it easier to put components in the `views` directory if we go down that path.
+* **Con**: It's no longer immediately clear that your class/object is a component. However, ViewComponents are often the only objects present in Rails view code, so the potential confusion is likely minimal.
+* **Con**: It will take time to migrate existing code to the new convention.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Change naming convention recommendation to remove `-Component` suffix.
+
+    *Joel Hawksley*
+
 * Update error messages to be more descriptive and helpful.
 
     *Joel Hawksley*
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -11,7 +11,7 @@ nav_order: 1
 
 - Components are subclasses of `ViewComponent::Base` and live in `app/components`. It's common practice to create and inherit from an `ApplicationComponent` that is a subclass of `ViewComponent::Base`.
 - Component module names are plural, as for controllers and views: `Users::Avatar`
-- Name components for what they render, not what they accept. (`Users::Avatar` instead of `Users::User`)
+- Name components after the UI element they render.
 
 ## Installation
 
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -10,9 +10,8 @@ nav_order: 1
 ## Conventions
 
 - Components are subclasses of `ViewComponent::Base` and live in `app/components`. It's common practice to create and inherit from an `ApplicationComponent` that is a subclass of `ViewComponent::Base`.
-- Component names end in -`Component`.
-- Component module names are plural, as for controllers and jobs: `Users::AvatarComponent`
-- Name components for what they render, not what they accept. (`AvatarComponent` instead of `UserComponent`)
+- Component module names are plural, as for controllers and views: `Users::Avatar`
+- Name components for what they render, not what they accept. (`Users::Avatar` instead of `Users::User`)
 
 ## Installation
 


joelhawksley:
@jaredcwhite I'd love to hear your thoughts on this change, as you had commented on #670.

jonspalmer:
I'm mostly unconvinced by this change. It's very un_railsy.

Consider that you could have this set of files:

app/components/users_component.rb
app/controllers/users_controller.rb
app/helpers/users_helper.rb

that seems pretty reasonable and consistent. 

What if everyone made the choice described here? We have three users.rb files different only because of directory that they're in? It probably also breaks the world as there are no namespaces. This decision documented here is only really possible because Rails convention is to use suffixes.

Spone:
I don't have a clear opinion yet. I hear the arguments made in the ADR. I definitely agree with @jonspalmer considerations.

I would add that:
- Not migrating existing components to the new naming scheme would mean cohabitation between `OldStuffComponent` and `NewStuff` = confusion. So migration is nearly mandatory (unless you don't plan to add new components to your project).
- The changes required to migrate will probably be a quick search-and-replace (in source code) and potentially a LOT of folders and files to rename. Could a migration task be provided?

> This is due to following the naming convention of views, which is to use the plural paths in the namespace. e.g. `users/index.html.erb` -> `Users::IndexComponent`/`Users::Index`.

- I don't know if it's necessarily the case in most projects using ViewComponent. It's not in mine :smile: 
- A quick diagnostics task could be helpful to test if the migration would create collisions in any given project. It may provide useful insights for this decision?

boardfish:
One thing I've been doing so far is keeping all components namespaced based on their purpose. As I've previously mentioned elsewhere, I'm following Atomic Design and using `A`toms, `M`olecules, and `O`rganisms. This is good for my use case because it's easily recognizable.

For most projects, the divide will be simpler than that, though - ViewComponents will be used for both components and views, so appropriate namespaces for those would be good. `View::Users::Index` and `Component::Button` would be potential examples. However, in the case of the `Component` namespace, we lose this benefit:

> Less boilerplate and better readability. [...] Once you learn that components are objects that can be rendered, the `Component` suffix loses a lot of its value.

Another question I have is how layouts might be considered under this structure.

boardfish (commented):
I worry that components being part of the top-level namespace might lead to conflicts without some good guidelines around how to name components and views.

One thing I've been doing so far is keeping all components namespaced based on their purpose. As I've previously mentioned elsewhere, I'm following Atomic Design and using `A`toms, `M`olecules, and `O`rganisms. This is good for my use case because it's easily recognizable, and keeps components contained to namespaces to prevent clashes. I plan to do the same with layouts and views.

For a lot of projects, `A`/`M`/`O` modules aren't necessary as that sort of granularity might cause more friction than good. They'll potentially have three main uses for ViewComponent: global components, views, and layouts. Theoretically, these could each have a namespace - let's say `Component`, `View`, and `Layout` to start with. However, because of the length of these names (particularly `Component`), we lose the benefit of making it quicker to call components, which is arguably the main goal of this ADR.

My gut tells me abbreviating the layouts (e.g. `V`iew and `C`omponent) would be unconventional, but I'll throw the idea out there regardless. Either way, namespaces with short (<5 character) names are worth consideration to solve for this IMO.

joelhawksley:
@spone @jonspalmer @boardfish thank you for the thoughtful feedback. You've helped me think more critically about this proposed change and understand how it could negatively affect others.

A different approach might make more sense: allowing the generators not to use the suffix as a configuration option turned off by default to preserve the existing behavior.

The motivation for this change was the experience we've had with Primer ViewComponents, where we've decided to remove the `-Component` suffix. However, all of those ViewComponents exist within the `Primer` namespace.

boardfish:
I still feel that the Component suffix should be abandoned entirely for the reasons described in the ADR. ViewComponent feels like a significant architecture decision in a Rails app — I don't believe folks that use the library will do it by halves. I expect a lot of components will make up the frontend at any level, and that means most users of the library will be likely to face the same issue, so the suffix still being present by default is potentially likely to cause friction. Disabling it with a flag would be a good first step to ease that pain for current users.

A namespace or something to that effect probably feels like the better long-term solution to me, though. While folks might not always be intending to make their own design system, I think using a namespace makes a lot of sense for the reasons I've described already. But it'll be difficult to decide on a sane default that brings the intended benefits of the ADR, maintains the ease of working with the library in the first instance, and considers the layers for which components can play a role.

I wonder if it'd be good to discuss this kind of thing on a call sometime.

boardfish:
I still feel that the Component suffix should be abandoned entirely for the reasons described in the ADR. ViewComponent feels like a significant architecture decision in a Rails app — I don't believe folks that use the library will do it by halves. I expect a lot of components will make up the frontend at any level, and that means most users of the library will be likely to face the same issue, so the suffix still being present by default is potentially likely to cause friction. Disabling it with a flag would be a good first step to ease that pain for current users.

A namespace or something to that effect probably feels like the better long-term solution to me, though. While folks might not always be intending to make their own design system, I think using a namespace makes a lot of sense for the reasons I've described already. But it'll be difficult to decide on a sane default that brings the intended benefits of the ADR, maintains the ease of working with the library in the first instance, and considers the layers for which components can play a role.

I wonder if it'd be good to discuss this kind of thing on a call sometime.

jonspalmer:
@joelhawksley I agree that you suggestion would be a great one. 

I agree with you that in larger systems namespaces are likely and can help. However, an extracted design system like Primer is not really the norm. Even if I was building out a design system in an app its not a given that I'd pick a namespace just for the design system. I might/have chosen other patterns to organize and group the code. Also even with a design system there will likely be components the aren't part of it but support some specific function. In my experience those components get namespaces under their functional area where collisions are more likely without the Component suffix.

@boardfish I disagree with the statement that folks won't do things by halves. Anyone with an existing Rails codebase wanting to try out ViewComponents is very likely to try it in a fewer places and see how it fits. They might have hundreds or thousands of views and partials to convert. That might take months or years depending on the organizational priorities. Its imperative IMO to minimize the experimentation cost for those folks. Having to worry about name spacing that components to avoid naming collisions is only going to be an impediment to adoption.

@joelhawksley One thought - its always struck me as unexpected to have to include the component suffix in the template file names (because Rails does not require `_controller` in regular views). Is it an option to drop that requirement? I actually thing that would make a big difference and make ViewComponents more consistent with Rails conventions.

boardfish:
Those are some good points @jonspalmer. I think my ideals here are influenced by the bias I've built up around implementing ViewComponent in my app, so I can understand if they don't extend to more cases. 

It might be good to have the suffix as a configurable default, so that the decision around how component names are generated only needs to be made once and not each time the generator is run.

Spone:
> It might be good to have the suffix as a configurable default, so that the decision around how component names are generated only needs to be made once and not each time the generator is run.

I agree, and I think that any flag that is added to the generator should also be configurable once, project-wide (see #623).

jonspalmer:
Looking at this from a slightly different perspective how do we think this should look if we were hooking into Rails scaffolding?

Suppose our push is that VIew Components should be used to replace the view layer of Rails entirely. Controller-actions should have a dedicated ViewComponent that they render directly. Today Rails scaffold looks like this:

```sh
bin/rails generate scaffold HighScore game:string score:integer
    invoke  active_record
    create    db/migrate/20190416145729_create_high_scores.rb
    create    app/models/high_score.rb
    invoke    test_unit
    create      test/models/high_score_test.rb
    create      test/fixtures/high_scores.yml
    invoke  resource_route
     route    resources :high_scores
    invoke  scaffold_controller
    create    app/controllers/high_scores_controller.rb
    invoke    erb
    create      app/views/high_scores
    create      app/views/high_scores/index.html.erb
    create      app/views/high_scores/edit.html.erb
    create      app/views/high_scores/show.html.erb
    create      app/views/high_scores/new.html.erb
    create      app/views/high_scores/_form.html.erb
    invoke    test_unit
    create      test/controllers/high_scores_controller_test.rb
    create      test/system/high_scores_test.rb
    invoke    helper
    create      app/helpers/high_scores_helper.rb
    invoke      test_unit
    invoke    jbuilder
    create      app/views/high_scores/index.json.jbuilder
    create      app/views/high_scores/show.json.jbuilder
    create      app/views/high_scores/_high_score.json.jbuilder
    invoke  assets
    invoke    scss
    create      app/assets/stylesheets/high_scores.scss
    invoke  scss
    create    app/assets/stylesheets/scaffolds.scss
```

What would that look like with ViewComponents?


```sh
bin/rails generate scaffold HighScore game:string score:integer
    invoke  active_record
    create    db/migrate/20190416145729_create_high_scores.rb
    create    app/models/high_score.rb
    invoke    test_unit
    create      test/models/high_score_test.rb
    create      test/fixtures/high_scores.yml
    invoke  resource_route
     route    resources :high_scores
    invoke  scaffold_controller
    create    app/controllers/high_scores_controller.rb
    invoke    components
    create      app/components/high_scores_component.rb # index
    create      app/components/high_scores_component.html.erb
    create      app/components/high_score_component.rb #show
    create      app/components/high_score_component.html.erb
    create      app/components/high_score_form_component.rb #new/edit
    create      app/components/high_score_form_component.html.erb
    invoke    test_unit
    create      test/controllers/high_scores_controller_test.rb
    create      test/components/high_scores_component_test.rb
    create      test/components/high_score_component_test.rb
    create      test/components/high_score_form_component_test.rb
    create      test/system/high_scores_test.rb
    invoke    helper
    create      app/helpers/high_scores_helper.rb
    invoke      test_unit
    invoke    jbuilder
    create      app/views/high_scores/index.json.jbuilder
    create      app/views/high_scores/show.json.jbuilder
    create      app/views/high_scores/_high_score.json.jbuilder
    invoke  assets
    invoke    scss
    create      app/assets/stylesheets/high_scores.scss
    invoke  scss
    create    app/assets/stylesheets/scaffolds.scss
 ```
 
 In the above it seems pretty natural/desirable to have a suffix.

**Comment by boardfish** [comment]
IMO the most intuitive thing would be to reflect the structure of current Rails views as closely as possible, e.g.

```
app/views/high_scores/show/...
app/views/high_scores/show.rb
app/views/high_scores/index/...
app/views/high_scores/index.rb
```

...and so on. Note that the components are stored in `app/views` as we'd hope to see through in #618. The corresponding component constants would be under the `HighScores` namespace in the top level. 

As a nice side-effect, plural namespaces would potentially lead to less clashes as constants tend to be named singularly.

jonspalmer:
@boardfish I understand the logic of that suggestion but honestly a plethora of index.html.erb littering a large code base is one of the most irritating things about the Rails view layer as grepping by file name just sucks. I'm unconvinced that having lots of index.rb files representing components is particularly great. I do see how the folder structure is an advantage.

boardfish:
That's a valid concern, but I think changes to the default view scaffold are more of a question for Rails than ViewComponent. In the meantime, I think it makes some sense for us to follow the existing structure, since that creates the least surprise. I'll refer back to this discussion in #618 as it'll help to guide that, but it's tangential to the ADR. 

That having been said, I do feel that the decisions on these two are coupled together. What we choose to do here affects the scaffold output as we'd hope to see in #618. Or, from a better angle, the best path to #618 influences the structure of component names in this PR.

joelhawksley:
@boardfish @Spone @jonspalmer thank you all for the thoughtful discussion here!

I'm going to close this proposal for now.

After giving it some more thought, I don't feel strongly enough about making these changes at this point. I appreciate the perspective you all have provided. 

I'll likely follow up with a light update to the documentation about how the suffix might be removed in some use cases.

PR #1008, opened 2021-07-16
Rendering slot collections returns the array of slots

jonspalmer:
### Summary

Rendering slot collections doesn't return the array of slots. This is asymmetric with the singular slot rendering and make it impossible to call methods like `with_content` on the rendered slots.

With this fix we can do things like this which previously would have raised an exception

```ruby
component.items([{ highlighted: false }, { highlighted: true }, { highlighted: false }])
 .each_with_index do |slot, index|
    slot.with_content("My Item #{index+1}")
  end
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Rendering slot collections returns the array of slots
+
+    *Jon Palmer*
+
 * Update error messages to be more descriptive and helpful.
 
     *Joel Hawksley*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,7 +7,7 @@ title: Changelog
 
 ## main
 
-* Rendering slot collections returns the array of slots
+* Setting a collection slot with the plural setter (`component.items(array)` for `renders_many :items`)  returns the array of slots.
 
     *Jon Palmer*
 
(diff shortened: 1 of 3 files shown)


jonspalmer:
It's perhaps a separate issue but I'd love to discuss how we might have other ways to pass content to slots rendered as collections. I don't think the block route is practical, there can only be one. We could take an array of blocks but at this point aren't you just better off calling the singular version multiple times?

I wonder if we might support a magic `content` key:

```ruby
render_inline SlotsV2DelegateComponent.new do |component|
  component.items(
    [
      { highlighted: false, content: "My Item 1" }, 
      { highlighted: true,  content: "My Item 2"  },
      { highlighted: false, content: "My Item 3"  }
    ]
  )
end
```
Thoughts?

joelhawksley:
> I wonder if we might support a magic content key

I've thought about this a lot. If we moved towards an attributes API, we could in theory implement a default `content` argument on all ViewComponents. Could be pretty cool!

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Rendering slot collections returns the array of slots
```
```suggestion
* Setting a collection slot with the plural setter (`component.items(array)` for `renders_many :items`)  returns the array of slots.
```

joelhawksley (approved):

PR #1009, opened 2021-07-16
Add slot_type slotable helpers

jonspalmer:
### Summary

Add three helper methods that allow introspection on what slots the component renders

```ruby
class SlotsV2Component < ViewComponent::Base
  renders_one :header
  renders_many :items
end
```
```ruby
SlotsV2Component.renders_one?(:header) # -> true
SlotsV2Component.renders_one?(:items) # -> false
SlotsV2Component.renders_one?(:item) # -> false

SlotsV2Component.renders_many?(:header) # -> false
SlotsV2Component.renders_many?(:items) # -> # true
SlotsV2Component.renders_many?(:item) # -> # false

SlotsV2Component.renders_many_item?(:header) # -> false
SlotsV2Component.renders_many_item?(:items) # -> # false
SlotsV2Component.renders_many_item?(:item) # -> # true
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,7 +7,7 @@ title: Changelog
 
 ## main
 
-* Add slot_type slotable helpers.
+* Add slot_type helper method.
 
     *Jon Palmer*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -17,7 +17,7 @@ title: Changelog
 
 * Setting a collection slot with the plural setter (`component.items(array)` for `renders_many :items`)  returns the array of slots.
 
-    *Jon Palmer*    
+    *Jon Palmer*
 
 * Update error messages to be more descriptive and helpful.
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add slot_type slotable helpers.
+
+    *Jon Palmer*
+
 * Update error messages to be more descriptive and helpful.
 
     *Joel Hawksley*
--- a/lib/view_component/slotable_v2.rb
+++ b/lib/view_component/slotable_v2.rb
@@ -142,6 +142,17 @@ module ViewComponent
         register_slot(slot_name, collection: true, callable: callable)
       end
 
+      def slot_type(slot_name)
+        registered_slot = registered_slots[slot_name]
+        if registered_slot
+          registered_slot[:collection] ? :collection : :single
+        else
+          plural_slot_name = ActiveSupport::Inflector.pluralize(slot_name).to_sym
+          plural_registered_slot = registered_slots[plural_slot_name]
+          plural_registered_slot&.fetch(:collection) ? :collection_item : nil
+        end
+      end
+
       # Clone slot configuration into child class
       # see #test_slots_pollution
       def inherited(child)
--- a/test/view_component/slotable_v2_test.rb
+++ b/test/view_component/slotable_v2_test.rb
@@ -363,4 +363,20 @@ class SlotsV2sTest < ViewComponent::TestCase
 
     assert_includes error.message, "It looks like a block was provided after calling"
   end
+
+  def test_slot_type_single
+    assert_equal(:single, SlotsV2Component.slot_type(:title))
+  end
+
+  def test_slot_type_collection
+    assert_equal(:collection, SlotsV2Component.slot_type(:tabs))
+  end
+
+  def test_slot_type_collection_item?
+    assert_equal(:collection_item, SlotsV2Component.slot_type(:tab))
+  end
+
+  def test_slot_type_nil?
+    assert_nil(SlotsV2Component.slot_type(:junk))
+  end
 end


jonspalmer:
I think I'd also like a combination that ORs these three methods. Thoughts on the name? `renders?` feels wrong. Perhaps `renders_slot?(slot_name)`?

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add render_xyz? slotable helpers.
+
+    *Joel Hawksley*
```
Spone, line 12:
```suggestion
    *Jon Palmer*
```

;)

lib/view_component/slotable_v2.rb
```diff
@@ -142,6 +142,19 @@ def renders_many(slot_name, callable = nil)
...
 
+      def renders_one?(slot_name)
+        registered_slots.dig(slot_name, :collection) == false
+      end
+
+      def renders_many?(slot_name)
+        registered_slots.dig(slot_name, :collection) == true
+      end
+
+      def renders_many_item?(slot_name)
```
Spone, line 153: I find this naming a bit confusing compared to the other two... but I'm not sure I have a better idea...

Spone:
> Perhaps `renders_slot?(slot_name)`?

:+1:

lib/view_component/slotable_v2.rb
jonspalmer: I had the same thought. Not great but..

joelhawksley (commented):
👋🏻 interesting! I'm curious about what purpose these helpers are meant to serve. Could you provide code comment documentation that explains their utility? That might help us come up with more descriptive naming per @spone's comment.

jonspalmer:
Sure my use case is ViewComponent Storybook. Specifically adding the slot API to write stories for slot components. I want to do two related things a) validate that the stories match the components they're documenting b) provide an API like ViewComponents for specifying the slots something like https://github.com/jonspalmer/view_component_storybook/pull/68#issuecomment-881144769

To do either of those things its necessary to ask if the component has a slot of that name. Most likely I'm writing a method_missing handler that would intercept slot methods etc.

Thinking about it that way its possible that we could have a simple API like this of the form:

```ruby
SlotsV2Component.slot_type(slot_name) #=> :many, :one, :many_item 
```
or something like that. Still has the same naming challenges.

I could of course have these helpers in ViewComponentStorybook but it seems to me that `registered_slots` shouldn't be part of the public API of ViewComponents.

joelhawksley:
@jonspalmer thanks for providing that context.

I wonder if it might make sense to expose `registered_slots` in some form, allowing for consumers to interrogate it as they see fit. 

Perhaps we could do something like:

```ruby
SlotsV2Component.slots_configuration

# that returns:

{
  header: {
    type: :renders_one,
  },
  posts: {
    type: :renders_many,
    setter: :post
  }
}
```

My thinking here is that this approach might avoid the naming awkwardness we're running into here.

jonspalmer:
Yeah I can see that would work. It's considerably more awkward to figure out it 'post' is a slot that I can call and it means consumers still need their own wrappers to get boolean answers back.

BTW for what it's worth I do actually have working code that would be a little better with some version of these apis:

https://github.com/jonspalmer/view_component_storybook/blob/5f4888f28daf5cead995e6f9a184ab45a1b539e9/lib/view_component/storybook/story_config.rb#L36

lib/view_component/slotable_v2.rb
```diff
@@ -142,6 +142,23 @@ def renders_many(slot_name, callable = nil)
...
+        plural_slot_name = ActiveSupport::Inflector.pluralize(slot_name).to_sym
+        plural_slot_name != slot_name && renders_many?(plural_slot_name)
+      end
+
+      def renders_slot?(slot_name)
```
joelhawksley, line 158: @jonspalmer based on the context provided, do you need all of these methods, or just `renders_slot`? If it's just the latter, I think we could inline the other methods into this one to avoid the naming question for now.

joelhawksley (commented):
👋🏻 thanks for the context @jonspalmer! I think we can take the direction you're suggesting in this PR.

lib/view_component/slotable_v2.rb
jonspalmer: I definitely need a way to get the type of the slot back. Important to be able to distinguish between a "one" slot and a singular setter of a many slot. So either three methods or perhaps just this

```ruby
SlotsV2Component.render_slot?(slot_name) #=>  :one, :many, :many_item, nil
```

The only naming issue there is `:many_item` Got any  better ideas for that?

Once we settle on the names I can refactor the PR>

joelhawksley: Gotcha. How about:

```ruby
SlotsV2Component.slot_type(slot_name) #=>  :single, :collection, :collection_item, nil
```

Since we're returning a symbol or nil, I think a regular method name vs. a predicate is more appropriate.

jonspalmer: Sold. I will make it so

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add slot_type slotable helpers.
```
```suggestion
* Add slot_type helper method.
```

joelhawksley (approved):

joelhawksley:
Thanks @jonspalmer! FYI, you should be able to open PRs without using a fork since you are a committer ❤️

Issue #909, opened 2021-05-05
Support slot rendering from controllers


th-ad:
## Feature request
Currently the `render_component` method in `ViewComponent::RenderingComponentHelper` does not accept a block. This prevents slots from being rendered in controllers.

### Motivation
With the rise in popularity of HTML over the wire, this seems like a valuable feature to have in order to render more complex components.

```ruby
class GalleryController < ApplicationController
  def index
    render_component(GalleryComponent.new) do |gallery|
      gallery.image
      ...
    end
  end
end
```

I saw the initial issue/PR for enabling component rendering in controllers (#235 ), but wasn't sure if there was a technical limitation with adding block support. Definitely happy to take on the issue if it's something people are interested in.

Edit: Now that I'm looking closer I also see [this](https://github.com/github/view_component/pull/235/commits/473beedb342bafc53557383f5a8c840c94645917) commit, which intentionally removed block support.

lfalcao:
@th-ad there's any difference between using controller or view, besides less files? :thinking:

boardfish:
@lfalcao I'd say there is. If you're rendering a component directly from the controller, then you get the benefit of encapsulation. So if you were rendering a component from a view file, you'd still have to pass the instance variables from the controller, e.g.:

```rb
class FooController < ApplicationController
  def foo
    @foo = 'Hello world'
  end
```

```erb
<%= render View::Foo.new(foo: @foo) %>
```



#618 exists with the aim of making it possible to render components in place of traditional Action View templates. It'd be helpful to use slots here to make it easier to reuse a 'view' component between actions, or change components of the page based on controller logic, e.g.

```rb
def foo
  render(View::Foo.new) do |c|
    c.button(disabled: !logged_in?)
  end
end
```

Slots will be really key here in ensuring that logic stays at the controller level and doesn't get coupled to the component itself.

boardfish:
I've put together a test case for this in #1010. However, my efforts to try and fix it have been in vain - I'm not sure of the timing on it, but I think that the feature was originally introduced back when the library needed to patch Rails to function. Now that `render_in` is supported directly in Rails itself, it's Rails' implementation that bars us from using blocks in the controller.

joelhawksley:
> Now that render_in is supported directly in Rails itself, it's Rails' implementation that bars us from using blocks in the controller.

[...]

Issue #918, opened 2021-05-06
Using #with_content on pass through slots throws 'undefined method `with_content` for nil:NilClass'


dkimot:
In lieu of a pr I have a suspected cause, detailed at the end of this issue.

### Steps to reproduce

Using `with_content` on a pass through slot with no arguments, like what is documented [here](https://viewcomponent.org/guide/slots.html#with_content).

### Expected behavior

The content argument passed into `with_content` is rendered as the slot.

### Actual behavior

The error undefined method `with_content` for nil:NilClass is raised.

### System configuration
**Rails version**:
6.1.3.1

**Ruby version**:
3.0.0

**Gem version**:
2.31.1

#### Suspected cause

`SlotableV2`'s defined method for a slot, seen [here](https://github.com/github/view_component/blob/main/lib/view_component/slotable_v2.rb#L70), checks to see if the slots caller passed in args or a block and uses that to determine the response.

If you call the slot without args or a block it's value is the result of `get_slot(slot_name)` which is the value nil returned on [line 202](https://github.com/github/view_component/blob/main/lib/view_component/slotable_v2.rb#L202) of that file.

If I pass args into the slot `<% modal.header(classes: nil).with_content(title) %>` it renders the slot correctly.

#### Potential solution

I'm happy to open a PR to fix this, but I'm not sure what the solution should be. To get access to `with_content` the slot's method will need to return an instance of `SlotV2` whether or not arguments/a block are passed in. I don't know if that means the parts of `set_slot responsible for initializing the slot need to be extracted to a method both `get_slot` and `set_slot` call or if `get_slot` should just be removed completely. I could also be way off base with my proposed solution.

dkimot:
I realized that this wasn't occurring on collection slots when passing in content to the singular form of that slot (ie a `menu_item` of the slot `menu_items). I went and checked and found that the singular version of a collection slot always returns the result of `set_slot`, see [here](https://github.com/github/view_component/blob/main/lib/view_component/slotable_v2.rb#L126).

jonspalmer:
I ran into this exact problem working on https://github.com/jonspalmer/view_component_storybook/pull/68
I like the proposed solution a great deal.

jonspalmer:
@BlakeWilliams Thinking about this a bit more I'm not convinced that your solution will work. I'm worried that it forces all slots to have to be set before they are read. Otherwise any optional slot will get instantiated incorrectly when it's read at render time. 

A few other thoughts on solutions:
[...]

PR #1010, opened 2021-07-19
Add test case for controller rendering with block

boardfish:
### Summary

As part of investigation into #909, I wrote a failing test case. It checked that a block passed when rendering a component from the controller is used in the final render output.

We've got a workaround now by calling `#tap` on the component instance itself, so this test serves as documentation of that.

### Other Information

As mentioned on #909, the ability to use blocks in controller rendering was removed in 473beedb342bafc53557383f5a8c840c94645917. It's likely that this was why it wasn't then supported by Rails when it was implemented over there.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add test case for rendering a ViewComponent with slots in a controller.
+
+    *Simon Fish*
+
 * Set maximum line length to 120.
 
     *Joel Hawksley*
--- a/test/view_component/integration_test.rb
+++ b/test/view_component/integration_test.rb
@@ -38,6 +38,15 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_includes inline_response, baseline_response
   end
 
+  def test_rendering_component_in_a_controller_with_block
+    # Currently unsupported.
+
+    get "/controller_inline_with_block"
+    assert_select("div", "bar")
+    assert_select("div", { count: 0, text: "baz" })
+    assert_response :success
+  end
+
   def test_template_changes_are_not_reflected_on_new_request_when_cache_template_loading_is_true
     # cache_template_loading is set to true on the initializer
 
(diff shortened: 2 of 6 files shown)


jonspalmer:
I find it weird that `render_inline` in the test suite includes support for blocks when controller rendering does not.

boardfish:
In line with discussion on the original issue, I'm going to update this to use `tap` and not pass any content blocks.

I'm not sure it's good for `tap` to be used like this in the long term, though, due to its inconsistency against how components get called in the view. So I'll look into seeing if I can get `render_in` to take a block in Rails itself.

boardfish:
Alright, I think this is good to go now.

jonspalmer:
The `tap` convention is practical but kinda ugly. I also believe that folks are likely to get bit hard by #918 when using this controller approach with slots that only take content.

boardfish:
Agreed on both counts - `tap`'s by no means a long-term solution, but as you say, it's practical and a good holdover until we resolve the original issue, so I feel it makes sense to document it by way of this test.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add test case for rendering from the controller with a block
```
```suggestion
* Add test case for rendering a ViewComponent with slots in a controller.
```

test/view_component/integration_test.rb
```diff
@@ -38,6 +38,16 @@ def test_rendering_component_in_a_controller
     assert_includes inline_response, baseline_response
   end
 
+  def test_rendering_component_in_a_controller_with_block
+    # Note that in order to do this, you must `tap` the component instance.
+
+    get "/controller_inline_with_block"
```
joelhawksley, line 44: IMO we should remove this comment. If we want to document this approach, we should do so in the component, or in the docs themselves.

```suggestion
    get "/controller_inline_with_block"
    
```

joelhawksley (approved):
Thanks for capturing this test case!

boardfish:
Good to merge if you're happy with it. :+1:

PR #1013, opened 2021-07-20
set max line length to 120

joelhawksley:
I've set our maximum line length to 120, what I think is a reasonable default.


srt32 (approved):
Works for me. Anything that limits trailing inline conditionals is a win in my book!

PR #1024, opened 2021-08-03
Add inline template example to benchmark script

andrewjtait:
### Summary

After trying to run some benchmarks to compare components that use an ERB file with components that use a `call` method, I discussed the results here: https://github.com/github/view_component/discussions/1014

It was then considered if we should add this example to the benchmark script, so I have raised this PR in case you want it included.

### Other Information

Here are the benchmark results after this change:

```
          component:    14.495k i/100ms
             inline:    13.457k i/100ms
            partial:     1.770k i/100ms
          component:    150.094k (± 4.5%) i/s -      1.507M in  10.064767s
             inline:    184.791k (± 4.5%) i/s -      1.857M in  10.071958s
            partial:     18.655k (± 4.8%) i/s -    187.620k in  10.081517s

Comparison:
             inline::   184791.1 i/s
          component::   150094.5 i/s - 1.23x  (± 0.00) slower
            partial::    18655.1 i/s - 9.91x  (± 0.00) slower
```

Also, I should mention that it looks as though the benchmark scripts have been broken since this change https://github.com/github/view_component/commit/e2f0379b1bd94b74064eedb29c4017fceb99d054 in which the sandbox app was moved to a different directory.

The main benchmark script was an easy fix as I just had to update the location of `environment.rb`, however the other benchmark scripts require a bit more work to fix so I have not included them in this change. I will raise them as a separate issue.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -19,6 +19,10 @@ title: Changelog
 
     *Ryo.gift*
 
+* Add inline template example to benchmark script.
+
+    *Andrew Tait*
+
 ## 2.36.0
 
 * Add `slot_type` helper method.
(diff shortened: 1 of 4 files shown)


andrewjtait:
I have raised the issue for the other broken benchmark scripts here https://github.com/github/view_component/issues/1025

joelhawksley (approved):
🙏🏻 Thanks for your contributions ❤️

PR #1027, opened 2021-08-03
Run benchmarks in CI

joelhawksley:
@andrewjtait recently pointed out that our benchmarks were broken.
In order to prevent this from happening again, let's run them in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,26 @@ jobs:
   test:
     needs: lint
     runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Setup Ruby
+      uses: ruby/setup-ruby@v1
+      with:
+        ruby-version: 2.7.2
+    - uses: actions/cache@v2
+      with:
+        path: vendor/bundle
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +58,26 @@ jobs:
...
+      with:
+        path: vendor/bundle
+        key: gems-build-rails-main-ruby-2.7.2-${{ hashFiles('**/Gemfile.lock') }}
+    - name: Run benchmarks
+      run: |
+        bundle config path vendor/bundle
+        bundle update
+        bundle exec rake benchmark
+        bundle exec rake translatable_benchmark
+        bundle exec rake slotable_benchmark
```
BlakeWilliams, line 80: Non-blocking, but It would be awesome if we could expose the results of these benchmarks as a build artifact.

BlakeWilliams (approved):

.github/workflows/ci.yml
joelhawksley: It would be! I'd also like to see us put in a basic stopgap for failing a build if they change too drastically.

PR #1030, opened 2021-08-07
Add --stimulus flag to component generator

### LINKED ISSUE #627: Add a `--stimulus` flag to generate a Stimulus controller for the component
## Feature request

As documentation mentions how to use Stimulus (see #481 and #485), I figured we could go a step further by adding [Stimulus](https://stimulus.hotwire.dev) support to the component generator.

Similar to `--inline` or `--sidecar`, it would be nice to allow a `--stimulus flag` when working with Stimulus:

```
bin/rails generate component Example --stimulus                 
Running via Spring preloader in process 15839
      create  app/components/example_component.rb
      invoke  test_unit
      create    test/components/example_component_test.rb
      invoke  slim
      create    app/components/example_component.html.slim
      create    app/components/example_component_controller.js
```

The content of the generated file (`app/components/example_component_controller.js`) could be:

```js
import { Controller } from "stimulus";

export default class extends Controller {
  connect() {
    console.log("Hello, Stimulus!", this.element);
  }
}
```

It would also be great if the generated template could include the correct `data-controller` attribute (the `parameterize`d name of the component, using `--` as a separator is the component is namespaced).

### Motivation

When adding a Stimulus controller to a component, you have to create the controller JS file and them add some boilerplate in it, then you have to add the `data-controller` attribute to the component root. This feature would make it faster by automating these two steps.

websebdev:
### Summary
Add `--stimulus` flag to the component generator.
This will create a basic Stimulus controller alongside the component.

Example:
```
$ rails generate component User --stimulus
      create  app/components/user_component.rb
      invoke  rspec
      create    spec/components/user_component_spec.rb
      invoke  stimulus
      create    app/components/user_component_controller.js
      invoke  erb
      create    app/components/user_component.html.erb
```

Closes #627

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add `--stimulus` flag to the component generator. Generates a Stimulus controller alongside the component.
+
+    *Sebastien Auriault*
+
 * Add test case for conflict with internal `@variant` variable.
 
    *David Backeus*
--- a/docs/index.md
+++ b/docs/index.md
@@ -172,6 +172,7 @@ ViewComponent is built by over a hundred members of the community, including:
 <img src="https://avatars.githubusercontent.com/Matt-Yorkley?s=64" alt="Matt-Yorkley" width="32" />
 <img src="https://avatars.githubusercontent.com/ryogift?s=64" alt="ryogift" width="32" />
 <img src="https://avatars.githubusercontent.com/andrewjtait?s=64" alt="andrewjtait" width="32" />
+<img src="https://avatars.githubusercontent.com/websebdev?s=64" alt="websebdev" width="32" />
 
 <hr />
 
--- a/lib/rails/generators/abstract_generator.rb
+++ b/lib/rails/generators/abstract_generator.rb
@@ -11,11 +11,19 @@ module ViewComponent
     private
 
     def destination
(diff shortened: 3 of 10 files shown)


joelhawksley (commented):
Looks great! Might we add a global option for this setting? We should also add documentation for this new functionality ❤️

websebdev:
@joelhawksley Done.
I added the global option `config.view_component.generate_stimulus_controller`. When set to `true`, the generator will automatically generate the stimulus controller. I'm not sure if that's what you had in mind though.

Also added documentation for both the new config option and the `---stimulus` flag.

Commit: https://github.com/github/view_component/pull/1030/commits/9346c714d6328e542218899993d3bffbd3081b9f

docs/CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ title: Changelog
 ## main
 
 * Add `--stimulus` flag to the component generator. Generates a Stimulus controller alongside the component.
+* Add config option `config.view_component.generate_stimulus_controller` to always generate a Stimulus controller.
```
joelhawksley, line 11: @Spone does this config attribute naming work for you?

Spone: Yep 👍👍

joelhawksley (approved):
Thank you for your contribution! ❤️

Spone:
@joelhawksley I noticed the documentation for the new `generate_stimulus_controller` option is missing from the documentation: https://viewcomponent.org/api.html

Got any idea why?

PR #1031, opened 2021-08-09
Update slots.md

ollie-nye:
### Summary

Fix bug in docs for passthrough `renders_many` slot call. Needed a block argument instead of parameter argument.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,7 +7,7 @@ title: Changelog
 
 ## main
 
-* Fix error in documentation for `render_many` passthrough slots
+* Fix error in documentation for `render_many` passthrough slots.
 
     *Ollie Nye*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Fix error in documentation for `render_many` passthrough slots
+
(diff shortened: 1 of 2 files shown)


boardfish:
@joelhawksley I looked at this one with @ollie-nye just now. This section of the docs uses passthrough slots declared with `renders_many :posts`, but passing arguments to the slot doesn't change the output. Passing a block does, though.

docs/guide/slots.md
```diff
@@ -44,7 +44,7 @@ To render a `renders_many` slot, iterate over the name of the slot:
   <% end %>
 
   <% @posts.each do |post| %>
-    <%= c.post(post: post) %>
+    <%= c.post { post } %>
```
boardfish, line 47:
```suggestion
    <%= c.post do %>
      <%= link_to post.name, post.url %>
    <% end %>
```

Adding some markup in here might better communicate that this is a passthrough slot that'll render the given markup.

boardfish (changes requested):

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Fix error in documentation for `render_many` passthrough slots
```
```suggestion
* Fix error in documentation for `render_many` passthrough slots.
```

joelhawksley (approved):

PR #956, opened 2021-06-09
refresh slots documentation to diminish emphasis on slot types

joelhawksley:
In their existing state, the slots documentation puts too much emphasis on the "Delegate" and other types that folks generally found confusing.

PR #1036, opened 2021-08-10
Increase clarity around purpose and use of slots

boardfish:
### Summary

I emailed @joelhawksley after discovering #1031 with @ollie-nye, since I'd identified some patterns in how folks are using slots. The aim of this PR is to make it clearer that slots render content and don't pass data.

This might make the docs a little more verbose and revert some of #956 by reintroducing passthrough slots as an official term, but I think that's important to make the distinction.

### Other Information

I also removed the comment under lambda slots stating that they "render the returned string". This isn't 100% true, because a component can also be the return value of a lambda slot's lambda, and that gets rendered in.

### DIFF
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -54,7 +54,14 @@ To render a `renders_many` slot, iterate over the name of the slot:
 <% end %>
 ```
 
-The link to "My blog" will render inside the `<h1>` tags, and links to each `BlogPost` will be rendered beneath.
+The rendered output will look like this:
+
+```erb
+<h1><a href="/">My blog</a></h1>
+
+<a href="/blog/first-post">First post</a>
+<a href="/blog/second-post">Second post</a>
+```
 
 ## Component slots
 
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -6,12 +6,16 @@ parent: Guide
 
 # Slots
 
-In addition to the `content` accessor, ViewComponents accept content through slots, enabling multiple blocks of content to be passed to a single ViewComponent.
+In addition to the `content` accessor, ViewComponents accept content through slots. Slots are dedicated areas of a
+component that render content, including other components.
 
 Slots are defined with `renders_one` and `renders_many`:
 
 - `renders_one` defines a slot that will be rendered at most once per component: `renders_one :header`
 - `renders_many` defines a slot that can be rendered multiple times per-component: `renders_many :posts`
+
+If you don't specify a second argument to these methods, you'll create a **passthrough slot**. Arbitrary content can be
+rendered inside these slots. If you need to render arbitrary markup or components, these may be the way to go.
 
 For example:
 
@@ -39,7 +43,7 @@ To render a `renders_many` slot, iterate over the name of the slot:
 ```erb
 <%# index.html.erb %>
 <%= render BlogComponent.new do |c| %>
-  <% c.header(classes: '') do %>
+  <% c.header do %>
     <%= link_to "My blog", root_path %>
   <% end %>
 
@@ -51,9 +55,13 @@ To render a `renders_many` slot, iterate over the name of the slot:
 <% end %>
 ```
 
+The link to "My blog" will render inside the `<h1>` tags, and links to each `BlogPost` will be rendered beneath.
+
 ## Component slots
 
-It's also possible to have a slot be a ViewComponent itself by passing in a second argument to `renders_one` and `renders_many`:
+Slots can also render other components. Pass the name of the component as the second argument. Component slots are
+useful when you want to render a specific component, but still allow full control of the arguments that are passed to
+it.
 
 ```ruby
 # blog_component.rb
@@ -107,18 +115,22 @@ end
 
 ## Lambda slots
 
-It's also possible to define a slot as a lambda that returns content to be rendered (either a string or a ViewComponent instance). Lambda slots are useful for working with helpers like `content_tag` or as wrappers for another ViewComponent with specific default values:
+Slots can be defined as lambdas that returns content to be rendered. Lambda slots are useful when you're outputting
+basic markup that doesn't warrant its own component yet.
 
 ```ruby
 class BlogComponent < ViewComponent::Base
-  # Renders the returned string
   renders_one :header, -> (classes:) do
+    # This isn't complex enough to be its own component yet, so we'll use a
+    # lambda slot. If it gets much bigger, it should be extracted out to a
+    # ViewComponent and rendered here with a component slot.
     content_tag :h1 do
       link_to title, root_path, { class: classes }
     end
   end
 
-  # Returns a ViewComponent that will be rendered in that slot with a default argument.
+  # They can also be used as wrappers for another ViewComponent with preset
+  # default values
   renders_many :posts, -> (title:, classes:) do
     PostComponent.new(title: title, classes: "my-default-class " + classes)
   end
@@ -141,7 +153,8 @@ end
 
 ## Rendering collections
 
-Collection slots (declared with `renders_many`) can also be passed a collection:
+`renders_many` slots can also be passed a collection. Instead of repeatedly calling the singular slot name (e.g.
+`c.link`), you can pass the arguments as an array.
 
 ```ruby
 # navigation_component.rb
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -13,8 +13,10 @@ Slots are defined with `renders_one` and `renders_many`:
 - `renders_one` defines a slot that will be rendered at most once per component: `renders_one :header`
 - `renders_many` defines a slot that can be rendered multiple times per-component: `renders_many :posts`
 
-If you don't specify a second argument to these methods, you'll create a **passthrough slot**. Arbitrary content can be
-rendered inside these slots. If you need to render arbitrary markup or components, these may be the way to go.
+If you don't specify a second argument to these methods, you'll create a
+**passthrough slot**. Any content you pass through can be rendered inside these
+slots. If you don't want to strictly specify what should go in the slot you're
+making, these may be the way to go.
 
 For example:
 
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -6,8 +6,7 @@ parent: Guide
 
 # Slots
 
-In addition to the `content` accessor, ViewComponents accept content through slots. Slots are dedicated areas of a
-component that render content, including other components.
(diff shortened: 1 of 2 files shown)


docs/guide/slots.md
```diff
@@ -39,7 +43,7 @@ To render a `renders_many` slot, iterate over the name of the slot:
 ```erb
 <%# index.html.erb %>
 <%= render BlogComponent.new do |c| %>
-  <% c.header(classes: '') do %>
```
boardfish, line 42: At this point in the docs, component slots haven't been introduced, so passing these arguments doesn't do anything.

docs/guide/slots.md
```diff
@@ -6,13 +6,17 @@ parent: Building ViewComponents
 
 # Slots
 
-In addition to the `content` accessor, ViewComponents accept content through slots, enabling multiple blocks of content to be passed to a single ViewComponent.
+In addition to the `content` accessor, ViewComponents accept content through slots. Slots are dedicated areas of a
+component that render content, including other components.
```
joelhawksley, line 10:
```suggestion
In addition to the `content` accessor, ViewComponents can accept content through slots. Think of slots as a way to render multiple blocks of content, including other components.
```

docs/guide/slots.md
```diff
@@ -6,13 +6,17 @@ parent: Building ViewComponents
...
 
 Slots are defined with `renders_one` and `renders_many`:
 
 - `renders_one` defines a slot that will be rendered at most once per component: `renders_one :header`
 - `renders_many` defines a slot that can be rendered multiple times per-component: `renders_many :posts`
 
+If you don't specify a second argument to these methods, you'll create a **passthrough slot**. Arbitrary content can be
+rendered inside these slots. If you need to render arbitrary markup or components, these may be the way to go.
+
 For example:
```
joelhawksley, line 20:
```suggestion
Arbitrary content can then be passed to the slots when rendering the component:
```

docs/guide/slots.md
```diff
@@ -51,9 +55,13 @@ To render a `renders_many` slot, iterate over the name of the slot:
 <% end %>
 ```
 
+The link to "My blog" will render inside the `<h1>` tags, and links to each `BlogPost` will be rendered beneath.
```
joelhawksley, line 58: I wonder if it would be simpler to just put what HTML this example renders here instead of describing it in words.

```suggestion
The link to "My blog" will render inside the `<h1>` tags, and links to each `BlogPost` will be rendered beneath.
```

docs/guide/slots.md
```diff
@@ -51,9 +55,13 @@ To render a `renders_many` slot, iterate over the name of the slot:
...
+
 ## Component slots
 
-It's also possible to have a slot be a ViewComponent itself by passing in a second argument to `renders_one` and `renders_many`:
+Slots can also render other components. Pass the name of the component as the second argument. Component slots are
+useful when you want to render a specific component, but still allow full control of the arguments that are passed to
+it.
```
joelhawksley, line 64:
```suggestion
Slots can also render other components by passing the name of the component as the second argument:
```

> enabling useful when you want to render a specific component, but still allow full control of the arguments that are passed to it.

I'm not sure what you mean by `full control` here. Can you explain another way?

docs/guide/slots.md
```diff
@@ -107,18 +115,22 @@ end
 
 ## Lambda slots
 
-It's also possible to define a slot as a lambda that returns content to be rendered (either a string or a ViewComponent instance). Lambda slots are useful for working with helpers like `content_tag` or as wrappers for another ViewComponent with specific default values:
+Slots can be defined as lambdas that returns content to be rendered. Lambda slots are useful when you're outputting
+basic markup that doesn't warrant its own component yet.
```
joelhawksley, line 119: IMO this loses some of the clarity and specificity of the original text. Maybe we could:

```suggestion
It's also possible to define a slot as a lambda that returns content to be rendered (either a string or a ViewComponent instance). Lambda slots are useful in cases where writing another component may be unnecessary, such as working with helpers like `content_tag` or as wrappers for another ViewComponent with specific default values:
```

docs/guide/slots.md
```diff
@@ -107,18 +115,22 @@ end
...
+    # lambda slot. If it gets much bigger, it should be extracted out to a
+    # ViewComponent and rendered here with a component slot.
     content_tag :h1 do
       link_to title, root_path, { class: classes }
     end
   end
 
-  # Returns a ViewComponent that will be rendered in that slot with a default argument.
+  # They can also be used as wrappers for another ViewComponent with preset
+  # default values
```
joelhawksley, line 133:
```suggestion
  # It's also possible to return another ViewComponent with preset default values:
```

docs/guide/slots.md
```diff
@@ -141,7 +153,8 @@ end
 
 ## Rendering collections
 
-Collection slots (declared with `renders_many`) can also be passed a collection:
+`renders_many` slots can also be passed a collection. Instead of repeatedly calling the singular slot name (e.g.
+`c.link`), you can pass the arguments as an array.
```
joelhawksley, line 157:
```suggestion
`renders_many` slots can also be passed a collection, using the plural setter (`links` in this example):
```

docs/guide/slots.md
boardfish: By "full control", I'm getting at the idea that lambda slots allow you to set defaults for, restrict, and modify the arguments passed to components, whereas component slots call the component with exactly the arguments you give it.

docs/guide/slots.md
boardfish: I think it's important that we keep the term "passthrough slot" in our glossary. I think it'll make it easier for users of the library to communicate solutions to problems. 

If the naming is an issue, it's just hit me that "content slot" could be a potential alternative here.

docs/guide/slots.md
boardfish: Agreed, I'll make the change.

docs/guide/slots.md
joelhawksley: Maybe we remove the term `arbitrary` then? I'm wondering whether the term `passthrough` is useful for helping people understand this feature or not. I recognize that I am not a good judge of this question as the creator of the framework and thus have different context than most people.

```suggestion
Content can then be passed to the slots when rendering the component:
```

docs/guide/slots.md
joelhawksley: I think that explanation would be nice to incorporate in place of `full control` 😄

docs/guide/slots.md
boardfish: I think you're right on both counts - arbitrary feels too vague, and passthrough accurately describes how the slot functions but doesn't fully demonstrate its value.

I've done some more thinking around naming here and I feel like I'm getting close to a potential solution:

- `template slots` is one option. It's trying to convey that this is a slot that forms part of a template (in the dictionary definition sense, not the view file sense). The fact that I had to clarify there is a bit of a red flag, but I feel like folks will reach for the dictionary definition first and thus understand that this is a slot into which anything can be rendered. 
- `content slots` is another, but then again, `content` is practically a (whatever-we're-calling-this) slot in its own right. Whether that causes confusion or actually bolsters understanding of the feature is probably dependent on the reader, but my gut feeling says it'll cause more confusion.

boardfish: I've updated this section as follows:

> If you don't specify a second argument to these methods, you'll create a
**passthrough slot**. Any content you pass through can be rendered inside these
slots. If you don't want to strictly specify what should go in the slot you're
making, these may be the way to go.

How does that sound?

boardfish: And on reflection, I think the naming conflicts with those alternatives I suggested are too strong for them to be considered.

boardfish:
This is ready for review again. I can't seem to re-request it.

Spone (approved):
These changes are welcome to make the slots easier to understand 👍

docs/guide/slots.md
```diff
@@ -6,13 +6,18 @@ parent: Guide
...
 Slots are defined with `renders_one` and `renders_many`:
 
 - `renders_one` defines a slot that will be rendered at most once per component: `renders_one :header`
 - `renders_many` defines a slot that can be rendered multiple times per-component: `renders_many :posts`
 
+If you don't specify a second argument to these methods, you'll create a
+**passthrough slot**. Any content you pass through can be rendered inside these
+slots. If you don't want to strictly specify what should go in the slot you're
+making, these may be the way to go.
+
```
joelhawksley, line 20:
```suggestion
If a second argument is not provided to these methods, a **passthrough slot** is registered. Any content passed through can be rendered inside these slots without restriction.

```

docs/guide/slots.md
```diff
@@ -51,9 +56,20 @@ To render a `renders_many` slot, iterate over the name of the slot:
 <% end %>
 ```
 
+The rendered output will look like this:
```
joelhawksley, line 59:
```suggestion
Returning:
```

docs/guide/slots.md
```diff
@@ -51,9 +56,20 @@ To render a `renders_many` slot, iterate over the name of the slot:
...
+<a href="/blog/first-post">First post</a>
+<a href="/blog/second-post">Second post</a>
+```
+
 ## Component slots
 
-It's also possible to have a slot be a ViewComponent itself by passing in a second argument to `renders_one` and `renders_many`:
+Slots can also render other components. Pass the name of a component as the second argument to define a component slot.
+
+Any arguments you pass when calling a component slot will be used to initialize the component and render it. You can also pass a block to set that component's content and slots.
```
joelhawksley, line 72:
```suggestion
Arguments passed when calling a component slot will be used to initialize the component and render it. A block can also be passed to set the component's content.
```

joelhawksley (approved):

docs/guide/slots.md
```diff
@@ -6,13 +6,15 @@ parent: Guide
...
 
-In addition to the `content` accessor, ViewComponents accept content through slots, enabling multiple blocks of content to be passed to a single ViewComponent.
+In addition to the `content` accessor, ViewComponents can accept content through slots. Think of slots as a way to render multiple blocks of content, including other components.
 
 Slots are defined with `renders_one` and `renders_many`:
 
 - `renders_one` defines a slot that will be rendered at most once per component: `renders_one :header`
 - `renders_many` defines a slot that can be rendered multiple times per-component: `renders_many :posts`
 
+If a second argument is not provided to these methods, a **passthrough slot** is registered. Any content passed through can be rendered inside these slots without restriction.
```
Spone, line 16: To make the Vale linter happy:

```suggestion
If a second argument isn't provided to these methods, a **passthrough slot** is registered. Any content passed through can be rendered inside these slots without restriction.
```

PR #1041, opened 2021-08-16
Deprecate Base#with_variant

camertron:
### Summary

Deprecates `Base#with_variant`. Variants are an often misunderstood feature that use the Action Pack variant information present in the request (i.e. phone, tablet, etc) to render a component differently depending on the requesting device. Normally one passes in a list of acceptable variants in an Action View template, eg:

```ruby
render(MyComponent.new, variants: [:phone, :tablet])
```

The variant can be overridden by calling `#with_variant` on the component, eg:

```ruby
  def call
    render(OtherComponent.new.with_variant(:tablet))
  end
end
```

It's important to note that these are two distinct `render` functions (the second does not accept a `:variants` option). Since components don't have access to the request object, it doesn't really make sense for them to decide which variant to render (that's probably why `:variants` isn't available in the first place). The consensus on the team was that overriding the variant is something that should really only be done in a test environment and not supported otherwise.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -218,9 +218,15 @@ module ViewComponent
 
     # Use the provided variant instead of the one determined by the current request.
     #
+    # @deprecated Define separate per-variant templates instead. Will be removed in v3.0.0.
     # @param variant [Symbol] The variant to be used by the component.
     # @return [self]
     def with_variant(variant)
+      ActiveSupport::Deprecation.warn(
+        "`with_variant` is deprecated and will be removed in ViewComponent v3.0.0.\n\n" \
+        "Define separate per-variant templates instead."
+      )
+
       @__vc_variant = variant
 
       self
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -218,7 +218,7 @@ module ViewComponent
 
     # Use the provided variant instead of the one determined by the current request.
     #
-    # @deprecated Define separate per-variant templates instead. Will be removed in v3.0.0.
+    # @deprecated Will be removed in v3.0.0.
     # @param variant [Symbol] The variant to be used by the component.
     # @return [self]
     def with_variant(variant)
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -223,8 +223,7 @@ module ViewComponent
     # @return [self]
     def with_variant(variant)
       ActiveSupport::Deprecation.warn(
-        "`with_variant` is deprecated and will be removed in ViewComponent v3.0.0.\n\n" \
-        "Define separate per-variant templates instead."
+        "`with_variant` is deprecated and will be removed in ViewComponent v3.0.0."
       )
 
       @__vc_variant = variant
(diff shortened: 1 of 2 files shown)


lib/view_component/base.rb
```diff
@@ -218,9 +218,15 @@ def format
 
     # Use the provided variant instead of the one determined by the current request.
     #
+    # @deprecated Define separate per-variant templates instead. Will be removed in v3.0.0.
     # @param variant [Symbol] The variant to be used by the component.
     # @return [self]
     def with_variant(variant)
+      ActiveSupport::Deprecation.warn(
+        "`with_variant` is deprecated and will be removed in ViewComponent v3.0.0.\n\n" \
+        "Define separate per-variant templates instead."
```
joelhawksley, line 227: I'm not sure if this is the guidance we want to provide. Perhaps we could leave this line out for now and see if folks ask for guidance on alternatives?

camertron: Ok sounds good! 😄

lib/view_component/base.rb
```diff
@@ -218,9 +218,14 @@ def format
 
     # Use the provided variant instead of the one determined by the current request.
     #
+    # @deprecated Define separate per-variant templates instead. Will be removed in v3.0.0.
```
joelhawksley, line 221:
```suggestion
    # @deprecated Will be removed in v3.0.0.
```

joelhawksley (approved):

lib/view_component/base.rb
camertron: Woops, thanks!

lib/view_component/base.rb
Sohair63: what is the alternative for this?

BlakeWilliams: @Sohair63 I think it depends on your specific use-case, but the general guidance is:

* https://guides.rubyonrails.org/layouts_and_rendering.html#the-variants-option is what powers the variant functionality in view component. (The only change we made is disabling explicitly setting variants on a component.)
* You can pass an argument to the component so it renders different content.
* You can conditionally render a different component based on some condition in your code.

coorasse:
We are using variants, right now, to define two different templates for the same component. 
Noticing the Deprecation Warning we realised that we are abusing the variant concept.
The idea is simple: the same component renders something in a page, and something completely different in another page.
We keep the same component, and just define two variants. How do we achieve that otherwise?

coorasse:
Thanks for the input. 
Complexity does not matter too much in my opinion: the point is that we need to pass an additional parameter `mode`, which basically decides the template to render. We then have to use a `if...else` somewhere to decide which template to render. All of this was already possible with `with_variant` which was very convenient and allowing a component to have multiple templates. That's not possible anymore, but needs to be implemented "manually". It would be nice to have an alternative 😄

joelhawksley:
@coorasse I think that's a good idea for a feature proposal then! https://github.com/github/view_component/pull/1096 has some overlap with what you're looking to do, perhaps you could check that out and provide commentary there?

PR #1046, opened 2021-08-23
Prevent nil from being returned from a Slot V2's to_s method

### LINKED ISSUE #1044: view_component raises an error when a slot capture block returns nil (using Slim)
### Steps to reproduce
Run the newly generated rails app here: https://github.com/yjukaku/view_component_nil_error and open localhost:3000/test

For reference, here's an example of code that triggers this issue:

```
# test_component.rb
class TestComponent < ViewComponent::Base
  renders_one :my_slot
end

# test_component.html.slim
| Hello!
= my_slot

# test.html.slim
= render TestComponent.new do |c|
  = c.my_slot do
    - something_that_returns_nil
```

### Expected behavior

The component should render, with no content in the slot.

### Actual behavior
An error is raised: `no implicit conversion of nil into String`

```
temple (0.8.2) lib/temple/utils.rb:38:in `escapeHTML'
temple (0.8.2) lib/temple/utils.rb:38:in `escape_html'
temple (0.8.2) lib/temple/utils.rb:21:in `escape_html_safe'
app/views/test/test.html.slim:4
actionview (6.1.4.1) lib/action_view/helpers/capture_helper.rb:45:in `block in capture'
actionview (6.1.4.1) lib/action_view/helpers/capture_helper.rb:209:in `with_output_buffer'
actionview (6.1.4.1) lib/action_view/helpers/capture_helper.rb:45:in `capture'
view_component (2.38.0) lib/view_component/base.rb:247:in `content'
view_component (2.38.0) lib/view_component/slotable_v2.rb:222:in `get_slot'
view_component (2.38.0) lib/view_component/slotable_v2.rb:72:in `block in renders_one'
app/components/test_component.html.slim:3:in `call'
view_component (2.38.0) lib/view_component/compiler.rb:83:in `render_template_for'
view_component (2.38.0) lib/view_component/base.rb:102:in `render_in'
actionview (6.1.4.1) lib/action_view/helpers/rendering_helper.rb:43:in `render'
app/views/test/test.html.slim:1
actionview (6.1.4.1) lib/action_view/base.rb:247:in `public_send'
```

**Please note: The line number in the exception trace corresponding to `test.html.slim file` is off by one, for some reason, so keep that in mind while debugging.**

### System configuration
**Rails version**:
Rails 6.1.4.1

**Ruby version**:
ruby 3.0.2p107 (2021-07-07 revision 0db68f0233) [x86_64-darwin19]

**Gem version**:
view_component-2.38.0

---

The example above is a bit contrived, but here's a more likely example:

```
# test.html.slim
= render TestComponent.new do |c|
  = c.my_slot
    - if user_signed_in?
       | Welcome!
```

In that case, the `if` condition evaluating to false implies a `nil` return.

yjukaku:
### Summary

Closes #1044 

When using `slim-rails` (and probably other temple-based languages), it's possible that a block defined in a Slim view evaluates to `nil`. When that block gets processed in a SlotV2's `to_s` method, that `nil` from the block can end up as the return value.

Returning `nil` from a call to `to_s` doesn't make sense and causes errors to be raised when rendering the view, so this patch prevents that from happening and instead returns an empty string.

### Other Information

I don't think the test I included is a good way to test this, but I can't think of a way to get the "real" issue to happen. `render_inline` doesn't raise an error like a Rails app using Slim will, because the temple gem used by Slim is what's raising the actual error I'm seeing in our app.

Perhaps instead of the test, I can include a comment next to the actual code change explaining that @content can be set to `nil` with certain template languages?

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -15,7 +15,7 @@ title: Changelog
 
     *Nicolas Brousse*
 
-* Fix bug when using Slim and writing a slot whose block evaluates to `nil`
+* Fix bug when using Slim and writing a slot whose block evaluates to `nil`.
 
     *Yousuf Jukaku*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,10 @@ title: Changelog
 
     *Nicolas Brousse*
 
+* Fix bug when using Slim and writing a slot whose block evaluates to `nil`
+
+    *Yousuf Jukaku*
+
 ## 2.39.0
 
 * Clarify documentation of `with_variant` as an override of Action Pack.
--- a/test/view_component/slotable_v2_test.rb
+++ b/test/view_component/slotable_v2_test.rb
@@ -272,6 +272,17 @@ class SlotsV2sTest < ViewComponent::TestCase
     assert_selector("p", text: "Footer part 2")
   end
 
+  def test_slot_with_block_content_returning_nil
+    component = SlotsV2Component.new
+    slot = component.title nil
+    slot.__vc_content_block = Proc.new { nil }
+    # Render the component so that the slot has a view_context
+    render_inline component
+
+    # No matter what, the slot shouldn't return nil from `to_s`
+    assert_not_nil(slot.to_s)
+  end
+
   def test_lambda_slot_with_missing_block
     render_inline(SlotsV2Component.new(classes: "mt-4")) do |component|
       component.footer(classes: "text-blue")
--- a/test/view_component/slotable_v2_test.rb
+++ b/test/view_component/slotable_v2_test.rb
@@ -272,17 +272,6 @@ class SlotsV2sTest < ViewComponent::TestCase
     assert_selector("p", text: "Footer part 2")
   end
 
-  def test_slot_with_block_content_returning_nil
-    component = SlotsV2Component.new
-    slot = component.title nil
-    slot.__vc_content_block = Proc.new { nil }
-    # Render the component so that the slot has a view_context
-    render_inline component
-
-    # No matter what, the slot shouldn't return nil from `to_s`
-    assert_not_nil(slot.to_s)
-  end
-
   def test_lambda_slot_with_missing_block
     render_inline(SlotsV2Component.new(classes: "mt-4")) do |component|
       component.footer(classes: "text-blue")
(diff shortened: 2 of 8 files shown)


yjukaku:
@joelhawksley WDYT?

docs/CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ title: Changelog
 
     *Nicolas Brousse*
 
+* Fix bug when using Slim and writing a slot whose block evaluates to `nil`
```
joelhawksley, line 18:
```suggestion
* Fix bug when using Slim and writing a slot whose block evaluates to `nil`.
```

test/view_component/slotable_v2_test.rb
```diff
@@ -272,6 +272,17 @@ def test_slot_with_block_content
...
+  def test_slot_with_block_content_returning_nil
+    component = SlotsV2Component.new
+    slot = component.title nil
+    slot.__vc_content_block = Proc.new { nil }
+    # Render the component so that the slot has a view_context
+    render_inline component
+
+    # No matter what, the slot shouldn't return nil from `to_s`
+    assert_not_nil(slot.to_s)
+  end
```
joelhawksley, line 284: This test _feels_ a little too coupled to implementation details. Might we instead add a case to `integration_test.rb` that mimics the failure mode in an actual slim file?

test/sandbox/app/views/integration_examples/empty_slot.html.erb
```diff
@@ -2,3 +2,8 @@
   <% component.slot(:title) do %>
   <% end %>
 <% end %>
+
+<%= render(EmptySlotV2ComponentWithSlim.new) do |component| %>
```
yjukaku, line 6: At first I added an entirely new route, view, and component (`/nil_slot`, `nil_slot.html.erb`, and `NilSlotComponent.rb`), but I think this is clearer: I copied the EmptySlotComponent, updated it to use SlotV2, and then re-used the same integration test to test both view components. WDYT?

yjukaku:
Just FYI, I think this also happens when using ERB, if you write a VC using SlotV2 like this:

```ruby
  renders_one :title

  def call
    title
  end
end
```

and use it like this:

```erb
<%= render MyComponent.new do |c| %>
  <%= c.title do %>
  <% end %>
<% end %>
```

I think it's because the `title` returns `nil` in the `call` method without this PR's changes, so then the entire VC will return `nil` when `.to_s` is called.

joelhawksley (approved):
👏🏻

yjukaku:
Awesome, thanks for merging!

Issue #1029, opened 2021-08-07
Previews: Sample source code doesn't show if view used


AlecRust:
### Steps to reproduce

1. Set `config.view_component.show_previews_source = true`
2. Create a preview i.e.
    `test/components/previews/alert_preview.rb`
    ```ruby
    class AlertPreview < ViewComponent::Preview
      def info
        render Alert.new(type: :info, message: 'This is a info alert.')
      end
    end
    ```
3. Navigate to preview e.g. `/rails/view-components/alert/info`
4. See that sample source code is correctly shown under the component preview
5. Move the preview into an ERB file as-per [docs](https://github.com/github/view_component/blob/main/docs/guide/previews.md#preview-templates) i.e.
    `test/components/previews/alert_preview.rb`
    ```ruby
    class AlertPreview < ViewComponent::Preview
      def info; end
    end
    ```
    `test/components/previews/alert_preview/info.html.erb`
    ```erb
    <%= render Alert.new(type: :info, message: 'This is a info alert') %>
    ```
6. Navigate to preview e.g. `/rails/view-components/alert/info`
7. See that sample source code is **not** shown under the component preview

### Expected behavior

When I enable `config.view_component.show_previews_source` I expect the sample source code to show under the component preview regardless of how the preview files are structured.

### Actual behavior

**When a view file is used for the preview** if `config.view_component.show_previews_source = true` is set the sample source code is not shown under the component preview.

### Further details

**When a view file is used for the preview** this gem's `app/views/view_components/preview.html.erb` is not used to render the page, instead only the view file e.g. `test/components/previews/alert_preview/info.html.erb` is rendered that doesn't include the sample source code output: https://github.com/github/view_component/blob/fe121163f6d4d2962e6c3e16a3d83296f3ce4f77/app/views/view_components/preview.html.erb#L7-L9

I guess ideally the content of `test/components/previews/alert_preview/info.html.erb` would be loaded into `app/views/view_components/preview.html.erb` so that `app/views/view_components/preview.html.erb` can be relied upon as the template for all previews?

This issue also means I can't easily set the `<title>` of the "show preview" page as if I set this in `app/views/view_components/preview.html.erb` it will only apply if a view file isn't being used for the preview.

### System configuration
**Rails version**: 6.1.4

**Ruby version**: 2.7.4

**Gem version**: 2.36.0

joelhawksley:
@johannesengl might you have some thoughts on this issue? I think we should have the source preview work in this case as well.

[...]

PR #1047, opened 2021-08-23
[1029] Preview the source of templates when exclusively used

edwinthinks:
### Summary

This PR addresses #1029 by modifying the `preview.html.erb` and associated code to properly display the source whether the preview uses a template or not.

Screen Shot 2021-08-23 at 6 01 42 PM

Thanks, @joelhawksley for the pairing session!

### Other Information

I noticed that ability to set `locals` via the query parameters doesn't work. For example, if you decided to do something like call `rails/view_components/previews/default?title='Meow'`

```
# preview_method
def default(title: 'Foobar')
   @title = title
end

# preview_template
<%=  render(CardComponent.new(title: @title)) %>
```

It wouldn't work and cause a raised exception. I'am wondering if this is a situation that we'd be interested in supporting?


test/view_component/integration_test.rb
```diff
@@ -320,18 +320,22 @@ def test_preview_from_multiple_preview_paths
...
   end
 
-  def test_renders_preview_source
+  def test_renders_preview_source_without_template
     get "/rails/view_components/preview_component/default"
 
     assert_select ".view-component-source-example h2", "Source:"
     assert_select ".view-component-source-example pre.source code"
+    refute_match '&lt;%=', response.body
+    refute_match '%&gt', response.body
```
edwinthinks, line 329: Wondering if there is a better way to ensure that the source is or isn't template code?

app/views/view_components/preview.html.erb
```diff
@@ -1,7 +1,11 @@
...
-  <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
+<% if @render_args[:component] %>
+  <% if ViewComponent::Base.render_monkey_patch_enabled || Rails.version.to_f >= 6.1 %>
+    <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
+  <% else %>
+    <%= render_component(@render_args[:component], &@render_args[:block]) %>
+  <% end %>
 <% else %>
-  <%= render_component(@render_args[:component], &@render_args[:block]) %>
+  <%= render template: @render_args[:template], locals: @render_args[:locals] || {} %>
```
edwinthinks, line 8: Looks like in Rails 5.x this is causing some issues in displaying the `preview_source`. My guess is that in app/views/view_components/_preview_source.html.erb:
```
<link href="<%= asset_path('prism.css', skip_pipeline: true) %>" media="screen" rel="stylesheet" type="text/css">
<div class="view-component-source-example">
  <h2>Source:</h2>
  <pre class="source">
    <% if @render_args[:component] %>
      <code class="language-ruby">
        <%= h @preview.preview_source(@example_name) %>
      </code>
    <% else %>
      <% template = @view_renderer.lookup_context.find_template(@render_args[:template]) %>
      <code class="language-<%= prism_language_name(template: template) %>">
        <%= h template.source %>
      </code>
    <% end %>
  </pre>
</div>
<script type="text/javascript" src="<%= asset_path('prism.min.js', skip_pipeline: true) %>"></script>
```

The `@view_renderer` doesn't know about the rendering when using this new way to render the template.

Thoughts?

edwinthinks:
**Update**

I've run into some issues trying to get this PR to work on Rails 5.x which caused source previews to not render. I'am not 100% sure why it was occurring but I suspect it has something to do with the monkey patching. I did find a way to get around it by fetching the template using the `view_paths` when working on Rails versions < 6.1 and it seems to work on all versions specified in CI now.

However now...

@joelhawksley I'am getting a failing test on `CI / pvc` and I'am not entirely sure why. I've pulled down the primer repo and ran the tests on my own branch locally which doesn't seem to trigger this test that is failing in CI:
```
==========================================================================
markdown#default failed check.
==========================================================================
[Screenshot Image]: /home/runner/work/view_component/view_component/primer_view_components/demo/tmp/screenshots/failures_test_accessibility_of_doc_examples.png
/home/runner/work/view_component/view_component/primer_view_components/vendor/bundle/ruby/2.7.0/gems/ferrum-0.11/lib/ferrum/target.rb:21: warning: instance variable @page not initialized
E

Error:
IntegrationDocExamplesAxeTest#test_accessibility_of_doc_examples:
Timeout::Error: execution expired
```

When you get a chance can you provide some guidance on what could be going wrong here?

Thanks!

app/views/view_components/_preview_source.html.erb
```diff
@@ -7,9 +7,13 @@
...
       </code>
     <% else %>
-      <% template = @view_renderer.lookup_context.find_template(@render_args[:template]) %>
+      <% template = find_template_source(lookup_context: @view_renderer.lookup_context, template_identifier: @render_args[:template]) %>
       <code class="language-<%= prism_language_name(template: template) %>">
-        <%= h template.source %>
+        <% if template.respond_to?(:source) %>
+          <%= h template.source %>
+        <% else %>
+          <%= h File.read(template) %>
```
edwinthinks, line 15: I'am looking for feedback if this is an ok approach. I've never really used `File` in a template before and don't know if we see any consequences. Based on how this is written, I think it isn't an issue but just wanted to make sure.

joelhawksley: This makes me think that we should change `find_template_source` to simply return the source of the template, and not an object or file path as it appears to now. What do you think?

edwinthinks: @joelhawksley are you suggesting that we wouldn't even need to do `template.source` like in line 13? I think that might add an extra layer of complexity since the method to determine the extension or type is expecting an object.

I'am going to try to modify it in such a way that might be more verbose but clearer in the end. I do think we shouldn't cause any confusion about wither or not `template` is an object or a file path. 

I'll see what I can do in terms of simplifying

edwinthinks: @joelhawksley hey there! I've made some changes to simplify things. My main goal was to reduce the complexity of the template and try to arrange it in the helper method. This way -- looking at the preview template doesn't require the reader to know exactly how the template source or language is being derived.

Let me know how this looks

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Display preview source on previews that exclusively use templates
```
```suggestion
* Display preview source on previews that exclusively use templates.
```

app/controllers/view_components_controller.rb
```diff
@@ -34,7 +34,7 @@ def previews
       opts = {}
       opts[:layout] = layout if layout.present? || layout == false
       opts[:locals] = locals if locals.present?
-      render template, opts # rubocop:disable GitHub/RailsControllerRenderLiteral
```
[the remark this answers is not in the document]
edwinthinks: Good catch! Removing!

docs/index.md
```diff
@@ -175,10 +175,12 @@ ViewComponent is built by over a hundred members of the community, including:
 <img src="https://avatars.githubusercontent.com/ryogift?s=64" alt="ryogift" width="32" />
 <img src="https://avatars.githubusercontent.com/andrewjtait?s=64" alt="andrewjtait" width="32" />
 <img src="https://avatars.githubusercontent.com/websebdev?s=64" alt="websebdev" width="32" />
+<img src="https://avatars.githubusercontent.com/edwinthinks?s=64" alt="edwinthinks" width="32" />
 <img src="https://avatars.githubusercontent.com/xkraty?s=64" alt="xkraty" width="32" />
 <img src="https://avatars.githubusercontent.com/wdrexler?s=64" alt="wdrexler" width="32" />
 <img src="https://avatars.githubusercontent.com/mattwr18?s=64" alt="mattwr18" width="32" />
 
+
```
edwinthinks, line 183:
```suggestion
```

edwinthinks:
@joelhawksley hey there -- I was wondering if you can give some clues as to why the `pvc` CI step is failing. https://github.com/github/view_component/pull/1047/checks?check_run_id=3674817827

joelhawksley:
> @joelhawksley hey there -- I was wondering if you can give some clues as to why the `pvc` CI step is failing. https://github.com/github/view_component/pull/1047/checks?check_run_id=3674817827

@edwinthinks I see this in the output:

```
NoMethodError: undefined method `execute_script' for #<Capybara::Cuprite::Browser:0x0000558394d64120>
```

Are you able to run the https://github.com/primer/view_components test suite against your changes locally? That might make it easier to debug.

edwinthinks:
Thanks @joelhawksley ! I'll see if i can try that out soon :)

juanmanuelramallo (approved):
Changes look good to me! Thank you!

I'll leave the approval/merge decision to @joelhawksley though as I've not been actively working on the gem lately.

edwinthinks:
Thanks @juanmanuelramallo . Please let me know @joelhawksley if there anything left to do here :)

docs/CHANGELOG.md
```diff
@@ -25,6 +29,7 @@ title: Changelog
 
     *Simon Fish*
 
+
```
edwinthinks, line 32:
```suggestion
```

joelhawksley (approved):

PR #1051, opened 2021-08-31
Error out in the CI if docs/api.md has to be regenerated

dmarcoux:
### Summary

Regenerating the API docs can be easily forgotten whenever creating/updating YARD comments. With the CI, this can be prevented.

### Other Information

I based the new job on the existing `benchmark` job.

Committing the changes in the API docs could be done, but I think it's better to let the PR submitter handle this since it can vary from a PR to another depending on the number of commits.


joelhawksley (approved):
Thanks, Dany!

PR #1055, opened 2021-09-02
Fix with_collection_parameter to be inherited from parent component

### LINKED ISSUE #1054: with_collection_parameter inherit
Hello!

### Steps to reproduce

```ruby
module Datalist
  class Base < ApplicationComponent
    with_collection_parameter :record
    attr_accessor :record

    def initialize(record:)
      self.record = record
    end
  end
end

module Datalist
  class User < ::Datalist::Base
  end
end
```

### Expected behavior

By running
```ruby
= render ::Datalist::User.with_collection(@users)
```
I expect `DatalistUser` to inherit `with_collection_parameter :record` and I get this error
```ruby
The initializer for Datalist::User does not accept the parameter `user`, which is required in order to render it as a collection.
```

### Actual behavior

But I do need to specify it again
```ruby
module Datalist
  class User < ::Datalist::Base
    with_collection_parameter :record
  end
end
```

So I'm guessing the extended class override the collection parameter with the default one and I'm not sure is intended.
Thanks
Chris

### System configuration
**Rails version**: Rails 6.1.4.1

**Ruby version**: 2.7.4

**Gem version**: 2.38.0

xkraty:
### Summary

Fixes #1054

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,8 @@ title: Changelog
 
 * Fix collection parameter to be inherited from parent component
 
+    *Will Drexler*, *Christian Campoli*
+
 * Add components to `rails stats` task.
 
     *Nicolas Brousse*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,7 +7,7 @@ title: Changelog
 
 ## main
 
-* Fix collection parameter to be inherited from parent component
+* Fix `with_collection_parameter` to be inherited from parent component
 
     *Will Drexler*, *Christian Campoli*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,8 @@ title: Changelog
 
 ## main
 
+* Fix collection parameter to be inherited from parent component
+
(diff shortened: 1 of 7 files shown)


docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Fix `with_collection_parameter` to be inherited from parent component
```
```suggestion
* Fix bug where `with_collection_parameter` did not inherit from parent component.
```

test/sandbox/app/components/collection_iteration_extend_override_component.html.erb
```diff
@@ -0,0 +1,6 @@
+<figure data-index="<%= @iteration.index %>" class="<%= "first" if @iteration.first? %>">
+  <img src="<%= @override.url %>" alt="<%= @override.title %>">
+  <figcaption>
+    Photo.<%= @counter %> - <%= @override.caption %>
+  </figcaption>
+  <figure>
```
joelhawksley, line 6:
```suggestion
<figure>
```

joelhawksley (approved):

joelhawksley:
Thank you!

PR #1058, opened 2021-09-04
Allow query parameters in `with_request_url`

javierm:
### Summary

Query parameters were ignored when using the `with_request_url` test helper. With these changes, that's no longer the case :smile:.

### Other Information

My knowledge of the internals of Rails routing is very limited, so if there's a more elegant solution than using `set_header` and `parse_query`, please let me know! :wink:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Allow query parameters in `with_request_url` test helper.
+
+    *Javi Martín*
+
 * Replace antipatterns section in the documentation with best practices.
 
     *Blake Williams*
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -137,6 +137,19 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+It's also possible to set query parameters:
+
+```ruby
+class ExampleComponentTest < ViewComponent::TestCase
+  def test_with_request_url
+    with_request_url "/products/42?locale=en" do
+      render_inline ExampleComponent.new # contains i.e. `link_to "Recent", url_for(request.query_parameters.merge(filter: "recent"))`
+      assert_link "Recent", href: "/?locale=en&filter=recent"
+    end
+  end
+end
+```
+
 ## RSpec configuration
 
 To use RSpec, add the following:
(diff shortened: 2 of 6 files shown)


docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Allow query parameters in `with_request_url`
```
```suggestion
* Allow query parameters in `with_request_url` test helper.
```

docs/guide/testing.md
```diff
@@ -137,6 +137,19 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+You can also set `request.query_parameters`, using the same method:
```
joelhawksley, line 140: It's also possible to set query parameters:

lib/view_component/test_helpers.rb
```diff
@@ -73,12 +73,15 @@ def with_controller_class(klass)
 
     def with_request_url(path)
       old_request_path_parameters = request.path_parameters
+      old_request_query_parameters = request.query_parameters
       old_controller = defined?(@controller) && @controller
 
       request.path_parameters = Rails.application.routes.recognize_path(path)
+      request.set_header("action_dispatch.request.query_parameters", request.send(:parse_query, path.split("?")[1]))
```
joelhawksley, line 80: Did you look into a way to pull this off with a public API? I'd really prefer we avoid using `send`.

joelhawksley (commented):
👋🏻 looking good! Thank you for the nice tests and docs ❤️

lib/view_component/test_helpers.rb
javierm: I agree 100% :+1:. The other day I didn't find anything and hoped there was a way to set query parameters similar to `#path_parameters=` that most people knew about.

I'll have another look :+1:.

javierm:
@joelhawksley Thank you very much for your review! :pray:

I've [applied your suggestions](https://github.com/github/view_component/compare/14bcf013e..96bc25f69). Also added another assertion for the case of several query parameters, to make sure the `&` is correctly treated as the separator.

If you've got any more comments, they're welcome :wink:.

joelhawksley (approved):

joelhawksley:
Thank you for your contribution, Javi!

Issue #904, opened 2021-05-04
Add paved path for unit testing components with behaviors


joelhawksley:
### Problem statement

> How have y'all at Github approached testing view components that have an interactive component to them? We've mainly been unit testing each view component we add and then writing feature specs for browser interactions, but I'm wondering if there is a better way to approach it - RailsConf Q&A

### Proposed solution

In Primer ViewComponents, we use system tests powered by cuprite and ViewComponent Previews for testing behaviors. As this approach matures, we should aim to move of much of it as possible into ViewComponent.

joelhawksley:
@edwinthinks thank you for your interest!

I could imagine us implementing a new base test case for browser-based tests. Something like Rails System tests, but catered to ViewComponent.

I could imagine this functionality being driven by some sort of `render_in_browser` test helper that accepts a component instance.

edwinthinks:
Thanks that is helpful context.

Though I'am wondering how something like the `render_in_browser` would work in practice? Specifically, how would it handle the layout (which probably contains the snippets of code to load the JS for the interaction). My initial idea was to accept a layout as a parameter that would generate the corresponding HTML in a temporary file and have capybara load it up. Thoughts?

Wondering if you'd like to brainstorm and pair on this in the next few weeks?

edwinthinks:
Opened up https://github.com/github/view_component/pull/1061 which I think is close-ish to what you were saying @joelhawksley. 

It does appear to work in some situations but I'am not entirely sure what the ideal way to handle assets is and also to make this compatible for Rails versions lower than 6.1. 

Excited to see this through and see if it helps others with testing components easier :)

jonspalmer:
Commented in #1061 but I feel pretty strongly that these kinds of tests should be a flavor of system test and leverage all the existing machinery for running system tests against multiple browsers, Selenium Grid etc. Problems with assets, differences with webpacker or not etc are already well solved in system tests. It would also encourage using a "production layout" for the css/js. What we need to make these tests simple to write is some  mechanism for defining the html endpoint (we do this with previews) and a likely some helpers to dry up the test declarations (ie something that auto 'gets' the html content for a named VC

edwinthinks:
Thanks, @jonspalmer; 

I'll be honest I don't have a lot of experience with system tests as it compares to browser tests. *Could you send me some resources to explain the difference? Thanks!* But if using system tests is going to simply things going forward then I think that is a good route to go.

[...]

PR #1061, opened 2021-09-05
[904] Enable testing component JS interactions using new with_rendered_component_path method

edwinthinks:
### Summary

Related to #904

I was able to add a method called `with_rendered_component_path` as a proof of concept to be able to test JS interactions of components within a browser. This PR adds a method that takes a component and creates a Tempfile containing the resultant HTML. This resultant HTML is then loaded in the browser via capybara. See the new test: `test_simple_js_interaction_in_browser` to see how it would function.

I believe this will be particularly useful in cases you would want to test a component in isolation that it works properly with JS!

Love to get feedback if this is on track or if this is way off hehe.

### Other Information
N/A


boardfish (commented):
> Do we want to make it so you don't need to specify page as it does in this PR?

That could be useful. One pattern I like to use in component tests is:

```ruby
RSpec.describe SomeComponent do
  subject { rendered_component }
  before { render component }
  let(:component) { described_class.new }

  it { is_expected.to have_css('div') }
end
```

`render component` having its output memoized to `rendered_component` is what makes this tick. So if `render_in_browser` had the same behavior, it'd feel intuitive. Potentially, this could even be reachable from the same accessor, as I'm not sure why you'd use both `render` and `render_in_browser` in the same test.

edwinthinks:
> That could be useful. One pattern I like to use in component tests is:

Makes sense! I'll see what can be done about that.

> render component having its output memoized to rendered_component is what makes this tick. So if render_in_browser had the same behavior, it'd feel intuitive. Potentially, this could even be reachable from the same accessor, as I'm not sure why you'd use both render and render_in_browser in the same test.

Yeh, I think one way to approach that I think can be used is using a different base class for specific tests like this. Something like a ViewComponentTestInBrowser (not the actual name I'am proposing haha) like thing.

test/view_component/view_component_browser_test.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ViewComponentBrowserTest < ViewComponent::BrowserTestCase
+  if Rails.version.to_f >= 6.1
```
edwinthinks, line 6: Ideally, we'd remove this conditional since we do want to enable loading the layout for Rails <= 6.1. If possible, love to see if there is any thoughts on how to make this work in those cases.

jonspalmer:
I like the feature set here. However, I wonder if it might be more natural to run 'in browser' by leveraging a flavor of Rails system tests. That's a more established existing pattern for doing this kind of thing. It would have the advantage of supporting many other testing strategies (like running the system test suites against multiple flavors of headless browser or against Selenium grid etc.) Rails System tests handle Capybara already what is missing is an 'end point' that renders the html etc. but we have that with Previews. I wonder if we can combine those two to get the features you present here.

edwinthinks:
> I like the feature set here. However, I wonder if it might be more natural to run 'in browser' by leveraging a flavor of Rails system tests. That's a more established existing pattern for doing this kind of thing. It would have the advantage of supporting many other testing strategies (like running the system test suites against multiple flavors of headless browser or against Selenium grid etc.) Rails System tests handle Capybara already what is missing is an 'end point' that renders the html etc. but we have that with Previews. I wonder if we can combine those two to get the features you present here.

Hey, @jonspalmer thanks for the feedback! I do think piggy-backing on an established pattern is the way to go. Perhaps I can utilize some ideas from this PR to be able to access a temporary endpoint used to contain the HTML of test cases. I think this would address other issues as well.

I'll see what I can muster up :)

test/view_component/view_component_system_test.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
```
edwinthinks, line 5: @jonspalmer hey there! I tried to go for the system approach. This is still a work in progress but wanted to make sure I'am on the right track on what you proposed in your comment.

edwinthinks:
Update -- here is what it looks like in practice for Rails 5.2 using rspec for a component that displays hidden text on hover.
```rb
  # added this in spec_helper.rb, but 
  config.include ViewComponent::SystemTestHelpers, type: :js_component
```

```rb
RSpec.describe 'Testing JS interactions for View Components', type: :js_component, js: true do
  describe 'Checkbox' do
    subject {
      visit_rendered_in_browser(component, layout: 'application_v2')
    }
    let(:component) { CheckboxComponent.new(label: label, name: 'checkbox[]', hover_text: hover_text, value: 'yes', checked: false) }
    let(:layout) { 'application_2' }
    let(:label) { 'A awesome label' }
    let(:hover_text) { 'Hovered' }

    before do
      subject
    end

    it 'should display the hover text only when hovered over' do
      assert find('label', text: hover_text, visible: false)
      page.find('label', text: label).hover
      assert find('label', text: hover_text, visible: true)
    end
  end
end
```

I think this works pretty well!

boardfish:
Better yet:

```ruby
RSpec.describe CheckboxComponent, type: :js_component, js: true do
  subject { page.find('label', text: 'Hovered', visible: false) }

  let(:component) { described_class.new(label: 'An awesome label', name: 'checkbox[]', hover_text: 'Hovered', value: 'yes', checked: false) }

  before { visit_rendered_in_browser(component, layout: 'application_v2') }

  context 'when not hovered' do
    it { is_expected.not_to be_visible }
  end

  context 'when hovered' do
    before { element.hover }
  
    it { is_expected.to be_visible }
  end
end
```

This should all work, assuming `find` with `visible: false` captures visible elements too.

Hopefully the `type` and `js: true` can go away based on the folder in which the test is stored.

docs/CHANGELOG.md
```diff
@@ -7,10 +7,15 @@ title: Changelog
...
 
+* Add render_in_browser method to enable testing JS interactions in components
+
+    *Edwin Mak*
+
 * Replace antipatterns section in the documentation with best practices.
 
     *Blake Williams*
 
+
```
edwinthinks, line 18:
```suggestion
```

boardfish (commented):
I'm probably not best placed to decide whether or not this gets merged, but I'm really happy with how it looks. 

> Hopefully the type and js: true can go away based on the folder in which the test is stored.

That's my only real feedback - things already work in such a way that `spec/components` contains tests of type `component`. It'd be nice if the type and `js: true` could be implied.

I suppose that'd have to be a different top-level folder. I'm not 100% sure `spec/js_component` would be the right name for it - maybe `interactive_components` or something in that vein?

dmarcoux:
With the changes from this PR, testing accessibility of view components would be possible by using [axe-core-rspec](https://github.com/dequelabs/axe-core-gems/tree/dc71acabc3dd8531b2051b64071643f6c5058254/packages/axe-core-rspec) and [axe-core-capybara](https://github.com/dequelabs/axe-core-gems/blob/dc71acabc3dd8531b2051b64071643f6c5058254/packages/axe-core-capybara/README.md) from [axe-core-gems](https://github.com/dequelabs/axe-core-gems). This would be amazing.

docs/CHANGELOG.md
```diff
@@ -203,6 +209,7 @@ title: Changelog
     *Matthew Rider*
 
 * Fix bug where `with_collection_parameter` didn't inherit from parent component.
+>>>>>>> cb5b2da1be2ad73d7ebd828bfbd016acd09c9f58
```
edwinthinks, line 212:
```suggestion
```

docs/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ title: Changelog
 
 ## main
 
+* Add visit_rendered_in_browser method to enable testing JS interactions in components
+
+    *Edwin Mak*
+
+* Fix bug where `with_collection_parameter` did not inherit from parent component.
+
```
edwinthinks, line 15:
```suggestion
```

app/helpers/preview_helper.rb
```diff
@@ -60,8 +60,6 @@ def prism_language_name_by_template(template:)
   def prism_language_name_by_template_path(template_file_path:)
     language = template_file_path.gsub(".html", "").split(".").last
 
-    return FALLBACK_LANGUAGE unless AVAILABLE_PRISM_LANGUAGES.include? language
-
-    language
+    render "view_components/preview_source"
```
joelhawksley, line 63: Did you mean to make this change?

app/views/view_components/preview.html.erb
```diff
@@ -1,13 +1,11 @@
-<% if @render_args[:component] %>
-  <% if ViewComponent::Base.render_monkey_patch_enabled || Rails.version.to_f >= 6.1 %>
-    <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
-  <% else %>
-    <%= render_component(@render_args[:component], &@render_args[:block]) %>
-  <% end %>
+<% @render_args ||= render_args %>
```
joelhawksley, line 1: What necessitated this change?

app/views/view_components/preview.html.erb
```diff
@@ -1,13 +1,11 @@
...
+<% if ViewComponent::Base.render_monkey_patch_enabled || Rails.version.to_f >= 6.1 %>
+  <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
 <% else %>
   <%= render template: @render_args[:template], locals: @render_args[:locals] || {} %>
 <% end %>
 
-<% if ViewComponent::Base.show_previews_source %>
+<% # Hide the source because it doesn't appear to be significant to system tests %>
```
joelhawksley, line 8: I think we should remove this comment as the `hide_preview_source` param feels more generic than just for this use-case.

```suggestion
```

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add visit_rendered_in_browser method to enable testing JS interactions in components
```
joelhawksley, line 10: How about this naming?

```suggestion
* Add `visit_rendered_component_in_browser` method to enable testing JS interactions in components.
```

lib/view_component/engine.rb
```diff
@@ -129,6 +129,13 @@ class Engine < Rails::Engine # :nodoc:
             internal: true
           )
 
+          get(
+            "view_components_system_test_entrypoint",
+            to: "#{preview_controller}#system_test_entrypoint",
+            as: :preview_view_component_entry_point,
```
joelhawksley, line 135: Why is this `as` defined? I think the default naming is likely more clear, as this isn't for Previews.

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      # Write to temporary file to contain fully rendered component
+      # within a browser
+      file = Tempfile.new(["rendered_#{component.class.name}", ".html"], "tmp/")
+      file.write(html)
+      file.rewind
+
+      # NOTE - not entirely sure how this would work
+      # given that the application may have their own capybara
+      # instance running
```
joelhawksley, line 35: Do you feel this comment is still accurate?

.gitignore
```diff
@@ -12,7 +12,8 @@
 /spec/examples.txt
 /test/tmp/
 /test/version_tmp/
-/tmp/
+/tmp/*
+!tmp/.gitkeep
```
joelhawksley, line 16: Why do we need this file? Couldn't we create/destroy this directory as needed locally?

joelhawksley (commented):
❤️ thanks for keeping this PR going @edwinthinks. I'd love to see some docs added so that we can get a feel for how this fits into the rest of the the framework. 

I'd also like to see @jonspalmer chime in if he's around 😄

app/helpers/preview_helper.rb
edwinthinks: Nope this was a merge conflict issue. Woops!

.gitignore
edwinthinks: Good point. I don't think this is necessary since the `visit_rendered_component_in_browser` method is now using `Tempfile` which would erase the temporary file.

lib/view_component/system_test_helpers.rb
edwinthinks: Hmm, it shouldn't be, since I believe the system test suite should handle this and my original concern with the comment isn't relevant. I'll remove it

lib/view_component/engine.rb
edwinthinks: Makes sense!

docs/guide/testing.md
```diff
@@ -199,3 +199,44 @@ To use component previews:
 # config/application.rb
 config.view_component.preview_paths << "#{Rails.root}/spec/components/previews"
 ```
+
+## Testing Interactive Components
```
edwinthinks, line 203: @joelhawksley added this documentation. Let me know if this is unclear. Thank you!

app/views/view_components/preview.html.erb
edwinthinks: This was added because `render_args` the method needed to be called to fetch the render_args. Without this, the system test helpers would not be able to determine what the `@render_args` should be when generating the temporary template to test against.

docs/guide/testing.md
```diff
@@ -199,3 +199,42 @@ To use component previews:
...
+
+    assert find("[data-hidden-field]", visible: false)
+    find("[data-button]", text: "Click Me To Reveal Something Cool").click
+    assert find("[data-hidden-field]", visible: true)
+  end
+end
+```
+
+### Handling components without JS included
+A common pattern is to include assets in a layout rather than directly inside the component. In those cases, you can define a `layout` in the `visit_rendered_component_in_browser` so that the component is rendered with those assets:
```
boardfish, line 224: Would it be possible to add a configuration option for the default layout to use?

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def page
+      Capybara.current_session
+    end
+
+    def visit_rendered_component_in_browser(component, **options)
+      layout = options[:layout] || false
+
+      opts = {
+        layout: layout,
```
boardfish, line 16: Would it be possible to add a configuration option for the default layout to use? It might even be better to allow defining the [controller to use](https://viewcomponent.org/guide/testing.html#configuring-the-controller-used-in-tests) instead so that the layout can be inherited from that.

I made the same comment elsewhere - if you intend on doing so, you may need to update that documentation too.

boardfish (approved):

lib/view_component/system_test_helpers.rb
edwinthinks: Hmm, I'am not sure tbh. I see its utility of adding a default but haven't yet hit that use case yet. What do you think about holding off on that change until we have a use-case and issue that requests that?

camertron:
Hey @edwinthinks! Is this PR ready to go?

app/controllers/view_components_controller.rb
```diff
@@ -37,6 +37,10 @@ def previews
     end
   end
 
+  def system_test_entrypoint
+    render file: "./tmp/#{params[:file]}", status: 200
```
BlakeWilliams, line 41:
```suggestion
    return head :not_authorized if Rails.env.production?
    render file: "./tmp/#{params[:file]}", status: 200
```

Can we guard this in production just in case this is accidentally exposed? It may also be worth trying to sanitize `params[:file]` to prevent accidental file system traversal.

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+        }
+      }
+      html = controller.render_to_string("view_components/preview", opts)
+
+      # Add 'tmp/' directory if it doesn't exist to store the rendered component html
+      Dir.mkdir("tmp/") unless Dir.exist?("tmp/")
+
+      # Write to temporary file to contain fully rendered component
+      # within a browser
+      file = Tempfile.new(["rendered_#{component.class.name}", ".html"], "tmp/")
```
BlakeWilliams, line 32: Thinking about alternatives, is it possible to use some kind of static asset middleware to serve HTML files that we write to a specific directory? e.g. `./tmp/view_component_integrations`

I think that would help us mitigate some security issues and maybe allow us to remove some complexity.

edwinthinks:
Apologies! I've been unable to make any changes recently. If anyone wants to take over this, feel free :)

app/controllers/view_components_controller.rb
edwinthinks: Great point! Yes, I think both suggestions are ideal to avoid any unintentional security holes

P.S - I'll try to put some effort into getting this fixed up and ready for review again :)

lib/view_component/system_test_helpers.rb
edwinthinks: I'am not entirely sure about the asset middleware but I believe we can use a specific directory :).

I'll make adjustments to use a specific directory. But could you elaborate a bit more on the static assets middleware? I'am afraid I'am not too versed in making middleware.

edwinthinks:
@Lukeghenco 

> Just a general question here, but will we have the ability to mock/stub API calls that a dynamic ViewComponent might make during an interaction with the systems in place here?

I believe it is possible as long as you are able to mock it in other contexts like a normal system test. The method added here just makes it so you can jump to a temp page with the component rendered. I don't see why you wouldn't be able to mock :)

> I'm also wondering how one might test forms in this flow. From the DSL I see in this PR, would we have access to other Capybara helpers like this? (testing a potential codespace creation flow in this example)

You would use the same approach as any system test. As mentioned above, this just puts the component onto a temporary webpage that your test suite can interact with.

edwinthinks:
@BlakeWilliams I think I addressed your comments. Please let me know if anything is missing or if I misunderstood something.

This is gonna be great :)

joelhawksley:
@edwinthinks sorry for not getting back to you sooner! I've been giving this PR a lot of thought and I _think_ there might be a better direction to take this feature that would be more cohesive with existing Rails and ViewComponent mental models. 

What I think we should do is move towards having a `visit_preview` method much like the `render_preview` method added in https://github.com/github/view_component/pull/1347. By reusing the preview infrastructure, it will be easier to debug these integration tests without having to boot up a test case and pause its execution.

More generally, I think that the more we can center the workflow of building ViewComponents around previews, the better. With tools like Lookbook taking the usefulness of previews to new levels, I think we have an opportunity here to build something really compelling! 

What do you think?

edwinthinks:
Hey @joelhawksley, no worries at all.

To make sure I understand, this option with `visit_preview` would essentially have specs/system tests jump to the existing preview instead of rendering when the test is made?

Timmitry:
@joelhawksley I'll try to make a point against that suggestion 😉 

It certainly would be nice to have the possibility to JS-test your preview components. Especially for presentational components (= generic components that receive simple arguments, like Dropdowns, Tablists, Buttons, ...), that might be enough.

However, some components are a lot more complex, and only work with persisted active record objects. They can therefore unfortunately not be integrated into a preview (as the preview would then need some kind of database setup and teardown).

The current solution allows me to render my component with any kind of setup in the test environment, e.g. I could do some `FactoryBot.create(:article)` beforehand, and then do a `ArticlePageComponent.new(article: article)`, and then give this to `visit_rendered_component_in_browser`. That is a lot of flexibility that I would not like to miss out on 🙂 

What do you think?

joelhawksley:
@Timmitry oooh, I hadn't thought of that argument specifically, but we _have_ looked into using FactoryBot in previews, but using the `build` strategy that does not touch the database. Perhaps we need to experiment there a bit further before we lean to heavily on previews. 

Let me think on this 🤔

edwinthinks:
Hey there, I was wondering if there was any update on this PR and if I should make any adjustments?

joelhawksley:
Hi ed! Sorry I’ve taken so long to get back to you. I’ve had other things on my plate and wanted to wait to give this PR my full attention. I want us to have this feature in V3, a milestone we should be hitting soon.

I hope to have that attention soon. Again I'm sorry to drag this one on, but it's from a place of wanting to get it right ❤️

edwinthinks:
@joelhawksley no worries at all! Let me know if there is anything I can update :). This is exciting!

joelhawksley:
@edwinthinks it looks like these browser tests are flaky. Would you be up for looking into that? I'll dig into a review of the code in the meantime.

app/controllers/view_components_controller.rb
```diff
@@ -4,4 +4,18 @@
...
 class ViewComponentsController < Rails::ApplicationController # :nodoc:
   include ViewComponent::PreviewActions
+
+  def system_test_entrypoint
+    if Rails.env.production?
+      render body: "Unauthorized", status: 403
+    else
+      render file: "./tmp/view_component_integrations/#{file_param[:file]}", status: 200
+    end
+  end
```
joelhawksley, line 14: I think we can be even more strict here, only turning on this route in test:

```suggestion
  if Rails.env.test?
    def system_test_entrypoint
      render file: "./tmp/view_component_integrations/#{file_param[:file]}", status: 200
    end
  end
```

app/views/view_components/preview.html.erb
```diff
@@ -1,3 +1,5 @@
+<% @render_args ||= render_args %>
```
joelhawksley, line 1: Why do we need this memoization? Why not reference `render_args` directly on 13?

Gemfile
```diff
@@ -10,6 +10,11 @@ gem "rails", rails_version == "main" ? {git: "https://github.com/rails/rails", r
 
 gem "rspec-rails", "~> 5"
 
+# Utilize to test interactions
+gem "cuprite", "~> 0.13"
+gem "puma", "~> 5.4"
+gem "selenium-webdriver"
+
```
joelhawksley, line 17: I'm wondering if there's a way we can avoid adding these as gem dependencies. Folks may want to use other browser drivers, for example. (We can't use cuprite due to incompatibilities in `axe-core`, for example)

```suggestion
```

joelhawksley: (we could still add them as dev dependencies for the sake of testing this feature in CI)

lib/view_component/engine.rb
```diff
@@ -123,6 +123,12 @@ class Engine < Rails::Engine # :nodoc:
             internal: true
           )
 
+          get(
+            "view_components_system_test_entrypoint",
+            to: "#{preview_controller}#system_test_entrypoint",
+            internal: true
+          )
+
```
joelhawksley, line 131: Can we conditionally add this route only in `test`?

Gemfile
joelhawksley: My thinking is that we'd have something like:

`visit(rendered_component_path(MyComponent.new, layout: "special"))`

joelhawksley: My thinking is that we'd have something like:

`visit(rendered_component_path(MyComponent.new, layout: "special"))`, that way we'd let folks handle `visit` however they please, limiting ViewComponent's responsibility to providing the route to visit.

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+        }
+      }
+      html = controller.render_to_string("view_components/preview", opts)
+
+      # Add './tmp/view_component_integrations/' directory if it doesn't exist to store the rendered component html
+      FileUtils.mkdir_p("./tmp/view_component_integrations/") unless Dir.exist?("./tmp/view_component_integrations/")
+
+      # Write to temporary file to contain fully rendered component
+      # within a browser
+      file = Tempfile.new(["rendered_#{component.class.name}", ".html"], "tmp/view_component_integrations")
```
joelhawksley, line 32: If we render a component twice with different arguments, wouldn't one clobber the other here?

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+      file.write(html)
+      file.rewind
+
+      filename = file.path.split("/").last
+
+      # Visit the file that contains the HTML
+      visit "/view_components_system_test_entrypoint?file=#{filename}"
+
+      # Erase temporary file
+      file.unlink
```
joelhawksley, line 42: One downside to moving way from handling `visit` is that it would be trickier to handle erasing temporary files. Perhaps we could yield the entrypoint instead? Allowing for something like this:

```ruby
with_rendered_component_path(MyComponent.new, layout: "special") do |path|
  visit(path)
end
```

app/controllers/view_components_controller.rb
edwinthinks: Good idea!

Gemfile
edwinthinks: I think the best we could do from just a quick glance is to put them as dev dependencies. You are right - people may have different preferences.

app/views/view_components/preview.html.erb
edwinthinks: Hehe, I don't actually remember. I think I had a good reason but I'll try without it and see if I can remember.

lib/view_component/system_test_helpers.rb
edwinthinks: @joelhawksley I believe it wouldn't because the output of `Tempfile` creates a unique string at the end. For example:
```
[1] pry(main)> Tempfile.new(["hello.html"])
=> #<File:/var/folders/ll/np7j8vfd05q58297qqhhbs1w0000gp/T/hello.html20220915-66507-tjp6pm>
[2] pry(main)> Tempfile.new(["hello.html"])
=> #<File:/var/folders/ll/np7j8vfd05q58297qqhhbs1w0000gp/T/hello.html20220915-66507-9gc2ki>
[3] pry(main)> Tempfile.new(["hello.html"])
=> #<File:/var/folders/ll/np7j8vfd05q58297qqhhbs1w0000gp/T/hello.html20220915-66507-128ga2h>
[4] pry(main)> Tempfile.new(["hello.html"])
=> #<File:/var/folders/ll/np7j8vfd05q58297qqhhbs1w0000gp/T/hello.html20220915-66507-19cjlew>
[5] pry(main)> Tempfile.new(["hello.html"])
=> #<File:/var/folders/ll/np7j8vfd05q58297qqhhbs1w0000gp/T/hello.html20220915-66507-gd95yb>
[6] pry(main)> Tempfile.new(["hello.html"])
=> #<File:/var/folders/ll/np7j8vfd05q58297qqhhbs1w0000gp/T/hello.html20220915-66507-17xubsr>
```

joelhawksley: Oh, that's great 👍🏻

lib/view_component/system_test_helpers.rb
edwinthinks: I like this idea and the `file.unlink` could be in the ensure block so we permanently delete it after it's been used. Let me try that out.

test/sandbox/test/view_component_system_test.rb
```diff
@@ -5,39 +5,22 @@
 class ViewComponentSystemTest < ViewComponent::SystemTestCase
   driven_by :cuprite
 
-  def test_protection_unauthorized_access_in_production_env
```
edwinthinks: Removed this since the route isn't drawn outside of Rails.env.test

edwinthinks:
@joelhawksley hey there - I've made the updates but I'am running into this seemingly flakiness with system tests:
```
ViewComponentSystemTest#test_simple_js_interaction_in_browser_with_layout:
Ferrum::TimeoutError: Timed out waiting for response. It's possible that this happened because something took a very long time (for example a page load was slow). If so, setting the :timeout option to a higher value might help.
```
Do you have any suggestions to fix this issue? I'am not sure if making the timeout longer ideal.

.gitignore
```diff
@@ -12,7 +12,6 @@
 /spec/examples.txt
 /test/tmp/
 /test/version_tmp/
-/tmp/
```
joelhawksley, line 15: Do we need this change?

Gemfile
```diff
@@ -10,6 +10,13 @@ gem "rails", rails_version == "main" ? {git: "https://github.com/rails/rails", r
 
 gem "rspec-rails", "~> 5"
 
+group :test do
+  # Utilize to test UI interactions
```
joelhawksley, line 14:
```suggestion
  # For testing UI interactions
```

Gemfile
```diff
@@ -10,6 +10,13 @@ gem "rails", rails_version == "main" ? {git: "https://github.com/rails/rails", r
...
+  gem "cuprite", "~> 0.13"
+  gem "puma", "~> 5.4"
+  gem "selenium-webdriver"
```
joelhawksley, line 17:
```suggestion
  gem "selenium-webdriver", "~> 4.1"
```

app/controllers/view_components_controller.rb
```diff
@@ -4,4 +4,16 @@
...
+    def system_test_entrypoint
+      render file: "./tmp/view_component_integrations/#{file_param[:file]}", status: 200
+    end
+  end
+
+  private
+
+  def file_param
+    params.permit(:file)
+  end
```
joelhawksley, line 18:
```suggestion
  if Rails.env.test?
    def system_test_entrypoint
      render file: "./tmp/view_component_integrations/#{params.permit(:file)[:file]}", status: 200
    end
  end
```

app/views/view_components/preview.html.erb
joelhawksley: Still needs resolved ❤️

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add visit_rendered_component_in_browser method to enable testing JS interactions in components
```
joelhawksley, line 13: I'll write this here and assume you'll handle the other changes, but I think `with_rendered_component_path` is a good name for this new API ❤️ 

```suggestion
* Add `with_rendered_component_path` method to enable testing components in browser tests.
```

docs/guide/testing.md
```diff
@@ -220,3 +220,43 @@ To use component previews:
 # config/application.rb
 config.view_component.preview_paths << "#{Rails.root}/spec/components/previews"
 ```
+
+## Testing Interactive Components
+
+Use the `visit_rendered_component_in_browser` helper method in your system tests to test interactivity of components.
```
joelhawksley, line 226: Just a note to update the docs in general ❤️

lib/view_component/engine.rb
```diff
@@ -122,6 +122,16 @@ class Engine < Rails::Engine # :nodoc:
             internal: true
           )
 
+          if Rails.env.test?
+            # This route is used by the system test to render the preview
+            # and allow for assertions against interactivity.
```
joelhawksley, line 127:
```suggestion
```

lib/view_component/engine.rb
```diff
@@ -122,6 +122,16 @@ class Engine < Rails::Engine # :nodoc:
...
+            get(
+              "view_components_system_test_entrypoint",
```
joelhawksley, line 129:
```suggestion
              "system_test_entrypoint",
```

lib/view_component/engine.rb
```diff
@@ -122,6 +122,16 @@ class Engine < Rails::Engine # :nodoc:
...
+              to: "#{preview_controller}#system_test_entrypoint",
+              internal: true
```
joelhawksley, line 131:
```suggestion
              to: "#{preview_controller}#system_test_entrypoint"
```

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module SystemTestHelpers
+    include TestHelpers
+    include Capybara::DSL
+
+    def page
```
joelhawksley, line 8: What do we need this for? I could see it conflicting with consumers' tests.

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def with_rendered_component_in_browser(component, **options, &block)
+      layout = options[:layout] || false
+
+      opts = {
+        layout: layout,
+        locals: {
+          render_args: {
+            component: component,
+            template: "view_components/preview",
```
joelhawksley, line 20: Do we need this? We're already referencing the template on line 25.

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+            hide_preview_source: true
+          }
+        }
+      }
+      html = controller.render_to_string("view_components/preview", opts)
+
+      # Add './tmp/view_component_integrations/' directory if it doesn't exist to store the rendered component html
+      FileUtils.mkdir_p("./tmp/view_component_integrations/") unless Dir.exist?("./tmp/view_component_integrations/")
```
joelhawksley, line 28: I'm curious, why create this folder? Why not just write a `Tempfile` and not worry about specifying the path?

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+      file = Tempfile.new(["rendered_#{component.class.name}", ".html"], "tmp/view_component_integrations")
+      file.write(html)
+      file.rewind
+
+      filename = file.path.split("/").last
+      path = "/view_components_system_test_entrypoint?file=#{filename}"
+
+      yield path
+    ensure
+      # Erase temporary file
```
joelhawksley, line 41:
```suggestion
```

joelhawksley (commented):
👋🏻 thanks for your work on ❤️ 

I'm expecting to done one more round of feedback once these items are addressed.

.gitignore
edwinthinks: Nope. I think this was leftover. It important to ignore any tmp files we might generate but shouldn't be the case since we `.unlink` or delete any tmp files generated.

lib/view_component/system_test_helpers.rb
edwinthinks: It's been a while, but I recalled this was suggested earlier so that `assert` or any other capybara commands operate on `page`. Its a bit of syntactic sugar :) so you don't have to keep specifying `page` and can do something like:
```
assert find("[data-hidden-field]", visible: false) 
```
Without having to specify the `page`

joelhawksley: Interesting. I guess what I'm wondering now that I re-read this is whether we need a `capybara` dependency at all. I think we should try to avoid one. I think we can just have this feature provide a temporary URL to visit in whatever system test stack folks are using.

app/views/view_components/preview.html.erb
edwinthinks: I remember why now @joelhawksley; The `@render_args` actually isn't defined when using the `render_to_string` method in lib/view_component/system_test_helpers.rb. It seems like the method is available but no the `@` variable version. Let me know if there is a better way :)

lib/view_component/system_test_helpers.rb
edwinthinks: Hey @joelhawksley; ideally, we want to avoid the dependency on general support. However, I feel like capybara is pretty standard, and I'am wondering if it makes sense to release with capybara support and then introduce wider support?

lib/view_component/system_test_helpers.rb
edwinthinks: Yeh it doesn't really matter anymore since the files will go away anyways. `/tmp` by itself should be sufficient :)

lib/view_component/system_test_helpers.rb
edwinthinks: Yep not necessary, it looks like this is specifically used when component is not provided. I'll remove.

edwinthinks:
Hey @joelhawksley I think I've addressed most of the comments. I'am still having a weird issue in which would seem to introduce a flaky spec (yikes!). Like some help figuring this out or find a better way. Here is that failure - https://github.com/ViewComponent/view_component/actions/runs/3169430375/jobs/5161337359

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+        }
+      }
+      html = controller.render_to_string("view_components/preview", opts)
+
+      # Add './tmp/view_components/' directory if it doesn't exist to store the rendered component html
+      FileUtils.mkdir_p("./tmp/view_components/") unless Dir.exist?("./tmp/view_components/")
+
+      # Write to temporary file to contain fully rendered component
+      # within a browser
+      file = Tempfile.new(["rendered_#{component.class.name}", ".html"], "tmp/view_components/")
```
edwinthinks, line 31: I don't fully understand this - but it seems like outputting Tempfile directly to `tmp/` without another directory starts to cause test failures. It appeared that the file wasn't there when the spec ran which seems odd to me.

However - I do think having a dedicated directory does make sense imo as many other gems can create files in that `tmp` directory.

lib/view_component/system_test_helpers.rb
joelhawksley: @edwinthinks I think we included capybara at one point but that some folks had issues with it. I think that's why we have this section in the docs: https://viewcomponent.org/guide/testing.html#without-capybara. I think I'd prefer to not included `capybara` in the package. We can have it as a dev dependency in our test suite, though 👍🏻

joelhawksley:
> Hey @joelhawksley I think I've addressed most of the comments. I'am still having a weird issue in which would seem to introduce a flaky spec (yikes!). Like some help figuring this out or find a better way. Here is that failure - https://github.com/ViewComponent/view_component/actions/runs/3169430375/jobs/5161337359

I'd be OK with trying the advice of the error message to increase `:process_timeout`.

Gemfile
```diff
@@ -10,6 +10,13 @@ gem "rails", rails_version == "main" ? {git: "https://github.com/rails/rails", r
 
 gem "rspec-rails", "~> 5"
 
+group :test do
+  # For testing UI interactions
```
joelhawksley, line 14:
```suggestion
```

Gemfile
```diff
@@ -10,6 +10,13 @@ gem "rails", rails_version == "main" ? {git: "https://github.com/rails/rails", r
...
+  gem "cuprite"
+  gem "puma"
+  gem "selenium-webdriver"
```
joelhawksley, line 17: Can you add coarse dependency pointers for these? Just to the major version is fine: `~> 1` etc.

app/controllers/view_components_controller.rb
```diff
@@ -4,4 +4,10 @@
 
 class ViewComponentsController < Rails::ApplicationController # :nodoc:
   include ViewComponent::PreviewActions
```
joelhawksley, line 6: Looking at this now, I think there was a reason we moved all of the actions to a module. Mind putting this new action in `PreviewActions`?

app/controllers/view_components_controller.rb
```diff
@@ -4,4 +4,10 @@
...
+
+  if Rails.env.test?
+    def system_test_entrypoint
+      render file: "./tmp/view_components/#{params.permit(:file)[:file]}", status: 200
```
joelhawksley, line 10: It's a nitpick, but isn't `status: 200` the default?

app/views/view_components/preview.html.erb
joelhawksley: This is fine for now, resolving 👍🏻

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add with_rendered_component_in_browser method to enable testing JS interactions in components
```
```suggestion
* Add `with_rendered_component_path` method to enable testing JS interactions in components
```

I thought I made this suggestion already, but maybe I didn't 😅 I think we should make it clear that we are not providing a browser.

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
...
+Use the `with_rendered_component_in_browser` helper method in your system tests to test interactivity of components.
+
+```rb
+require "test_helper"
+
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
+  driven_by :cuprite
+
+  def test_simple_js_interaction_in_browser_without_layout
+    with_rendered_component_in_browser(SimpleJavascriptInteractionWithJsIncludedComponent.new) do |page|
```
joelhawksley, line 248: So this would be the API:

```suggestion
    with_rendered_component_path(SimpleJavascriptInteractionWithJsIncludedComponent.new) do |path|
```

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
 # config/application.rb
 config.view_component.preview_paths << "#{Rails.root}/spec/components/previews"
 ```
+
+## Testing Interactive Components
+
+Use the `with_rendered_component_in_browser` helper method in your system tests to test interactivity of components.
```
joelhawksley, line 239:
```suggestion
Use the `with_rendered_component_path` helper method in system tests:
```

app/controllers/view_components_controller.rb
edwinthinks: Yup! I think I can remove it since it isn't really providing any value.

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add with_rendered_component_path method to enable testing JS interactions in components
```
```suggestion
* Add with_rendered_component_path method to enable testing JS interactions in components.
```

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
 # config/application.rb
 config.view_component.preview_paths << "#{Rails.root}/spec/components/previews"
 ```
+
+## Testing Interactive Components
+
+Use the `with_rendered_component_path` helper method in your system tests to test interactivity of components.
```
joelhawksley, line 239:
```suggestion
Use `with_rendered_component_path` in system tests to test interactive components:
```

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
...
+
+```rb
+require "test_helper"
+
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
+  driven_by :cuprite
+
+  def test_simple_js_interaction_in_browser_without_layout
+    with_rendered_component_path(SimpleJavascriptInteractionWithJsIncludedComponent.new) do |page|
+      visit page
```
joelhawksley, line 249: `page` is a reserved word in most system tests:

```suggestion
    with_rendered_component_path(SimpleJavascriptInteractionWithJsIncludedComponent.new) do |path|
      visit(path)
```

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
...
+
+      assert find("[data-hidden-field]", visible: false)
+      find("[data-button]", text: "Click Me To Reveal Something Cool").click
+      assert find("[data-hidden-field]", visible: true)
+    end
+  end
+end
+```
+
+### Handling components without JS included
```
joelhawksley, line 259:
```suggestion
### Testing components with external JavaScript
```

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
...
+
+A common pattern is to include assets in a layout rather than directly inside the component. In those cases, you can define a `layout` in the `visit_rendered_component_in_browser` so that the component is rendered with those assets:
```
joelhawksley, line 261:
```suggestion
For components that depend on JavaScript provided by a layout, provide the `layout` argument:
```

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
...
+
+```rb
+require "test_helper"
+
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
+  driven_by :cuprite
+
+  def test_simple_js_interaction_in_browser_with_layout
+    with_rendered_component_path(SimpleJavascriptInteractionWithoutJsIncludedComponent.new, layout: 'application') do |page|
+      visit page
```
joelhawksley, line 271:
```suggestion
    with_rendered_component_path(SimpleJavascriptInteractionWithoutJsIncludedComponent.new, layout: 'application') do |path|
      visit(path)
```

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
...
+
+      assert find("[data-hidden-field]", visible: false)
+      find("[data-button]", text: "Click Me To Reveal Something Cool").click
+      assert find("[data-hidden-field]", visible: true)
```
joelhawksley, line 275:
```suggestion
      assert(find("[data-hidden-field]", visible: false))
      find("[data-button]", text: "Click Me To Reveal Something Cool").click
      assert(find("[data-hidden-field]", visible: true))
```

docs/guide/testing.md
```diff
@@ -233,3 +233,47 @@ To use component previews:
...
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
+  driven_by :cuprite
+
+  def test_simple_js_interaction_in_browser_without_layout
+    with_rendered_component_path(SimpleJavascriptInteractionWithJsIncludedComponent.new) do |page|
+      visit page
+
+      assert find("[data-hidden-field]", visible: false)
+      find("[data-button]", text: "Click Me To Reveal Something Cool").click
+      assert find("[data-hidden-field]", visible: true)
```
joelhawksley, line 253:
```suggestion
      assert(find("[data-hidden-field]", visible: false))
      find("[data-button]", text: "Click Me To Reveal Something Cool").click
      assert(find("[data-hidden-field]", visible: true))
```

lib/view_component/engine.rb
```diff
@@ -121,11 +121,17 @@ class Engine < Rails::Engine # :nodoc:
             internal: true
           )
 
+          if Rails.env.test?
+            get(
+              "system_test_entrypoint",
+              to: "#{preview_controller}#system_test_entrypoint"
+            )
```
joelhawksley, line 128:
```suggestion
            get("system_test_entrypoint", to: "#{preview_controller}#system_test_entrypoint")
```

joelhawksley (commented):
More polish, getting close! ❤️

docs/guide/testing.md
edwinthinks: Good point!

lib/view_component/engine.rb
```diff
@@ -121,11 +121,17 @@ class Engine < Rails::Engine # :nodoc:
...
+          end
+
           get(
             "#{options.preview_route}/*path",
             to: "#{preview_controller}#previews",
-            as: :preview_view_component,
-            internal: true
```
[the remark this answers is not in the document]
edwinthinks: Oph - I think this may have been added by accident in a merge conflict. Let me undo this change - I feel like it isn't a part of this work :)

lib/view_component/system_test_helpers.rb
edwinthinks: @joelhawksley picking up on this. I agree on the assessment on not including capybara as a dependency. I'am working on making the dependency only on our testing and not on the actual gem itself.

edwinthinks:
@joelhawksley thanks for the feedback! I believe I've addressed all of them. One thing I modified is to remove Capybara as a requirement - [d29948b](https://github.com/ViewComponent/view_component/pull/1061/commits/d29948bb3f373ac49f1014e1bc11cf85acd9985b). Let me know what you think

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add with_rendered_component_path method to enable testing JS interactions in components.
```
```suggestion
* Add `with_rendered_component_path` method to enable testing JS interactions in components.
```

test/sandbox/test/view_component_system_test.rb
```diff
@@ -0,0 +1,27 @@
...
+      visit page
+
+      assert find("[data-hidden-field]", visible: false)
+      find("[data-button]", text: "Click Me To Reveal Something Cool").click
+      assert find("[data-hidden-field]", visible: true)
+    end
+  end
+
+  def test_simple_js_interaction_in_browser_with_layout
+    with_rendered_component_path(SimpleJavascriptInteractionWithJsIncludedComponent.new, layout: "application") do |page|
```
joelhawksley, line 19:
```suggestion
    with_rendered_component_path(SimpleJavascriptInteractionWithoutJsIncludedComponent.new, layout: "application") do |page|
```

docs/guide/testing.md
```diff
@@ -233,3 +233,43 @@ To use component previews:
...
 ```
+
+## Testing Interactive Components
+
+Use `with_rendered_component_path` in system tests to test interactive components:
+
+```rb
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
+  driven_by :cuprite
+
```
joelhawksley, line 244:
```suggestion
```

docs/guide/testing.md
```diff
@@ -233,3 +233,43 @@ To use component previews:
...
+```
+
+### Testing components with external JavaScript
+
+For components that depend on JavaScript provided by a layout, provide the `layout` argument:
+
+```rb
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
+  driven_by :cuprite
+
```
joelhawksley, line 264:
```suggestion
```

joelhawksley (approved):

test/sandbox/test/view_component_system_test.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
+  driven_by :cuprite
+
+  def test_simple_js_interaction_in_browser_without_layout
+    with_rendered_component_path(SimpleJavascriptInteractionWithJsIncludedComponent.new) do |page|
```
joelhawksley, line 9: Can you add a test that uses slots as well as a test that passes arguments to the component?

edwinthinks: @joelhawksley I added a basic test for the passing arguments. But I can't seem to figure out how to create a component with slots that I can pass into the helper. I see in test/sandbox/test/slotable_v2_test.rb that `render_inline` accepts and provides a block where slots can be added but not sure exactly how to handle it without using `render_inline`

Wondering if you can point me in the right direction here... or if we need to improve the helper to handle this use case?

joelhawksley: The added test looks good! I do think we need one to show that slots work (or that they don't, and that they should).

Something like:

```ruby
c = SlotsV2Component.new(classes: "mt-4") do |component|
  component.title do
    "This is my title!"
  end
with_rendered_component_path(c) do |path|
  visit path
  assert_selector(".title", text: "This is my title!")
end
```

edwinthinks: @joelhawksley hey there, I tried that out and it doesn't seem to be working. Here is what I tried;
```
  def test_components_with_slots
    c = SlotsV2Component.new(classes: "mt-4") do |component|
      component.title do
        "This is my title!"
      end
    end

    with_rendered_component_path(c) do |path|
      visit path
      assert_selector(".title", text: "This is my title!")
    end
  end
```

And the snapshot of the failed spec looks like it isn't picking up the slots.
Captura de Pantalla 2022-11-03 a la(s) 4 58 39 p m

It looks like the slots aren't being rendered

joelhawksley: Damn. @BlakeWilliams might you be able to have a look here?

BlakeWilliams: Ah, it looks like the example code has a mistake that's relatively common:

```ruby
c = SlotsV2Component.new(classes: "mt-4") do |component|
  component.title do
    "This is my title!"
  end
end

# needs to be 
render_inline(SlotsV2Component.new(classes: "mt-4")) do |component|
  component.title do
    "This is my title!"
  end
end
```

You can't pass blocks to `.new` because components control the initializer, and we rely on the block being passed to `render_inline` instead (same goes for `render` in views). Unfortunately you'd need `with_rendered_component_path` to use the block being passed to it to set slot information via blocks. I think we might need to rethink the API here.  Maybe something like:

```ruby
render_inline_and_visit(SlotsV2Component.new(classes: "mt-4")) do |component|
  component.title do
    "This is my title!"
  end
 end
```

edwinthinks: @BlakeWilliams thanks for taking a look at this. In the example you provided, as a new way to think of the API would the `render_inline_and_visit` also return a block with `path` like in the current API in this PR?

BlakeWilliams: I don't think so, since it's really "one way or the other" when it comes to being able to set slots or providing path in the block. You *could* probably combine the two in the same block but I think that might be mixing concerns and make the intention of the tests less clear so I'd likely avoid that approach.

edwinthinks: I think it might make sense to break away from the block concept since slots already have a use of a block. So perhaps, maybe something like:
```
path = render_inline_and_visit(SlotsV2Component.new(classes: "mt-4")) do |component|
  component.title do
    "This is my title!"
  end
end

visit path
# do your tests stuff with `path`
```

I think it makes sense perhaps to use this approach for both slots and non slot components so there isn't too many variations for how this helper is used with or without slots.

Thoughts @joelhawksley @BlakeWilliams ? Appreciate the feedback!

BlakeWilliams: I think that makes sense.

I am curious why we require consumers to call `visit` explicitly. Is there a reason for requiring consumers of the API to call `visit` explicitly instead of us calling it for them?

edwinthinks: > I think that makes sense.

@joelhawksley thoughts? Wanted to get your opinion on this since you mentioned the block pattern. Wondering if the approach I proposed of still works?

> I am curious why we require consumers to call visit explicitly. Is there a reason for requiring consumers of the API to call visit explicitly instead of us calling it for them?

I think was originally suggested so that we don't forcibly require someone to have capybara. From what I understand explicitly calling `visit` affords people who don't use capybara to access the component to be tested using other methods. Does that make sense?

BlakeWilliams: > I think was originally suggested so that we don't forcibly require someone to have capybara. From what I understand explicitly calling `visit` affords people who don't use capybara to access the component to be tested using other methods. Does that make sense?

I'd like to see us enumerate on that a little bit. What are some realistic scenarios where this would happen?

I imagine we could have our own "ViewComponent.interactions_test_driver` or something similar that could be used to hook into the `visit` part of `render_inline_and_visit`. If we went down that route we'd the flexibility we want (assuming we do want it) and the public API would stay lean/clear.

joelhawksley: > I am curious why we require consumers to call visit explicitly. Is there a reason for requiring consumers of the API to call visit explicitly instead of us calling it for them?

I guided @edwinthinks in this direction because I wanted to avoid a dependency on `visit` (and thus `Capybara`) in ViewComponent. We've previously gotten feedback and have accommodated folks that do not use `capybara`.

I don't mind having the method return a path. But then how would we clean up the temp files? The block approach allows us to use `ensure` to remove them.

joelhawksley: @BlakeWilliams your response got me thinking - What if we set `path` in an instance variable on the test instead of yielding it? Then we could access it in the test via an attr_accessor...

edwinthinks: Hmm to address this original problem with not being able to introduce slots into the current implementation. Is it possible to use `render_inline` and pass that directly to the `with_rendered_component_path`? Right now `with_rendered_component_path` is using `render_to_string` to take the HTML output and place it on a temp file.  That is, the `with_rendered_component_path` takes a component or HTML and puts it in an accessible tmp route

> I think that's fair but I'm curious how the API ergonomics differ when we focus on flexibility vs usability. Personally I feel like we should start at usability then work backwards towards flexibility. e.g. In this case maybe we could implement some kind of adapter that's hot-swappable between capybara and other solutions.

I generally agree with the usability - however I'am not sure how different it is using `visit path`. The only thing I find that could cause friction is the use of a `block` like we have now. This makes me think either yielding or the `attr_accesor` @joelhawksley you mentioned makes sense.

I auger to do a SPIKE to see how this looks without `block` syntax and support slots by refactoring the `with_rendered_component_path` to accept the output of render_inline to allow for slots.

BlakeWilliams: > @BlakeWilliams your response got me thinking - What if we set `path` in an instance variable on the test instead of yielding it? Then we could access it in the test via an attr_accessor...

I like that idea, but the only downside is that we miss having a `block` that can ensure the tmp file is deleted.

edwinthinks: > I like that idea, but the only downside is that we miss having a block that can ensure the tmp file is deleted.

From what I understand temp files automatically get deleted when garbage is collected, but isn't for sure to happen which can lead to leaving behind excess past temp files. Wondering though if the chances of this happening is often enough that it should influence the implementation. 

Ideally we do ensure temp files are deleted, but just wondering and thinking "outside" the box here :)

joelhawksley: Thanks for thinking this through @edwinthinks. Given all of these compromises, I'm 👍🏻 to having `capybara` as a dependency for this feature. All I ask is that was conditionally include this helper if `capybara` is present (or perhaps that a `visit` method is defined?) and document it as such.

edwinthinks: No problemo :). To verify that we are aligned are we saying we are going to breakaway from the block constant so we can avoid having to call "visit"? Moreover, I think there is still some work to be done to allow this to work with slots.

Sorry been a bit out of the loop! Might be worth catching up on a synchronous call to catch up on the direction for this PR @joelhawksley ?

joelhawksley: Sure thing, email me at joelhawksley@github.com with your availability for next week?

edwinthinks: Hey there Joel, sorry for the late reply. I'll email you with some times, not likely I can do this week or the next. Its holiday season :)

test/sandbox/test/view_component_system_test.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
```
edwinthinks, line 5: @joelhawksley let me know what you think about this API change based on our conversation. Once this is looking good, I'll update the corresponding README.

Gemfile.lock
```diff
@@ -268,26 +284,29 @@ GEM
       webrick (~> 1.7.0)
     yard-activesupport-concern (0.0.1)
       yard (>= 0.8)
-    zeitwerk (2.6.4)
+    zeitwerk (2.6.6)
 
 PLATFORMS
-  ruby
+  x86_64-darwin-21
```
joelhawksley, line 290: Do we need this change?

Gemfile.lock
```diff
@@ -298,4 +317,4 @@ DEPENDENCIES
   yard-activesupport-concern (~> 0.0.1)
 
 BUNDLED WITH
-   2.2.33
+   2.2.32
```
joelhawksley, line 320: Same with this change.

test/sandbox/test/view_component_system_test.rb
```diff
@@ -0,0 +1,55 @@
...
+  def test_simple_js_interaction_in_browser_without_layout
+    inline_component = render_inline(SimpleJavascriptInteractionWithJsIncludedComponent.new)
+
+    with_inline_rendered_component_path(inline_component) do |path|
+      visit path
+
+      assert find("[data-hidden-field]", visible: false)
+      find("[data-button]", text: "Click Me To Reveal Something Cool").click
+      assert find("[data-hidden-field]", visible: true)
+    end
```
joelhawksley, line 17: 🤔 this look good to me, but I'm wondering if we might want to recommend nesting `render_inline` inside the helper call instead of assigning it to `inline_component`? This feels a tad verbose.

I also wonder about calling this `with_rendered_fragment_path`. What do you think of that name?

joelhawksley (commented):
👏🏻

test/sandbox/test/view_component_system_test.rb
joelhawksley: Let's do `with_rendered_component_path` 👍🏻

edwinthinks:
@joelhawksley all changed up!

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add `with_rendered_component_path` method to enable testing JS interactions in components.
```
```suggestion
* Add `with_rendered_component_path` for writing component system tests.
```

docs/guide/testing.md
```diff
@@ -233,3 +233,37 @@ To use component previews:
 # config/application.rb
 config.view_component.preview_paths << "#{Rails.root}/spec/components/previews"
 ```
+
+## Testing Interactive Components
```
joelhawksley, line 237:
```suggestion
## Component system tests
```

docs/guide/testing.md
```diff
@@ -233,3 +233,37 @@ To use component previews:
...
+
+Use `with_rendered_component_path` with `render_inline` in system tests to test interactive components:
```
joelhawksley, line 239:
```suggestion
Use `with_rendered_component_path` with `render_inline` to system test components:
```

docs/guide/testing.md
```diff
@@ -233,3 +233,37 @@ To use component previews:
...
+
+      assert(find("[data-hidden-field]", visible: false))
+      find("[data-button]", text: "Click Me To Reveal Something Cool").click
+      assert(find("[data-hidden-field]", visible: true))
+    end
+  end
+end
+```
+
+For components that depend on JavaScript provided by a layout, provide the `layout` argument:
```
joelhawksley, line 255:
```suggestion
For components that depend on a layout, provide the `layout` argument:
```

docs/guide/testing.md
```diff
@@ -233,3 +233,37 @@ To use component previews:
...
+
+```rb
+class ViewComponentSystemTest < ViewComponent::SystemTestCase
+  def test_simple_js_interaction_in_browser_with_layout
+    with_rendered_component_path(render_inline(SimpleJavascriptInteractionWithoutJsIncludedComponent.new), layout: 'application') do |path|
+      visit(path)
+
+      assert(find("[data-hidden-field]", visible: false))
+      find("[data-button]", text: "Click Me To Reveal Something Cool").click
+      assert(find("[data-hidden-field]", visible: true))
```
joelhawksley, line 265:
```suggestion
      # ...
```

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  module SystemTestHelpers
+    include TestHelpers
+
+    #
+    # Returns a block that can be used to visit the path of the inline rendered
+    # component.
```
joelhawksley, line 9:
```suggestion
    # Returns a block that can be used to visit the path of the inline rendered component.
```

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,47 @@
...
+    # @param fragment [Nokogiri::Fragment] The fragment returned from `render_inline` on instance of a view component.
```
joelhawksley, line 10:
```suggestion
    # @param fragment [Nokogiri::Fragment] The fragment returned from `render_inline`.
```

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,47 @@
...
+    # @param layout [String] The layout to use for the inline rendered component.
```
joelhawksley, line 11:
```suggestion
    # @param layout [String] The (optional) layout to use.
```

lib/view_component/system_test_helpers.rb
```diff
@@ -0,0 +1,47 @@
...
+        }
+      }
+
+      html = controller.render_to_string("view_components/preview", opts)
+
+      # Add './tmp/view_components/' directory if it doesn't exist to store the rendered component html
+      FileUtils.mkdir_p("./tmp/view_components/") unless Dir.exist?("./tmp/view_components/")
+
+      # Write to temporary file to contain fully rendered component
+      # within a browser
```
joelhawksley, line 32:
```suggestion
```

joelhawksley (approved):

joelhawksley:
@edwinthinks I can't thank you enough for your patience and hard work towards getting this patch landed. I'm so pleased with where we ended up ❤️

edwinthinks:
@joelhawksley it has been a pleasure! Thank you for working with me to get this to the finish line, couldn't have done it without you.

jonspalmer:
This is huge 🥳

developius:
This is absolutely fantastic. I implemented a very similar solution over the last few weeks (without realising this was on the horizon) and was going to merge today, but saw the new release which included this 😍

We're using rspec and our component specs live in `spec/components`. Capybara doesn't fire up the Rails server because it doesn't know we need it for our component specs, which was true until now.
Is there a way to get the server to boot without moving our specs to `spec/features`?

edwinthinks:
@developius you don't need to move them to get access to this. However, you do need to add this in your `rails_helper` to make it work as intended:
```
  config.include ViewComponent::SystemTestHelpers, type: :component
  config.include Capybara::DSL, type: :component
```

I'll open a PR to include this documentation that is missing for RSpec users :)

Timmitry:
I've been subscribed to this PR on Github since the beginning, and am very glad that it is finally merged. This is a fantastic feature, thank you so much @edwinthinks and @joelhawksley 💙

PR #1063, opened 2021-09-08
Add basic documentation for test helpers

joelhawksley:
### Summary

This PR introduces preliminary API documentation for `ViewComponent::TestHelpers`.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -151,6 +151,42 @@ namespace :docs do
           f.puts("_#{method.tag(:deprecated).text}_")
         end
       end
+
+      f.puts
+      f.puts("## ViewComponent::TestHelpers")
+
+      registry.
+        get("ViewComponent::TestHelpers").
+        meths.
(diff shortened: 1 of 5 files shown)


Rakefile
```diff
@@ -151,6 +151,42 @@ namespace :docs do
           f.puts("_#{method.tag(:deprecated).text}_")
         end
       end
+
+      f.puts
+      f.puts("## ViewComponent::TestHelpers")
+
+      registry.
```
BlakeWilliams, line 158: Does it make sense to start thinking about extracting a documentation generator class since we repeat this logic a few times?

BlakeWilliams (approved):
Looks good!

Rakefile
joelhawksley: Totally. We have similar code in Primer ViewComponents as well that I think we should consider abstracting...

PR #1065, opened 2021-09-09
Allow disabling of collection parameter validation

patrickarnett:
### Summary
This PR adds a config option to disable `validate_collection_parameter` checks.

### Other Information

It feels like an overstep that this gem would require a particular pattern when defining a component's `initialize` method. As long as I set the correct instance variables, why shouldn't I be able to use an options hash if I wish?

When using dry-initializer, specifically, this validation results in useless `initialize` definitions like
```ruby
  extend Dry::Initializer

  option :item
  option :another_option

  def initialize(item:, **opts)
    super
  end
end
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Allow disabling of `validate_collection_parameter` checks.
+
+    *Patrick Arnett*
+
 * Replace antipatterns section in the documentation with best practices.
 
     *Blake Williams*
--- a/docs/api.md
+++ b/docs/api.md
@@ -71,6 +71,14 @@ Set a custom default layout used for preview index and individual previews:
 
     config.view_component.default_preview_layout = "component_preview"
 
+### #enforce_collection_parameter
+
+Enable or disable enforcement of collection parameter as keyword argument in component initializer:
+
+    config.view_component.enforce_collection_parameter = false
+
+Defaults to `true`.
+
 ### #generate_stimulus_controller
 
 Always generate a Stimulus controller alongside the component:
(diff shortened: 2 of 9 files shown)


joelhawksley:
@patrickarnett thanks for opening this PR ❤️ 

Might there be a way to have this functionality play nicely with Dry::Initializer? It does provide a helpful feedback mechanism to consumers that I'd rather avoid turning off.

patrickarnett:
Hi @joelhawksley. The changes in this PR would not affect default behavior for current or new installs.

As best I can tell, the validation provides guidance for users who may be unfamiliar with `::with_collection`, and I'm not necessarily opposed to that. For some users like myself, however, the validation is an annoyance; my view components would work just fine without writing a redundant `initialize` method if only I could tell ViewComponent that I know what I'm doing.

patrickarnett:
It occurred to me last night that this PR still provides no means of including counters or iterators without defining `initialize`.

joelhawksley:
@patrickarnett that's a good point. Is there a way for us to introspect what Dry::Initializer has configured?

Spone:
> @patrickarnett that's a good point. Is there a way for us to introspect what Dry::Initializer has configured?

Maybe this? https://dry-rb.org/gems/dry-initializer/3.0/attributes/

patrickarnett:
I wouldn't want to get too specific to Dry::Initializer.

It seems the validation isn't entirely effective; if the initializer has the expected collection parameter as a positional argument, rather than a keyword argument, then validation will pass, but initialization will raise an ArgumentError. @joelhawksley What are your thoughts on checking for one of:

- A `key` or `keyreq` argument with the expected name
- Any `keyrest`, `rest`, or `opt` argument

With any of those, I believe we could be confident that the object is at least made available to the initializer.

joelhawksley:
@patrickarnett that seems fair to me 👍🏻

Issue #201, opened 2020-02-01
Rendering components outside of view context


dennyf:
I'm trying to find if there is a way to render a component outside of the view context. I'm working on a feature to register a custom tag in my markdown processor - so basically, I want to replace the custom tag/shortcode with the markup generated by a component.
So, I'm wondering if there is a way to call the component to generate the markup from another class, outside the view context?
Thanks!

dylnclrk:
Not sure if I understand correctly, and maybe this is too hacky, but could you set up a fake view context in much the same way that `[render_inline](https://github.com/github/actionview-component/blob/3b822bb07192e5c4bc5d0a8f9dd2f5b3e393b433/lib/action_view/component/test_helpers.rb#L6)` does?

joelhawksley:
@dennyf - @dylnclrk is probably on the right track. 

I'd love to see a PR for this. Let me know if you'd like to pair program on it 😄

dennyf:
@dylnclrk @joelhawksley Thank you both! I did something similar too, the following code worked fine for a very simple component :
```
component = MyComponent.new(title: "My Component")
component.render_in(ActionView::Base.new)
```
but it felt too hacky and probably it's not a good idea to call `render_in` directly.
I'll dig into the ActionView Component code to better understand how it works internally - I'll submit a PR if I can get to find a good way to implement this.

joelhawksley:
@dennyf `render_in` is the public interface for components - it's definitely the right way in!

What if we had changed `render_in` from:

```ruby
def render_in(view_context, *args, &block)
```

to:

```ruby
def render_in(view_context = ActionView::Base.new, *args, &block)
```

I'm not sure if that would work in all cases, but I think we'd be fine to use that kind of fallback.

dennyf:
@joelhawksley  Thanks, that's good to hear! Actually I just noticed that Rails is throwing a warning when initializing `ActionView::Base` with no arguments:
```
DEPRECATION WARNING: ActionView::Base instances should be constructed with a lookup context, assignments, and a controller.
```
so I guess it won't be as simple as that, I think it will be an overkill to initialize it with all the required parameters... I'll see if I can get any other ideas.

joelhawksley:
@dennyf It might be possible to provide all of those things, FWIW. Let me know if you want to have a look at this together: joelhawksley@github.com

dennyf:
[...]

PR #1068, opened 2021-09-10
Add "how to render a component to a string" to FAQ

Spone:
### Summary

It's not obvious that you cannot use `render` or `render_to_string` to render a ViewComponent from inside a controller action.

### Other Information

The `#render_in` method is private, but I can't find any other way to achieve this. I think it's a good idea to mention this in the FAQ (especially to help people debug `AbstractController::DoubleRenderError`), but maybe add a disclaimer? Or consider making the method public?

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,7 +14,6 @@ title: Changelog
 * Allow query parameters in `with_request_url` test helper.
 
     *Javi Martín*
-    
 * Add "how to render a component to a string" to FAQ.
 
     *Hans Lemuet*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ title: Changelog
 * Allow query parameters in `with_request_url` test helper.
 
     *Javi Martín*
+    
+* Add "how to render a component to a string" to FAQ
+
+    *Hans Lemuet*
 
 ## 2.40.0
 
@@ -66,7 +70,7 @@ title: Changelog
 
 * Add test case for conflict with internal `@variant` variable.
 
-   *David Backeus*
+    *David Backeus*
 
 * Document decision to not change naming convention recommendation to remove `-Component` suffix.
 
@@ -130,7 +134,7 @@ title: Changelog
 
 * Ensure consistent indentation with Rubocop.
 
-    *Joel Hawksley
+    *Joel Hawksley*
 
 * Bump `activesupport` upper bound from `< 7.0` to `< 8.0`.
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,7 +11,7 @@ title: Changelog
 
     *Javi Martín*
     
-* Add "how to render a component to a string" to FAQ
+* Add "how to render a component to a string" to FAQ.
 
     *Hans Lemuet*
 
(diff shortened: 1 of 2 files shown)


docs/CHANGELOG.md
```diff
@@ -66,7 +70,7 @@ title: Changelog
 
 * Add test case for conflict with internal `@variant` variable.
 
-   *David Backeus*
+    *David Backeus*
```
vpereira, line 73: looks like this change is irrelevant to the PR?

docs/CHANGELOG.md
```diff
@@ -130,7 +134,7 @@ title: Changelog
 
 * Ensure consistent indentation with Rubocop.
 
-    *Joel Hawksley
+    *Joel Hawksley*
```
vpereira, line 137: looks like this change is irrelevant to the PR?

docs/CHANGELOG.md
Spone: Yes, I was just fixing the changelog indentation at the same time

vpereira: Ok, normally in that case we should add these "unrelated" changes to a separate commit, like "Fix indentation"

Spone:
Here is a related comment: https://github.com/github/view_component/issues/201#issuecomment-582413280 where @joelhawksley explains that:

> `render_in` is the public interface for components - it's definitely the right way in!

docs/CHANGELOG.md
joelhawksley: I'm fine with landing this formatting fix :)

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ title: Changelog
 * Allow query parameters in `with_request_url` test helper.
 
     *Javi Martín*
+    
+* Add "how to render a component to a string" to FAQ
```
joelhawksley, line 14:
```suggestion
* Add "how to render a component to a string" to FAQ.
```

docs/CHANGELOG.md
joelhawksley: We squash commits on merge, so I'm not too worried about atomic commits unless it's a big PR that might be best reviewed step-wise. Even then, we'd probably want to multiple PRs ❤️

docs/frequently-asked-questions.md
```diff
@@ -13,6 +13,24 @@ Yes. ViewComponent is tested against ERB, Haml, and Slim, but it should support
...
+class PagesController < ApplicationController
+  def index
+    # Does not work: triggers a `AbstractController::DoubleRenderError`
+    # @reusable_icon = render IconComponent.new('close')
+    
+    # Does not work: renders the whole index view as a string
+    # @reusable_icon = render_to_string IconComponent.new('close')
+  
+    # Works: renders the component as a string
+    @reusable_icon = IconComponent.new('close').render_in(view_context)
```
joelhawksley, line 30: Woah, I didn't even think about this use case. That is tricky!

joelhawksley (approved):

docs/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ title: Changelog
 * Allow query parameters in `with_request_url` test helper.
 
     *Javi Martín*
+
```
```suggestion

```

Spone:
@joelhawksley what do you think about documenting `#render_in` in https://viewcomponent.org/api.html#instance-methods?

joelhawksley:
@Spone sure thing! The API docs need some attention for sure ❤️

Issue #1062, opened 2021-09-07
Support different formats


vpereira:
Reading the code https://github.com/github/view_component/blob/0fd7cbf087248f3ce9c0becb4a00f598e361e3ea/lib/view_component/base.rb#L416 and documentation, I'm unsure we are able to use different formats like `:xml` or `:text`, right?

### Problem:

I have a component that I want to render as `:html` and `:text`. For that, I created two views for my component, (i.e `foo_component.html.haml and foo_component.text.haml`. Somehow I expected to be able to pass the format to `render`, however I'm not able to, right?

### Possible solution:

To render text "inline", I should have a method `call` to my class where I bypass any assigned template and render a string inline

#### Update
looks like I cannot have a template (i.e `foo_component.html.haml`) *and* implement the `call` method to deal with the inline rendering in the same class

Is that correct or did I miss anything? Any idea how to deal with that?

joelhawksley:
@vpereira I'd be open to moving to support other formats. Could you open a PR that adds a test component with an API you think makes sense?

vpereira:
> @vpereira I'd be open to moving to support other formats. Could you open a PR that adds a test component with an API you think makes sense?

Hi  @joelhawksley, thank you for your answer :)

So I will have to think about it. We have so many scenarios,  basically we have:

a) regular components 
b) XML components (we have an XML API, where we use a lot of partials)
c) helpers with some `raw()` calls embedded, which we wanted to extract to partials (with different formats), and consequently it would become a component itself.

the first case, is straight forward. For XML endpoints, would just be possible if `ViewComponent` would support XML format. The last case, I'm kind of solving it like https://github.com/openSUSE/open-build-service/pull/11583, where for html format, i render a template and for text format, i call a simple method (and bypass any template).  BTW the "OBS" is probably one of the oldest rails code base being actively developed. We are on rails 6.1 and it exists since 2006. 

For which scenario you would like to see a test component? for the XML API, or a component supporting many formats?

joelhawksley:
@vpereira thanks for sharing the example PR! How cool to see ViewComponent in use in OBS ❤️ 

I think it would be best to prototype something that supports both `raw` and `html` like that PR would benefit from.

reeganviljoen:
@vpereira I have started a discussion around how view_component handles formats because their are other issues with similar interests, and hopefully that will help shed some light on this feature

simkim:
Would be helpfull to have an html and turbo_stream format

[...]

PR #1070, opened 2021-09-13
Add test component for multiple formats support

vpereira:
As asked by @joelhawksley on issue #1062, I'm opening a PR with a test component with multiple formats (html and text)

Feel free to guide me if something is missing.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -15,6 +15,10 @@ title: Changelog
 
     *Javi Martín*
 
+* Add test component for multiple formats support.
+
+    * Victor Pereira*
+
 ## 2.40.0
 
 * Replace antipatterns section in the documentation with best practices.
--- a/test/view_component/multiple_formats.rb
+++ b/test/view_component/multiple_formats.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ViewComponentMultipleFormatsTest < ViewComponent::TestCase
+  def test_render_html
+    render_inline(MyComponentMultipleFormat.new, format: :html)
+
+    assert_selector("div", text: "hello,world!")
+  end
+
+  def test_render_text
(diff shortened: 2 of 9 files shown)


docs/CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ title: Changelog
 
     *Javi Martín*
 
+* Add test component for multiple formats support
+    * Victor Pereira*
```
joelhawksley, line 19:
```suggestion
* Add test component for multiple formats support.

    * Victor Pereira*
```

test/view_component/multiple_formats.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ViewComponentMultipleFormatsTest < ViewComponent::TestCase
+  def test_render_html
+    render_inline(MyComponentMultipleFormat.new, format: :html)
```
joelhawksley, line 7: Wouldn't the format be set by the request? I'm guessing we'd need to add a `with_request_format` test helper.

joelhawksley (commented):
Thanks for adding this test case! We should pair on making it pass 😄 

Email me: joelhawksley@github.com

test/view_component/multiple_formats.rb
vpereira: Ok, I started with a skeleton code/pseudo code.

What I'm thinking to do is:

- check if there is a view for `#{component}.#{format}.#{extension}`. if it exists, set it as view.

vpereira:
@joelhawksley any reason why `view_component` isn't using https://github.com/rails/rails/blob/main/actionview/lib/action_view/lookup_context.rb to find the templates?

joelhawksley:
🤦🏻 I had completely forgotten about this PR and basically re-implemented it with @jaredmoody! For now, I'm going to close this as stale as we are tracking the need for this feature here: https://github.com/ViewComponent/view_component/issues/1670.

PR #1071, opened 2021-09-13
Add #render_in to API docs

Spone:
### Summary

Removed the `@private` YARD tag in order for the `#render_in` instance method to appear in the API docs.

### Other Information

Continued from https://github.com/github/view_component/pull/1068?notification_referrer_id=MDE4Ok5vdGlmaWNhdGlvblRocmVhZDI0MDIyMDE5MjE6Mzg1MjQ%3D#issuecomment-918589884

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -52,31 +52,14 @@ coupling that inhibits encapsulation & reuse, often making testing difficult.
 
 Override to determine whether the ViewComponent should render.
 
-### #render_in(view_context, &block)
+### #render_in(view_context, &block) → [String]
 
 Entrypoint for rendering components.
 
-view_context: ActionView context from calling view
-block: optional block to be captured within the view context
+- `view_context`: ActionView context from calling view
+- `block`: optional block to be captured within the view context
 
-returns HTML that has been escaped by the respective template handler
-
-Example subclass:
-
-app/components/my_component.rb:
-class MyComponent < ViewComponent::Base
-  def initialize(title:)
-    @title = title
-  end
-end
-
-app/components/my_component.html.erb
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -52,6 +52,32 @@ coupling that inhibits encapsulation & reuse, often making testing difficult.
...
+  end
+end
+
+app/components/my_component.html.erb
+<span title="<%= @title %>">Hello, <%= content %>!</span>
+
+In use:
+<%= render MyComponent.new(title: "greeting") do %>world<% end %>
+returns:
+<span title="greeting">Hello, world!</span>
```
Spone, line 79: I'm not sure we should keep the whole comment. Maybe we can remove everything after `respective template handler`?

docs/api.md
```diff
@@ -52,6 +52,32 @@ coupling that inhibits encapsulation & reuse, often making testing difficult.
 
 Override to determine whether the ViewComponent should render.
 
+### #render_in(view_context, &block)
```
joelhawksley, line 55: These docs are meant to be auto-generated from YARD comments. Perhaps we could update the contributing guidelines / add a comment to the top of this file pointing folks in that direction?

docs/api.md
joelhawksley: I agree!

docs/api.md
Spone: I'm not sure what you mean. In this PR the docs are generated from YARD comments. I ran `bundle exec rake docs:build` before committing, as per https://viewcomponent.org/CONTRIBUTING.html#documentation

joelhawksley (approved):

PR #1073, opened 2021-09-17
Add parent option

### LINKED ISSUE #1067: Add --parent option to component generator
## Feature request

The `ComponentGenerator` implements a [`#parent_class`](https://github.com/github/view_component/blob/8ffe1d96ef52995ba0a12efff48e33b02ac5279a/lib/rails/generators/component/component_generator.rb#L34-L36) private method that picks the right parent class (whether `ApplicationComponent` is defined or not).

I'd like to be able to pass `--parent CustomBaseComponent`, so the generated component looks like that:

```rb
class ExampleComponent < CustomBaseComponent
  # ...
end
```

### Motivation

A similar `--parent` option [exists in Rails](https://github.com/rails/rails/blob/f95c0b7e96eb36bc3efc0c5beffbb9e84ea664e4/activerecord/lib/rails/generators/active_record/model/model_generator.rb#L14).

It would also be useful when implementing our component generator in [view_component-form](https://github.com/pantographe/view_component-form) :)

Spone:
### Summary

* Add `--parent` generator option to specify the parent class.
* Add config option `config.view_component.component_parent_class` to change it project-wide.
* Add tests

### Other Information

Close #1067

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,11 @@ title: Changelog
 
 ## main
 
+* Add `--parent` generator option to specify the parent class.
+* Add config option `config.view_component.component_parent_class` to change it project-wide.
+
+    *Hans Lemuet*
+
 * Update docs to add example for using Devise helpers in tests.
 
     *Matthew Rider*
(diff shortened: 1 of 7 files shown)


joelhawksley (approved):

joelhawksley:
Wow! Thank you for the clean PR 🙇🏻

PR #1074, opened 2021-09-18
Generator --parent option: add more tests and fix edge case

Spone:
### Summary

(Continued from #1073)

I found an edge case: when you both have `ApplicationComponent` defined in your project and configure a `component_parent_class`, `ApplicationComponent` was always used.

I added more tests, fixed the issue and refactored `ComponentGenerator#parent_class`.


Spone:
@joelhawksley I'm not sure it makes sense to add a CHANGELOG entry in that case

Issue #1069, opened 2021-09-11
Helpers not reloaded in development?


jdelStrother:
If I have a ViewComponent that calls a helper method (eg from app/helpers/application_helper.rb), the helper is normally reloaded on each page load, as expected.

However, if I load the preview page (eg http://localhost:3000/rails/view_components/example_component/default), the helper methods appear to be the helpers as they were at application-boot-time. Changes to app/helpers/application_helper.rb have no effect, and I have to restart my rails server before they're picked up.

Changes to the view-component code (eg app/components/example_component.rb) _do_ get reloaded, however.

Is this expected behaviour? Some side-effect of inheriting the preview controller from Rails::ApplicationController?

joelhawksley:
@jdelStrother interesting! Would you be up for writing an integration test that reproduces this behavior?

jdelStrother:
@joelhawksley sure, try https://github.com/github/view_component/pull/1075 ?

PR #1075, opened 2021-09-20
Fix helpers not reloading in previews

jdelStrother:
### Summary

Helpers don't seem to get reloaded when working with component previews (https://github.com/github/view_component/issues/1069).

This PR adds 2 tests.
[One](https://github.com/github/view_component/compare/main...jdelStrother:helper-reloads?expand=1#diff-c7634fb9ae4108c9cbf3eaebf452a2cdd2d87466586cab6923a007f36f6f34d8R86) demonstrating that helpers _are_ successfully reloaded when components are rendered through a 'normal' controller.
[The other](https://github.com/github/view_component/compare/main...jdelStrother:helper-reloads?expand=1#diff-c7634fb9ae4108c9cbf3eaebf452a2cdd2d87466586cab6923a007f36f6f34d8R109) attempts to do the same via `ViewComponentsController#previews`, but the test fails since the helper isn't reloaded.

Unfortunately I had to set `config.cache_classes=false` to get this working, which causes a couple of other test failures. I'm not sure there's a good way to only set that for the two tests I added, unless we want to spawn a brand new test process for them.

### DIFF
--- a/test/sandbox/test/rendering_test.rb
+++ b/test/sandbox/test/rendering_test.rb
@@ -419,6 +419,8 @@ class RenderingTest < ViewComponent::TestCase
   def test_compiles_unrendered_component
     skip "this might not still be compiled if it's been reloaded since boot"
     assert UnreferencedComponent.compiled?
+    # Or a possible alternative is booting a fresh process:
+    # assert_equal "true\n", `ruby -r bundler/setup -r ./test/sandbox/config/environment.rb -e "puts UnreferencedComponent.compiled?"`
   end
 
   def test_compiles_components_without_initializers
--- a/test/sandbox/test/rendering_test.rb
+++ b/test/sandbox/test/rendering_test.rb
@@ -417,10 +417,12 @@ class RenderingTest < ViewComponent::TestCase
   end
 
   def test_compiles_unrendered_component
+    skip "this might not still be compiled if it's been reloaded since boot"
     assert UnreferencedComponent.compiled?
   end
 
   def test_compiles_components_without_initializers
+    MissingInitializerComponent.compile
     assert MissingInitializerComponent.compiled?
   end
 
--- a/test/sandbox/test/rendering_test.rb
+++ b/test/sandbox/test/rendering_test.rb
@@ -931,8 +931,8 @@ class RenderingTest < ViewComponent::TestCase
     # undo the changes made by self.class.compile and friends, forcing a compile the next time
     # #render_template_for is called. This shouldn't be necessary except in the test environment,
     # since eager loading is turned on here.
-    Object.send(:remove_const, :MyComponent)
-    Object.send(:remove_const, :InheritedWithOwnTemplateComponent)
+    Object.send(:remove_const, :MyComponent) if defined?(MyComponent)
+    Object.send(:remove_const, :InheritedWithOwnTemplateComponent) if defined?(InheritedWithOwnTemplateComponent)
 
     load "test/sandbox/app/components/my_component.rb"
     load "test/sandbox/app/components/inherited_with_own_template_component.rb"
--- a/test/sandbox/test/rendering_test.rb
+++ b/test/sandbox/test/rendering_test.rb
@@ -417,14 +417,18 @@ class RenderingTest < ViewComponent::TestCase
   end
 
   def test_compiles_unrendered_component
-    skip "this might not still be compiled if it's been reloaded since boot"
+    # The UnreferencedComponent will get compiled at boot,
+    # but that might have been thrown away if code-reloading is enabled
+    skip unless Rails.env.cache_classes?
+
     assert UnreferencedComponent.compiled?
-    # Or a possible alternative is booting a fresh process:
-    # assert_equal "true\n", `ruby -r bundler/setup -r ./test/sandbox/config/environment.rb -e "puts UnreferencedComponent.compiled?"`
   end
 
   def test_compiles_components_without_initializers
-    MissingInitializerComponent.compile
+    # MissingInitializerComponent will get compiled at boot,
+    # but that might have been thrown away if code-reloading is enabled
+    skip unless Rails.env.cache_classes?
+
     assert MissingInitializerComponent.compiled?
   end
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -71,10 +71,12 @@ jobs:
       run: |
         bundle config path vendor/bundle
         bundle update
-        bundle exec rake test spec
+        # Code-reloading isn't compatible with simplecov, so we need to run once
+        # to collect coverage, and again to test reloads.
+        MEASURE_COVERAGE=true bundle exec rake test spec
+        ENABLE_RELOADING=true bundle exec rake test spec
       env:
         RAISE_ON_WARNING: 1
-        MEASURE_COVERAGE: true
         RAILS_VERSION: ${{ matrix.rails_version }}
         CAPTURE_PATCH_ENABLED: ${{ matrix.mode == 'capture_patch_enabled' && 'true' || 'false' }}
     - name: Upload coverage results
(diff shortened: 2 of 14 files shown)


joelhawksley:
@jdelStrother thank you for providing test cases 🙏🏻 

@juanmanuelramallo would you be up for having a look into this?

juanmanuelramallo:
Hey @jdelStrother thanks for the failing tests! 

Please have a look at [this commit](https://github.com/github/view_component/pull/1075/commits/f601018c3c31f82750b98914a9644f7f64feb100) and let me know what you think. This commit solves the reloading issue with helpers. (cc @joelhawksley)

We'd still need to solve the other tests that fail due to the cache_classes configuration.

app/controllers/view_components_controller.rb
```diff
@@ -2,12 +2,14 @@
 
 require "rails/application_controller"
 
-class ViewComponentsController < Rails::ApplicationController # :nodoc:
+class ViewComponentsController < ActionController::Base # :nodoc:
```
camertron, line 5: I spent some time yesterday looking into this as well and this line fixed the failing test. However, it made another test fail. Do you know why the inheritance matters here?

juanmanuelramallo: The other failures in this branch seem to be related to the cache_classes configuration change. I wasn't able to dynamically update that config at runtime and make it work as expected.

I think the actual issue is that rails/application_controller is not reloaded by zeitwerk, perhaps because we require it manually and/or we don't have it present in the autoload_paths. BTW I just noticed we should remove the require line as well.

jdelStrother:
> Hey @jdelStrother thanks for the failing tests! 
> 
> 
> 
> Please have a look at [this commit](https://github.com/github/view_component/pull/1075/commits/f601018c3c31f82750b98914a9644f7f64feb100) and let me know what you think. This commit solves the reloading issue with helpers. (cc @joelhawksley)
> 
> 

Thanks - looks promising. I'm away for a week though, will try it on my return

jdelStrother:
Confirming this does fix the reloading issues for me. However, I found a possible alternate fix...

It looks like Rails only re-includes reloaded helpers in direct subclasses of ActionController::Base (https://github.com/rails/rails/blob/f1f86603a96b965e28839e418a5712325bf9ca38/actionpack/lib/action_controller/railties/helpers.rb#L18), which is why @juanmanuelramallo's fix works.

So, ViewComponentsController could do something like this:

```diff
diff --git a/app/controllers/view_components_controller.rb b/app/controllers/view_components_controller.rb
index fbc6e32..e3e1452 100644
--- a/app/controllers/view_components_controller.rb
+++ b/app/controllers/view_components_controller.rb
@@ -8,6 +8,7 @@ class ViewComponentsController < Rails::ApplicationController # :nodoc:
   around_action :set_locale, only: :previews
   before_action :find_preview, only: :previews
   before_action :require_local!, unless: :show_previews?
+  helper :all

   if respond_to?(:content_security_policy)
     content_security_policy(false)
```


which ensures that when ViewComponentsController is reloaded, it re-includes the app's helper modules. Of course, that assumes that we want to include `:all` helper modules... have you heard from any users wanting to only include specific helpers in ViewComponentsController ?

juanmanuelramallo:
I like your approach better. Less changes and way simpler. 

Do you mind reverting my commits in favor of the change you proposed?

jdelStrother:
Couple of thoughts - 

* By adding `helper :all` to ViewComponentsController, we'll be including the helpers twice - a non-reloadable version in `Rails::ApplicationController`, and a reloadable version in ViewComponentsController. Not ideal, but I think it's probably fine in practice.
* What do you want to do about the tests? I think the options are:
  * Leave reloading helpers untested
  * Disable cache_classes for the entire test suite, and see if we can fix the failing tests that that has produced
  * Make the test suite much more complicated and have it fork off entire new rails tests processes with different settings in test.rb

MrHubble:
Anything I can do to help with this? Changes to my view-component ERB code (eg app/components/example_component.html.erb) do **not** get reloaded unless I restart the server.

joelhawksley:
@jdelStrother would you be interested in picking this PR back up?

jdelStrother:
Sure. Any thoughts on my latest comment?

test/sandbox/test/rendering_test.rb
```diff
@@ -418,10 +418,12 @@ def test_validations_component
   end
 
   def test_compiles_unrendered_component
+    skip "this might not still be compiled if it's been reloaded since boot"
     assert UnreferencedComponent.compiled?
   end
```
jdelStrother, line 423: Not sure if it's worth keeping this. UnreferencedComponent will get compiled on boot (because eager_load=true) but it might get reloaded before this test runs (because cache_classes=false).

I assume it's trying to test this line - https://github.com/ViewComponent/view_component/blob/964a2c15f74588819ea64ff69d12dfb0de80fe3b/lib/view_component/engine.rb#L68




If we reeaaally want to test that unreferenced components get compiled on boot, we could start a new rails process within the test:

```ruby
  def test_compiles_unrendered_component
    assert_equal "true\n", `ruby -r bundler/setup -r ./test/sandbox/config/environment.rb -e "puts UnreferencedComponent.compiled?"`
  end
```

jdelStrother:
OK, I've rebased & fixed up the test failures with `cache_classes=false`. (Well, mostly. One of them I've marked with `skip` - see comments)

test/sandbox/test/rendering_test.rb
camertron: Since it's unreferenced, why would it get reloaded?

camertron (changes requested):
Hey @jdelStrother I took some time to go through this in detail today and I think it's almost ready. I was going to take a crack at the test failures but there were some conflicts merging main and I wasn't confident I could do it without making a mistake. Could you merge main?

test/sandbox/test/rendering_test.rb
jdelStrother: It'll get eager-loaded during boot due to https://github.com/ViewComponent/view_component/blob/d6b385e0757e6fb1cebabc7f222fdf51a29be24c/test/sandbox/config/environments/test.rb#L45

and then reloaded whenever we call `app.reloader.reload!`

jdelStrother:
> Something's going on with code coverage tho

Mm. I notice that spec_helper & test_helper use the same command_name - won't that result in them overwriting each others' coverage results?

Sadly, that's not the real issue. I think the real issue is that simplecov doesn't support coverage of reloaded files - https://github.com/simplecov-ruby/simplecov/issues/389 - and preview_helper.rb is getting reloaded. I'm struggling to figure out a good way around it.

camertron:
> I notice that spec_helper & test_helper use the same command_name - won't that result in them overwriting each others' coverage results?

Huh, yes it will, although it doesn't seem to be? Not sure what's going on here.

> I think the real issue is that simplecov doesn't support coverage of reloaded files

You're absolutely right. I messed around with this for a long time today and wasn't able to get anywhere. I tried dumping the coverage results before reloading, but for some reason that doesn't work (it identifies a bunch of uncovered lines that are definitely covered).

As mentioned in the issue you linked, this appears to be a problem in Ruby's coverage module. Or rather, it's expected behavior we can't do anything about. Whatever we would do is likely to be pretty hacky too 😢 

What if we ran the test suite multiple times, once for coverage and once to test reloading behavior? We could skip the test you added if `config.cache_classes` is `true`.

jdelStrother:
> > I notice that spec_helper & test_helper use the same command_name - won't that result in them overwriting each others' coverage results?
> Huh, yes it will, although it doesn't seem to be? Not sure what's going on here.

It's only setting the command name if `ENV["RUBY_VERSION"]` is set, which it never is AFAICT. So simplecov by default is naming them "RSpec" and "Unit test" - see #1750 

Not sure what's up with the failing tests, will try & take a closer look later. Bad MiniTest version ??

PR #1085, opened 2021-10-01
Add failing test for memoized helpers

BlakeWilliams:
Currently helpers do not re-use the same view_context which can cause subtle bugs when using ivars in components.

This adds a failing test for this unexpected behavior.


lib/view_component/base.rb
```diff
@@ -173,7 +178,14 @@ def helpers
...
 
-      @__vc_helpers ||= controller.view_context
+      # Attempt to re-use the original view_context passed to the first
+      # component rendered in the rendering pipeline. This prevents the
+      # instantiation of a new view_context via `controller.view_context` which
+      # always returns a new instance of the view context class.
+
+      # This allows ivars to remain persisted when using the same helper via
+      # `helpers` across multiple components and partials.
+      @__vc_helpers ||= original_view_context || controller.view_context
```
BlakeWilliams, line 188: We looked at upstreaming a change to `rails/rails` that would memoize `controller.view_context` but the intention seemed to be that a single instance should be instantiated and threaded throughout the render chain. I'm curious if anyone has more context or thoughts about attempting to upstream that to resolve the issue with the intention of eventually reverting this change.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Fix bug where `helpers` would instantiate and use a new `view_context` in each component.
+
+    *Blake Williams, Ian C. Anderson*
```
```suggestion
    *Blake Williams*, *Ian C. Anderson*
```

lib/view_component/base.rb
```diff
@@ -28,6 +28,8 @@ class Base < ActionView::Base
     class_attribute :content_areas
     self.content_areas = [] # class_attribute:default doesn't work until Rails 5.2
 
+    attr_accessor :original_view_context
```
joelhawksley, line 31: Is this threadsafe? I wonder if CurrentAttributes would be a more Rails-y way to pull this off.

BlakeWilliams: Thinking about it a bit more, I don't think it's much different than the existing code since we set `@view_context` as a property on the component in `render_in`, both aren't thread safe.

PR #1090, opened 2021-10-07
Add failing test for default form builder

boardfish:
### Summary

When form helpers are called directly, the default form builder is not used and must be specified. For example, calling this:

```
<%= fields do |f| %>
  <%= f.text_field :hello_world %>
<% end %>
```

does not use the default form builder. Either manually specifying the `builder:` argument to `fields` or calling `helpers.fields` (as I have in the tests here) works, though.

I feel that this is a little unintuitive - it's odd that form helpers do work directly within ViewComponent but do not inherit the default form builder specified by the controller. I feel that they should either raise an error or be delegated to `helpers` by default.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add failing test for default form builder
+
+    *Simon Fish*
+
 ## 2.46.0
 
 * Add thread safety to the compiler.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,7 +7,7 @@ title: Changelog
 
 ## main
 
-* Add failing test for default form builder
+* Add failing test for default form builder and documentation around known issue.
 
     *Simon Fish*
 
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -12,9 +12,10 @@ ViewComponent [isn't currently compatible](https://github.com/github/view_compon
 ## Inconsistent controller rendering behavior between Rails versions
 
 In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout.
-## Forms do not use the default form builder
 
-Calls to form helpers such as `form_with` in ViewComponents [do not use the default form builder](https://github.com/github/view_component/pull/1090#issue-753331927). This is by design, as it allows global state to change the rendered output of a component. Instead, consider passing a form builder into form helpers via the `builder` argument:
+## Forms don't use the default form builder
+
+Calls to form helpers such as `form_with` in ViewComponents [don't use the default form builder](https://github.com/github/view_component/pull/1090#issue-753331927). This is by design, as it allows global state to change the rendered output of a component. Instead, consider passing a form builder into form helpers via the `builder` argument:
 
 ```html.erb
 <%= form_for(record, builder: CustomFormBuilder) do |f| %>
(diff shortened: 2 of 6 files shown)


boardfish:
I'd appreciate it if someone could add a `hacktoberfest-accepted` label to the PR. Thanks!

boardfish:
@joelhawksley Did you have any further thoughts on that, or do we still want to go with delegation?

joelhawksley:
@boardfish thanks for the ping. Thinking about this some more, I think we should document the existing incompatibility and push folks towards passing in a builder explicitly. 

Allowing global state to change the rendered output of a component seems like a tricky thing to reason about and I think we should tread carefully in making changes that support such behavior.

boardfish:
Good to know. I'll wrap this up with that in mind soon.

boardfish:
@joelhawksley I've rebased, and documented the issue in 0dfac565d6228697dbd2f6a7dcae561d55579c7e. Hopefully that makes this good to go.

Spone (approved):

docs/known_issues.md
```diff
@@ -12,3 +12,13 @@ ViewComponent is [not currently compatible](https://github.com/github/view_compo
 ## Inconsistent controller rendering behavior between Rails versions
 
 In versions of Rails < 6.1, rendering a ViewComponent from a controller does not include the layout.
+
+## Forms do not use the default form builder
+
+Calls to form helpers such as `form_with` from ViewComponent [will not use the default form builder](https://github.com/github/view_component/pull/1090#issue-753331927). This is by design, as it allows global state to change the rendered output of a component. Instead, you may want to consider passing a form builder into form helpers via the `builder` argument.
```
joelhawksley, line 18:
```suggestion
Calls to form helpers such as `form_with` in ViewComponents [do not use the default form builder](https://github.com/github/view_component/pull/1090#issue-753331927). This is by design, as it allows global state to change the rendered output of a component. Instead, consider passing a form builder into form helpers via the `builder` argument:
```

joelhawksley (approved):

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add failing test for default form builder
```
```suggestion
* Add failing test for default form builder and documentation around known issue.
```

Spone:
Yes, I'll merge!

rmacklin (commented):
@boardfish I've added a few questions, but I might just be misunderstanding things - apologies if that's the case

docs/known_issues.md
```diff
@@ -12,3 +12,13 @@ ViewComponent [isn't currently compatible](https://github.com/github/view_compon
 ## Inconsistent controller rendering behavior between Rails versions
 
 In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout.
+
+## Forms don't use the default form builder
+
+Calls to form helpers such as `form_with` in ViewComponents [don't use the default form builder](https://github.com/github/view_component/pull/1090#issue-753331927). This is by design, as it allows global state to change the rendered output of a component. Instead, consider passing a form builder into form helpers via the `builder` argument:
```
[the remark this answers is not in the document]
pboling: Agree, the phrasing is awkward.

```suggestion
Calls to form helpers such as `form_with` in ViewComponents [don't use the default form builder](https://github.com/github/view_component/pull/1090#issue-753331927). This is by design, as it would otherwise allow global state to change the rendered output of a component. Instead, consider passing a form builder into form helpers via the `builder` argument:
```

boardfish:
Thanks for your feedback @rmacklin - I'll create a follow-up PR to address these issues.

PR #1096, opened 2021-10-13
Feature/multiple templates arguments

### LINKED ISSUE #387: Support multiple views

jsolas:
Summary
Fixes #387

This PR builds on the branch by @joelhawksley, @fermion, @fsateler and adds the ability to specify arguments for the generated functions, based on what was discussed https://github.com/github/view_component/issues/387#issuecomment-926668950. In this PR:

1- Each template would define a method called `render_<template_name>(locals = {})`
2- We will not be defining the locals inside the generated method like rails does. They would need to be accessed with locals[:foo]

The logic being that one may want to use templates to extract sections, without wanting to build a full component. For example, I could have a Table component, and I would want to have the row definition in a sidecar template. This requires being able to pass in each item for every iteration:
```ruby
  def initialize(models)
    @models = models
  end
end
```
```erb
<%# table_component.html.erb %>
<table>
  <% @models.each do |model| %>
    <%= render_row model: model %>
  <% end %>
</table>
```
```erb
<%# row.html.erb %>
<tr>
  <%# I can use `model` here %>
  <td><%= locals[:model].name  %></td>
</tr>
```

### DIFF
--- a/adr/0003-multiple-templates-per-component.md
+++ b/adr/0003-multiple-templates-per-component.md
@@ -1,0 +1,88 @@
+# 1. Allow multiple templates 
+
+Date: 2021-10-13
+
+## Status
+
+Proposed.
+
+## Context
+
+As components become larger (for example, because you are implementing a whole page), it becomes
+useful to be able to extract sections of the view to a different file. ActionView has
+partials, and ViewComponent lacks a similar mechanism. 
+
+ActionView partials have the problem that their interface is not introspectable. Data
+may be passed into the partial via ivars or locals, and it is impossible to know
+which without actually opening up the file. Additionally, partials are globally
+invocable, thus making it difficult to detect if a given partial is in use or not,
+and who are its users.
+
+## Considered Options
+
+* Introduce component partials to components
+* Keep components as-is
+
+### Component partials
+
+Allow multiple ERB templates available within the component, and make it possible to
+invoke them from the main view.Templates are compiled to methods in the format `render_#{template_basename}(locals = {})`
+
+**Pros:**
+* Better performance due to lack of GC pressure and object creation
+* Reduces the number of components needed to express a more complex view.
+* Extracted sections are not exposed outside the component, thus reducing component library API surface.
+
+**Cons:**
+* Another concept for users of ViewComponent to learn and understand.
+* Components are no longer the only way to encapsulate behavior.
+
+### Partial components by [fstaler](https://github.com/fsateler)
+
+In this approach the template arguments are previously defined in the `template_arguments` method
+
+[See here](https://github.com/github/view_component/pull/451):
+
+**Pros:**
+* Better performance due to lack of GC pressure and object creation
+* Reduces the number of components needed to express a more complex view.
+* Extracted sections are not exposed outside the component, thus reducing component library API surface.
+
+**Cons:**
+* Another concept for users of ViewComponent to learn and understand.
+* Components are no longer the only way to encapsulate behavior.
+* `call_foo` api feels awkward and not very Rails like
+* Declare templates and their arguments explicitly before using them
+
+### Keeping components as-is
+
+**Pros:**
+* The API remains simple and components are the only way to encapsulate behavior.
+* Encourages creating reusable sub-components.
+
+**Cons:**
+* Extracting a component results in more GC and intermediate objects.
+* Extracting a component may result in tightly coupled but split components.
+* Creates new public components thus expanding component library API surface.
+
+## Decision
+
+We will allow having multiple templates in the sidecar asset. Each asset will be compiled to
+it's own method `render_<template_name>`. I think it is simple, similar to rails and meets what is expected
+
+## Consequences
+
+This implementation has better performance characteristics over both an extracted component
+and ActionView partials, because it avoids creating intermediate objects, and the overhead of
+creating bindings and `instance_exec`. 
+Having explicit arguments makes the interface explicit.
+
+TODO: The following are consequences of the current approach, but the approach might be extended
+to avoid them:
+
+The interface to render a sidecar partial would be a method call, and depart from the usual 
+`render(*)` interface used in ActionView.
+
+The generated methods cannot have arguments with default values.
+
+The generated methods are public, and thus could be invoked by a third party.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -113,6 +113,10 @@ title: Changelog
 
     *Joel Hawksley*
 
+* Add support for multiple templates.
+
+    *Rob Sterner*, *Joel Hawksley*
+
 ## 2.36.0
 
 * Add `slot_type` helper method.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -36,6 +36,10 @@ title: Changelog
 
     *Hans Lemuet*
 
+* Add support for multiple templates.
+
+    *Rob Sterner*, *Joel Hawksley*
+
 ## 2.40.0
 
 * Replace antipatterns section in the documentation with best practices.
@@ -112,10 +116,6 @@ title: Changelog
 * Run benchmarks in CI.
 
     *Joel Hawksley*
-
-* Add support for multiple templates.
-
-    *Rob Sterner*, *Joel Hawksley*
 
 ## 2.36.0
 
--- a/docs/guide/templates.md
+++ b/docs/guide/templates.md
@@ -92,3 +92,36 @@ Component subclasses inherit the parent component's template if they don't defin
 class MyLinkComponent < LinkComponent
 end
 ```
+
+#### Multiple templates
+
+ViewComponents can render multiple templates defined in the sidecar directory:
+
+```
+app/components
+├── ...
+├── test_component.rb
+├── test_component
+|   ├── list.html.erb
+|   └── summary.html.erb
+├── ...
+```
+
+Templates are compiled to methods in the format `call_#{template_basename}`, which can then be called in the component.
+
+```ruby
+class TestComponent < ViewComponent::Base
+  def initialize(mode:)
+    @mode = mode
+  end
+
+  def call
+    case @mode
+    when :list
+      call_list
+    when :summary
+      call_summary
+    end
+  end
+end
+```
--- a/docs/guide/templates.md
+++ b/docs/guide/templates.md
@@ -93,11 +93,11 @@ class MyLinkComponent < LinkComponent
 end
 ```
 
-#### Multiple templates
+## Multiple templates with arguments
 
-ViewComponents can render multiple templates defined in the sidecar directory:
+ViewComponents can render multiple templates defined in the sidecar directory and send arguments to it:
 
-```
+```console
 app/components
 ├── ...
 ├── test_component.rb
@@ -107,7 +107,7 @@ app/components
 ├── ...
 ```
 
-Templates are compiled to methods in the format `call_#{template_basename}`, which can then be called in the component.
+Templates are compiled to methods in the format `render_#{template_basename}(locals = {})`, which can then be called in the component.
 
 ```ruby
 class TestComponent < ViewComponent::Base
@@ -118,9 +118,9 @@ class TestComponent < ViewComponent::Base
   def call
     case @mode
     when :list
-      call_list
+      render_list number: 1
     when :summary
-      call_summary
+      render_summary string: "foo"
     end
   end
 end
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -331,7 +331,7 @@ module ViewComponent
         # view files in the same directory as the component
         sidecar_files = Dir["#{directory}/#{component_name}.*{#{extensions}}"]
 
-        sidecar_directory_files = Dir["#{directory}/#{component_name}/#{filename}.*{#{extensions}}"]
+        sidecar_directory_files = Dir["#{directory}/#{component_name}/*.*{#{extensions}}"]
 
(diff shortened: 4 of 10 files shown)


adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,73 @@
...
+ActionView partials have the problem that their interface is not introspectable. Data
+may be passed into the partial via ivars or locals, and it is impossible to know
+which without actually opening up the file. Additionally, partials are globally
+invocable, thus making it difficult to detect if a given partial is in use or not,
+and who are its users.
+
+## Considered Options
+
+* Introduce component partials to components
+* Keep components as-is
```
fsateler, line 24: I wonder if we should add #451 as a considered option.

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,73 @@
...
+creating bindings and `instance_exec`. 
+Having explicit arguments makes the interface explicit.
+
+TODO: The following are consequences of the current approach, but the approach might be extended
+to avoid them:
+
+The interface to render a sidecar partial would be a method call, and depart from the usual 
+`render(*)` interface used in ActionView.
+
+The generated methods are only invokable via keyword arguments
```
fsateler, line 69: This is no longer true

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,73 @@
...
+
+The generated methods cannot have arguments with default values.
```
fsateler, line 71: No, but they can be simulated by the view.

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,73 @@
...
+
+The generated methods are public, and thus could be invoked by a third party.
```
fsateler, line 73: This could be changed, but not sure if it's worth it?

lib/view_component/compiler.rb
```diff
@@ -43,16 +43,29 @@ def compile(raise_errors: false)
...
+          end
 
         if component_class.instance_methods.include?(method_name.to_sym)
           component_class.send(:undef_method, method_name.to_sym)
         end
 
-        component_class.class_eval <<-RUBY, template[:path], -1
-          def #{method_name}
+        component_class.class_eval <<-RUBY, template[:path], -2
+          def #{method_name}(locals = {})
```
fsateler, line 63: I wonder if `locals = {}` or `**locals` should be preferred. If the first, we should invoke `ruby2_keywords`

test/view_component/view_component_test.rb
```diff
@@ -888,4 +899,17 @@ def test_output_postamble
 
     assert_text("Hello, World!")
   end
+
+  private
+
+  def modify_file(file, content)
```
fsateler, line 905: Is this actually necessary? It might be a remnant of a previous rebase.

fsateler (changes requested):
Thanks for picking this up @jsolas! I've added a few comments.

lib/view_component/compiler.rb
jsolas: I chose the second option

jsolas:
Hello @BlakeWilliams, @camertron, @joelhawksley, i tried to do what was discussed in https://github.com/github/view_component/issues/387#issuecomment-926668950, I would like to receive your opinions and see if there is a possibility of moving forward with this. Thanks.

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,88 @@
+# 1. Allow multiple templates
```
joelhawksley, line 1: We need a name for this feature, I think. What about:

```suggestion
# 3. Template parts
```

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,88 @@
+# 1. Allow multiple templates 
+
+Date: 2021-10-13
+
+## Status
+
+Proposed.
```
joelhawksley, line 7:
```suggestion
Accepted
```

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Context
+
+As components become larger (for example, because you are implementing a whole page), it becomes
+useful to be able to extract sections of the view to a different file. ActionView has
+partials, and ViewComponent lacks a similar mechanism.
```
joelhawksley, line 13:
```suggestion
As components become larger (for example, implementing a whole page), it can become useful to be able to extract sections of the view to a different file. ActionView has partials, and ViewComponent lacks a similar mechanism. 
```

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,88 @@
...
+
+As components become larger (for example, because you are implementing a whole page), it becomes
+useful to be able to extract sections of the view to a different file. ActionView has
+partials, and ViewComponent lacks a similar mechanism. 
+
+ActionView partials have the problem that their interface is not introspectable. Data
+may be passed into the partial via ivars or locals, and it is impossible to know
+which without actually opening up the file. Additionally, partials are globally
+invocable, thus making it difficult to detect if a given partial is in use or not,
+and who are its users.
```
joelhawksley, line 19:
```suggestion
ActionView partials have the problem that their interface is not introspectable. Data may be passed into the partial via ivars or locals, and it is impossible to know which without actually opening up the file. Additionally, partials are globally invocable, thus making it difficult to detect if a given partial is in use or not, and who are its users.
```

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Considered Options
+
+* Introduce component partials to components
+* Keep components as-is
+
+### Component partials
+
+Allow multiple ERB templates available within the component, and make it possible to
+invoke them from the main view.Templates are compiled to methods in the format `render_#{template_basename}(locals = {})`
```
joelhawksley, line 29:
```suggestion
Allow multiple ERB templates available within the component, and make it possible to invoke them from the main view. Templates are compiled to methods in the format `render_#{template_basename}(locals = {})`.
```

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,88 @@
...
+ActionView partials have the problem that their interface is not introspectable. Data
+may be passed into the partial via ivars or locals, and it is impossible to know
+which without actually opening up the file. Additionally, partials are globally
+invocable, thus making it difficult to detect if a given partial is in use or not,
+and who are its users.
+
+## Considered Options
+
+* Introduce component partials to components
+* Keep components as-is
```
joelhawksley, line 24: Let's just remove these lines in favor of the section headers below.

```suggestion
```

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,88 @@
...
+
+**Cons:**
+* Extracting a component results in more GC and intermediate objects.
+* Extracting a component may result in tightly coupled but split components.
+* Creates new public components thus expanding component library API surface.
+
+## Decision
+
+We will allow having multiple templates in the sidecar asset. Each asset will be compiled to
+it's own method `render_<template_name>`. I think it is simple, similar to rails and meets what is expected
```
joelhawksley, line 71:
```suggestion
We will allow having multiple templates in the sidecar asset. Each asset will be compiled to its own method `render_<template_name>`. I think it is simple, similar to rails and meets what is expected.
```

adr/0003-multiple-templates-per-component.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Consequences
+
+This implementation has better performance characteristics over both an extracted component
+and ActionView partials, because it avoids creating intermediate objects, and the overhead of
+creating bindings and `instance_exec`. 
+Having explicit arguments makes the interface explicit.
```
joelhawksley, line 78:
```suggestion
This implementation has better performance characteristics over both an extracted component and ActionView partials, because it avoids creating intermediate objects, and the overhead of creating bindings and `instance_exec`. Having explicit arguments makes the interface explicit.
```

adr/0003-multiple-templates-per-component.md
joelhawksley: I think it would be nice to introduce them as private to start.

docs/CHANGELOG.md
```diff
@@ -36,6 +36,10 @@ title: Changelog
 
     *Hans Lemuet*
 
+* Add support for multiple templates.
+
+    *Rob Sterner*, *Joel Hawksley*
```
joelhawksley, line 41:
```suggestion
    *Jose Solás Moreno*, *Felipe Sateler*, *Rob Sterner*, *Joel Hawksley*
```

docs/guide/templates.md
```diff
@@ -92,3 +92,36 @@ Component subclasses inherit the parent component's template if they don't defin
 class MyLinkComponent < LinkComponent
 end
 ```
+
+## Multiple templates with arguments
```
joelhawksley, line 96:
```suggestion
## Template parts
```

docs/guide/templates.md
```diff
@@ -92,3 +92,36 @@ Component subclasses inherit the parent component's template if they don't defin
...
+
+ViewComponents can render multiple templates defined in the sidecar directory and send arguments to it:
```
joelhawksley, line 98:
```suggestion
ViewComponents can render multiple templates defined in the sidecar directory:
```

docs/guide/templates.md
```diff
@@ -92,3 +92,36 @@ Component subclasses inherit the parent component's template if they don't defin
...
+app/components
+├── ...
+├── test_component.rb
+├── test_component
+|   ├── list.html.erb
+|   └── summary.html.erb
+├── ...
+```
+
+Templates are compiled to methods in the format `render_#{template_basename}(locals = {})`, which can then be called in the component.
```
joelhawksley, line 110:
```suggestion
Template parts can then be rendered by calling `render_#{template_basename}`. Template parts accept a single positional argument that is accessible in the template as `locals`, defaulting to an empty hash if no argument is passed.
```

lib/view_component/base.rb
```diff
@@ -331,7 +331,7 @@ def _sidecar_files(extensions)
         # view files in the same directory as the component
         sidecar_files = Dir["#{directory}/#{component_name}.*{#{extensions}}"]
 
-        sidecar_directory_files = Dir["#{directory}/#{component_name}/#{filename}.*{#{extensions}}"]
+        sidecar_directory_files = Dir["#{directory}/#{component_name}/*.*{#{extensions}}"]
```
joelhawksley, line 334: What happens if there is a slot/subcomponent with a template in the Sidecar directory? How can we prevent/manage conflicts between the two?

joelhawksley (commented):
Looking good! I think we're on the right track here so I've marked the ADR as accepted.

fsateler:
@joelhawksley I feel this proposal is not very railsy. In particular, the loop in the initializer feels like boilerplate. I don't really like the API.

joelhawksley:
@fsateler I think it would also be possible to use the plural slot setter if you didn't want to decompose `items`, like this:

```ruby
# table_component.rb
  renders_many :rows, private: true

  def initialize(items)
    rows(items)
  end
end
```

The difference between ☝🏻 and the proposal in this PR is that it remains inside of the existing Slots API, which I think would help reduce potential confusion around its introduction and use.

joelmoss:
Has this moved along at all recently? I have a real need to be able to dynamically determine the template to render within a component, and this PR seems to enable that. Would be great to get something out, and I would be happy to help.

fsateler:
Not a problem on my side! If this helps things move forward, please go ahead!

fsateler:
Hi! I've been thinking about @joelhawksley 's proposal for a new API. I don't think it's going to work, not without a lot of refactoring:

1. `with_<slotname>` creates a `SlotV2` object for each item in the collection. 
2. This creation is eager, because at render time we iterate through  `@_set_slots[slot_name]`, which must have been prepopulated by `with_<slotname>`. This means the `SlotV2` object creation is eagerly done.

This means we will create an intermediate array of intermediate objects, which is what we wanted to avoid.

Moreover, we *still* would need a lower-level api for actually rendering the template (which would be used by the slot interface).   Please lets move forward with this as the lower-level API :pray:

joelhawksley:
Closing per https://github.com/ViewComponent/view_component/pull/451#issuecomment-2613411129

PR #1098, opened 2021-10-14
Support component generator for tailwindcss-rails

dixpac:
This commit adds support for [tailwindcss-rails](https://github.com/rails/tailwindcss-rails) engine.

This is my first contribution so I hope I did this correctly.
I've spinend up demo app with `view_component` and `tailwindcss-rails`and it works.
But, I may have missed something :)

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -6,16 +6,20 @@ end
 
 appraise "rails-6.0" do
   gem "rails", "~> 6.0.0"
+  gem "tailwindcss-rails", "~> 2.0"
 end
 
 appraise "rails-6.1" do
   gem "rails", "~> 6.1.0"
+  gem "tailwindcss-rails", "~> 2.0"
 end
 
 appraise "rails-7.0" do
   gem "rails", "~> 7.0.0"
+  gem "tailwindcss-rails", "~> 2.0"
 end
 
 appraise "rails-head" do
   gem "rails", github: "rails/rails", branch: "main"
+  gem "tailwindcss-rails", "~> 2.0"
 end
--- a/gemfiles/rails_6.0.gemfile
+++ b/gemfiles/rails_6.0.gemfile
@@ -4,5 +4,6 @@ source "https://rubygems.org"
 
 gem "capybara", "~> 3"
 gem "rails", "~> 6.0.0"
+gem "tailwindcss-rails", "~> 2.0"
 
 gemspec path: "../"
--- a/lib/rails/generators/tailwindcss/component_generator.rb
+++ b/lib/rails/generators/tailwindcss/component_generator.rb
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+require "rails/generators/erb/component_generator"
+
+module Tailwindcss
+  module Generators
+    class ComponentGenerator < Erb::Generators::ComponentGenerator
+      source_root File.expand_path("templates", __dir__)
+    end
+  end
+end
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add generators to support `tailwindcss-rails`.
+
+    *Dino Maric*, *Hans Lemuet*
+
 * Add a namespaced component example to docs.
 
     *Hans Lemuet*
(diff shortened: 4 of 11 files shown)


joelhawksley:
@dixpac thank you for your contribution! @Spone has handled code reviews for generators in the past, so I've passed this one to him ❤️

Spone:
Hi @dixpac! Thanks for your PR! I'm not much of a Tailwind user, can you elaborate on what's the expected behavior for the generator? I read the diff quickly and I'm not sure I understand how this differs from the regular generator?

toddsiegel:
This would certainly be useful for those of us using tailwindcss-rails. 

For others using view_component and tailwindcss-rails this workaround will keep you going.

`$ bin/rails g component <Name> --template-engine=erb`

lib/rails/generators/tailwindcss/component_generator.rb
```diff
@@ -0,0 +1,24 @@
...
+      def engine_name
+        "erb"
+      end
+
+      def copy_view_file
+        super
+      end
+    end
+  end
+end
```
Spone, line 24: The tests pass when I remove this (everything I removed is inherited from `Erb::Generators::ComponentGenerator`).

```suggestion
module Tailwindcss
  module Generators
    class ComponentGenerator < Erb::Generators::ComponentGenerator
      source_root File.expand_path("templates", __dir__)
    end
  end
end
```

I'd rather keep the generator as lean as possible if its behavior is the same as the Erb generator. What do you think?

Spone:
@dixpac can you please rebase to resolve the conflicts, and try to fix the tests that do not pass for Rails 5?

lib/rails/generators/tailwindcss/component_generator.rb
dixpac: Yup, makes perfect sense

dixpac:
@Spone sure I will try. Rails 5 will be a bit harder since tw-rails is rails >= 6, but we will make it work somehow :)

VSPPedro:
What do you think about using [Appraisal](https://github.com/thoughtbot/appraisal)?

You will be able to install tailwindcss-rails only for rails 6.

What do you think?

joelhawksley:
👋🏻 I think it's worth pursuing conditionally adding this support only in Rails >= 6, as long as we make it clear in the documentation. The Rails asset delivery story has certainly changed a bit from 5>6>7 and I think it's not unreasonable to expect folks to be reasonably up to date to use this functionality.

Spone:
I tried [this](https://stackoverflow.com/a/5177807/1789900) without success (works locally, but not in CI).

@joelhawksley would you be open to adding Appraisal? I think it's the cleanest way to implement conditional dependencies. It would also make locally running tests for all versions of Rails much easier.

joelhawksley:
@Spone I'm totally fine with adding Appraisal ❤️

pboling:
Worth noting that there may be complixity getting Appraisal to fully work with GithubActions: https://github.com/thoughtbot/appraisal/issues/173

Spone:
@pboling I haven't looked into it in detail yet, but I know we're using Appraisal + Github Actions to test https://github.com/pantographe/view_component-form without issues.

coder2000:
Is still useful with the new tailwindcss-rails?

dixpac:
@coder2000 yes. If you want to "attack" adding Appraisal that would be awesome

Spone:
Hi @dixpac, Appraisal is now setup and I used it in your PR to manage the conditional dependency to `tailwindcss-rails`.

Appraisals
```diff
@@ -6,16 +6,20 @@ end
 
 appraise "rails-6.0" do
   gem "rails", "~> 6.0.0"
+  gem "tailwindcss-rails", "~> 0.5.1"
```
Spone, line 9: @dixpac I've seen that `tailwindcss-rails` is now in version [2.0.3](https://github.com/rails/tailwindcss-rails/releases). Do you think we should bump the required version here (and below)?

dixpac: Yup, we should bump to >=1. You can also go with ~> 2 because I'm not sure that anyone uses version 1 anymore. 
But, for generators both 1 and 2 are the same.

dixpac:
👏🏼  @Spone thank you!

dixpac:
@Spone version bumped, conflicts fixed, squashed and I've added you as a co-author :)

gemfiles/rails_6.0.gemfile
```diff
@@ -4,5 +4,6 @@ source "https://rubygems.org"
 
 gem "capybara", "~> 3"
 gem "rails", "~> 6.0.0"
+gem "tailwindcss-rails", "~> 2.0"
```
dixpac, line 7: I think 2 as a dependency is what we are looking for, since `tailwindcss-rails` version 1 was not complete and it was missing some of the major tailwindcss parts. That is solved with version 2, so I can't really see anyone using version 1 :)

Spone (approved):

PR #1100, opened 2021-10-19
Test lamda slots with no arguments

manuelpuyol:

### DIFF
--- a/test/view_component/slotable_v2_test.rb
+++ b/test/view_component/slotable_v2_test.rb
@@ -383,6 +383,14 @@ class SlotsV2sTest < ViewComponent::TestCase
     assert_includes error.message, "It looks like a block was provided after calling"
   end
 
+  def test_renders_lamda_slot_with_no_args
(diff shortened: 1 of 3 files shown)


test/view_component/slotable_v2_test.rb
```diff
@@ -383,6 +383,14 @@ def test_raises_if_using_both_block_content_and_with_content
     assert_includes error.message, "It looks like a block was provided after calling"
   end
 
+  def test_renders_lamda_slot_with_no_args
+    render_inline(SlotsV2WithEmptyLambdaComponent.new) do |c|
+      c.item { "Item 1" }
+      c.item { "Item 2" }
+      c.item { "Item 3" }
+    end
```
joelhawksley, line 391: Might we add an assertion for what should be rendered?

joelhawksley (commented):
Thanks for filing this case @manuelpuyol! How do you think we should address it?

PR #1101, opened 2021-10-19
Forward keyword arguments from slot wrapper to component instance

camertron:
### Summary

It looks like `SlotV2`'s `method_missing` method doesn't forward keyword arguments to the component instance. This PR fixes that.

### DIFF
--- a/lib/view_component/slot_v2.rb
+++ b/lib/view_component/slot_v2.rb
@@ -86,8 +86,8 @@ module ViewComponent
     #   end
     # end
     #
-    def method_missing(symbol, *args, &block)
-      @__vc_component_instance.public_send(symbol, *args, &block)
+    def method_missing(symbol, *args, **kwargs, &block)
+      @__vc_component_instance.public_send(symbol, *args, **kwargs, &block)
     end
 
     def html_safe?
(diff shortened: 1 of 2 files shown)


BlakeWilliams (approved):
LGTM!

camertron:
@joelhawksley hmm... I think so, lemme give it a shot.

lib/view_component/slot_v2.rb
```diff
@@ -86,8 +86,14 @@ def to_s
     #   end
     # end
     #
-    def method_missing(symbol, *args, &block)
-      @__vc_component_instance.public_send(symbol, *args, &block)
+    if RUBY_VERSION >= "2.7.0"
```
joelhawksley, line 89: Do you think we need to document this difference or is this assumed due to differences between Ruby versions?

BlakeWilliams: A bit late to the party, but I think we could use [`ruby2_keywords`](https://ruby-doc.org/core-2.7.3/Module.html#ruby2_keywords-method) here.

I think that would be nice to avoid the `RUBY_VERSION` check.  I believe all that's needed is to change the signature to only accept args, so `def method_missing(symbol, *args, &block)` then below it call `ruby2_keywords(:method_missing) if respond_to?(:ruby2_keywords, true)`.

Thoughts?

camertron: Oh wow I didn't know about `ruby2_keywords` (or if I did, I forgot about it!) Super handy!

PR #1107, opened 2021-10-25
Extract ViewComponent::Core class

BlakeWilliams:
This extracts a new class, `ViewComponent::Core` which implements the
minimum amount of functionality to have a functioning component library.

The difference between `ViewComponent::Core` and `ViewComponent::Base`
is that Base includes extra functionality and compatibility with Rails,
such as slots, built-in route helpers, access to `request` and
`controller`, etc.

This new `Core` class is marked as private and should still be
considered experimental.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -57,7 +57,7 @@ namespace :docs do
       meths.
       select do |method|
         !method.tag(:private) &&
-          method.path.include?("ViewComponent::Base") &&
+          method.path.match?("ViewComponent::(Base|Core)") &&
           method.visibility == :public
       end.sort_by { |method| method[:name] }
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Extract `ViewComponent::Core` class from `ViewComponent::Base`.
+
+    *Blake Williams*
+
 * Fix bug where calling lambda slots without arguments would break in Ruby < 2.7.
 
     *Manuel Puyol*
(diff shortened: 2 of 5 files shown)


joelhawksley:
@jaredcwhite I'd love to see us add some sort of test(s) that put us in the right direction towards better supporting Bridgetown with this change. Might you be up for suggesting some?

jaredcwhite:
I'm revisiting this now after some time has passed, and I think what I'm starting to wonder is if there could be some way to have a framework-agnostic component base class (aka one that's not directly subclassing `ActionView::Base`). The particular view layer used then becomes an implementation detail, pluggable and customizable across different frameworks and not just within Rails.

Alternatively—and this is a much broader topic I realize—is to somehow come up with an external code interface which defines the basics of a "view component" (whether that's actual code in the form of a mixin or something more theoretical like a spec), and then ensure the `ViewComponent::Core/Base` conforms to that interface.

The rationale for all this being in the Bridgetown framework, we have a `Bridgetown::Component` base class which is unapologetically inspired by ViewComponent and shares some common semantics, and rather than that being purely "by happenstance" because we're fans, it'd be awesome for those semantics to be enforced to a certain degree by a common interface recognized in the Ruby community as a whole. The "Rack" of component-based view architecture if you will.

Thoughts?

joelhawksley:
@jaredcwhite I like it! We've toyed with removing our dependency on `AV::Base` anyways, and I think @camertron has a branch that is _nearly_ passing without it.

joelhawksley:
Closing as stale.

Issue #1112, opened 2021-10-29
Implicit dependency on `::ActionDispatch::Static` breaks Rails environment


grekko:
## Summary

ViewComponent recently added support to view the source code of a preview (see https://github.com/github/view_component/pull/667).

For this ViewComponent bundles `prism` which is served via a `ActionDispatch::Static`-middleware, inserted here https://github.com/github/view_component/blob/928f86c6c6ce048d791f2f5874f662780b3bed7e/lib/view_component/engine.rb#L106

The call to [`app.middleware.insert_before`](https://github.com/rails/rails/blob/83217025a171593547d1268651b446d3533e2019/actionpack/lib/action_dispatch/middleware/stack.rb#L100-L103) will fail though, if there is no `ActionDispatch::Static` class already defined in the Rails middleware stack.

E.g. our Rails staging environment has configured `config.public_file_server.enabled = false`, which tells Rails to [not add](https://github.com/rails/rails/blob/90357af08048ef5076730505f6e7b14a81f33d0c/railties/lib/rails/application/default_middleware_stack.rb#L25-L29) `ActionDispatch::Static` to the default middleware stack.

Trying to render the middleware stack will cause an error:

```sh
RAILS_ENV=staging bundle exec rails middleware
No such middleware to insert before: ActionDispatch::Static
/var/www/path-to-app/shared/bundle/ruby/2.7.0/gems/actionpack-5.2.6/lib/action_dispatch/middleware/stack.rb:108:in `assert_index'
/var/www/path-to-app/shared/bundle/ruby/2.7.0/gems/actionpack-5.2.6/lib/action_dispatch/middleware/stack.rb:75:in `insert'
/var/www/path-to-app/shared/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/configuration.rb:71:in `block in merge_into'
/var/www/path-to-app/shared/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/configuration.rb:70:in `each'
/var/www/path-to-app/shared/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/configuration.rb:70:in `merge_into'
/var/www/path-to-app/shared/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/engine.rb:509:in `block in app'
/var/www/path-to-app/shared/bundle/ruby/2.7.0/gems/railties-5.2.6/lib/rails/engine.rb:506:in `synchronize'
[...]

PR #1116, opened 2021-11-01
engine.rb: Change ActionDispatch::MiddlewareStack#insert_before -> #use

grekko:
### Summary

See related issue #1112 

The call to `ActionDispatch::MiddlewareStack#insert_before` expected an instance of `ActionDispatch::Static` being present in the middleware stack.

Calling `ActionDispatch::MiddlewareStack#use` instead will instead append `ActionDispatch::Static`-middleware to the "end" of the middleware stack without any dependency.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -5,6 +5,12 @@ title: Changelog
 
 # Changelog
 
+## main
+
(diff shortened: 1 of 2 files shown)


joelhawksley (approved):

joelhawksley:
Thanks for the fix!

PR #1117, opened 2021-11-01
Require engine automatically

camertron:
### Summary

Currently anyone using the framework has to require `view_component/engine` manually, usually by putting `gem "view_component", require: "view_component/engine"` in the Gemfile or requiring it specifically in application.rb. Forgetting to do so can lead to weird reloading problems and other subtle issues as evidenced in #560.

This PR adds the relevant `require` to avoid these issues.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,10 @@ title: Changelog
 
     *Gregory Igelmund*
 
+* Require `view_component/engine` automatically
+
+    *Cameron Dutro*
+
 ## 2.42.0
 
 * Add logo files and page to docs.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,7 +11,7 @@ title: Changelog
 
     *Gregory Igelmund*
 
-* Require `view_component/engine` automatically
+* Require `view_component/engine` automatically.
 
     *Cameron Dutro*
 
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -19,7 +19,7 @@ nav_order: 1
 In `Gemfile`, add:
 
 ```ruby
-gem "view_component", require: "view_component/engine"
+gem "view_component"
 ```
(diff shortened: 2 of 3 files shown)


docs/CHANGELOG.md
```diff
@@ -5,6 +5,12 @@ title: Changelog
 
 # Changelog
 
+## main
+
+* Require `view_component/engine` automatically
```
```suggestion
* Require `view_component/engine` automatically.
```

joelhawksley (approved):
Do any of the docs need to be updated to reflect this change?

Spone:
That's a great improvement for newcomers! :clap: @camertron

maxshelley:
@camertron @joelhawksley Hi both, a massive thank you for this change 🙌 Just a quick heads up that I bumped into an issue which is that it seems that merging this PR updated the docs i.e it removed the `, require: "view_component/engine"`,  however the corresponding change hasn't yet made it into the latest version. So when I ran `bundle install` and rendered my first component, I was confused as to why it wasn't then updating.

Not sure I have a suggesting for how situations like that can be mitigated, but could potentially be confusing for newcomers in the gap between PRs being merged and a new version being released.

Was livestreaming when I bumped into this, so if you'd like to see me writing this comment (which is super meta), you can do [here](https://youtu.be/g5Qi2A3wQxc).

Spone:
@maxshelley I guess we should either add a disclaimer to the docs or release a new version now (or both?)

@camertron @joelhawksley by the way, what about adding an explicit and more visible mention at the top of the changelog that the engine `require` can be removed from the Gemfile after upgrading?

joelhawksley:
@maxshelley thanks for flagging this. I think there is definitely some room for us to improve our release process. 

In the meantime, I've published v2.43.0 that includes the change from this PR. (It turns out we hadn't shipped the release notes for v2.42.0 either 😬)

camertron:
@maxshelley oops that's my bad, I totally thought we had released this in 2.42 😅

PR #1121, opened 2021-11-04
Add failing test showing ERB extra spaces

manuelpuyol:

### DIFF
--- a/test/sandbox/app/components/content_component.html.erb
+++ b/test/sandbox/app/components/content_component.html.erb
@@ -1,0 +1,1 @@
+<div><%= content %></div>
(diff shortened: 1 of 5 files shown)


boardfish:
Hey @manuelpuyol! This PR seems linked to what I experienced in #913 - you may want to check out that thread. The solution I put together for that trims trailing whitespace from rendered ERB templates, but what you're experiencing here with ViewComponent is internal to your rendered component itself, and probably needs its own solution.

[As @joelhawksley mentioned on the linked PR](https://github.com/primer/view_components/pull/897#discussion_r758774570), I agree that the `trimmed_content` helper you put together feels like a framework feature and would do well to resolve this. Maybe it's a render mode that, by default, follows the `strip_trailing_whitespace` config option from Action View?

PR #1141, opened 2021-11-17
Add require method_source to avoid `NoMethodError` in `ViewComponent#previews`

yhirano55:
### Summary

If options.show_previews_source is enabled, we need to require "method_source" in engine.

If not, it cause `NoMethodError` in `ViewComponent#previews`.


docs/CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ title: Changelog
 
     *James Hart*
 
+* Add `require "method_source"` if it options.show_previews_source is enabled
```
joelhawksley, line 22:
```suggestion
* Add `require "method_source"` if it options.show_previews_source is enabled.
```

joelhawksley (approved):

yhirano55:
[additional info]

In case of using test_unit, method source is called from `Rails::TestUnit::Runner` in railties.

https://github.com/rails/rails/blob/1a240cf83e03527567dea68f3c6bea5301d90113/railties/lib/rails/test_unit/runner.rb#L4

But if we use RSpec, it'll not be called from anywhere. So we needed this PR's change. Thanks.

PR #1147, opened 2021-11-18
Clear cache in MethodSource to apply the change of preview code without app server restart

yhirano55:
### Summary

Currently, we can show preview codes in component previews if it enables `show_previews_code` option.

However, the change of preview class's code is not applied to screen without app server restart.

It's not convenient, isn't it?

So I configured to clear cache in `MethodSource` per request using `to_prepare` block.

### Usecase

For example, suppose you have a code like the following:

```ruby
class FooComponentPreview < ViewComponent::Preview
  def default
    render(FooComponent.new)
  end
end
```

In preview, it will display like this:

Screenshot from 2021-11-19 00-28-29

Then, change the code while the server is running. 

```diff
  class FooComponentPreview < ViewComponent::Preview
    def default
+     # This is comment
      render(FooComponent.new)
    end
  end
```

Unfortunately, this change will not apply until we restart the server.

With this fix, the code changes will be applied to the screen without restarting the server.

Screenshot from 2021-11-19 00-39-05

Thanks :pray:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -39,6 +39,10 @@ title: Changelog
 
     *Yoshiyuki Hirano*
 
+* Clear cache in MethodSource to apply the change odf preview code without app server restart.
+
+    *Yoshiyuki Hirano*
+
 ## 2.43.1
 
 * Remove unnecessary call to `ruby2_keywords` for polymorphic slot getters.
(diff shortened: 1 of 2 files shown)


joelhawksley (approved):
Fantastic! Happy to land this as-is. Would you be interested in opening a PR for a test for this change?

Spone (approved):

PR #1152, opened 2021-11-19
Remove `require "rails/generators/test_case"` in generator tests

yhirano55:
### Summary

`require "rails/generators/test_case"` is called from rails/test_help on railtie gem. So we don't need to call them on each test files, removed them.

https://github.com/rails/rails/blob/5-2-stable/railties/lib/rails/test_help.rb#L11

---

In the process of following up on #1147, I noticed it while reading the code and opened a pull request. If you feel this is just like cosmetic change, would you please close it :pray:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Remove `require "rails/generators/test_case"` in generator tests
+
+    *Yoshiyuki Hirano*
+
 ## 2.44.0
 
 * Rename internal accessor to use private naming.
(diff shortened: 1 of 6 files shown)


Spone (approved):
Nice catch!

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Remove `require "rails/generators/test_case"` in generator tests
```
```suggestion
* Remove `require "rails/generators/test_case"` in generator tests.
```

joelhawksley (approved):

Issue #869, opened 2021-04-20
GitHub Actions tests sometimes fail with the error "undefined local variable or method `call'" in a view component


adrianthedev:
Hello,

I'm using `view_component` in [Avo](https://github.com/avo-hq/avo) quite a lot (about 95 components). Sometimes the tests fail when being run in GitHub Actions.
The error is `ActionView::Template::Error:
       undefined local variable or method 'call' for #<Avo::PanelComponent:0x000055865888ba10>`.
This error is never raised in my local environment.
**Most times I hit "Re-run jobs" and the tests pass.**

 - [The test](https://github.com/avo-hq/avo/pull/365/checks?check_run_id=2391333278) in a PR made especially for this issue. Ruby 2.6, Rails 6.0
 - [Another test](https://github.com/avo-hq/avo/pull/362/checks?check_run_id=2391322152) Ruby 2.6, Rails 6.0
 - [Another test](https://github.com/avo-hq/avo/runs/2347770232) Ruby 2.6, Rails 6.1
 - [The actual PR](https://github.com/avo-hq/avo/pull/365).

It seems it only happens for `PanelComponent`.

 - [`PanelComponent`](https://github.com/avo-hq/avo/blob/test/view-component-call-issue/app/components/avo/panel_component.rb) source code
 - [`ResourceIndexComponent`](https://github.com/avo-hq/avo/blob/test/view-component-call-issue/app/components/avo/views/resource_index_component.html.erb#L1) the place where the error is raised.

### Steps to reproduce

I can't reproduce it locally. It only happened on GA and sporadically.

### Expected behavior
The tests should pass without raising this error.

### Actual behavior
Raises this error sporadically. It's not raised every time. 

### System configuration
**Rails version**:
Tested with 6.0 and 6.1

**Ruby version**:
Tested with 2.6 and 2.7

**Gem version**:
view_component (2.28.0)

adrianthedev:
This is still an issue on my end. 
Did anyone encounter this before? Is there anyone that can point me in a direction to troubleshoot it?

BlakeWilliams:
👋 @joelhawksley and I briefly looked at that component and nothing stood out. Since it's happening sporadically it's difficult to say what the problem might be.

Have you ruled out the test order potentially causing the issue? If you re-run the test-suite using the same seed, does that reproduce the issue?

adrianthedev:
[...]

PR #1161, opened 2021-11-23
Compiler redefines .call in thread-safe way

horiaradu:
### Summary

Redefine component's `call` method in a thread-safe way, by making `Compiler` acquire a lock (Monitor), per each component class.

fixes https://github.com/github/view_component/issues/869


lib/view_component/compiler.rb
```diff
@@ -79,11 +81,13 @@ def define_render_template_for
 
       component_class.class_eval <<-RUBY, __FILE__, __LINE__ + 1
         def render_template_for(variant = nil)
-          if variant.nil?
-            call
-          #{variant_elsifs}
-          else
-            call
+          self.class.lock.synchronize do
```
BlakeWilliams, line 84: We likely don't want to use the mutex in the render path for performance reasons, especially since this is usually only an issue in the test environment.

Happy to take a look later, but I'm curious what breaks without this code.

BlakeWilliams (commented):
Thanks for taking this on! I had a question about performance in the render path but this seems like the right direction.

lib/view_component/compiler.rb
horiaradu: Without adding the thread-safety in render, the build was failing, see [here](https://github.com/github/view_component/runs/4299758830?check_suite_focus=true). And I believe it's right because, if we only add the lock acquisition when we redefine `call`, then, while the code which redefines `call` is running, we can still have `render` calls being made and those can fail.

If performance is an issue, would maybe moving the compilation from render time to loading time be a better idea?

BlakeWilliams: > If performance is an issue, would maybe moving the compilation from render time to loading time be a better idea?

I had a similar thought and have been thinking on some of the trade-offs. The big one that comes to mind is that we'd need to somehow determine which components to compile, otherwise we may unnecessarily slow down test suite boot time. That wouldn't be great for cases where you want to run a single test or subset of tests.

horiaradu: > > If performance is an issue, would maybe moving the compilation from render time to loading time be a better idea?
> 
> I had a similar thought and have been thinking on some of the trade-offs. The big one that comes to mind is that we'd need to somehow determine which components to compile, otherwise we may unnecessarily slow down test suite boot time. That wouldn't be great for cases where you want to run a single test or subset of tests.

Makes sense...

I see that people start to mention issues appearing in `development` mode as well. Does it make sense to put this locking mechanism behind a check for `env == test || development`?

horiaradu: Also, the reason why I switched from `Mutex` to `Monitor` is [this](https://github.com/github/view_component/runs/4300334626?check_suite_focus=true) failing build.

lib/view_component/base.rb
```diff
@@ -31,6 +31,10 @@ class Base < ActionView::Base
 
     attr_accessor :__vc_original_view_context
 
+    # Lock required to be obtained before compiling the component
+    class_attribute :lock
+    self.lock = Monitor.new
```
BlakeWilliams, line 36:
```suggestion
    class_attribute :__vc_compiler_lock
    self.__vc_compiler_lock = Monitor.new
```

Can we follow the "internal attribute" naming convention here?

BlakeWilliams: Alternative thought, would it be possible to put the monitor in the compiler instead of on this specific component instance? Each component should have a single compiler instance, as defined on https://github.com/github/view_component/pull/1161/files#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9eR412

lib/view_component/compiler.rb
```diff
@@ -40,28 +40,30 @@ def compile(raise_errors: false)
         component_class.validate_collection_parameter!
       end
 
-      templates.each do |template|
-        # Remove existing compiled template methods,
-        # as Ruby warns when redefining a method.
-        method_name = call_method_name(template[:variant])
+      component_class.lock.synchronize do
```
BlakeWilliams, line 43: Can we move this to be just under the `return if compiled?` check to cover the entire method? 

Additionally, I think we may want to implement something like `CompilerCache.invalidate_component!(component_klass)` so that any other callers will block on the monitor instead of trying to continue rendering due to the `return if compiled?` call on the first line of this method.

lib/view_component/compiler.rb
BlakeWilliams: Using a `Monitor` makes sense to me. 👍 

I just made a few suggestions in the latest review round. I don't think it will make things 100% thread safe, but should minimize the risk and at least allow us to remove the `synchronize` here.

Similar to your env idea, I'm wondering if we need to go a level higher and think about introducing different modes to the compiler. A consistent compiler, for dev + test, and a fast compiler, for production. I don't think that idea should block this PR, but we do want to keep the performance benefits we see in production while also making dev+test consistent in threaded environments.

lib/view_component/base.rb
horiaradu: I will look into it. Thanks a lot for the suggestions!

lib/view_component/compiler.rb
horiaradu: Good idea. Will do it!

horiaradu: Done :)

lib/view_component/base.rb
horiaradu: Awesome idea! This simplified things a bit, since I didn't have to take care of ensuring that each subcass of `ViewComponent::Base` had its own lock instance. Didn't realise that each component class had its own compiler instance!

horiaradu:
I've added your suggestions. Thanks a lot!

I've removed the lock from `render_template_for` and I will see what the tests say.

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Compiler redefines `call` in a thread-safe way.
```
BlakeWilliams, line 10:
```suggestion
* Add thread safety to the compiler
```

BlakeWilliams (approved):

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add thread safety to the compiler
```
BlakeWilliams, line 10:
```suggestion
* Add thread safety to the compiler.
```

PR #1167, opened 2021-11-24
Remove `config/boot.rb` in Sandbox

yhirano55:
### Summary

This is a tiny change.

The `view_component/lib` load path has already appended in [gemspec](https://github.com/github/view_component/blob/da341f82a18f48d252a0fedcd1b8310f05ad5c75/view_component.gemspec#L5), and rubygems has already called in [bundler/rubygems_integration](https://github.com/rubygems/rubygems/blob/master/bundler/lib/bundler/rubygems_integration.rb#L3).

Therefore, the sandbox app doesn't need to require boot file, and removed it.

### For maintainers

Currently, it is difficult to rewrite and test the configuration for each test case. I'm preparing to make it possible (I'm referring to [railties test](https://github.com/rails/rails/blob/main/railties/test/isolation/abstract_unit.rb), but it's a bit difficult.) In the process, I noticed this case, so I opened a pull request.

If you don't need it, please feel free to close it. Thanks :pray:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Remove `config/boot.rb` in Sandbox.
+
+    *Yoshiyuki Hirano*
+
 ## 2.45.0
 
 * Remove internal APIs from API documentation, fix link to license.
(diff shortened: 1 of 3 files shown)


yhirano55:
The pvc workflow failed. Since this change is only for the sandbox app, I don't think it will affect primer :thinking:  However, it would not be good if we broke it, so please retry or close it as it is a trivial change :sweat_smile:

Issue #1134, opened 2021-11-12
ViewComponent logo hard to read in dark mode


joelhawksley:
## Feature request

@dylanatsmith do you think we could come up with a text color that passes contrast checks in both dark and light modes on GitHub?

### Motivation

It's hard to read the ViewComponent text in the README in dark mode:

<img width="737" alt="Screen Shot 2021-11-12 at 2 40 29 PM" src="https://user-images.githubusercontent.com/1940294/141537966-17d44f98-2416-4499-a3b3-cf27addccb49.png">

dylanatsmith:
Oof, good catch.

I know we discussed single-colour logo options (e.g. all red) and initially ruled it out because we liked the red + black and red + white combos. But that would solve for this and probably look better than red + medium grey or whatever.

From the original colour combination doc:

![image](https://user-images.githubusercontent.com/6905903/141973206-92cf4c43-d39d-46b0-841a-c3ecf96293fd.png)

boardfish:
I'm a sucker for red background and white text, but white background, red logo, black text is a close second.

joelhawksley:
@dylanatsmith I think the red background, white text option looks sharp. Would you mind PR'ing an update to update the README logo to use that look?

dylanatsmith:
Here's another potential option: [one SVG with a media query inside of it](https://p-76F1Jq.t4.n0.cdn.getcloudapp.com/items/NQu7WkE1/2d0bfcd9-916a-43ad-bb05-bf694fd26ce7.gif) that dynamically changes text colour depending on OS settings.

The downside of this is that there's no guarantee the OS settings are in sync with someone's GitHub user preference.

Spone:
> The downside of this is that there's no guarantee the OS settings are in sync with someone's GitHub user preference.

That would be awesome to encapsulate the logic in the SVG :)

Related community request: https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981 (without a working solution yet)

PR #1169, opened 2021-11-24
Readme logo update

dylanatsmith:
### Summary

Takes advantage of GitHub’s newly supported functionality to show/hide images in Markdown depending on which GH theme is served as per user settings.

* [Feature documentaion](https://docs.github.com/en/github/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to)
* [Community request thread](https://github.community/t/support-theme-context-for-images-in-light-vs-dark-mode/147981/77)

Further optimises the solution to https://github.com/github/view_component/issues/1134 shipped in https://github.com/github/view_component/pull/1143.

Note: There doesn't seem to be a way, at least in GFM, to specify image size. So using exising assets they'll be full-width unless we create smaller ones for this purpose.


Spone:
Great that this feature shipped!

> Note: There doesn't seem to be a way, at least in GFM, to specify image size. So using exising assets they'll be full-width unless we create smaller ones for this purpose.

It's a bit large for the README, don't you think?

image

What about changing the `width="735" height="84" viewBox="0 0 735 84"` values in the existing assets? Since it's SVG there will be no quality loss if people want to use them at larger size in other contexts.

dylanatsmith:
It's nearly comical on desktop for sure.

Changing the canonical SVG is an option. I don’t _think_ there’d be any side effects. I don’t love the idea of adding those attributes to satisfy this one use case, just on principle, but I’m not opposed.

I think we do one of the following:
1. Leave it full-width in the readme
2. Change the size of the canonical SVG
3. Add additional, readme-specific assets like we’re currently using

I’m leaning toward option 3 at the moment but I’m open.

BlakeWilliams:
I'm 👍 on option 3, using assets specifically for the README makes sense to me.

joelhawksley:
I'm fine with option 3 as well 👍🏻

joelhawksley (approved):

PR #1175, opened 2021-12-01
Add with_SLOT_NAME_content helper for slots

willcosgrove:
### Summary

This PR adds a new method for every `renders_one` slot that lets you set the content of the slot directly on the parent instance.

```rb
  renders_one :header
  renders_one :subtitle
end
```

```erb
<%= render BlogComponent.new.with_header_content("My blog").with_subtitle_content("Ramblings of a ViewComponent developer") %>
```

### Other Information

This API was suggested by @joelhawksley after looking at a component of mine that was attempting to allow setting the content of two slots both via arguments to the component initializer, and via a block.

```rb
class DisabledComponent < ApplicationComponent
  renders_one :enabled
  renders_one :disabled

  def initialize(enabled: nil, disabled: nil)
    self.enabled { enabled } if enabled
    self.disabled { disabled } if disabled
  end
end
```

```erb
<!-- disabled_component.html.erb -->
<span class="group-disabled:hidden contents"><%= enabled %></span>
<span class="hidden group-disabled:contents"><%= disabled %></span>
```

```erb
<%= render DisabledComponent.new(enabled: "Apply", disabled: "Applying...") %>
<!-- Or for a more complex use case -->
<%= render DisabledComponent.new do |c| %>
  <% c.enabled do %>
    Apply 
  <% end %>

  <% c.disabled do %>
    <i class="fa fa-spinner fa-animate-spin" />
    Applying...
  <% end %>
<% end %>
```

The suggested framework solution for supporting a pattern like this was to add a helper on the component instance that allowed setting the slot content via a method argument. Here, I propose `with_#{slot_name}_content` as a possible option.

With the changes in this PR, my above component definition would become:

```rb
class DisabledComponent < ApplicationComponent
  renders_one :enabled
  renders_one :disabled
end
```
And the usage of the component could then become:

```erb
<%= render DisabledComponent.new.with_enabled_content("Apply").with_disabled_content("Applying...") %>
```

And the complex use case example would retain the same block based API as before.


BlakeWilliams:
Thanks for opening this!

I think this approach makes sense, but I think there's a few points worth discussing. I think we'll want to support `renders_many` slots in addition to the existing support for `renders_one` slots.

I think there's also one drawback to the API, which is that arguments can't be passed to the slots which narrows the possible use-cases for this functionality. 

We've been discussing changing the slots API to have `with_#{slot_name}` as an explicit setter, so that `with_content` can be called on Slots that define no arguments.

e.g. with the change to the slots API, the following could work:

```erb
<%= render DisabledComponent.new do |c| %>
  <% c.with_enabled.with_content('enabled') %>
  <% c.with_disabled.with_content('disabled') %>
<% end %>
```

There's some overlap there, so I'm curious if that would help this particular use case.

joelhawksley:
One thing I noticed today: I think this would close https://github.com/ViewComponent/view_component/issues/931.

camertron (approved):
Cool 😎

PR #1176, opened 2021-12-02
Update ruby to the latest versions

VSPPedro:
### Summary

The purpose of this PR is to make the CI use the latest versions of Ruby - https://www.ruby-lang.org/en/downloads/releases/

What do you guys think about dropping the support for ruby 2.5.8? I can help with that.

I'm not entirely sure if I should update the CHANGELOG for something like this.


joelhawksley:
@VSPPedro thanks for the patch! What is your reasoning around suggesting we drop 2.5.8? I generally feel like dropping support for a Ruby version would be a breaking change. I figure we might as well support as many versions as possible, as I know some folks are using ViewComponent on Ruby 2.4, for that matter 😄

joelhawksley:
Also, regarding the changelog: I like to recognize all contributions to the project, even if they aren't material to the actual runtime ❤️

VSPPedro:
I'm happy to be able to help with this amazing project - even with simple things.

About dropping support for 2.5, sorry I should have been more clearer. I've noticed that the version 2.5 is no longer supported: https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-5-9-released/, but I understand the reason to keep it.

Would you like me to update the rails version in this PR too?

VSPPedro:
Ops, the CI is failing when ruby 2.5.9 and rails 6.1.3.1. I'm working on that.

Spone:
> Ops, the CI is failing when ruby 2.5.9 and rails 6.1.3.1. I'm working on that.

Re-running the test suite fixed the issue.

@VSPPedro can you please resolve the conflicts on `docs/CHANGELOG.md`?

Spone:
I guess we could also drop the patch-level part of the version:

```diff
ruby_version: [2.5.9, 2.6.9, 2.7.5, 3.0.3]
ruby_version: [2.5, 2.6, 2.7, 3.0]
```

This way we won't need to update this every time a new patch version is released.

What do you think @VSPPedro?

VSPPedro:
@Spone, I'm fine with that. Done!

Spone (approved):

joelhawksley (approved):

PR #1178, opened 2021-12-03
Make ViewComponent::Collection a collection

sammyhenningsson:
### Summary

Despite the name `ViewComponent::Collection`, is a thing that renders a collection, rather than being a collection itself. This is a bit confusing and prevents some use cases were we need to operate on the collection in more ways than just rendering them.

I think my use case (see background below) is quite unique. However I do still believe that it could be useful to be able to do more with collections, than to render them 😄  

### Background

My current use case is that I have users that has some set of data and they should be able to select how this data is presented. This will be done through selecting a template implemented using a view component that gets instantiated with the users data. In order to be sure that the view component (template) can render I've added validations on them. Something like:
```ruby
# app/components/application_component.rb                                                                                                   
  include ActiveModel::Validations                                                                                                          
end   
                                                                                                                                            
# app/components/site_component.rb
class SiteComponent < ApplicationComponent
  attr_reader :site                                                            
  validates :site, :name, presence: true                                                                                                    
  validate :valid_sections
        
  def initialize(site:)                                                                                                                     
    @site = site                                                                                                                            
  end 
        
  def name                                                                                                                                  
    site.name                                                                                                                               
  end 
        
  def sections                                                                                                                              
    SectionComponent.with_collection(site.sections)                                                                                         
  end 
                                                                                                                                            
  private 
        
  def valid_sections
    sections.each do |section|                                                                                                              
      errors.add(:sections, :invalid) unless section.valid?                                                                                 
    end                                                                                                                                     
  end                                                                                                                                       
end

# app/components/section_component.rb
class SectionComponent < ApplicationComponent                                                                                               
  validates :section, :foobar, presence: true                                                                                               
                                                                                                                                            
  def initialize(section:)
    @section = section
  end
  
  def foobar                                                                                                                                
    site.foobar
  end
end
```

The reason that I need to validate the data at runtime, is because I will add more and more templates that my users can select. But they need to know which ones they can use and what data they need to change to make use of the ones that are currently invalid.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,7 @@ title: Changelog
 # Changelog
 
 ## main
+
 * Make `ViewComponent::Collection` act like a collection of view components
 
     *Sammy Henningsson*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -6,6 +6,9 @@ title: Changelog
 # Changelog
 
 ## main
+* Make `ViewComponent::Collection` act like a collection of view components
+
+    *Sammy Henningsson*
 
 * Update `@param` of `#render_inline` to include `ViewComponent::Collection`.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -183,7 +183,7 @@ ViewComponent is built by over a hundred members of the community, including:
 <img src="https://avatars.githubusercontent.com/yykamei?s=64" alt="yykamei" width="32" />
 <img src="https://avatars.githubusercontent.com/xkraty?s=64" alt="xkraty" width="32" />
 <img src="https://avatars.githubusercontent.com/xronos-i-am?s=64" alt="xronos-i-am" width="32" />
-<img src="https://avatars.githubusercontent.com/sammyhenningsson?s=64" alt="dylanatsmith" width="32" />
+<img src="https://avatars.githubusercontent.com/sammyhenningsson?s=64" alt="sammyhenningsson" width="32" />
 
 ## Who uses ViewComponent?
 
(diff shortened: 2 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ title: Changelog
 # Changelog
 
 ## main
+* Make `ViewComponent::Collection` act like a collection of view components
```
Spone, line 9:
```suggestion

* Make `ViewComponent::Collection` act like a collection of view components
```

docs/index.md
```diff
@@ -178,6 +178,7 @@ ViewComponent is built by over a hundred members of the community, including:
 <img src="https://avatars.githubusercontent.com/wdrexler?s=64" alt="wdrexler" width="32" />
 <img src="https://avatars.githubusercontent.com/mattwr18?s=64" alt="mattwr18" width="32" />
 <img src="https://avatars.githubusercontent.com/dylanatsmith?s=64" alt="dylanatsmith" width="32" />
+<img src="https://avatars.githubusercontent.com/sammyhenningsson?s=64" alt="dylanatsmith" width="32" />
```
Spone, line 181:
```suggestion
<img src="https://avatars.githubusercontent.com/sammyhenningsson?s=64" alt="sammyhenningsson" width="32" />
```

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Make `ViewComponent::Collection` act like a collection of view components
```
```suggestion
* Make `ViewComponent::Collection` act like a collection of view components.
```

joelhawksley (approved):

joelhawksley:
@sammyhenningsson looks good! I'll merge once the conflicts are resolved ❤️

PR #1183, opened 2021-12-07
Add `--locale` flag to the component generator

bobmaerten:
### Summary

Add --translations flag to the component generator. This will create a basic translations file alongside the component.

Example:

```
$ rails generate component User --locales
      create  app/components/user_component.rb
      invoke  rspec
      create    spec/components/user_component_spec.rb
      invoke  erb
      create    app/components/user_component.html.erb
      invoke  locales
      create    app/components/user_component.en.yml
        gsub    app/components/user_component.en.yml
      create    app/components/user_component.fr.yml
        gsub    app/components/user_component.fr.yml
```

```yml
// user_component.en.yml
en:
  hello: "Hello world!"

// user_component.fr.yml
fr:
  hello: "Hello world!"
```

### Other Information

This will also work with --sidecar flag.

The code style and this very issue description is heavily inspired by @websebdev in #1030!


Spone:
Hi @bobmaerten thanks for contributing!

I'm wondering if the name for this option should be `translations` or `locales`? Rails Guides seem to [use the name "locale files"](https://guides.rubyonrails.org/i18n.html#organization-of-locale-files).

How would you manage multiple locales? Rails usually use [one file per locale](https://guides.rubyonrails.org/i18n.html#providing-translations-for-internationalized-strings), so I guess in your example the file name could be `user_component.en.yml` instead of `user_component.yml`?

It's probably best to rely on `I18n.available_locales` to figure out which locales are used in the project, then generate one file for each.

What do you think?

bobmaerten:
Hey @Spone, I sticked to what's provided by view_component. So I use the word "translations" as it's quoted this way in the docs.

The documentation is not very explicit, as it appears `view_component` only rely on a single file for translations/locales (https://viewcomponent.org/guide/translations.html). It would probably be smarter to stick with rails-i18n conventions though and have as many files as available locales if we can support it.

I extracted this code from what we implemented locally in our main app. One file with all translations in locales we use (mostly _en_ and _fr_) for each component, as there should not be so many translations in a single component. We also extract provided attributes from the component generator to add translations lines for them. But in a previous PR I made here, I was told to make a more simple approach first, rather than complex logic/opinionated choices. So here is my first iteration ;)

Should I iterate on rails configured locales and create as many files?

bobmaerten:
Here some updates:
  - rename generator to `locales_generator.rb`
  - rely on `I18n.available_translations` for creating locale files
  - adapted flags and documentation

bobmaerten:
I did try to find a way to allow passing locales as a flag param, like `rails generate component Hello --locales=de,en` with no real success. Seems that Thor `class_option` and Rails `hook_for` are very closely related.

Spone:
> I sticked to what's provided by view_component. So I use the word "translations" as it's quoted this way in the docs.

Makes sense, but it may be clearer to use "locale" as you did in your updates. Would you mind mentioning the `--locale` flag in the `translations` docs page as well?

> It would probably be smarter to stick with rails-i18n conventions though and have as many files as available locales if we can support it.

Yes, I agree. One caveat though, if you add `rails-i18n` to your project *without* defining `I18n.available_translations`, the generator may create dozens of files. We could add a warning in the generator docs, such as:

> Make sure `I18n.available_locales` is defined in your project, so that only the necessary locale files are created when generating a component.

----

> I did try to find a way to allow passing locales as a flag param, like rails generate component Hello --locales=de,en with no real success. Seems that Thor class_option and Rails hook_for are very closely related.

I'm not sure we need that. Most of the time, you want to translate your component in all `available_locales`. If not, it's easy to deleted unwanted files.

docs/guide/generators.md
```diff
@@ -82,6 +82,25 @@ bin/rails generate component Example title --stimulus
...
+      invoke  locales
+      create    app/components/example_component.en.yml
+        gsub    app/components/example_component.en.yml
+      create    app/components/example_component.fr.yml
+        gsub    app/components/example_component.fr.yml
+      invoke  erb
+      create    app/components/example_component.html.erb
+```
+
+To always generate locale files, set `config.view_component.generate_locales = true`. Make sure I18n.available_locales is defined in your project, so that only the necessary locale files are created when generating a component.
```
Spone, line 102:
```suggestion
To always generate locale files, set `config.view_component.generate_locales = true`. Make sure `I18n.available_locales` is defined in your project, so that only the necessary locale files are created when generating a component.
```

lib/view_component/base.rb
```diff
@@ -279,6 +279,14 @@ def content_evaluated?
     #
     mattr_accessor :generate_stimulus_controller, instance_writer: false, default: false
 
+    # Always generate a locale file(s) alongside the component:
+    #
+    #     config.view_component.generate_locales = true
+    #
+    # Defaults to `false`. Relies on I18n.available_locales, be sure to configure them in your Rails project before use.
```
Spone, line 286:
```suggestion
    # Defaults to `false`. Relies on `I18n.available_locales`, be sure to configure them in your Rails project before use.
```

lib/rails/generators/locales/component_generator.rb
```diff
@@ -0,0 +1,32 @@
...
+      include ViewComponent::AbstractGenerator
+
+      source_root File.expand_path("templates", __dir__)
+      class_option :sidecar, type: :boolean, default: false
+
+      def create_locale_file
+        I18n.available_locales.each do |locale|
+          dest_with_path = destination(locale)
+          template "component.yml", dest_with_path
+          gsub_file dest_with_path, /\Aen:\n/, "#{locale}:\n"
```
elia, line 17: nit: gsub can become fragile over time, thoughts on using an instance variable instead?

```rb
I18n.available_locales.each do |locale|
  @locale = locale
  template "component.yml", destination(locale)
end
```

And the locale template be `component.yml.tt`:

```erb
<%= @locale %>:
  hello: "Hello world!"
```

Alternatively it could also dump a hash to YAML:

```rb
create_file "component.#{locale}.yml", { locale.to_s => { "hello" => "Hello world!" } }.to_yaml
```

elia (commented):
I wish there was a way to have all the translations in a single file, as I find them easier to be kept in synch for small components, not a blocker though.

I left just one nit on the file creation in the generator.

Spone:
> I wish there was a way to have all the translations in a single file, as I find them easier to be kept in synch for small components, not a blocker though.

How would you approach that? With a config option (one file per locale vs. a single file for all locales)?

elia:
> How would you approach that? With a config option (one file per locale vs. a single file for all locales)?

Yes, a config option would probably work, and than can still rely on `available_locales` and fallback to just `en` for all those apps that use I18n just to keep the copy separated from the template.

@bobmaerten one more thing, I think we can stop depending on the presence of `available_locales` if we fall back to `[:en]`.

bobmaerten:
Hello folks.
I put all your nice advices into few more commits.
Now the generator is named `locale_generator` and respond to the `--locale` flag.
By default, only the single translation file is generated when triggered, with all `I18n.available_locales` configured entries, falling back on`[:en]`. A config option is available for those who prefer splitted files (`config.view_component.generate_splitted_locale_files`).
I got rid of the template file, using a `create_file` from a build YAML content inside the generator.
The generator also take main component arguments for building translation keys. We often want to translate parameters we use, not a placeholder `hello:` key. I had the same approach in #1194.
I added test cases for all of this and updated docs and API as requested by the linters.

bobmaerten:
There's a failing test in 6.0.3.6/2.5 setup I can't figure out. I did try to use backward compatible ruby, though :'(

joelhawksley:
@bobmaerten I re-ran your tests and it looks like they are green. This is both a relief and not a great sign about the reliability of our test suite 😬

bobmaerten:
Flakiness is Evil ;)

docs/CHANGELOG.md
```diff
@@ -39,6 +39,10 @@ title: Changelog
 
     *Simon Fish*
 
+* Add `--locales` flag to the component generator. Generates as many locale files as defined in `I18n.available_locales`, alongside the component.
```
Spone, line 42:
```suggestion
* Add `--locale` flag to the component generator. Generates as many locale files as defined in `I18n.available_locales`, alongside the component.
```

docs/api.md
```diff
@@ -85,6 +85,22 @@ Set a custom default layout used for preview index and individual previews:
 
     config.view_component.default_preview_layout = "component_preview"
 
+### #generate_locale
+
+Always generate translation file alongside the component:
```
Spone, line 90:
```suggestion
Always generate translations file alongside the component:
```

docs/guide/generators.md
```diff
@@ -82,6 +82,23 @@ bin/rails generate component Example title --stimulus
...
+      invoke  test_unit
+      create    test/components/example_component_test.rb
+      invoke  locale
+      create    app/components/example_component.yml
+      invoke  erb
+      create    app/components/example_component.html.erb
+```
+
+To always generate locale files, set `config.view_component.generate_locale = true`.
+If your prefer having your translations in distinct locale files you can set `config.view_component.generate_splitted_locale_files = true` in order to generate as many files as configured in `I18n.available_locales`.
```
Spone, line 100:
```suggestion

If your prefer having your translations in distinct locale files you can set `config.view_component.generate_splitted_locale_files = true` in order to generate as many files as configured in `I18n.available_locales`.
```

docs/api.md
```diff
@@ -85,6 +85,22 @@ Set a custom default layout used for preview index and individual previews:
...
+
+    config.view_component.generate_locale = true
+
+Defaults to `false`.
+
+### #generate_splitted_locale_files
```
Spone, line 96: What about renaming the option to use the word `distinct`, as you did in the generators docs? It feels clearer to me.

```suggestion
### #generate_distinct_locale_files
```

docs/guide/translations.md
```diff
@@ -23,6 +23,20 @@ en:
...
+Translations can be splitted into several locale files for convenience.
+
+```yml
+# app/components/example_component.en.yml
+en:
+  hello: "Hello world!"
+
+# app/components/example_component.fr.yml
+fr:
+  hello: "Bonjour monde !"
```
Spone, line 35: ;)

```suggestion
  hello: "Bonjour le monde !"
```

docs/guide/translations.md
```diff
@@ -23,6 +23,20 @@ en:
...
+```
+
+These files can be automatically generated by the component generator when the `--locales` flag is specified.
```
Spone, line 38:
```suggestion
These files can be automatically generated by the component generator when the `--locale` flag is specified.
```

lib/view_component/base.rb
```diff
@@ -279,6 +279,22 @@ def content_evaluated?
...
+    #
+    mattr_accessor :generate_locale, instance_writer: false, default: false
+
+    # Always generate one translations file per available locale:
+    #
+    #     config.view_component.generate_splitted_locale_files = true
+    #
+    # Defaults to `false`. One file will be generated for each configured `I18n.available_locales`.
+    #
+    mattr_accessor :generate_splitted_locale_files, instance_writer: false, default: false
```
Spone, line 296: What about renaming the option to use the word `distinct`, as you did in the generators docs? It feels clearer to me.

```suggestion
    mattr_accessor :generate_distinct_locale_files, instance_writer: false, default: false
```

Spone:
Thanks for your continued effort @bobmaerten, I think we're nearly there :)

lib/view_component/base.rb
bobmaerten: Seems right to me. Thanks for your hawk-coherence-eye! ;)

bobmaerten:
> Thanks for your continued effort @bobmaerten, I think we're nearly there :)

Thank for all you quick and fine-tuning review comments. This generator, now with the config options, can completely replace the custom-one we currently use on our main project.

Spone (approved):

docs/guide/translations.md
```diff
@@ -23,6 +23,20 @@ en:
   hello: "Hello world!"
 ```
 
+Translations can also be defined in distinct per-locale files for convenience.
```
joelhawksley, line 26:
```suggestion
Translations can also be defined in per-locale files:
```

joelhawksley (approved):

elia (approved):
This looks great! 🤩

PR #1184, opened 2021-12-07
Separate slot setter and getter method

BlakeWilliams:
This introduces an explicit setter for slots in the form of
`with_#{slot_name}`. This makes it more clear when slots are being set
vs accessed but also allows chaining when slots have no arguments.

Before this change, the following would not work:

```ruby
component.header.with_content("my header!")
```

This fails due to `header` being passed no arguments and no block,
resulting in it acting as a getter, returning `nil` since the slot was
not defined. This is due to the framework assuming that slots will always 
be passed either a block or arguments and predates the `with_content` 
method.

Now, the above can be implemented using `with_*`:

```ruby
component.with_header.with_content("my header!")
```


BlakeWilliams:
I didn't add a deprecation notice in this PR, but I think we'll likely decide to deprecate using `#{slot_name}` as a setter and instead utilize it only as a getter. That seems like a good follow-up for this change since it will give us time to try out the new API and come up with a migration plan.

lib/view_component/slotable_v2.rb
```diff
@@ -103,15 +108,15 @@ def renders_one(slot_name, callable = nil)
       # = Setting sub-component content
       #
       # Consumers of the component can set the content of a slot by calling a
-      # helper method with the same name as the slot. The method can be
-      # called multiple times to append to the slot.
+      # helper method with the same name as the slot prefixed with `with`. The
```
Spone, line 111: To be extra clear... ?

```suggestion
      # helper method with the same name as the slot prefixed with `with_`. The
```

BlakeWilliams: Good call, thank you!

lib/view_component/slotable_v2.rb
```diff
@@ -56,16 +56,21 @@ module SlotableV2
       # = Setting sub-component content
       #
       # Consumers of the component can render a sub-component by calling a
-      # helper method with the same name as the slot.
+      # helper method with the same name as the slot prefixed with `with`.
```
BlakeWilliams, line 59:
```suggestion
      # helper method with the same name as the slot prefixed with `with_`.
```

jonspalmer:
@BlakeWilliams I'd agree that the API proposed here has the right ergonomics and we should aim for that vs compromising it for the sake of upgrade ease. For me the trick is going to be to provide the right upgrade path. Perhaps a mechanism to opt in to the old behavior allowing consumers to upgrade legacy components at their leisure.

sarhm:
> @BlakeWilliams I'd agree that the API proposed here has the right ergonomics and we should aim for that vs compromising it for the sake of upgrade ease. For me the trick is going to be to provide the right upgrade path. Perhaps a mechanism to opt in to the old behavior allowing consumers to upgrade legacy components at their leisure.

lib/view_component/slotable_v2.rb
sarhm: > To be extra clear... ?
> 
> ```suggestion
>       # helper method with the same name as the slot prefixed with `with_`. The
> ```

sarhm:
> Hi @BlakeWilliams! I'm wondering if you considered keeping the setter unchanged, but namespacing the getter.
> 
> My starting point here is trying to ease the migration of large codebases to the new syntax.
> 
> I think the getter is nearly always used internally to the ViewComponent, while the setter is usually called in a lot of places around the app, wherever you render an instance of the ViewComponent with the given slot.
> 
> So if only the getter is changed, you would only have to update each ViewComponent using slots (and its template(s)).
> Whereas if the setter is changed, it could mean updating hundreds of files.
> The former is probably easier to reason about (and maybe even to automate):
> - for each component:
>   - open the class
>   - if there are `renders_one` / `renders_many`
>     - find the slots getters in the component class and/or templates and rename them
>   - else, do nothing
> 
> In terms of syntax, what about something like this:
> 
> ```ruby
>   renders_one :header
> end
> c = MyComponent.new
> 
> # setter
> c.header { "Hello world!" }
> c.header.with_content("Hello world!")
> 
> # getter (suggestions)
> c.slots.header # or 
> c.header_slot # or 
> c.header_content
> ```

docs/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Add `with_*` slot API for defining slots
```
BlakeWilliams, line 12:
```suggestion
* Add `with_*` slot API for defining slots.
```

joelhawksley (approved):

Issue #1198, opened 2021-12-13
Make it easier for folks to provide us with example repos


boardfish:
> @joelhawksley I've thought about that too, but there are a couple of changes I'd make to improve the experience there. First off, I think we could provide a repo template for this purpose, and automate keeping it up to date so that all the setup is just a click away. We could probably leverage Codespaces to help with that.
> ~~Second, I'm waiting to merge a change into `bundler` that'll let you specify a `require` path, so that if folks do need to build from scratch to more easily add support for other libraries like Hotwire and Tailwind, they can `require: 'view_component/engine'` without having to update their Gemfile to do so. So the optimal solution might be a couple of hops away, but this is definitely a good thing for me to drop in before we consider those.~~

_Originally posted by @boardfish in https://github.com/github/view_component/issues/1159#issuecomment-982096344_

boardfish:
The latter's unnecessary now thanks to #1117, but I'd like to work on the former this week and see how far I can take it. If anyone has any helpful ideas or wants to contribute, you're welcome to reply to this thread or get in touch with me.

PR #1199, opened 2021-12-13
Replicate failure instructions

boardfish:
### Summary

Adds some basic instructions to `CONTRIBUTING.md` to help folks provide us with example repos. Also adds a reference to this file to the bug report issue template to encourage folks to check it out.

#1198 

### Other Information

Some of the more advanced stuff I mentioned on that issue is yet to come, but this is a small step we can make right now that'll be helpful.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.md
+++ b/.github/ISSUE_TEMPLATE/bug-report.md
@@ -3,7 +3,12 @@ name: ViewComponent bug report
 about: Create a report to help us improve ViewComponent
 ---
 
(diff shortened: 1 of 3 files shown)


boardfish:
One thing I also noticed while I was working on this was that both the bug report template and the PR template could potentially do with timestamps to say when the guidance was last updated, so that folks don't forget to check them out. The fact that the templates link to living documents is a blessing and a curse - what's there will always be up to date, but it's on the user to check that, which might reduce the likelihood that they check it out.

boardfish:
Now part of #1200.

Issue #1177, opened 2021-12-02
Create a separate development and production compiler mode


BlakeWilliams:
Context taken from: https://github.com/github/view_component/pull/1161#discussion_r755484984

> Similar to your env idea, I'm wondering if we need to go a level higher and think about introducing different modes to the compiler. A consistent compiler, for dev + test, and a fast compiler, for production. I don't think that idea should block this PR, but we do want to keep the performance benefits we see in production while also making dev+test consistent in threaded environments.

The idea is that we could have a development mode for the compiler, which adds the `Monitor.synchronize` call in the generated render methods to prevent the compiler from overwriting the call method while rendering. https://github.com/github/view_component/pull/1161#discussion_r755484984 got us very close to thread safety, but without the extra sync we're still likely susceptible to race conditions in test/dev (just much less so).

The reason behind two different modes is that production code doesn't need the synchronization since it's compiled at boot time and will not have a chance to run into issues due to a race.

horiaradu:
Is this urgent? If not, I would like the opportunity to take a crack at it :)

BlakeWilliams:
@horiaradu No urgency at all, feel free to assign yourself and take it on. 🙂

horiaradu:
I can't assign myself (I think because of rights), but maybe u can? Otherwise, it's not an issue, I have the tab open :)

BlakeWilliams:
No problem, I went ahead and assigned it to you so nobody else picks it up.

PR #1202, opened 2021-12-13
Improved threadsafety

horiaradu:
### Summary

* add blocking and non blocking compiler mode (blocking in Rails dev & test mode)
* add monitor check in render method to ensure thread safety
* fixes https://github.com/github/view_component/issues/1177


lib/view_component/compiler.rb
```diff
@@ -5,70 +5,90 @@ class Compiler
...
 
+    # Compiler mode. Can be either:
+    # * blocking (default in Rails development and test mode)
+    # * non_blocking (default in Rails production mode)
+    attr_accessor :mode
+
     def initialize(component_class)
       @component_class = component_class
       @__vc_compiler_lock = Monitor.new
+      @mode = defined?(Rails.root) && (Rails.env.development? || Rails.env.test?) ? :blocking : :non_blocking
```
boardfish, line 16: Why `defined?(Rails.root)` specifically? Is the intent here to check whether we're running within the context of a full Rails app, or to check that Rails is installed at all?

test/view_component/view_component_test.rb
```diff
@@ -933,6 +933,26 @@ def test_each_component_has_a_different_lock
...
+
+  def test_compilation_in_non_blocking_mode
+    ViewComponent::CompileCache.cache.delete(MyComponent)
+    MyComponent.compiler.mode = :non_blocking
+
+    render_inline(MyComponent.new)
+
+    assert_selector("div", text: "hello,world!")
+
+    MyComponent.compiler.mode = :blocking
```
boardfish, line 953: Probably a good idea to put this in an `ensure` so that a failure means other tests don't run in non-blocking mode.

Might also be good to provide a support method to this effect, assuming we'll need wider testing of this mode:

```ruby
def with_compiler_mode(mode, component)
  previous_mode = component.compiler.mode
  yield(component)
ensure
  component.compiler.mode = previous_mode
end

File: lib/view_component/compiler.rb (hunk 1 above)
Comment (boardfish) [reply lib/view_component/compiler.rb:16]: I'm inclined to agree, but it looks to me like we're using that same check a lot elsewhere, so I guess it makes sense to keep it around here.

**Review by boardfish — APPROVED** [review]

**Comment by joelhawksley** [comment]
@BlakeWilliams mind providing a review here?

File: lib/view_component/compiler.rb [hunk 3]
```diff
@@ -5,70 +5,90 @@ class Compiler
...
     end
 
     def compiled?
       CompileCache.compiled?(component_class)
     end
 
+    def blocking?
+      mode == :blocking
```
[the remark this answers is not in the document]
horiaradu: Renamed to DEVELOPMENT_MODE and PRODUCTION_MODE

lib/view_component/compiler.rb
```diff
@@ -5,70 +5,90 @@ class Compiler
     # Lock required to be obtained before compiling the component
     attr_reader :__vc_compiler_lock
 
+    # Compiler mode. Can be either:
+    # * blocking (default in Rails development and test mode)
+    # * non_blocking (default in Rails production mode)
+    attr_accessor :mode
```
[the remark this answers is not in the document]
horiaradu: I think it's a great idea!

lib/view_component/compiler.rb
horiaradu: The other way around, but yes! Good idea

horiaradu:
@boardfish @joelhawksley @BlakeWilliams I have addressed the above comments. Thanks a lot for the feedback! I think this approach is much better than my initial one. Let me know what you think :)

horiaradu:
I dunno how to manage to test that initialisation condition for development/production mode, since I can't really put the engine in production mode.

boardfish:
I assume by that you mean testing what was added in `lib/view_component/engine.rb`. Perhaps you could do it by stubbing `Rails.env` to return `'development'.inquiry` or `'production'.inquiry`?

horiaradu:
> I assume by that you mean testing what was added in `lib/view_component/engine.rb`. Perhaps you could do it by stubbing `Rails.env` to return `'development'.inquiry` or `'production'.inquiry`?

Lol! obvious. Thanks!

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add blocking compiler mode, used in Rails development and testing mode
```
```suggestion
* Add blocking compiler mode, used in Rails development and testing mode..
```

lib/view_component/compiler.rb
```diff
@@ -85,7 +104,21 @@ def define_render_template_for
         "elsif variant.to_sym == :#{variant}\n    #{call_method_name(variant)}"
       end.join("\n")
 
-      component_class.class_eval <<-RUBY, __FILE__, __LINE__ + 1
+      if development?
+        component_class.class_eval <<-RUBY, __FILE__, __LINE__ + 1
+        def render_template_for(variant = nil)
```
joelhawksley, line 109: Might we be able to extract the shared parts of this heredoc? I could see these two branches falling out of sync.

BlakeWilliams (approved):
Looks great!

lib/view_component/compiler.rb
horiaradu: Good idea. Done!

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add blocking compiler mode for us use in Rails development and testing modes, improving thread safety.
```
boardfish, line 10:
```suggestion
* Add blocking compiler mode for use in Rails development and testing modes, improving thread safety.
```

lib/view_component/compiler.rb
```diff
@@ -14,18 +22,22 @@ def compiled?
...
+      with_lock do
         CompileCache.invalidate_class!(component_class)
 
         subclass_instance_methods = component_class.instance_methods(false)
 
         if subclass_instance_methods.include?(:with_content) && raise_errors
           raise ViewComponent::ComponentError.new(
             "#{component_class} implements a reserved method, `#with_content`.\n\n" \
-            "To fix this issue, change the name of the method."
+          "To fix this issue, change the name of the method."
```
Spone, line 40:
```suggestion
            "To fix this issue, change the name of the method."
```

lib/view_component/compiler.rb
```diff
@@ -5,6 +5,14 @@ class Compiler
     # Lock required to be obtained before compiling the component
     attr_reader :__vc_compiler_lock
 
+    # Compiler mode. Can be either:
+    # * blocking (default in Rails development and test mode)
+    # * non_blocking (default in Rails production mode)
```
Spone, line 10: Can you clarify why `blocking` and `non_blocking` are mentioned in the comment? In the latest version, the constants values are `:development` and `:production`.

boardfish: I guess this comment must not have been updated after that change was made.

horiaradu: Yea! I have updated the comment, let me know if it's more clear now.

Spone: Great, thank you!

horiaradu:
@Spone I have addressed the comments. Let me know if it's ok.

horiaradu:
I saw that I was getting failed builds and the root cause was [this test](https://github.com/github/view_component/pull/1202/commits/7c5c4025d48279a2fe6c6310fe7a4ea86a5ae02a). The problem was that, after the test was executed, the compiler remained in PRODUCTION mode which was causing the multithreaded test to fail.

After I've removed it, I am failing the coverage check.

What I could do, in order to keep the test, is to add a before hook for each test, to ensure the compiler mode is DEVELOPMENT.

What do you think?

BlakeWilliams:
> I saw that I was getting failed builds and the root cause was [this test](https://github.com/github/view_component/pull/1202/commits/7c5c4025d48279a2fe6c6310fe7a4ea86a5ae02a). The problem was that, after the test was executed, the compiler remained in PRODUCTION mode which was causing the multithreaded test to fail.

Could we capture the current state around the test and set reset it once the test is complete? Something like:

```ruby
  def test_sets_the_compiler_mode_in_production_mode
    original_compiler_mode = ViewComponent::Compiler.mode
    begin
      Rails.env.stub :development?, false do
        Rails.env.stub :test?, false do
          ViewComponent::Engine.initializers.find { |i| i.name == "compiler mode" }.run
          assert_equal ViewComponent::Compiler.mode, ViewComponent::Compiler::PRODUCTION_MODE
        end
      end
    ensure
      ViewComponent::Compiler.mode = original_compiler_mode
    end
  end
```

boardfish:
That's possible with [this test helper](https://github.com/github/view_component/pull/1202/files#diff-ba37813ca277c227a74a372479b7b05b7f3ff085d890ab708f80d62573efdb7aR135-R141), I should think.

horiaradu:
I have found the issue! I have to re-run the initializer after I reset the stub :). I've pushed the change.

test/view_component/integration_test.rb
```diff
@@ -558,4 +558,29 @@ def test_renders_a_preview_with_image_path
 
     assert_includes response.body, "images/foo.png"
   end
+
+  def test_sets_the_compiler_mode_in_production_mode
+    old_env = Rails.env
+    begin
+      Rails.env = "production"
```
boardfish, line 565:
```suggestion
      Rails.env = "production".inquiry
```

`Rails.env` is normally an `ActiveSupport::StringInquirer` - see [here](https://api.rubyonrails.org/classes/ActiveSupport/StringInquirer.html) for more details on that.

boardfish (approved):

BlakeWilliams (approved):

BlakeWilliams:
Thank you for adding this functionality!

PR #1208, opened 2021-12-15
Add test for a lambda slot with a block and no other arguments

### LINKED ISSUE #1133: lambda slot with `&block` argument requires an unrelated argument
@boardfish i never considered using fields_for which cuts it for me!! Thank you so much!! There is just onw weird thing i don't really understand - using `renders_one` as a lambda slot with `&block` argument requires me to define ANY arguments. Otherwise, it's gonna raise an `wrong number of arguments (given 1, expected 0)` exception. So i can't just do `renders_one :quick_filters, ->(&block) { fields_for(@q, nil, &block) }` - instead, i did:

```
renders_one :quick_filters, ->(dummy: nil, &block) { fields_for(@q, nil, &block) }
```

Thanks again !!!

_Originally posted by @mbajur in https://github.com/github/view_component/discussions/1126#discussioncomment-1617520_

---

Might be a misconception on my part here, but in order to pass content to a lambda slot, I figure you'd need to cascade a block down into the function the lambda is calling as described above. I'm not sure if (and also doubt) you can pass `content` instead of `&block` to `fields_for` in the above example.

I'll pull together a failing test case for this issue once I find time.

boardfish:
### Summary

Closes #1133 

### Other Information

I couldn't replicate @mbajur's original issue, but we now have a passing test to ensure it doesn't become a problem.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Add a test to ensure blocks can be passed into lambda slots without the need for any other arguments.
+
+    *Simon Fish*
+
 * Add linters for file consistency.
 
     *Simon Fish*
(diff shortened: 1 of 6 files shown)


joelhawksley (approved):
👏🏻

PR #1209, opened 2021-12-15
Update `.tool-versions` to latest Ruby version

cpjmcquillan:
### Summary

In #1176, `.tool-versions` was updated to be less specific about the ruby
version. asdf isn't smart enough to figure out which Ruby version to use,
though.

This updates the Ruby version specified in `.tool-versions` to the latest
release (3.0.3).


Spone:
Sorry about that! Thanks for the fix :)

Spone (approved):

Spone:
I'm not very familiar with asdf, but can we use `latest:3.0` as described here: https://asdf-vm.com/manage/versions.html#install-latest-stable-version

cpjmcquillan:
> I'm not very familiar with asdf, but can we use `latest:3.0` as described here: https://asdf-vm.com/manage/versions.html#install-latest-stable-version

You can install the latest version like that, but the problem is that asdf isn't clever enough to allow `3.0` in `.tool-versions` and a version matching `3.0.x` installed. If you only specify `3.0` in `.tool-versions`, asdf will raise an error when you try to run any ruby in that directory (e.g `bundle`). 

```
No preset version installed for command bundle
Please install a version by running one of the following:

asdf install ruby 2.6
```

If `.tool-versions` instead said `ruby non-existent-version` you could get the same error.

```
No preset version installed for command bundle
Please install a version by running one of the following:

asdf install ruby non-existent-version
```

Spone:
Thanks for your explanation :)

cpjmcquillan:
Here is a [useful explanation](https://github.com/asdf-vm/asdf-nodejs/issues/235#issuecomment-885809776) about why too

joelhawksley (approved):

PR #1211, opened 2021-12-16
Add the WEBrick gem to the docs app

cpjmcquillan:
### Summary

Since WEBrick is no longer bundled with Ruby in 3.0, and Jekyll depends on
WEBrick, the webrick gem needs to be added as a dependency to the docs
application in order to run `bundle exec jekyll serve` with Ruby 3.0+.

WEBrick has been added to a dependency of Jekyll on main in
https://github.com/jekyll/jekyll/pull/8524, but the change hasn't yet made it
into a release.


Spone (approved):

docs/Gemfile.lock
```diff
@@ -266,6 +267,7 @@ DEPENDENCIES
   tzinfo (~> 1.2)
   tzinfo-data
   wdm (~> 0.1.1)
+  webrick
 
 BUNDLED WITH
-   2.2.0
+   2.2.32
```
joelhawksley, line 273:
```suggestion
   2.2.0
```

joelhawksley (approved):

PR #1212, opened 2021-12-16
Allow component tests to define the current_user

cpjmcquillan:
### Summary

Authentication gems like [clearance] and [devise] define a `current_user` helper
method that can be used in controllers and views to access the currently signed
in user.

When building components that depend on the currently signed in user and
delegating `:current_user` to the Rails helpers, it is useful to be able to
define `current_user` in tests.

The `with_current_user` helper method defines a current_user method on the
controller so that these components can be tested easily.

[clearance]: https://github.com/thoughtbot/clearance
[devise]: https://github.com/heartcombo/devise

### Other Information

[Discussion](https://github.com/github/view_component/discussions/371) a while back indicated there could be appetite for a test helper like this
in `ViewComponent::TestHelpers`.

### DIFF
--- a/lib/view_component/test_helpers.rb
+++ b/lib/view_component/test_helpers.rb
@@ -126,6 +126,22 @@ module ViewComponent
       @controller = old_controller
     end
 
+    # Set the current_user in the given controller to the given value:
+    #
+    # ```ruby
+    # with_current_user(user) do
+    #  render_inline(MyComponent.new)
+    # end
+    # ```
+    #
+    # @param user [Object] The user to set as the current_user.
+    def with_current_user(user)
+      controller.define_singleton_method(:current_user) { user }
+      yield
+    ensure
+      controller.define_singleton_method(:current_user) { nil }
+    end
+
     # @private
     def build_controller(klass)
       klass.new.tap { |c| c.request = request }.extend(Rails.application.routes.url_helpers)
--- a/test/sandbox/app/controllers/integration_examples_controller.rb
+++ b/test/sandbox/app/controllers/integration_examples_controller.rb
@@ -3,6 +3,8 @@
 class IntegrationExamplesController < ActionController::Base
   layout false
 
+  helper_method :current_user
+
   def variants
     request.variant = params[:variant].to_sym if params[:variant]
   end
@@ -46,5 +48,11 @@ class IntegrationExamplesController < ActionController::Base
     products = [OpenStruct.new(name: "Radio clock"), OpenStruct.new(name: "Mints")]
 
     render(ProductComponent.with_collection(products, notice: "Today only"))
+  end
+
+  private
+
+  def current_user
+    nil
   end
 end
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Allow component tests to define the current user.
+
+    *Connor McQuillan*
+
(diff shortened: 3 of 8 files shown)


lib/view_component/test_helpers.rb
```diff
@@ -126,6 +126,22 @@ def with_request_url(path)
...
+    # end
+    # ```
+    #
+    # @param user [Object] The user to set as the current_user.
+    def with_current_user(user)
+      controller.define_singleton_method(:current_user) { user }
+      yield
+    ensure
+      controller.define_singleton_method(:current_user) { nil }
+    end
```
cpjmcquillan, line 143: I'm not convinced this is the best approach, but I'm haven't been able to think of other approaches yet that would work and don't depend on a particular testing framework.

test/sandbox/app/controllers/integration_examples_controller.rb
```diff
@@ -3,6 +3,8 @@
 class IntegrationExamplesController < ActionController::Base
   layout false
 
+  helper_method :current_user
```
cpjmcquillan, line 6: I assumed the default test controller was an ok place to define this, and I didn't want to obfuscate the intention of the test by creating a specific controller.

joelhawksley:
@cpjmcquillan doesn't devise already have a `sign_in` test helper? Is there any way we could adapt it to work with ViewComponent and avoid having to write our own test helper?

lib/view_component/test_helpers.rb
BlakeWilliams: I feel like this is a great application-level helper, but I'm not sure how it fits into the framework. In the GitHub codebase we've defined our own custom method for this that is specific to our application.

 Do you know if/how authentication library helper methods like Devise's [`sign_in`](https://www.rubydoc.info/gems/devise/Devise/Test/ControllerHelpers) helper work with ViewComponent? I think that could provide some guidance on how to move forward here.

primer-css:

cpjmcquillan:
Sorry I haven't found time to pick this up again yet.

Devise and Clearance both have test helpers that differ slightly in implementation. I think it would be nice if a ViewComponent test helper was authentication-gem-agnostic, but I'm not sure if that is achievable.

Ultimately some guidance in the docs with clear examples for common libraries might be sufficient. 

I'll try and pick this up in the next couple of weeks.

pupgray:
If anyone is curious about this, with Clearance at least, the standard ControllerHelpers does work in ViewComponent tests!
There's a few small tweaks to make in `test_helper.rb`

```ruby
require "clearance/test_unit"   # <-- bring in clearance's controller helpers

class ActiveSupport::TestCase
  # ...
  include Clearance::Testing::ControllerHelpers

  # ...

  def with_request_url_and_clearance(path)
    with_request_url path do
      @request.env[:clearance] = Clearance::Session.new(@request.env)
      yield
    end
  end
end
```

Then just use that helper when writing tests

```ruby
require "test_helper"

class HeaderComponentTest < ViewComponent::TestCase
  def test_header_renders_site_title
    with_request_url_and_clearance "/" do   #  <-- Helper used here
      sign_in                               #  <-- You should be able to use clearance's controller test helpers within this block
      render_inline(HeaderComponent.new)

      assert_selector('h1', text: 'Jackbattle')
    end
  end
end
```

I'm also quite new to this, but I got something working and thought I might share, your milage may vary.

cpjmcquillan:
Closing this because it's gone stale (both in my mind and on here), and I think application-specific helpers are probably sufficient.

If anyone ever wants to reopen this and run with it themselves, feel free!

Issue #1216, opened 2021-12-22
Add --sidecar as config option


t2:
Hello,

The `--sidecar` flag is very helpful for organizing our View Components. I would like to make this option a config variable so that we can make it a standard in our codebase without having to remember the `--sidebar`. How does this sound for a feature enhancement?

`config.view_component.sidecar_enabled = true`

joelhawksley:
@t2 I generally think that all generator options should allow for defaults to be set like that 👍🏻 

@Spone generally reviews generator-related things so perhaps he can chime in here?

g13ydson:
I created this PR #1218  that possibly solves this issue

t2:
@g13ydson you're right, this is what I was aiming for. Any movement on that merge?

Spone:
Let's keep the issue open until we merge the PR.

PR #1218, opened 2021-12-23
Add generate_sidecar as config option

g13ydson:
### Summary

Add `config.view_component.generate_sidecar` as config option

### Other Information

possibly resolves this issue #1216


boardfish:
I'm just wondering if there's a way to expose a layer of config for all generator options. There are other things like the parent class name that folks may want to set defaults for, so if we're going to introduce config for generator defaults, I think it should be resilient to new generator options. @Spone's got the smarts when it comes to generators, so let's see what they say. That's what I'm thinking, though.

Spone:
Thanks for the PR! I'll review this in a few days, I'm off for holidays.

g13ydson:
I set the default ruby version 3 to 3.0.3 as the CI was jumping to version 3.1 (I don't know why).
Some tests with Rails 6.1 + ruby 3.1 failed. Here we have an explanation: [https://gist.github.com/yahonda/2776d8d7b6ea7045359f38c10449937b](https://gist.github.com/yahonda/2776d8d7b6ea7045359f38c10449937b)

```
Run ruby/setup-ruby@v1
  with:
    ruby-version: 3
    bundler: default
    bundler-cache: false
    working-directory: .
    cache-version: 0
Modifying PATH
  Entries added to PATH to use selected Ruby:
    /opt/hostedtoolcache/Ruby/3.1.0/x64/bin
Downloading Ruby
  https://github.com/ruby/ruby-builder/releases/download/toolcache/ruby-3.1.0-ubuntu-20.04.tar.gz
```

.github/workflows/ci.yml
```diff
@@ -27,12 +27,12 @@ jobs:
     strategy:
       matrix:
         rails_version: [5.2.5, 6.0.3.6, 6.1.3.1, main]
-        ruby_version: [2.5, 2.6, 2.7, 3.0]
+        ruby_version: [2.5, 2.6, 2.7, 3.0.3]
```
Spone, line 30: Please revert these changes, as the issue will be solved separately, in #1224

Spone:
@g13ydson Please make sure to run `bundle exec rake docs:build` to regenerate docs/api.md to include changes in YARD comments from this pull request.

lib/view_component/base.rb
```diff
@@ -310,8 +310,15 @@ def content_evaluated?
...
-    mattr_accessor :component_parent_class,
-                   instance_writer: false
+    mattr_accessor :component_parent_class, instance_writer: false
+
+    # Always generate a component with a sidecar directory:
+    #
+    #     config.view_component.generate_sidecar = true
+    #
+    # Defaults to `false`.
+    #
```
Spone, line 320:
```suggestion
```

Spone:
> I'm just wondering if there's a way to expose a layer of config for all generator options. There are other things like the parent class name that folks may want to set defaults for, so if we're going to introduce config for generator defaults, I think it should be resilient to new generator options. @Spone's got the smarts when it comes to generators, so let's see what they say. That's what I'm thinking, though.

I agree with @boardfish: now that the generators have quite a few options, it would be interesting to refactor this, in a separate PR. Who wants to pair with me on this?

Spone (approved):

boardfish:
@Spone I'm up for pairing on that!

Spone:
> @Spone I'm up for pairing on that!

Pointer to the issue where we will work on this: #1229

PR #1221, opened 2021-12-28
[Fix] Inner Rack based methods return blank values fixed

stiig:
### Summary

We check the current page for the active class in our `ApplicationHelper` like this
```rb
...
return 'active' if link_path == request.path
...
```
but it doesn't work with the test helper `with_request_url`. This PR fixes this problem and makes the `request` variable look more like a real.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Return properly values for `request.path` and `request.query_string` methods.
+
+    *Vasiliy Matyushin*
+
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Return properly values for `request.path` and `request.query_string` methods.
```
BlakeWilliams, line 10:
```suggestion
* Return correct values for `request.path` and `request.query_string` methods when using the `with_request_url` test helper.
```

BlakeWilliams (approved):

Issue #1137, opened 2021-11-15
`turbo_stream` helpers not capturing correctly


boardfish:
I'm trying to update multiple parts of the page by rendering a component that uses multiple `turbo_stream.update` tags. This comes out mangled, with a subset of the expected HTML rendered followed by several empty `turbo-stream` tags. I'll put together a failing test case soon.

### Steps to reproduce

1. Create a component whose template renders `turbo-stream` tags using the helper provided by `turbo-rails`.
2. Render this in a controller action as follows:

```rb
respond_to { |format| format.turbo_stream { render ExampleComponent.new } }
```

### Expected behavior

The content given to each use of the `turbo_stream.update` helper should be rendered inside `turbo-stream` tags.

### Actual behavior

Some content is captured and rendered before empty `turbo-stream` tags.

### System configuration
**Rails version**: 7.0.0.alpha2

**Ruby version**: 2.7.4

**Gem version**: 2.40.0

boardfish:
I think this could be resolved by using a single output buffer if we come to do it again. In the meantime, [this](https://github.com/ViewComponent/view_component/issues/1322#issuecomment-1172315202) might work since it's akin to rendering a component in a controller context.

PR #1225, opened 2021-12-30
Setup Appraisal

Spone:
### Summary

This sets up [Appraisal](https://github.com/thoughtbot/appraisal) in order to facilitate testing multiple Rails versions.

### Other Information

This has been decided here: https://github.com/github/view_component/pull/1098#issuecomment-988295192 and will allow conditional dependencies (required for Tailwind support, see #1098)

PR #1227, opened 2022-01-03
Add support for rendering turbo stream tags

boardfish:
### Summary

Replicates #1137 with the aim of supporting #1106.

### Other Information

I'm not sure how to get this test to pass. I've tried using `helpers.tag.turbo_stream` and `helpers.tag.template` too, but neither seems to change the output.

Desired output (whitespace formatted for readability):

```html
<turbo-stream action="update" target="area1">
    <template>
        <span>Hello, world!</span>
    </template>
</turbo-stream>
```

Actual output (whitespace formatted for readability):

```html
<span>Hello, world!</span>
<turbo-stream action="update" target="area1">
    <template>
        &lt;span&gt;Hello, world!&lt;/span&gt;
    </template>
</turbo-stream>
```

Regardless of whether `helpers.tag.turbo_stream` or `turbo_stream.update` is used, what seems to be happening is that the block that is received is concatenated to the view as normal, then done again without being marked as `html_safe` within the tags.

Implementation of the `turbo_stream` tag builder starts around [here](https://github.com/hotwired/turbo-rails/blob/6f6d8c7cfb5f18fb7b71b47c7e73691b9e7d7e2e/app/models/turbo/streams/tag_builder.rb).This is the object returned by calling `turbo_stream` against the view.


boardfish:
Turbo doesn't support Rails < 6 - that seems to come right down to installation, since the builds are failing on dependencies. Was Appraisal ever added to our testing setup? I seem to remember seeing talk about it in another PR. I think that'd resolve this issue.

Spone:
> Turbo doesn't support Rails < 6 - that seems to come right down to installation, since the builds are failing on dependencies. Was Appraisal ever added to our testing setup? I seem to remember seeing talk about it in another PR. I think that'd resolve this issue.

Yes this is still a work in progress here: https://github.com/github/view_component/pull/1225

Feel free to review :)

Spone:
@boardfish #1225 is merged, you can rebase and add turbo as a conditional dependency!

boardfish:
@Spone I've tried to move things over to using Appraisal, but CI seems to be failing because it doesn't recognize `turbo-rails` as having been installed. I'm not having this problem locally when I run `bundle exec appraisal install && bundle exec appraisal rake` on a version that's compatible with `turbo-rails` (Ruby 3, `rails-head`). It doesn't look to me as though the tests are running via `appraisal` in CI, but I'm not confident around the source of the issue - would you be able to look into it?

Spone:
> It doesn't look to me as though the tests are running via `appraisal` in CI, but I'm not confident around the source of the issue - would you be able to look into it?

I guess we will have to wait for #1230 (see [this comment](https://github.com/github/view_component/pull/1230/#issuecomment-1039669413))

Spone:
We're now waiting for #1308 :)

reeganviljoen:
@Spone I see #1308 has been merged, are you still blocked on this ?

joelhawksley:
@Spone @boardfish @reeganviljoen any interest in taking this on, or should I close?

boardfish:
I'll rebase and see where that lands us – hopefully we're able to get it merged 🤞

reeganviljoen:
@joelhawksley i had to get investge this the other day for work reasons

Tldr the tag helpers use concat wich trips up the capture

boardfish:
Looks like we're subject to hotwired/turbo-rails#74 here. That's perhaps the final blocker – the purist in me thinks it's not worth bringing Action Cable in just to get this over the line, and that the capture compatibility patch ought to fix this. But equally I'm thinking we shouldn't be encouraging folks towards that if they can avoid it.

Also, while it's been a hot minute since I checked #1106, fixing this between ourselves and the `turbo-rails` folks would really help these two libraries to work together. They're both pretty core to the Rails experience now too, so the easier we can make it, the better.

joelhawksley:
@boardfish I'm fine adding ActionCable to our test stack as it's present in a lot of Rails apps ❤️

boardfish:
Awesome, I'll do that next – I expect that to get this passing. I might see if I can contribute to `turbo-rails` to get this fully resolved afterwards.

boardfish:
Alright, all set! We now require Action Cable in the sandbox app so that the Turbo Stream tag helper works. I also found that because of the age of this PR there was still some support for Rails versions we don't test against kicking around, so I got rid of that.

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add test coverage for use of `turbo_stream` helpers in components.
```
```suggestion
* Add test coverage for use of `turbo_stream` helpers in components when `capture_compatibility_patch_enabled` is `true`.
```

joelhawksley (approved):

PR #1230, opened 2022-01-06
Use Appraisal in CI

Spone:
Related to https://github.com/github/view_component/pull/1227#issuecomment-1005945458

In my initial setup of Appraisal (#1225), I forgot to specify which gemfile to use in the CI configuration.


.github/workflows/ci.yml
```diff
@@ -46,7 +49,7 @@ jobs:
     - uses: actions/cache@v2
       with:
         path: vendor/bundle
-        key: gems-build-rails-${{ matrix.rails_version }}-ruby-${{ matrix.ruby_version }}-${{ hashFiles('**/Gemfile.lock') }}
+        key: gems-build-rails-${{ matrix.rails_version }}-ruby-${{ matrix.ruby_version }}-${{ hashFiles('**/gemfiles/rails_${{ matrix.rails_version }}.gemfile.lock') }}
```
Spone, line 52: I'm not sure this works. :thinking:

boardfish:
Might it make sense to use Appraisal directly the same way it's mentioned in`CONTRIBUTING.md`, or does that complicate caching?

Spone:
> Might it make sense to use Appraisal directly the same way it's mentioned in`CONTRIBUTING.md`, or does that complicate caching?

Something like that?

```diff
- bundle exec rake
+ bundle exec appraisal rails-${{ matrix.rails_version }} rake
```

.github/workflows/ci.yml
boardfish: I think this part probably works alright. I'm inclined to believe it's setting `BUNDLE_GEMFILE` that isn't doing anything.

if caching was failing but the Gemfile was fine, GitHub Actions would try to install `turbo-rails`. I'm inclined to believe setting the Gemfile doesn't work, but caching still does, because there's no mention of `turbo-rails` or `tailwindcss-rails` in the logs, and the logs still say they're `using` a gem that has already been downloaded.

boardfish:
Yeah. Come to think of it, I don't think it would cause caching issues, since it'd be installing the gems to the same place. You're right to be setting the cache keys based on the Rails version in use, though.

Spone:
What do you think of this approach @joelhawksley & @BlakeWilliams?

Spone:
I finally got CI to pass, with appraisal running the specs. It's currently running a [fork](https://github.com/Spone/appraisal) of Appraisal, with this PR https://github.com/thoughtbot/appraisal/pull/174 merged. So we should probably wait for the a new version of Appraisal before going forward with this.

.github/workflows/ci.yml
Spone: `tailwindcss-rails` now appear in the logs of the latest runs.

Gemfile
```diff
@@ -8,6 +8,10 @@ rails_version = "#{ENV['RAILS_VERSION'] || 'main'}"
 gem "capybara", "~> 3"
 gem "rails", rails_version == "main" ? { git: "https://github.com/rails/rails", ref: "main" } : rails_version
 
+group :development, :test do
+  gem "appraisal", github: "excid3/appraisal", branch: "fix-bundle-env"
```
BlakeWilliams, line 12: Is there a reason we shouldn't be using `thoughtbot/appraisal`?

Spone: Yes because the fix we need is not yet merged: https://github.com/thoughtbot/appraisal/pull/174

.github/workflows/ci.yml
boardfish: It doesn't look like the call to `hashFiles` is returning anything.

boardfish (commented):
The call to `hashFiles` when setting the cache key isn't returning anything, so the cache can be overwritten by PRs.

Gemfile
boardfish: Is there a way we can keep tabs on where @excid3's fork is at to ensure it's not susceptible to upstream issues with Appraisal?

excid3: Just FYI, I am skipping appraisal in CI right now and just setting `BUNDLE_GEMFILE` via `matrix`.

https://github.com/excid3/noticed/blob/master/.github/workflows/ci.yml#L28

boardfish:
Seems like the multiple calls to the `cache` action might be interfering with each other here.

boardfish:
What needs doing to move this forward? Is there any way I can help? From what I can see it's probably just the gem caching that remains to be solved, but I'm not sure if you had any other plans @Spone.

Spone:
> What needs doing to move this forward? Is there any way I can help? From what I can see it's probably just the gem caching that remains to be solved, but I'm not sure if you had any other plans @Spone.

Yes I think the last blocking item is fixing the caching issue. If you have time to look into it that would be awesome!

Spone:
I'll close this, as it's superseeded by #1308.

PR #1245, opened 2022-01-13
Fix PVC CI

pouretrebelle:
### Summary

This was broken due to a change in Primer ViewComponents which now
requires running yarn in the docs directory to build the docs.
See https://github.com/primer/view_components/pull/970

[Example failed run](https://github.com/github/view_component/runs/4801553969?check_suite_focus=true)


BlakeWilliams (approved):
Thanks for fixing this!

PR #1249, opened 2022-01-19
Place all generator options under `generate` config option

### LINKED ISSUE #1229: Make generator defaults configurable
Now that the generators have quite a few options, it would be good to make their defaults configurable from the outside so that folks don't need to create their own generators as often.

We could start by looking at `lib/view_component/engine.rb`, where most of the library's config options are defined. It'd probably be a good idea to use [ActiveSupport::OrderedOptions](https://api.rubyonrails.org/classes/ActiveSupport/OrderedOptions.html) for this.

boardfish:
### Summary

Closes #1229 by placing all config options linked to generators under the `generate` config option. Behind the scenes, this uses [`ActiveSupport::OrderedOptions`](https://api.rubyonrails.org/classes/ActiveSupport/OrderedOptions.html), which inherits from `Hash` to provide method accessors. Like `Hash`, this can be initialized with a default value, which has been set to `false` to reflect the existing behavior - all config options are `false` except for `component_path`.

### Other Information

TODO:

- [x] Move `generate_stimulus_controller` under namespace
- [x] Move `generate_locale` under namespace
- [x] Move `generate_distinct_locale_files` under namespace
- [x] Move `view_component_path` under namespace
- [x] Move `component_parent_class` under namespace

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -64,131 +64,34 @@ namespace :docs do
 
     instance_methods_to_document = meths.select { |method| method.scope != :class }
     class_methods_to_document = meths.select { |method| method.scope == :class }
-
-    File.open("docs/api.md", "w") do |f|
-      f.puts("---")
-      f.puts("layout: default")
-      f.puts("title: API")
-      f.puts("nav_order: 3")
-      f.puts("---")
-      f.puts
-      f.puts("<!-- Warning: AUTO-GENERATED file, don't edit. Add code comments to your Ruby instead <3 -->")
-      f.puts
-      f.puts("# API")
-
-      f.puts
-      f.puts("## Class methods")
-
-      class_methods_to_document.each do |method|
-        suffix =
-          if method.tag(:deprecated)
-            " (Deprecated)"
-          end
-
-        types =
-          if method.tag(:return)&.types
-            " → [#{method.tag(:return).types.join(',')}]"
-          end
-
-        f.puts
-        f.puts("### #{method.sep}#{method.signature.gsub('def ', '')}#{types}#{suffix}")
-        f.puts
-        f.puts(method.docstring)
-
-        if method.tag(:deprecated)
-          f.puts
-          f.puts("_#{method.tag(:deprecated).text}_")
-        end
-      end
-
-      f.puts
-      f.puts("## Instance methods")
-
-      instance_methods_to_document.each do |method|
-        suffix =
-          if method.tag(:deprecated)
-            " (Deprecated)"
-          end
-
-        types =
-          if method.tag(:return)&.types
-            " → [#{method.tag(:return).types.join(',')}]"
-          end
-
-        f.puts
-        f.puts("### #{method.sep}#{method.signature.gsub('def ', '')}#{types}#{suffix}")
-        f.puts
-        f.puts(method.docstring)
-
-        if method.tag(:deprecated)
-          f.puts
-          f.puts("_#{method.tag(:deprecated).text}_")
-        end
-      end
-
-      f.puts
-      f.puts("## Configuration")
-
-      registry.
-        get("ViewComponent::Base").
-        meths.
-        select { |method| method[:mattr_accessor] }.
-        sort_by { |method| method[:name] }.
-        each do |method|
-        suffix =
-          if method.tag(:deprecated)
-            " (Deprecated)"
-          end
-
-        f.puts
-        f.puts("### #{method.sep}#{method.name}#{suffix}")
-
-        if method.docstring.length > 0
-          f.puts
-          f.puts(method.docstring)
-        end
-
-        if method.tag(:deprecated)
-          f.puts
-          f.puts("_#{method.tag(:deprecated).text}_")
-        end
-      end
-
-      f.puts
-      f.puts("## ViewComponent::TestHelpers")
-
-      registry.
+    configuration_methods_to_document = registry.get("ViewComponent::Base").meths.select { |method| method[:mattr_accessor] }
+    test_helper_methods_to_document = registry.
         get("ViewComponent::TestHelpers").
         meths.
         sort_by { |method| method[:name] }.
         select do |method|
           !method.tag(:private) &&
             method.visibility == :public
-        end.
-        each do |method|
-          suffix =
-            if method.tag(:deprecated)
-              " (Deprecated)"
-            end
-
-          types =
-            if method.tag(:return)&.types
-              " → [#{method.tag(:return).types.join(',')}]"
-            end
-
-          f.puts
-          f.puts("### #{method.sep}#{method.signature.gsub('def ', '')}#{types}#{suffix}")
-
-          if method.docstring.length > 0
-            f.puts
-            f.puts(method.docstring)
-          end
-
-          if method.tag(:deprecated)
-            f.puts
-            f.puts("_#{method.tag(:deprecated).text}_")
-          end
         end
+
+    require 'rails'
+    require 'action_controller'
+    require 'view_component'
+    require 'view_component/docs_builder_component'
+
+    docs = ActionController::Base.new.render_to_string(
+      ViewComponent::DocsBuilderComponent.new(
+        sections: [
+          ViewComponent::DocsBuilderComponent::Section.new(heading: 'Class methods', methods: class_methods_to_document),
+          ViewComponent::DocsBuilderComponent::Section.new(heading: 'Instance methods', methods: instance_methods_to_document),
+          ViewComponent::DocsBuilderComponent::Section.new(heading: 'Configuration', methods: configuration_methods_to_document, show_types: false),
+          ViewComponent::DocsBuilderComponent::Section.new(heading: 'ViewComponent::TestHelpers', methods: test_helper_methods_to_document)
+        ]
+      )
+    )
+
+    File.open("docs/api.md", "w") do |f|
+      f.puts(docs)
     end
   end
 end
--- a/Rakefile
+++ b/Rakefile
@@ -78,7 +78,6 @@ namespace :docs do
     require 'action_controller'
     require 'view_component'
     require 'view_component/docs_builder_component'
-
     docs = ActionController::Base.new.render_to_string(
       ViewComponent::DocsBuilderComponent.new(
         sections: [
@@ -88,7 +87,7 @@ namespace :docs do
           ViewComponent::DocsBuilderComponent::Section.new(heading: 'ViewComponent::TestHelpers', methods: test_helper_methods_to_document)
         ]
       )
-    )
+    ).chomp
 
     File.open("docs/api.md", "w") do |f|
       f.puts(docs)
--- a/Rakefile
+++ b/Rakefile
@@ -64,27 +64,34 @@ namespace :docs do
 
     instance_methods_to_document = meths.select { |method| method.scope != :class }
     class_methods_to_document = meths.select { |method| method.scope == :class }
-    configuration_methods_to_document = registry.get("ViewComponent::Base").meths.select { |method| method[:mattr_accessor] }
+    configuration_methods_to_document = registry.get("ViewComponent::Base").meths.select { |method|
+      method[:mattr_accessor]
+    }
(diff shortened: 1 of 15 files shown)


docs/api.md
```diff
@@ -71,54 +71,48 @@ _Will be removed in v3.0.0._
...
+Defaults to `app/components`.
 
-Always generate a Stimulus controller alongside the component:
+Parent class for generated components
 
-    config.view_component.generate_stimulus_controller = true
+    config.view_component.generate.parent_class = "MyBaseComponent"
 
-Defaults to `false`.
+Defaults to "ApplicationComponent" if defined, "ViewComponent::Base" otherwise.
```
boardfish, line 115: Something's interesting here - this default is mandated not by the config, but by the generator. I wouldn't say it's a huge problem, but it does mean that right now, `config.view_component.component_parent_class` returns `nil` if it hasn't explicitly been set.

docs/api.md
```diff
@@ -71,54 +71,48 @@ _Will be removed in v3.0.0._
...
 
-    config.view_component.generate_distinct_locale_files = true
+Defaults to `false` unless otherwise stated.
 
-Defaults to `false`.
+Always generate a component with a sidecar directory:
 
-One file will be generated for each configured `I18n.available_locales`.
-Fallback on `[:en]` when no available_locales is defined.
+    config.view_component.generate.sidecar = true
```
joelhawksley, line 88: 🤔 I wonder if a noun would be more appropriate for this namespace:

`config.view_component.generator.sidecar`

boardfish: I considered that, but I think naming it `generate` makes it quite clear that we're `config`uring whether to `generate` the `sidecar` directory, `distinct_locale_files`, and other things. It makes the responsibility of this area of config explicit - whether to generate certain things. 

Granted there are a couple of things that don't fit that mold, (`view_component_path` and `parent_component`), but that pushes me to wonder whether they should really be part of the namespace. The former example definitely has uses outside of generators. The latter I'm thinking could be renamed `inheriting_from`, so to say you're `generate`ing a component `inheriting_from` `ApplicationComponent`, but there's also the argument that what we have already is clear.

boardfish: With this in mind, I think I should drop ceed17032b57346ce8eff5bf302c834b403ee1f1 and 92e8b834e8f66310ef574f5f338a9bb96b3a95b7. Does that sound correct?

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Place all generator options under `config.generate` namespace,
```
```suggestion
* Place all generator options under `config.generate` namespace.
```

docs/api.md
```diff
@@ -71,54 +71,48 @@ _Will be removed in v3.0.0._
...
 
 Set a custom default layout used for preview index and individual previews:
 
     config.view_component.default_preview_layout = "component_preview"
 
-### #generate_distinct_locale_files
+### #generate
 
-Always generate as many translations files as available locales:
+Configuration for generators
```
joelhawksley, line 82:
```suggestion
Configuration for generators.
```

docs/api.md
```diff
@@ -71,54 +71,48 @@ _Will be removed in v3.0.0._
...
 
-    config.view_component.generate_distinct_locale_files = true
+Defaults to `false` unless otherwise stated.
```
joelhawksley, line 84: This doesn't seem right. Isn't this just a namespace, not an option?

docs/api.md
```diff
@@ -71,54 +71,48 @@ _Will be removed in v3.0.0._
...
 
-Defaults to `false`.
+Always generate as many translations files as available locales:
 
-### #generate_sidecar
+    config.view_component.generate.distinct_locale_files = true
 
-Always generate a component with a sidecar directory:
+One file will be generated for each configured `I18n.available_locales`.
+Fallback on `[:en]` when no available_locales is defined.
```
joelhawksley, line 103:
```suggestion
One file will be generated for each configured `I18n.available_locales`, falling back to `[:en]` when `available_locales` is not defined.
```

docs/api.md
```diff
@@ -71,54 +71,48 @@ _Will be removed in v3.0.0._
...
+Path for component files
 
-Defaults to `false`.
+    config.view_component.generate.component_path = "app/my_components"
 
-### #generate_stimulus_controller
+Defaults to `app/components`.
 
-Always generate a Stimulus controller alongside the component:
+Parent class for generated components
```
joelhawksley, line 111: These docs feel hard to scan, as we no longer have headings folks can skim over. How do you think we could make this new format easy to skim?

docs/api.md
joelhawksley: That's interesting. I feel like we should fix the implementation or the docs, then 😄

docs/api.md
joelhawksley: I think your argument makes sense. Did you consider `generates`?

docs/api.md
boardfish: I see where you're coming from - this perhaps isn't phrased accurately to what I was intending. What I meant by this was that all options under the namespace default to `false` unless otherwise stated.

```suggestion
All options under this namespace default to `false` unless otherwise stated.
```

docs/api.md
boardfish: Not sure off the top of my head, but I'll investigate.

boardfish: The `docs:build` task currently builds that `.md` file using the `mattr_accessor`s on `ViewComponent::Base`. What I think I'll start to do now is write a component that can build that markdown, both to make any future changes to the generated docs more maintainable and to provide support for any further nested options we may introduce.

lib/view_component/base.rb
```diff
@@ -314,13 +287,33 @@ def content_evaluated?
...
-    #     config.view_component.generate_sidecar = true
+    #     config.view_component.generate.sidecar = true
+    #
+    # Always generate a Stimulus controller alongside the component:
+    #
+    #     config.view_component.generate.stimulus_controller = true
+    #
+    # Always generate translations file alongside the component:
+    #
+    #     config.view_component.generate_locale = true
```
Spone, line 305:
```suggestion
    #     config.view_component.generate.locale = true
```

docs/api.md
Spone: Maybe adding `###`-level headings for each option? Or displaying all these options in a table with 3 columns (option name, default value and description)?

boardfish: I went with the former about the same time as you sent this comment. It didn't actually require any changes to the implementation of our docs rake task, but I suppose it's worth keeping the changes I've made there.

Rakefile
```diff
@@ -64,131 +64,40 @@ namespace :docs do
...
-
-      f.puts
-      f.puts("## ViewComponent::TestHelpers")
+    require "rails"
+    require "action_controller"
+    require "view_component"
+    require "view_component/docs_builder_component"
+    docs = ActionController::Base.new.render_to_string(
+      ViewComponent::DocsBuilderComponent.new(
+        sections: [
```
joelhawksley, line 85: These could be slots 😄 (not blocking though)

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Place all generator options under `config.generate` namespace.
```
joelhawksley, line 10: Is this a breaking change? I'm wondering if we'd want to do a deprecation cycle for these updates where both the old and new options work.

lib/view_component/docs_builder_component.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module ViewComponent
+  class DocsBuilderComponent < Base
```
joelhawksley, line 4: Oh, this is just lovely 😍

docs/CHANGELOG.md
Spone: Yes, definitely a breaking change that requires a first release with a deprecation warning. And probably wait for the next major version to enforce the new syntax?

boardfish: Expected this — I'll make sure it's done before this is good to go. AFAIK this is the last blocking change that's been identified, so I'll re-request review from @Spone to be sure it's done correctly.

boardfish: Added these in d2e27655376cb30b787cafc4a759e102fbf207e3.

lib/view_component/base.rb
```diff
@@ -271,56 +288,58 @@ def content_evaluated?
...
-    #     config.view_component.generate_sidecar = true
+    #     config.view_component.generate.distinct_locale_files = true
     #
-    # Defaults to `false`.
+    # One file will be generated for each configured `I18n.available_locales`,
+    # falling back to `[:en]` when no available_locales is defined.
     #
-    mattr_accessor :generate_sidecar, instance_writer: false, default: false
+    mattr_accessor :generate, instance_writer: false, default: (ActiveSupport::OrderedOptions.new(false).tap do |c|
+      c.component_path = "app/components"
```
Spone, line 341: Shouldn't this reuse `config.view_component.view_component_path`?

lib/view_component/rails/tasks/view_component.rake
```diff
@@ -6,6 +6,6 @@ namespace :view_component do
   task statsetup: :environment do
     require "rails/code_statistics"
 
-    ::STATS_DIRECTORIES << ["ViewComponents", ViewComponent::Base.view_component_path]
+    ::STATS_DIRECTORIES << ["ViewComponents", ViewComponent::Base.generate.component_path]
```
Spone, line 9: I'm not sure this change is needed. The `view_component_path` setting still exists.

boardfish: That must've crept through - I thought I'd removed this when I dropped ceed17032b57346ce8eff5bf302c834b403ee1f1. I'll look into it.

lib/view_component/base.rb
boardfish: I don't think `generate` needs the `component_path` option at all - I'd have thought it got removed when I dropped ceed17032b57346ce8eff5bf302c834b403ee1f1.

lib/view_component/base.rb
```diff
@@ -271,56 +288,56 @@ def content_evaluated?
...
+    # #### #distinct_locale_files
+    #
+    # Always generate as many translations files as available locales:
     #
-    #     config.view_component.generate_sidecar = true
+    #     config.view_component.generate.distinct_locale_files = true
     #
-    # Defaults to `false`.
+    # One file will be generated for each configured `I18n.available_locales`,
+    # falling back to `[:en]` when no available_locales is defined.
```
Spone, line 338:
```suggestion
    # falling back to `[:en]` when no `available_locales` is defined.
```

lib/view_component/base.rb
```diff
@@ -40,6 +40,23 @@ def self._after_compile
       # noop
     end
 
+    # @!macro [attach] deprecated_generate_mattr_accessor
+    #   @method generate_$1
+    #   @deprecated Use `#generate.$1` instead. Will be removed in v3.0.0.
+    def self._deprecated_generate_mattr_accessor(name)
+      define_singleton_method("generate_#{name}".to_sym) do
```
Spone, line 47: This part of the code is missing test coverage (see CI).

Spone (approved):

Spone:
@boardfish I disabled auto-merge to have time to discuss this first: https://github.com/github/view_component/issues/1229#issuecomment-1036138490

docs/guide/generators.md
```diff
@@ -36,6 +36,8 @@ bin/rails generate component Sections::Example title content
 
 ## Options
 
+You can specify options when running the generator. To alter the default values project-wide, define the configuration settings described in [API docs](/api.html#configuration).
```
Spone, line 39: What do you think @boardfish? Feel free to rewrite this if unclear :)

boardfish:
Awesome, that sounds great. Merge if you're happy with it 👌

PR #1240, opened 2022-01-12
Correct path in example test command

mhw:
### Summary

The test files are in `test/view_component` now; this fixes the example to reflect that.

PR #1233, opened 2022-01-07
:wrench: Add "Wecasa" to "Who uses ViewComponent?" section

WaKeMaTTa:
### Summary

Add `Wecasa` to the **Who uses ViewComponent?** list.

PR #1250, opened 2022-01-20
Automate changelog maintenance

### LINKED ISSUE #1182: Automating changelog maintenance
After #1181, I'd like to suggest automating the maintenance of ViewComponent's changelog.

While I love a good hand-crafted changelog, I think the conventions already in place in this repository make it worthwhile to pursue changelog automation:
- All PRs are required to have a changelog entry.
  - I think a manually curated changelog makes sense in projects where the maintainers prefer to leave out the changes that are scoped to the internals of the repository. But [we recognize all contributions to the project, even if they aren't material to the actual runtime](https://github.com/github/view_component/pull/1176#issuecomment-984902386).
- PRs are typically squash-merged into `main`

Those conventions make `git log` a good fit for changelog generation.

Automating the changelog management would solve a few issues:
- `CHANGELOG.md` is a frequent source of merge conflicts among the PRs that aren't merged in a short window.
- It's possible to miss an entry in the GitHub Release description because a change was listed under the wrong version when merging `CHANGELOG.md`. (#1181)
- We have `/CHANGELOG.md` as a symlink to `/docs/CHANGELOG.md`, which doesn't work for some tools (https://github.com/github/view_component/issues/1170).
  - We can address this by generating the changelog content in both locations (which would remain in sync since the updates would be totally automated).
- Maintainers who modify a PR's changelog entry before merging could be more efficient (https://github.com/github/view_component/pull/1161#pullrequestreview-821693727). Instead of adding a `suggestion` and committing it (two steps, with an extra CI run), maintainers can just edit the PR title or the squash commit message.
- Currently, the changelog doesn't link to PRs. PRs often have additional context, though, so it'd be nice to make them easier to access directly from the changelog entries.
  - When squash-merging, the PR number is automatically appended to the generated commit's subject, so generating the changelog from `git log` would have that covered.

---

Thoughts?

boardfish:
### Summary

Closes #1182 by introducing a `git-chglog` GitHub Action, which is configured from `.chglog/config.yml`.

The changelog will now look as follows:

> ## [v2.48.0] - 2022-01-14
> - Correct path in example test command ([#1240](https://github.com/github/view_component/issues/1240))
> - Fix PVC CI ([#1245](https://github.com/github/view_component/issues/1245))
> - Add generate_sidecar as config option ([#1218](https://github.com/github/view_component/issues/1218))
> - :wrench: Add "Wecasa" to "Who uses ViewComponent?" section ([#1233](https://github.com/github/view_component/issues/1233))
> ...

Versions pre-v2.34.0 will have their commit history preserved as originally written, because merge commits were used at the time, which meant the relevant commit messages weren't as informative.

There are a couple of caveats to this PR - we might be subject to git-chglog/git-chglog#15, and I've also opened git-chglog/git-chglog#199. We could link to the [contributors](https://viewcomponent.org/#contributors) section of the docs to fix the latter.


boardfish:
Seems as though `github_changelog_generator` is buggy. I'm going to see if I can replicate this with `release-drafter` instead: https://github.com/release-drafter/release-drafter

boardfish:
I've elected to go with [git-chglog](https://github.com/git-chglog/git-chglog#installation) instead - it's much faster, more lightweight, and seems more malleable. Right now, that does unfortunately come at the cost of contributors not getting credited alongside their changes, but I've raised an issue with them to see if that can be implemented.

I think that's because it doesn't query GitHub and instead runs with Git history, which is what makes it so fast.

.vale.ini
```diff
@@ -17,3 +17,7 @@ Microsoft.Passive = NO
 Microsoft.SentenceLength = NO
 Microsoft.Vocab = NO
 Microsoft.We = NO
+
+TokenIgnores = ({{ .*? -?}})
+proselint.Typography = NO
+Microsoft.Ellipses = NO
```
boardfish, line 23: Aware that this isn't something we want, but I'm not sure how to disable this for just `.chglog/CHANGELOG.tpl.md`. In that file, I think it makes sense for an ellipsis to be made up of individual periods so that it's possible to use it in Git as a range immediately, e.g. `2.28.0..HEAD`.

boardfish:
This'll need to be merged forcibly when the time comes. I'll hang fire until it's through review before asking for that, though.

BlakeWilliams:
> I've elected to go with [git-chglog](https://github.com/git-chglog/git-chglog#installation) instead - it's much faster, more lightweight, and seems more malleable. Right now, that does unfortunately come at the cost of contributors not getting credited alongside their changes, but I've raised an issue with them to see if that can be implemented.

Unfortunately I think that's a blocker for adopting this approach. I feel it's important to make sure contributors are credited appropriately and have their name on the changelog. Ideally we can display their full name in addition to their handle, but I think there's room for discussion there as long as contributors get some kind of credit. Is there another tool we could potentially investigate using here?

BlakeWilliams:
> I've elected to go with [git-chglog](https://github.com/git-chglog/git-chglog#installation) instead - it's much faster, more lightweight, and seems more malleable. Right now, that does unfortunately come at the cost of contributors not getting credited alongside their changes, but I've raised an issue with them to see if that can be implemented.

Unfortunately I think that's a blocker for adopting this approach. I feel it's important to make sure contributors are credited appropriately and have their name on the changelog. Ideally we can display their full name in addition to their handle, but I think there's room for discussion there as long as contributors get some kind of credit. Is there another tool we could potentially investigate using here?

boardfish:
I agree - I'll search for other options. So far, my understanding of tools I've looked at and tried to work with is that:

- `github_changelog_generator` meets our needs, but hammers GitHub's API to meet its goal and is volatile to failure as a result. When run against this repo, it also fails to categorize things by release correctly. Its maintainership is unclear for now. 
- `git-chglog` is fast, efficient, and outputs correctly when used with squash merges as we've been doing since v2.34.0, but it doesn't credit the author of the PR. This is something we'd ideally like to continue to do, which at present makes this solution hard to recommend.
- `release-drafter` generates a changelog on a per-release basis, which means we'd potentially have to write more manual logic to ensure the state of the changelog file is always as intended. We may also have to stop it from creating a GitHub release or integrate it with the release process.
- Other solutions exist in this space, but many are unmaintained.
- I also checked out a changelog SaaS whose name eludes me at the moment. While I feel like perhaps it'd meet our needs, reliance on services that aren't provided by GitHub itself might be too much commitment for this project.

My mindset here is to find a dependency we can rely on rather than needing to reinvent the wheel. Generating a changelog from pull request history feels like a problem that should already have a solution. I'll keep working on this — any support or recommendations are welcome. ✨

PR #1254, opened 2022-01-24
Add Ruby 3.1 and Rails 7.0 to CI

petergoldstein:
### Summary

Changes to CI configuration to add Ruby 3.1 and Rails 7.0 to the CI matrix, along with necessary changes to get everything green and on stable versions.

### Other Information

In addition to the straightforward changes required for the CI configuration, there are a few other changes included in this PR.

The first change was the `Gemfile.lock` in the base of the repo.  I updated it to be Rails 7.0.1 as opposed to 7.0.0.alpha.2.  This required including a few explicit `RAILS_VERSION` environment variables for certain steps in CI and linting.

Two additional items required to get the specs to green were changes to some argument handling and module prepending.

To disambiguate with keyword arguments, it was necessary to add an explicit hash in the `lib/view_component/instrumentation.rb` implementation of `render_in`.

The second issue, which was a bit more subtle, has to do with the change to how module prepending in Ruby 3.1 works IF the module is already in the ancestor chain.  This is discussed [here](https://rubyreferences.github.io/rubychanges/3.1.html#moduleprepend-behavior-change).

Because of initializer code in `lib/view_component/engine.rb`, the `ViewComponent::Instrumentation` module is prepended to `ViewComponent::Base` on engine load.  But the `prepend` call is made for a second time when `ViewComponent::Instrumentation` is explicitly included.  In Ruby 3.0 and earlier this second call is a no-op, but in Ruby 3.1 it adds a second `ViewComponent::Instrumentation` to the ancestor chain, resulting in a duplicate invocation of the notification.

My sense is that there's no reason for the `self.included` call in `ViewComponent::Instrumentation` (and in fact, no reason for a component to explicitly include `ViewComponent::Instrumentation` at all), but given my limited familiarity with the internals of the library I simply restored the Ruby 3.0 and earlier behavior for this case.


.github/workflows/ci.yml
```diff
@@ -95,6 +104,7 @@ jobs:
         bundle exec rake
       env:
         VIEW_COMPONENT_PATH: ../view_component
+        RAILS_VERSION: '~> 6.1.0'
```
BlakeWilliams, line 107: Is this change necessary?

If so, can it use the matrix version instead of a hardcoded version of Rails?

.github/workflows/ci.yml
```diff
@@ -115,3 +125,5 @@ jobs:
         bundle config path vendor/bundle
         bundle install --jobs 4 --retry 3
         bundle exec rake coverage:report
+      env:
+        RAILS_VERSION: '~> 7.0.0'
```
BlakeWilliams, line 129: Same comment above

Gemfile
```diff
@@ -7,3 +7,9 @@ rails_version = "#{ENV['RAILS_VERSION'] || 'main'}"
 
 gem "capybara", "~> 3"
 gem "rails", rails_version == "main" ? { git: "https://github.com/rails/rails", ref: "main" } : rails_version
+
+if RUBY_VERSION >= "3.1"
+  gem "net-imap", require: false
```
BlakeWilliams, line 12: What happens if we omit these gems?

lib/view_component/instrumentation.rb
```diff
@@ -5,14 +5,16 @@
 module ViewComponent # :nodoc:
   module Instrumentation
     def self.included(mod)
-      mod.prepend(self)
+      mod.prepend(self) unless ancestors.include?(ViewComponent::Instrumentation)
```
BlakeWilliams, line 8: Nice find here!

BlakeWilliams (commented):
I had a few questions but overall looks good, thanks for opening this!

.github/workflows/ci.yml
petergoldstein: The pvc task runs without a matrix, so we can't use a matrix value (unless we want to create a matrix for this task).   It seems to fall back to `main` as a RAILS_VERSION without it and with the updated lockfile I was getting errors about sprockets not being available.  Explicitly setting Rails to a 6.1.x version addressed that.

I'm open to alternate suggestions.

.github/workflows/ci.yml
petergoldstein: And largely the same response as above.  :)

The coverage task doesn't have a matrix, so we can't use a matrix value.  When I left it free-floating - so it used Rails main - I got errors.  Given it's a hard-coded Ruby 2.7, it made sense to me to use the highest Rails version that supports Ruby 2.7.

Again, open to alternate suggestions.

Gemfile
petergoldstein: So this is a known Rails bug, fixed on the Rails 7 branch (with Rails 7.0.1) but not on Rails 6.1.  You can find a discussion [here](https://github.com/mikel/mail/pull/1439#issuecomment-1002801221)

These gems were included as defaults in Ruby versions 3.0 and earlier, but are not included in 3.1+.  Rails depends on them transitively through the mail gem.  [Rails 7.0.1 fixed this](https://github.com/rails/rails/pull/44083) but it can't be fixed by Rails for 6.x because of the need to support Ruby 2.6, where this inclusion causes an error.

On a more practical note, if these are left out then Ruby 3.1 / Rails 6.1 builds fail with an error saying they can't find net-smtp.

lib/view_component/instrumentation.rb
petergoldstein: Thanks.  :)

I have to admit it took me a couple of hours to suss this out.  I was scratching my head looking at `caller` outputs and trying to figure out how they could be different between Ruby 3.0 and Ruby 3.1.  Very satisfying to get an answer.

.github/workflows/ci.yml
BlakeWilliams: Ah great point, I completely missed that. I didn't expand the diff up high-enough and thought it was in the same section. YAML 😄

BlakeWilliams (approved):

BlakeWilliams:
@petergoldstein Looks great 👍. thanks again for opening this!

PR #1259, opened 2022-01-31
Automate changelog maintenance as part of `script/release`

rmacklin:
### Summary

This PR is an attempt to address #1182 using the approach mentioned in https://github.com/github/view_component/issues/1182#issuecomment-988470999

### DIFF
--- a/.github/pull-request-template.md
+++ b/.github/pull-request-template.md
@@ -11,7 +11,4 @@ to keep the conversation linked together. -->
 
 <!-- If there's anything else that's important and relevant to your pull
 request, mention that information here. This could include
-benchmarks, or other information.
-
-If you are updating any of the CHANGELOG files or are asked to update the
-CHANGELOG files by reviewers, please add the CHANGELOG entry at the top of the file. -->
+benchmarks, or other information. -->
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -34,35 +34,6 @@ jobs:
       with:
         config: '.github/lint/markdown.json'
         args: 'docs/**/*.md'
-  changedfiles:
-    runs-on: ubuntu-latest
-    # Map a step output to a job output
-    outputs:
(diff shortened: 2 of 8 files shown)


boardfish:
Hi @rmacklin! I'd also like to contribute here - I've been working towards a solution in #1250. I think it would be ideal to choose a dependency here so that we can effectively just set it and forget it. So far, I've looked into several solutions including `github_changelog_generator`, `release-drafter`, and `git-chglog`. None of those met all of our needs, which seem to be:

- it produces correct output (`github_changelog_generator` didn't meet this)
- it is efficient and doesn't use more CI time than it should (`github_changelog_generator` didn't meet this)
- it adds to CHANGELOG.md
  `release-drafter` didn't meet this as I hoped - it generates the documentation for a single GitHub release, which it also provides to a CI pipeline as text. We'd need more manual logic to add that to the top of `CHANGELOG.md` at an appropriate time, so we could probably only do that per release rather than per merged PR. The solution given in this PR so far seems to work the same way.
- it credits those responsible for the work (`git-chglog` didn't meet this, but I have an issue open. There's been no response yet.)

We may need an original solution as you've devised here, but if you're able to find a suitable dependency that can fulfill the same role, I feel that may be preferable.

rmacklin:
Yeah, I feel you. My proposal was effectively to make `git` itself be that dependency. I did wrap it with a few lines of ruby scripting, but it still ends up being, in my opinion, very little to maintain. (Thus, I think it could still be "set it and forget it". But it also opens the door for more flexibility down the road, should we want that, compared to a purely templated changelog. As an example, the script could easily be altered to handle skipping commits that contain `[skip changelog]` or other logic.)

boardfish:
I've closed #1250 - seems like the way #1182 is going, we may need a middle ground between this and that approach.

I like where discussion is going with this - I think it's leading to a changelog solution that might be transferable across a lot of projects.

rmacklin:
Closing as stale

PR #1260, opened 2022-02-01
Add test showing that form helpers aren't used by View Components

sampart:
### Summary

This PR adds a test showing that form helpers aren't used when a View Component calls `form_tag`. CI will fail because of the failing test, but I'm envisaging that having the test will provide a helpful starting point for addressing this issue.

### Other Information

I'd expect form helpers to be picked up automatically. `form_tag_html` is an override of [this method](https://github.com/wycats/rails-api/blob/4aa40d1381fac5bc69bae6bb8e24dfb421997b40/vendor/rails/actionpack/lib/action_view/helpers/form_tag_helper.rb#L550-L553) which is called by `form_tag` directly [here](https://github.com/wycats/rails-api/blob/4aa40d1381fac5bc69bae6bb8e24dfb421997b40/vendor/rails/actionpack/lib/action_view/helpers/form_tag_helper.rb#L558) and indirectly [here](https://github.com/wycats/rails-api/blob/4aa40d1381fac5bc69bae6bb8e24dfb421997b40/vendor/rails/actionpack/lib/action_view/helpers/form_tag_helper.rb#L558).

### DIFF
--- a/test/sandbox/app/components/form_component.html.erb
+++ b/test/sandbox/app/components/form_component.html.erb
@@ -1,0 +1,1 @@
+<%= form_tag do %><% end %>
--- a/test/sandbox/app/components/form_component.rb
+++ b/test/sandbox/app/components/form_component.rb
@@ -1,0 +1,4 @@
+# frozen_string_literal: true
+
+class FormComponent < ViewComponent::Base
+end
--- a/test/sandbox/app/helpers/form_helper.rb
+++ b/test/sandbox/app/helpers/form_helper.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+module FormHelper
+  def form_tag_html(options = {})
+    safe_join [
+      "<span>Hello, World!</span>".html_safe,
+      super(options),
+    ]
+  end
+end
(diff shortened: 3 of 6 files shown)


camertron:
Thanks for the failing test @sampart. Is it related to another issue, and if so, would you mind linking to it in the description?

mhw:
I think part of the issue here is that helpers defined in `app/helpers` are handled differently from those defined in `action_view/helpers`. The ActionView helpers are included in to `ActionView::Base` and so are available directly within View Components (and their templates) as they are inherited from `ActionView::Base`. The ones from `app/helpers` are included in to an anonymous module that is then included in to an anonymous subclass of `ActionView::Base`, and on which the templates from `app/views` are compiled. This is why View Component provides the `helpers` method - it gives access to the original view context (the anonymous subclass of `ActionView::Base`) that the helpers from `app/helpers` are defined on.

Given this, I don't think redefining `form_tag_html` from `app/helpers` in a View Component context will work.

Problems with `form_tag` may be due to it using `capture` and issues that currently exist with not having a global output buffer across the original view context and the View Components. There's some work to address this [here](https://github.com/github/view_component/pull/974)

camertron:
@mhw makes sense, thanks for that additional context.

We chatted about this internally and decided to include all helpers in components starting with view_component v3.0.

primer-css:

joelhawksley:
Closing as stale ❤️

PR #1261, opened 2022-02-03
Updating preview page of docs to mention rspec

leighhalliday:
### Summary

When trying to get previews working I didn't realize that the `preview_paths` config setting had to be updated when working with rspec. I eventually found this on the `testing` page but I thought it would be better to colocate this with the preview docs because at this point I wasn't at the testing stage.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Small update to preview docs to include rspec mention
+
+    *Leigh Halliday*
+
 * Small improvements to collection iteration docs.
 
     *Brian O'Rourke*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,7 +7,7 @@ title: Changelog
 
 ## main
 
-* Small update to preview docs to include rspec mention
+* Small update to preview docs to include rspec mention.
 
     *Leigh Halliday*
 
--- a/docs/guide/previews.md
+++ b/docs/guide/previews.md
@@ -174,7 +174,7 @@ To render a source preview in a different place, use the view helper `preview_so
 
 ## Using with rspec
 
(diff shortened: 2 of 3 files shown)


juanmanuelramallo:
Isn't it already present? https://github.com/github/view_component/blob/main/docs/guide/testing.md?plain=1#L200

leighhalliday:
> Isn't it already present? https://github.com/github/view_component/blob/main/docs/guide/testing.md?plain=1#L200

Hey Juan! You're right, reason I brought it up was because I was working through the documentation on https://viewcomponent.org/guide/previews.html to get Previews set up, and didn't realize I had to go to the testing page to find the configuration I was missing. It took some fumbling around on my part to realize this and I wonder if others have been confused as well.

Maybe taking a step back, it seems like previews are both for testing & development environments, so I found it a bit odd that something displayed in the testing environment was written within the test or spec folder, that said I get it if system tests are their primary motivation.

juanmanuelramallo (approved):
Cool, makes sense!

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Small update to preview docs to include rspec mention
```
```suggestion
* Small update to preview docs to include rspec mention.
```

docs/guide/previews.md
```diff
@@ -171,3 +171,12 @@ config.view_component.show_previews_source = true
 ```
 
 To render a source preview in a different place, use the view helper `preview_source` from within your preview template or preview layout.
+
+## Using with rspec
+
+When using previews with rspec, `test/components` should be replaced with `spec/components` and the `preview_paths` setting must be updated.
```
joelhawksley, line 177:
```suggestion
When using previews with rspec, `test/components` should be replaced with `spec/components` and the `preview_paths` setting must be updated:
```

joelhawksley (approved):

PR #1262, opened 2022-02-07
Add separate slot getter/setter API proposal

BlakeWilliams:
This adds an ADR covering the reasons for breaking the existing slot API
by separating the getter/setter.

related to https://github.com/github/view_component/pull/1184


adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+Proposed
+
+## Context
+
+Currently, slots implement a single method for both getting and setting a slot.
+for example,  Given a slot named `header`:
+
+```ruby
+class MyComponent < ViewComponent::Base
+  has_one :header
```
Spone, line 20:
```suggestion
  renders_one :header
```

jonspalmer:
I love this change. The mil/with_content case has tripped me up many times and I suspect some of the inner workings of https://github.com/jonspalmer/view_component_storybook will simplify with these changes too.

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+either passing an argument or passing a block.
+
+This assumption hasn't remained valid. Specifically, the `with_content` breaks
+the assumption when you want to pass static content to a slot.
+
+e.g.
+
+```ruby
+class MyComponent < ViewComponent::Base
+  has_one :header
```
Spone, line 39:
```suggestion
  renders_one :header
```

camertron (approved):

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Consequences
+
+The largest consequence of this change is that we'll need to deprecate the old
+setter usage (`header { "Hello world!"}`) in favor of the new setter API
+(`with_header { "Hello world!" }`).
+
+I propose that we make at least one release with the new API and no deprecation
+warning followed by another release that includes the deprecation warning. This
+will give teams some time to migrate before running into deprecation warnings.
```
jonspalmer, line 100: Can we provide a DSL to retain the old behavior even once deprecated? Something backed by a class variable that could be flipped in the ApplicationComponent or in individual legacy components to allow users to defer the breaking change remediation?

sarhm:
> This adds an ADR covering the reasons for breaking the existing slot API
> by separating the getter/setter.
> 
> related to https://github.com/github/view_component/pull/1184

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+We've spoken about a few alternatives:
+
+* Making a special `NilSlot` class that responds to `with_content`.
+  * Results in extra allocations
+  * Can't treat the `NilSlot` as falsy. So `if header` would no longer work
+    even though you would expect it to.
+* Introducing an API for the `header.with_content("Hello world!")` pattern as explained above (like: `c.with_header_content("Hello world!")`):
+  * The API gap still exists and requires a specific work around for
+    `with_content`, leaving the gap for future API's.
+  * This API doesn't allow arguments to be passed 1-to-1 like the current setter API.
```
BlakeWilliams, line 90: Just bringing the conversation to the ADR, @Spone [commented in the PR](https://github.com/github/view_component/pull/1184#issuecomment-1031782586):

> Hi @BlakeWilliams! I'm wondering if you considered keeping the setter unchanged, but namespacing the getter.
> 
> My starting point here is trying to ease the migration of large codebases to the new syntax.
> 
> I think the getter is nearly always used internally to the ViewComponent, while the setter is usually called in a lot of places around the app, wherever you render an instance of the ViewComponent with the given slot.
> 
> So if only the getter is changed, you would only have to update each ViewComponent using slots (and its template(s)). Whereas if the setter is changed, it could mean updating hundreds of files. The former is probably easier to reason about (and maybe even to automate):
> 
> * for each component:
>   
>   * open the class
>   * if there are `renders_one` / `renders_many`
>     
>     * find the slots getters in the component class and/or templates and rename them
>   * else, do nothing
> 
> In terms of syntax, what about something like this:
> 
> ```ruby
>   renders_one :header
> end
> c = MyComponent.new
> 
> # setter
> c.header { "Hello world!" }
> c.header.with_content("Hello world!")
> 
> # getter (suggestions)
> c.slots.header # or 
> c.header_slot # or 
> c.header_content
> ```

I haven't explored this much since I think it's harder to find a different getter name than it is to find a different setter name that works well with the API. Building on your example, in Ruby I'd expect `header` to return the object since that aligns with `attr_reader` and the general getter/setter pattern in Ruby.

I like the idea of `c.slots`, but I think we'd want the API to be consistent and define the setter there too. I wonder if that approach would also help make the slots feature more clear to new users of the framework since it's been a relatively tough concept to teach. Somewhat related, but I sometimes think slots should be renamed to subcomponents, but that might be a different ADR.

adr/0004-slots-separate-getter-setter.md
BlakeWilliams: > Can we provide a DSL to retain the old behavior even once deprecated? Something backed by a class variable that could be flipped in the ApplicationComponent or in individual legacy components to allow users to defer the breaking change remediation?

Do you mean after we release 3.0 and remove the deprecated behavior? I'm a little hesitant since I think there will be plenty of time to resolve the deprecation warning as well as potential performance impacts.

I wonder if a middle ground would be to provide a snippet for a module that brings back the behavior? Thoughts being that it can be copy/pasted into applications and included on legacy components as-needed without requiring the library to officially support it.

We're also a good bit aways from 3.0 I believe, so it's possible circumstances will change by then and help guide us towards a specific solution here when we get there.

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+
+Date: 2022/02/07
+
+## Author
+
+Blake Williams
+
+## Status
+
+Proposed
```
joelhawksley, line 11:
```suggestion
Accepted
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Context
+
+Currently, slots implement a single method for both getting and setting a slot.
+for example,  Given a slot named `header`:
```
joelhawksley, line 16:
```suggestion
Currently, slots implement a single method for both getting and setting a slot. For example, given a slot named `header`:
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+
+```ruby
+class MyComponent < ViewComponent::Base
+  renders_one :header
+end
+
+c = MyComponent.new
+
+c.header { "Hello world!" } # sets the slot
+c.header # gets the header
```
joelhawksley, line 26: `slot` or `header`, pick one:

```suggestion
c.header { "Hello world!" } # sets the slot
c.header # gets the slot
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+This API was built with the assumption that a slot will always be defined by
+either passing an argument or passing a block.
```
joelhawksley, line 30:
```suggestion
This API was built with the assumption that a slot will always be set by passing an argument and/or passing a block.
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+
+This assumption hasn't remained valid. Specifically, the `with_content` breaks
+the assumption when you want to pass static content to a slot.
```
joelhawksley, line 33:
```suggestion
This assumption hasn't remained valid. Specifically, `with_content` breaks the assumption when you want to pass static content to a slot:
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+
+e.g.
+
```
joelhawksley, line 36:
```suggestion
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+# c.header returns nil because the getter path is being executed due to having
+# no arguments and no block passed: https://github.com/github/view_component/blob/main/lib/view_component/slotable_v2.rb#L70-L74
+#
+c.header.with_content("Hello world!") # undefined method `with_content' for nil:NilClass (NoMethodError)
+```
+
+The above example shows off the gap in the slots API via `with_content`, but it's
+likely that as the library continues to grow this gap will appear in other
+valid use-cases.
```
joelhawksley, line 52:
```suggestion
The above example shows off the gap in the slots API via `with_content`, but it's likely that as the library continues to grow this gap will appear in other valid use-cases.
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Decision
+
+To eliminate this gap in the API, I propose we split the slots API into a
+getter and setter. I think keeping the slot name as the getter makes the most
+sense, but the setter can be renamed to `with_#{slot_name}`.
```
joelhawksley, line 58:
```suggestion
Split the slots API into a getter and setter. Keeping the slot name as the getter makes the most sense, but the setter can be renamed to `with_#{slot_name}`.
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+# New API for setting slots
+c.with_header { "hello world" }
+
+# Now `with_content` is valid when defining slots
+c.with_header.with_content("Hello world!")
+```
+
+This keeps the getter and setter separate, allowing for more API flexibility
+when working with slots.
+
```
joelhawksley, line 78:
```suggestion
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,100 @@
...
+* Introducing an API for the `header.with_content("Hello world!")` pattern as explained above (like: `c.with_header_content("Hello world!")`):
+  * The API gap still exists and requires a specific work around for
+    `with_content`, leaving the gap for future API's.
+  * This API doesn't allow arguments to be passed 1-to-1 like the current setter API.
+
+## Consequences
+
+The largest consequence of this change is that we'll need to deprecate the old
+setter usage (`header { "Hello world!"}`) in favor of the new setter API
+(`with_header { "Hello world!" }`).
```
joelhawksley, line 96:
```suggestion
The largest consequence of this change is that we'll need to deprecate the old setter usage (`header { "Hello world!"}`) in favor of the new setter API (`with_header { "Hello world!" }`).
```

adr/0004-slots-separate-getter-setter.md
joelhawksley: I agree with @jonspalmer. We might want to delay v3 to make sure folks have plenty of time to migrate.

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,86 @@
+# Separate Slot Getters and Settings
+
+Date: 2022/02/07
```
joelhawksley, line 3:
```suggestion
Date: 2022/03/22
```

adr/0004-slots-separate-getter-setter.md
```diff
@@ -0,0 +1,86 @@
+# Separate Slot Getters and Settings
```
joelhawksley, line 1:
```suggestion
# Separate Slot Getters and Setters
```

joelhawksley (approved):

PR #1267, opened 2022-02-10
Use path instead of absolute_path for inside eval

HParker:
This prevents getting a nil absolute_path for code executed inside of an eval

### Summary

Absolute path is not always available on your callers, but path is more likely to be available. This especially fixes files that are evaluated which in ruby 3.1 never have absolute paths after this change in ruby: https://github.com/ruby/ruby/commit/64ac984129a7a4645efe5ac57c168ef880b479b2

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Fix path handling for evaluated view components
+
+    *Adam Hess*
+
 * Ensure i18n scope is a symbol to protect lookups.
 
     *Simon Fish*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -7,7 +7,7 @@ title: Changelog
 
 ## main
 
-* Fix path handling for evaluated view components
+* Fix path handling for evaluated view components that broke in Ruby 3.1.
 
     *Adam Hess*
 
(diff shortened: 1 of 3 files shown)


BlakeWilliams (commented):
This looks great!

Would you mind adding a test and a changelog entry for this change? 🙇

HParker:
Thanks @BlakeWilliams updated!

HParker:
oh blah changelog conflict, one sec...

test/view_component/base_test.rb
```diff
@@ -90,4 +90,20 @@ def test_does_not_render_additional_newline_with_render_in
...
+          @cta = cta
+          @title = title
+        end
+        private
+
+        attr_reader :cta, :title
+      end
+    RUBY
+
+    eval(source)
```
joelhawksley, line 107: Looks like we need to add a rubocop exception for this line for our linters to pass ❤️

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Fix path handling for evaluated view components
```
```suggestion
* Fix path handling for evaluated view components that broke in Ruby 3.1.
```

test/view_component/base_test.rb
joelhawksley, line 107:
```suggestion
    eval(source) # rubocop:disable Security/Eval
```

joelhawksley (approved):

Issue #1042, opened 2021-08-17
Include Translations support by default


joelhawksley:
## Feature request

Now that ViewComponent translations have been out for a while, I think we should include them by default. What do you think @elia?

### Motivation

We generally only leave features as opt-in until we're confident that they are stable.

elia:
Sounds good, I'd just like to ensure that adding the global translation support provided by view_component-contrib still works

joelhawksley:
@palkan might you be able to provide your perspective here given @elia's comment?

palkan:
Seems working fine: https://github.com/palkan/view_component-contrib/commit/5b92196d1d796399a2c2f24e78e09cdaca29b1c1 🚢

elia:
Great!
I'll prepare a PR shortly 👍

elia:
@joelhawksley I have one locally that's half-backed, guess it's time to finish it! 😅 👍

PR #1268, opened 2022-02-10
Move the `Translatable` module into `Base`

elia:
### Summary

Closes https://github.com/github/view_component/issues/1042


docs/guide/translations.md
```diff
@@ -4,7 +4,7 @@ title: Translations
 parent: Guide
 ---
 
-# Translations (experimental)
+# Translations
 
 To use experimental support for `I18n` translations, include `ViewComponent::Translatable`:
```
joelhawksley, line 9: I think we'll need to rework this section a bit 😄

lib/view_component/base.rb
```diff
@@ -37,7 +38,7 @@ class Base < ActionView::Base
     # For example, one might compile component-specific assets at this point.
     # @private TODO: add documentation
     def self._after_compile
-      # noop
+      super if defined?(super)
```
joelhawksley, line 41: Do we need this change?

lib/view_component/translatable.rb
```diff
@@ -21,7 +21,7 @@ def i18n_scope
       end
 
       def _after_compile
-        super
+        super if defined?(super)
```
joelhawksley, line 24: Or this one?

docs/guide/translations.md
elia: For sure! 😉

lib/view_component/base.rb
camertron: Hey @elia, this is _super_ (haha, get it?!) interesting! First, I assume by `AC::Base` you actually mean `VC::Base`, is that right?

I think what you've got here is fine, although it makes me a little nervous that some future module may forget to call `super` and therefore interfere with `Translatable`'s functionality. Fortunately our tests should catch something like that.

It looks like we're uncovering the fact that relying on the inheritance hierarchy to manage lifecycle events is inherently brittle. It would probably be easier and more flexible to implement them using something like ActiveModel's callbacks API. Something to think about for the future.

BlakeWilliams: > It looks like we're uncovering the fact that relying on the inheritance hierarchy to manage lifecycle events is inherently brittle. It would probably be easier and more flexible to implement them using something like ActiveModel's callbacks API. Something to think about for the future.

I think that's a good idea. What are y'alls thoughts on adding a hook to the compiler itself so that we can use it for translations and keep translation compilation logic out of components themselves?

camertron: I like it!

elia: > Hey @elia, this is super (haha, get it?!) interesting! First, I assume by AC::Base you actually mean VC::Base, is that right?

Indeed! Corrected the comment, thanks!

> It looks like we're uncovering the fact that relying on the inheritance hierarchy to manage lifecycle events is inherently brittle. It would probably be easier and more flexible to implement them using something like ActiveModel's callbacks API. Something to think about for the future.

Absolutely! I added a commit with a dedicated callback. Will make sense to handle the callback thing in a different PR.

docs/guide/instrumentation.md
```diff
@@ -17,9 +17,12 @@ config.view_component.instrumentation_enabled = true
...
 ```ruby
-ActiveSupport::Notifications.subscribe("!render.view_component") do |*args|
-  event = ActiveSupport::Notifications::Event.new(*args)
+ActiveSupport::Notifications.subscribe("!render.view_component") do |event|
   event.name    # => "!render.view_component"
   event.payload # => { name: "MyComponent", identifier: "/Users/mona/project/app/components/my_component.rb" }
+
+  ApplicationComponent.logger.debug \
+    "  Rendered #{bold}#{event.payload[:name]}#{clear}" \
+    " (Duration: #{event.duration.round(1)}ms | Allocations: #{event.allocations})"
```
joelhawksley, line 26: I think we can remove these lines. Perhaps this could be added in another PR?
```suggestion
```

joelhawksley (approved):

docs/guide/instrumentation.md
elia: Done! @joelhawksley you think we can merge? Or should be part of v3?

docs/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Move the `Translatable` module out of experimental mode.
```
```suggestion
* Include the `Translatable` module by default.
```

docs/guide/translations.md
```diff
@@ -4,18 +4,9 @@ title: Translations
...
-
-```ruby
-# app/components/example_component.rb
-module ExampleComponent < ApplicationComponent
-  include ViewComponent::Translatable
-end
-```
-
-Add a sidecar YAML file with translations for the component:
+To use `I18n` translations, add a sidecar YAML file with translations for the component:
```
joelhawksley, line 9:
```suggestion
To use `I18n` translations, add a sidecar YAML file:
```

docs/guide/translations.md
```diff
@@ -51,7 +42,7 @@ Global Rails translations are available as well:
 <%= t("my.global.translation") %>
 ```
 
-Global translations shadowed by sidecar translations can be accessed with `helpers` or `I18n`:
+Global translations can still be accessed via `helpers` or `I18n`:
```
joelhawksley, line 45:
```suggestion
Global translations can be accessed via `helpers` or `I18n`:
```

joelhawksley (approved):

joelhawksley:
Thanks @elia! I view this as a feature-add and am happy to land this now ❤️

PR #1269, opened 2022-02-10
Move preview controller logic into concern

sammyhenningsson:
### Make it easier to get ViewComponent previews into an existing app. 
If we want previews to be shown inside existing application layouts, which requires application logic. E.g. url helpers and helper methods like `current_user` etc. Then we need write a custom controller, and either write all the preview stuff from scratch or copy and paste from `ViewComponentsController`. If we instead move the content of `ViewComponentsController` inside a module, then we could just mixin this logic into the application code. Then applications only need something like this:
```ruby
#app/controllers/view_component_preview_controller.rb
class ViewComponentPreviewController < ApplicationController
  include ViewComponent::PreviewActions
end
```
```ruby
#config/application.rb
config.view_component.preview_controller = "ViewComponentPreviewController"
```
And we are free to use all the application specific stuff that comes with `ApplicationController`


docs/CHANGELOG.md
```diff
@@ -27,6 +27,10 @@ title: Changelog
 
     *Peter Goldstein*
 
+* Move preview logic to module, for easier app integration.
```
joelhawksley, line 30:
```suggestion
* Move preview logic to module for easier app integration.
```

joelhawksley (approved):

PR #1271, opened 2022-02-11
Update CI config to latest Rails 7.0 version

Spone:


Gemfile.lock
```diff
@@ -194,11 +167,32 @@ GEM
...
+      actionpack (= 7.0.2.2)
+      actiontext (= 7.0.2.2)
+      actionview (= 7.0.2.2)
+      activejob (= 7.0.2.2)
+      activemodel (= 7.0.2.2)
+      activerecord (= 7.0.2.2)
+      activestorage (= 7.0.2.2)
+      activesupport (= 7.0.2.2)
+      bundler (>= 1.15.0)
+      railties (= 7.0.2.2)
```
elia, line 183: If we use the latest Rails 7 for the top-level gemfile.lock maybe that should also be reflected in the Gemfile.

Also since we have Appraisal we might think of ditching support for the RAILS_VERSION env var and rely on the different gemfiles, it's as easy as switching from `export RAILS_VERSION=6.1` to `export BUNDLE_GEMFILE=gemfiles/rails_6.1.gemfile`. Thoughts on this?

Spone: Can you please have a look at https://github.com/github/view_component/pull/1230 and leave a review there?

Spone: > If we use the latest Rails 7 for the top-level gemfile.lock maybe that should also be reflected in the Gemfile.

Do you think I should update the Gemfile as following?

```diff
-rails_version = "#{ENV['RAILS_VERSION'] || 'main'}"
+rails_version = "#{ENV['RAILS_VERSION'] || '~> 7.0.0'}"
```

BlakeWilliams: my understanding is that we're using `main` so we can test against the latest Rails which has caught some potential bugs, both in ViewComponent and Rails

Spone: When running in CI, the `RAILS_VERSION` env var is already set with `main` by the versions matrix:

- https://github.com/github/view_component/blob/ci-config-updates/.github/workflows/ci.yml#L30
- https://github.com/github/view_component/blob/ci-config-updates/.github/workflows/ci.yml#L66

But when running the specs locally (where `RAILS_VERSION` is not defined usually), Bundler will complain that Rails is not up-to-date if we stay on `main` by default:

```
$ bundle exec appraisal rake                     
The git source https://github.com/rails/rails is not yet checked out. Please run `bundle install` before trying to start your application
```
```
$ bundle exec appraisal install
The git source https://github.com/rails/rails is not yet checked out. Please run `bundle install` before trying to start your application
```
```
$ bundle install
Fetching https://github.com/rails/rails
Fetching gem metadata from https://rubygems.org/.........
Resolving dependencies...
Bundler could not find compatible versions for gem "activesupport":
  In snapshot (Gemfile.lock):
    activesupport (= 7.0.2.2)

  In Gemfile:
    better_html (~> 1) was resolved to 1.0.16, which depends on
      actionview (>= 4.0) was resolved to 7.0.2.2, which depends on
        activesupport (= 7.0.2.2)

    rails was resolved to 7.1.0.alpha, which depends on
      activesupport (= 7.1.0.alpha)

Running `bundle update` will rebuild your snapshot from scratch, using only
the gems in your Gemfile, which may resolve the conflict.
```

Then you need to run `bundle update` to be able to run the `appraisal` commands. But currently, that will bump Rails to `7.1.0.alpha` in the `Gemfile.lock`.

That's why I would recommend the change above.

elia: @Spone I agree 💯, ideally local development should be as easy as:

- clone repo
- `script/bootstrap` (which should include `appraisal install`)
- `bundle exec rake` or `bundle exec appraisal rake` to run tests

One alternative to specifying `~> 7.0.0` is to use `> 0.a` which, by bundler `gemspec` rules, will be resolved to the minimal supported version.

I would also suggest to remove `Gemfile.lock` from the repo, that uncovered bugs for some of my gems that didn't work with the minimal supported version, forcing me to bump the requirement or update the code.

joelhawksley:
@Spone would you be up for bringing this PR back up to date?

PR #1273, opened 2022-02-11
Silence deprecation notices in CI

Spone:
### Summary

The CI logs include a lot of deprecation warnings. In order to make them more readable, let's silence these warnings.

I took a pretty basic approach: check for `ENV["CI"]` which is [defined automatically](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables).

### Other Information

This is an alternative to #586

PR #1277, opened 2022-02-13
Add ViewComponent::Deprecation

elia:
### Summary

This should allow consumers to see deprecation warnings that correctly mention the next version of VC.

Silencing is already a feature of AS::Deprecation

Closes #586 and closes #1273 (this was built starting from those PRs).

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,3 +25,7 @@ Layout/LineLength:
 
 Layout/SpaceBeforeBrackets:
   Enabled: true
+
+Style/FrozenStringLiteralComment:
+  Exclude:
+    - "**/*.jbuilder" # not yet supported inside jbuilder templates
--- a/lib/view_component.rb
+++ b/lib/view_component.rb
@@ -23,7 +23,7 @@ end
 # :nocov:
 if defined?(ViewComponent::Engine)
   ViewComponent::Deprecation.warn(
-    "This manually engine loading is deprecated and will be removed in v3.0.0. " \
+    "Manually loading the engine is deprecated and will be removed in v3.0.0. " \
     "Remove `require \"view_component/engine\"`."
   )
 elsif defined?(Rails::Engine)
--- a/lib/view_component.rb
+++ b/lib/view_component.rb
@@ -10,6 +10,7 @@ module ViewComponent
   autoload :Compiler
   autoload :CompileCache
   autoload :ComponentError
+  autoload :Deprecation
   autoload :Instrumentation
   autoload :Preview
   autoload :PreviewTemplateError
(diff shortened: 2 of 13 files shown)


.rubocop.yml
```diff
@@ -22,3 +23,7 @@ Layout/DotPosition:
 Layout/LineLength:
   Max: 120
   Enabled: true
+
+Style/FrozenStringLiteralComment:
+  Exclude:
+    - test/sandbox/app/components/jbuilder_component.json.jbuilder # not yet supported inside jbuilder templates
```
Spone, line 29: What about supporting any future .jsbuilder file as well?

```suggestion
    - **/*.jbuilder # not yet supported inside jbuilder templates
```

.rubocop.yml
```diff
@@ -9,6 +9,7 @@ AllCops:
     - gemfiles/**/*
     - test/tmp/**/*
     - vendor/bundle/**/*
+  NewCops: disable
```
Spone, line 12: I'm wondering if it's the right move.

lib/view_component/deprecation.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+require "active_support/deprecation"
+
+module ViewComponent
+  Deprecation = ActiveSupport::Deprecation.new("3.0", "ViewComponent")
```
Spone, line 6: We will need to bump the version here when releasing v3.0, right? If yes, we could add this step to `script/release`.

.rubocop.yml
elia: Moving this to a different PR

lib/view_component/polymorphic_slots.rb
```diff
@@ -25,23 +25,26 @@ def renders_many(slot_name, callable = nil)
       end
 
       def register_polymorphic_slot(slot_name, types, collection:)
+        unless types.empty?
```
Spone, line 28: This change feels unrelated, is it aiming at solving a warning?

elia: Yes, the warning message is reported in the commit, but basically the getter method was redefined for each type.
The `unless types.empty?` guard is there just to keep the exact same behavior, i.e. if `types` was empty the getter is not defined.

Spone (approved):

lib/view_component.rb
```diff
@@ -21,7 +22,7 @@ module ViewComponent
 
 # :nocov:
 if defined?(ViewComponent::Engine)
-  ActiveSupport::Deprecation.warn(
+  ViewComponent::Deprecation.warn(
     "This manually engine loading is deprecated and will be removed in v3.0.0. " \
```
camertron, line 26:
```suggestion
    "Manually loading the engine is deprecated and will be removed in v3.0.0. " \
```

camertron (approved):

maxbeizer:
👋 thanks for picking up what I started and never got back around to 🙇 💖

PR #1194, opened 2021-12-10
Allow passing component attributes to preview generator

bobmaerten:
### Summary

This PR allows the preview generator to get the provided component generator attributes.
Why?
Because, the default generated preview when passing attributes is not valid and fails tests, and it's not cool to generate failing tests files. ;)

### Other Information

This PR move preview generator tests to its own file as for other generators.
A config option has been added too, in order to enable project-wide preview generation.

PR #1278, opened 2022-02-14
Add frozen_string_literal statement to all generated .rb files

bobmaerten:
As @boardfish evoked in https://github.com/github/view_component/pull/1194#issuecomment-1036405220:

> it's relatively safe to say that generated components will be new code in new apps, which are likely to default to `frozen_string_literal: true`

Here are the generated .rb files updated to comply with ViewComponent self coding style.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -11,6 +11,10 @@ title: Changelog
 
     *Simon Fish*
 
+* Ensure all generated `.rb` files include `# frozen_string_literal: true` statement.
+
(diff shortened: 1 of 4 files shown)


Spone (commented):
Can you add a Changelog entry?

bobmaerten:
> Can you add a Changelog entry?

Damned, you quick! I was just adding a description to this PR ^^

Spone (approved):

BlakeWilliams (approved):

PR #1279, opened 2022-02-14
Disable new cops by default?

elia:
### Summary

Not sure if there's already a strategy, but each time new cops are added to rubocop we'll get a warning.

I'd be in favor of disabling by default and cherry-picking additions, or if there's an official gihtub set of cops follow that.
@Spone correctly [pointed out](https://github.com/github/view_component/pull/1277#discussion_r805672184) it might not be the best approach or necessarily be a shared perspective, so I move the commit here in order to reach a consensus.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -9,6 +9,7 @@ AllCops:
     - gemfiles/**/*
     - test/tmp/**/*
     - vendor/bundle/**/*
+  NewCops: disable
 
 Gemspec/OrderedDependencies:
   Enabled: true


camertron:
Hey @elia, thanks for bringing this up! Our current configuration decision to accept all new cops is intentional because, as you suspected, we're using the rubocop-github gem to ensure consistency across all our codebases. When a new cop is introduced, we want to make sure view_component conforms. Please feel free to reopen if you'd like to continue the discussion.

elia:
It's all good, I was annoyed by the rubocop warning 😅 and though it was worth checking 👍

PR #1281, opened 2022-02-14
Predicate methods for slots

### LINKED ISSUE #1275: Predicate methods for slots
## Feature request

I would like to suggest adding predicate methods for slots.

For instance:

```rb
  renders_one :header
  renders_many :posts
end
```

```erb
<%# blog_component.html.erb %>
<% if header? %>
  <h1><%= header %></h1>
<% end %>

<% if posts? %>
  <ol>
    <% posts.each do |post| %>
      <li><%= post %></li>
    <% end %>
  </ol>
<% end %>
```

They could also be used in `#render?`:

```rb
  renders_one :header
  renders_many :posts

  def render?
    header? || posts?
  end
end
```

### Motivation

Most of the time, when slots are optional, I find myself testing their presence with `#present?` because I want to render a wrapper element conditionally. I think we could add some syntactic sugar for this, similar to the `#{attribute}?` method [provided by ActiveRecord](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/attribute_methods/query.rb).

Spone:
### Summary

This implements predicate methods for slots, as described in #1275.

- [x] define predicate methods on slot declaration
- [x] add validation to make sure slot names do not end with `?`
- [x] add documentation
- [x] add changelog entry

### Open questions

- Do you think it's possible / a good idea to try to warn users when they define a `#{slot_name}?` method themselves, so they notice that the library now provide this feature?

### Other Information

Close #1275

### DIFF
--- a/lib/view_component/slotable_v2.rb
+++ b/lib/view_component/slotable_v2.rb
@@ -9,7 +9,7 @@ module ViewComponent
 
     RESERVED_NAMES = {
       singular: %i[content render].freeze,
-      plural: %i[contents].freeze,
+      plural: %i[contents renders].freeze,
     }.freeze
 
     # Setup component slot state
--- a/lib/view_component/slotable_v2.rb
+++ b/lib/view_component/slotable_v2.rb
@@ -75,6 +75,10 @@ module ViewComponent
         end
         ruby2_keywords(slot_name.to_sym) if respond_to?(:ruby2_keywords, true)
 
+        define_method "#{slot_name}?" do
+          send(slot_name).present?
+        end
+
         register_slot(slot_name, collection: false, callable: callable)
       end
 
@@ -138,6 +142,10 @@ module ViewComponent
               set_slot(slot_name, nil, **args, &block)
             end
           end
+        end
+
+        define_method "#{slot_name}?" do
+          send(slot_name).present?
         end
 
         register_slot(slot_name, collection: true, callable: callable)
--- a/lib/view_component/slotable_v2.rb
+++ b/lib/view_component/slotable_v2.rb
@@ -241,7 +241,9 @@ module ViewComponent
       def raise_if_slot_ends_with_question_mark(slot_name)
         if slot_name.to_s.ends_with?("?")
           raise ArgumentError.new(
-            "#{self} declares a slot named #{slot_name}, which ends with a question mark. This is not allowed because the ViewComponent framework already provides predicate methods ending in `?`.\n\n" \
+            "#{self} declares a slot named #{slot_name}, which ends with a question mark.\n\n"\
+            "This is not allowed because the ViewComponent framework already provides predicate "\
+            "methods ending in `?`.\n\n" \
             "To fix this issue, choose a different name."
           )
         end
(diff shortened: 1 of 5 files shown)


test/sandbox/app/components/slots_v2_component.html.erb
```diff
@@ -1,12 +1,16 @@
 <div class="card <%= @classes %>">
+  <% if title? %>
```
Spone, line 2: Should I create a separate sandbox component to test this behavior, instead of altering the existing one?

lib/view_component/slotable_v2.rb
```diff
@@ -227,6 +237,17 @@ def raise_if_slot_registered(slot_name)
           )
         end
       end
+
+      def raise_if_slot_ends_with_question_mark(slot_name)
```
BlakeWilliams, line 241: This is great to have! Big 👍

test/sandbox/app/components/slots_v2_component.html.erb
BlakeWilliams: I think we have other tests that cover getting slots, so I'm fine with this change 👍

BlakeWilliams (commented):
Nice! This makes sense to me 👍

boardfish (commented):
Love this ❤️

lib/view_component/slotable_v2.rb
```diff
@@ -215,6 +224,7 @@ def validate_singular_slot_name(slot_name)
           )
         end
 
+        raise_if_slot_ends_with_question_mark(slot_name)
```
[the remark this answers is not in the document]
Spone: Will add this

lib/view_component/slotable_v2.rb
```diff
@@ -140,6 +144,10 @@ def renders_many(slot_name, callable = nil)
           end
         end
 
+        define_method "#{slot_name}?" do
+          send(slot_name).present?
```
BlakeWilliams, line 148:
```suggestion
          get_slot(slot_name).present?
```

same comment as above, re: performance

camertron (approved):
Awesome! 🎉  Pending comments from @BlakeWilliams and co.

lib/view_component/slotable_v2.rb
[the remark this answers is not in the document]
Spone: Done in 34528e3

lib/view_component/slotable_v2.rb
```diff
@@ -7,6 +7,11 @@ module ViewComponent
   module SlotableV2
     extend ActiveSupport::Concern
 
+    RESERVED_NAMES = {
+      singular: %i[content render].freeze,
+      plural: %i[contents].freeze,
```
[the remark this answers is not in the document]
BlakeWilliams: non-blocking, but I'm 👍 on parity between the two.

BlakeWilliams (approved):
This looks great! ✨

joelhawksley:
@anthonyshchang - @Spone is right! I honestly don't know how we made it this far without a `content?` method but we should totally add one!

PR #1282, opened 2022-02-14
Update benchmarks to render more than a single time

BlakeWilliams:
Previously we would render a single component or a single partial from a
controller context to benchmark performance. This is valid, but isn't as
realistic as rendering more components/partials inside the rendering
path.

This updates the benchmarks to render multiple of the renderer type
being benchmarked in an attempt to get closer to simulating a real page
load.

* Updates both components to render another component 50 times
* Updates partial to render another partial 50 times


camertron (approved):
Nice!

PR #1287, opened 2022-02-15
Add Cometeer and Framework to users list

elia:
cc @filippoliverani


Spone:
Really awesome work on Cometeer :open_mouth: 

Should we start sorting the list alphabetically (while keeping Github on top)?

joelhawksley:
> Should we start sorting the list alphabetically (while keeping Github on top)?

Yes to alphabetically. No need to put GitHub on top ❤️

joelhawksley (approved):

PR #1290, opened 2022-02-21
Support returning arrays from translations

foca:
When i18n finds a YAML array as a value, it returns the Array as is. ActionView allows this, and the `#translate` helper in AV will correctly handle "HTML-safe" keys when the value is an Array, by calling `#html_safe` on each element.

However, ViewComponent assumes translations can only be strings, and raises an error when this is not the case.

This PR lifts the code from ActionView to handle HTML-safe translations so we can correctly escape both strings and arrays, to avoid surprising anyone that is extracting components from existing views that depend on this AV behavior.

### DIFF
--- a/lib/view_component/translatable.rb
+++ b/lib/view_component/translatable.rb
@@ -82,7 +82,7 @@ module ViewComponent
         end
 
         if HTML_SAFE_TRANSLATION_KEY.match?(key)
-          translated = translated.html_safe # rubocop:disable Rails/OutputSafety
+          translated = html_safe_translation(translated)
         end
 
         translated
@@ -95,6 +95,14 @@ module ViewComponent
     # Exposes .i18n_scope as an instance method
     def i18n_scope
       self.class.i18n_scope
+    end
+
+    def html_safe_translation(translation)
+      if translation.respond_to?(:map)
+        translation.map { |element| element.respond_to?(:html_safe) ? element.html_safe : element } # rubocop:disable Rails/OutputSafety
+      else
+        translation.respond_to?(:html_safe) ? translation.html_safe : translation # rubocop:disable Rails/OutputSafety
+      end
     end
   end
 end
--- a/lib/view_component/translatable.rb
+++ b/lib/view_component/translatable.rb
@@ -100,8 +100,10 @@ module ViewComponent
     def html_safe_translation(translation)
       if translation.respond_to?(:map)
         translation.map { |element| html_safe_translation(element) }
+      elsif translation.respond_to?(:html_safe)
+        translation.html_safe # rubocop:disable Rails/OutputSafety
       else
-        translation.respond_to?(:html_safe) ? translation.html_safe : translation # rubocop:disable Rails/OutputSafety
+        translation
       end
     end
   end
--- a/lib/view_component/translatable.rb
+++ b/lib/view_component/translatable.rb
@@ -100,10 +100,8 @@ module ViewComponent
     def html_safe_translation(translation)
       if translation.respond_to?(:map)
         translation.map { |element| html_safe_translation(element) }
-      elsif translation.respond_to?(:html_safe)
-        translation.html_safe # rubocop:disable Rails/OutputSafety
       else
-        translation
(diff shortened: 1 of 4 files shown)


elia (approved):
👍👍👍

lib/view_component/translatable.rb
```diff
@@ -96,5 +96,13 @@ def translate(key = nil, **options)
     def i18n_scope
       self.class.i18n_scope
     end
+
+    def html_safe_translation(translation)
+      if translation.respond_to?(:map)
+        translation.map { |element| element.respond_to?(:html_safe) ? element.html_safe : element } # rubocop:disable Rails/OutputSafety
```
joelhawksley, line 102: Could we simplify this line by:

```suggestion
        translation.map { |element| html_safe_translation(element) }
```

That way we'd keep the two branched in sync.

foca: Good callout! Applied :)

foca: In fact, simplified it a bit more: the ternary inside the `if` was a bit weird to begin with. Now it's a simple 3-branch if statement.

camertron (approved):

camertron:
Hey @foca, it looks like there's one untested branch of `html_safe_translation`, would you mind adding a test case for it?

foca:
> Hey @foca, it looks like there's one untested branch of `html_safe_translation`, would you mind adding a test case for it?

Yeah, been thinking what's the best solution here.

The branch is _essentially_ unreachable code, **as long as we're in Rails**. ActiveSupport [defines `html_safe?` on Object](https://github.com/rails/rails/blob/9d65858d7daa9e2310fa68476ff78550cf1fb714/activesupport/lib/active_support/core_ext/string/output_safety.rb#L121-L125), which means pretty much everything that can be returned from the i18n backend will always respond to `html_safe?` when called within Rails.

Digging through Rails' history, the change that introduced the `respond_to?(:html_safe)` [was introduced in 3.0](https://github.com/rails/rails/commit/f53a6d814966e0ea039bde680b0fc3814236641c) (all the way back in 2010). From the commit message and particularly the tests, it seems the intent of that `respond_to?` check was to handle Arrays. Since [explicit support for enumerables was added for Rails 6](https://github.com/rails/rails/commit/42c353705abf133b76dd989bb477af449db8f9fb), it seems that the case where `respond_to?(:html_safe)` became obsolete.

That is obviously depending on `active_support/core_ext/string/output_safety.rb` being loaded, which adds `Object#html_safe?`. Assuming ViewComponent is tied enough to ActionView that this is a non-problem, I think the best course of action is to just remove that code branch, and expect that anything that can feasibly be returned from the I18n backend will respond to `html_safe?`

WDYT?

(And also to open a PR on AV to clean up that conditional)

lib/view_component/translatable.rb
```diff
@@ -96,5 +96,13 @@ def translate(key = nil, **options)
     def i18n_scope
       self.class.i18n_scope
     end
+
+    def html_safe_translation(translation)
+      if translation.respond_to?(:map)
+        translation.map { |element| html_safe_translation(element) }
+      else
+        translation.html_safe # rubocop:disable Rails/OutputSafety
```
boardfish, line 104: @foca I think removing the conditional is a reasonable conclusion, so long as it's effectively documented. Maybe a comment to this effect will help:

```suggestion
        # It's assumed here that objects loaded by the i18n backend will respond to `#html_safe?`.
        # It's reasonable that if we're in Rails, `active_support/core_ext/string/output_safety.rb`
        # will provide this to `Object`.
        translation.html_safe # rubocop:disable Rails/OutputSafety
```

boardfish (approved):

lib/view_component/translatable.rb
foca: Agreed that it's a good idea to document it. Thanks.

elia (approved):

joelhawksley:
@foca I think we're right to remove the unreachable branch. If we run into a case where we need it, we'll have a better idea of how to test it!

PR #1295, opened 2022-03-02
Fix XSS vulnerability when using HTML-safe translations and interpolation arguments

camertron:
### Summary

Fix XSS vulnerability when using HTML-safe translations and interpolation arguments


lib/view_component/translatable.rb
```diff
@@ -66,6 +67,8 @@ def translate(key = nil, **options)
       return super unless i18n_backend
       return key.map { |k| translate(k, **options) } if key.is_a?(Array)
 
+      html_escape_translation_options!(options)
+
```
elia, line 71: I looked up the rails implementation and seems like this should be applied under `if HTML_SAFE_TRANSLATION_KEY.match?(key)` a few lines below, to avoid double escaping

elia (changes requested):

lib/view_component/translatable.rb
camertron: Can you describe a test case that would result in double-escaping? I'm having trouble thinking of when this would happen.

elia: I thought something like:

```yml
en:
  foo: "baz %{bar}"
```

```erb
<%= t('.foo', bar: '<br>') %>
```

in which the `bar` option gets escaped and then the erb tag will escape again, didn't try it tho

camertron: Ok that's what I thought. I don't understand why, but I wrote a test case for this and don't see any double-escaping. Just in case though, I added the necessary guard to the code to match Rails.

camertron:
NOTE: v2.49.1 and v2.31.2 have been manually released.

PR #1296, opened 2022-03-02
Fix XSS vulnerability when using HTML-safe translations and interpolation arguments

camertron:
### Summary

Fix XSS vulnerability when using HTML-safe translations and interpolation arguments


BlakeWilliams (approved):

camertron:
This can't be merged, but v2.49.1 has been released.

PR #1307, opened 2022-03-10
Use a global output buffer

camertron:
### The Problem

Since the beginning, ViewComponent has striven for deep integration with ActionView. One of the known issues however has been using Rails' form helpers with components. Consider this template for the `LabelComponent`:

```html+erb
<div>
  <%= form.label :published do %>
    <%= form.check_box :published %>
  <% end %>
</div>
```

Now let's render it from the `FormForComponent`:

```html+erb
<%= form_for Post.new, url: "/" do |form| %>
  <%= render LabelComponent.new(form: form) %>
<% end %>
```

The latest version of ViewComponent produces the following HTML:

```html
<form class="new_post" id="new_post" action="/" accept-charset="UTF-8" method="post">
  <input name="utf8" type="hidden" value="&#x2713;" autocomplete="off" />
  <div>
    <input name="post[published]" type="hidden" value="0" autocomplete="off" />
    <input type="checkbox" value="1" name="post[published]" id="post_published" />
    <label for="post_published"></label>
  </div>
</form>
```

Notice that the `<label>` tag appears below the inputs even though it's supposed to appear above them. This happens because components render to their own output buffer while the form helpers (and indeed the rest of ActionView) render to another one.

### Potential Solution

The solution proposed in this PR is to use a single, global output buffer. For now it continues to make sense for ViewComponents to render into their own buffer (mostly because they inherit from `ActionView::Base`). Accordingly I've introduced a data structure called the `OutputBufferStack` that is used in place of the usual `ActionView::OutputBuffer`. Components push new `OutputBuffer`s onto the stack whenever they are rendered. The buffer at the top of the stack is the one all `ActionView::Base`s write to, making it appear as if there is only one, global buffer. Instances of `OutputBufferStack` behave exactly like `OutputBuffer` instances, meaning they can be swapped in and out without any changes to ActionView. Moreover, only component classes need this behavior, so no Rails monkeypatches are required.

The global output buffer behavior is opt-in and available either by manually `prepend`ing the `ViewComponent::GlobalOutputBuffer` module into individual component classes, or globally by setting the `config.view_component.use_global_output_buffer` Rails configuration setting.

With the global output buffer enabled, the following HTML is produced for the example above:

```html
<form class="new_post" id="new_post" action="/" accept-charset="UTF-8" method="post">
  <input name="utf8" type="hidden" value="&#x2713;" autocomplete="off" />
  <div>
    <label for="post_published">
      <input name="post[published]" type="hidden" value="0" autocomplete="off" />
      <input type="checkbox" value="1" name="post[published]" id="post_published" />
    </label>
  </div>
</form>
```

It works! 🎉 

## Other Information

Unfortunately managing the global output buffer adds some overhead. My benchmarks show that it decreases performance by about 18%. While not ideal, I've tried to mitigate the performance hit by submitting [this PR](https://github.com/github/view_component/pull/1302) in tandem, which increases overall performance by about 15%. Here are the raw numbers for just this PR (run `bundle exec ruby performance/global_output_buffer_benchmark.rb`):

```
           component   239.000  i/100ms
  component with GOB   242.000  i/100ms
           component      2.820k (± 4.8%) i/s -     28.202k in  10.026443s
  component with GOB      2.339k (± 3.9%) i/s -     23.474k in  10.050467s

Comparison:
           component:     2819.8 i/s
  component with GOB:     2339.5 i/s - 1.21x  (± 0.00) slower
```

We hope to fully integrate the global output buffer in ViewComponent v3, which will mean merging all the code from `ViewComponent::GlobalOutputBuffer` into `ViewComponent::Base`. Doing so should result in fewer method calls and get us back that remaining 3% of performance.


lib/view_component/global_output_buffer.rb
```diff
@@ -0,0 +1,61 @@
...
+      if @output_buffer && other_buffer
+        @output_buffer.replace(other_buffer)
+      else
+        # :nocov:
+        @output_buffer = other_buffer
+        # :nocov:
+      end
+    end
+
+    def with_output_buffer(buf = nil)
```
BlakeWilliams, line 39: Do we need to override this?

If-so, is there any issue with partials using the existing `with_output_buffer` method with the output buffer stack?

BlakeWilliams (approved):
I left a few comments, but otherwise this LGTM!

test/test_helper.rb
```diff
@@ -55,6 +55,38 @@ def with_preview_controller(new_value)
...
 
+def module_parent_for(mod)
+  # activesupport >= 6
+  return mod.module_parent if mod.respond_to?(:module_parent)
+
+  # activesupport < 6
+  mod.parent
+end
+
+def with_global_output_buffer(components: [])
```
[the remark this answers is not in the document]
camertron: I like that! The only problem is the actionview compatibility tests will fail, but I guess we can skip them if `GLOBAL_OUTPUT_ENABLED` is set.

lib/view_component/global_output_buffer.rb
camertron: Yeah, because any view that calls `#capture` will call this method. The original temporarily swaps `self.output_buffer` whereas the buffer stack needs to be `#push`ed to. As far as I can tell, form helpers and the fragment caching stuff are the only things that `#capture` though. In other words if you're not doing either of those things, this code never gets invoked.

lib/view_component/slot_v2.rb
```diff
@@ -45,18 +45,12 @@ def to_s
           if defined?(@__vc_content_set_by_with_content)
             @__vc_component_instance.with_content(@__vc_content_set_by_with_content)
 
-            view_context.capture do
-              @__vc_component_instance.render_in(view_context)
-            end
+            @__vc_component_instance.render_in(view_context)
```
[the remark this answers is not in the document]
camertron: I... I think so? Tests pass and github.com looks good? But I don't understand why, which is kind of freaking me out.

percyhanna:
Unfortunately, this branch does not work with our use case, while the original PR, https://github.com/github/view_component/pull/974, does work fine. I have a simple test case in our application that was used to test the "fixed" behaviour to ensure the ViewComponent's would render correctly. This test case can be reduced to this simple reproduction:

```rb
# A helper method in our application
def my_helper_method(options, &block)
  content_tag(:dig, options, &block)
end
```

```erb
<%# Inside a ViewComponent %>
<%= my_helper_method({ foo: :bar }) do %>
  <p>Content inside helper method</p>
<% end %>
```

Unfortunately, the content does not render properly with this branch:

```
       expected: "<div> <p>Content inside helper method</p> </div>"
            got: "<p>Content inside helper method</p> <div> &lt;p&gt;Content inside helper method&lt;/p&gt; </div>"
```

camertron:
Thanks for reporting this @percyhanna :)

The problem you're running into is caused by the fact that view components are their own instances of `ActionView::Base`. In order to access helpers using the `helpers` method, instances of `ViewComponent::Base` hold on to a reference to the calling view context (a separate instance of `ActionView::Base`), which has all the helper modules included. This means that any helper method that calls `#capture` (as `#content_tag` does) will do so in the original view context and not in the view context of the component. PR #974 avoids this problem by monkeypatching Action View so that the output buffer is truly global, i.e. is shared by both Action View _and_ all view components. In this PR, I've scoped the changes so they only affect view components. The buffer is global in the sense that it is shared between the first component in the render hierarchy and any other components rendered within it, but local in the sense that it is not shared with the calling view context provided by Action View. Perhaps I should have titled the PR "Use a _global-ish_ output buffer."

View components have always had a sort of weird relationship with helpers. Normally, all the helpers in your app are (as far as I know) automatically included into the module that houses your compiled templates and therefore directly available inside your views. Historically however view components have favored defining methods on component classes themselves instead of relying on the application-wide grab bag of helper methods, which can not only override each other in unintuitive ways but also increase coupling between the component and Action View.

So, given all that, I'd say the issue is more related to helpers than it is to the global output buffer, although the two are certainly related. As things stand, you have three options:

1. Define the helper method on your component directly, eg:
    ```ruby
    class MyComponent < ApplicationComponent
      def my_helper_method
        ...
      end
    end
    ```
3. Include the specific helper module into each component you want to use the helpers in, eg:

    ```ruby
    class MyComponent < ApplicationComponent
      include MyHelper
    end
    ```
1. Wait for view_component v3.

Yes, that's right. We're considering including all helpers in view components by default in v3 😄

BlakeWilliams:
I think this is a valid edge case that we may have to handle on our end. I generally recommend using `helpers` over alternatives like `include MyHelper` since that allows you to re-use any memoization that may exist in the original view context.

When it comes to V3 and including all helpers by default, I think it's relatively likely we'd end up with something like:

```ruby
def method_missing(method, *args, **kwargs)
  return super unless helpers.respond_to?(method)
  helpers.send(method, *args, **kwargs)
end
```

The reason I think that approach is likely is due to the memoization/ivar problem mentioned above. [This PR that adds the original view context](https://github.com/github/view_component/pull/1085) has more details on the impacts of using/not using the default helpers instance. If we were to include all helpers on ViewComponent's themselves, each component would have its own ivars (and possible collisions, but that's another issue entirely). The "single renderer object" (AKA one object that inherits from `ActionView::Base`) could alleviate some of that if we included helpers there, but components would still need to use `method_missing`.

All that being said, I think the example provided is something that we have to expect users of the library to run into and expect to work. Before we commit to anything specific, I think we should verify the helper approach and global buffer approach we land on are compatible with one another.

camertron:
Hmm ok thanks for that additional context @BlakeWilliams. Let me see if I can get something working.

camertron:
Hey everyone, I ended up having to monkeypatch ActionView, but helpers should be working now. Can you confirm @percyhanna?

camertron:
Consensus on the team is to ship this as-is for now, then look into a more maintainable approach that doesn't monkeypatch ActionView. The implementation will likely come in the form of an upstream change to Rails (maybe a config option?) and a corresponding update to view_component. Enabling the global output buffer by default is now a stretch goal for v3. If we aren't confident when v3 release time rolls around, we'll punt it to v4.

lib/view_component/base.rb
```diff
@@ -65,12 +65,14 @@ def self._deprecated_generate_mattr_accessor(name)
     # Returns HTML that has been escaped by the respective template handler.
     #
     # @return [String]
-    def render_in(view_context, &block)
+    def render_in(view_context, global_buffer_in_use = false, &block)
```
BlakeWilliams, line 68:
```suggestion
    def render_in(view_context, global_buffer_in_use = false, &block)
```

I'm a little hesitant about modifying this method since it's implementing a hook from Rails. Thoughts on setting some variable on `ViewComponent::Base` instead? e.g. `ViewComponent.global_buffer_in_use?`?

lib/view_component/global_output_buffer.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    module ActionViewMods
+      def with_output_buffer(buf = nil)
+        # save a reference to the stack-based buffer
+        orig_buf = output_buffer
+        result = nil
+
+        super do
+          if orig_buf.respond_to?(:push)
```
BlakeWilliams, line 62: Can we measure the performance difference between this and the original `with_output_buffer` implementation? My only concern is making existing rendering paths slower (specifically for dotcom), but it's possible it's comparable.

.github/workflows/ci.yml
```diff
@@ -29,6 +29,7 @@ jobs:
       matrix:
         rails_version: ['5.2.6', '6.0.4.4', '6.1.4.4', '7.0.1', 'main']
         ruby_version: ['2.5', '2.6', '2.7', '3.0', '3.1']
+        use_global_output_buffer: ['true', 'false']
```
Spone, line 32: Do you think there is a way to make the name of the checks clearer?

image

ie. replacing `true` and `false` with something more meaningful

camertron: Oh man, that's a great idea, thanks!

lib/view_component/global_output_buffer.rb
camertron: It looks like things are slower with the patch, but the impact is negligible:

```
       without_patch     1.112k i/100ms
       without_patch     10.864k (± 4.6%) i/s -    108.976k in  10.055584s

Pausing here -- run Ruby again to measure the next benchmark...

          with_patch     1.077k i/100ms
          with_patch     10.629k (± 6.4%) i/s -    106.623k in  10.085791s

Comparison:
       without_patch:    10864.5 i/s
          with_patch:    10629.1 i/s - same-ish: difference falls within error
```

lib/view_component/base.rb
camertron: I think it should be ok to use an instance variable here.

BlakeWilliams (approved):

percyhanna:
Thanks for the thoughts and context for the changes. I can appreciate that having to reach inside `ActionView` to fix my use case is not ideal

>I think this is a valid edge case that we may have to handle on our end.

I would assume the same. While some of our helper methods could probably be either converted into their own view components, others really are just basic helper methods, so being able to use them within normal Rails views as well as view components would be ideal. For now, the solution I [shared in the other PR](https://github.com/github/view_component/pull/974#issuecomment-932818184) is working for us.

camertron:
@percyhanna it may not be ideal, but as @BlakeWilliams mentioned it's still important to address because we will continue to delegate to the original view context for helpers in v3 anyway.

The `unbind`/`bind` idea is really interesting, thanks for sharing :) As I understand things, even unbinding and binding won't re-use instance variables, but maybe that's not a concern for your use-case.

Would you be able to test this branch to see if it works for you? I'm planning on merging it on Monday.

percyhanna:
@camertron Unfortunately it looks like this branch still doesn't work with our use case. I tried calling our helper method directly in the component partial, e.g. `helpers.my_helper`, and I also used this in the base view component class, neither of which worked.

```ruby
def method_missing(method, *args, **kwargs, &block)
  if helpers.respond_to?(method)
    helpers.send(method, *args, **kwargs, &block)
  else
    super
  end
end
```

From some quick testing, I think that the `ivar`s _are_ being re-used, which was a little surprising to me, given how it's written. Perhaps it depends on which `ivar`s we are talking about: `ivar`s from the Rails helpers, or from inside the view component class?

The theory behind my "hack" code above was that it would execute the helper method from within the context of the view component, so that calls to lower-level rendering helpers like `content_tag` and `capture` would be executed in the view component context, rather than in the helper/view context.

camertron:
@percyhanna hmm interesting. Just to rule this out, did you set `config.view_component.use_global_output_buffer = true` in your Rails config, i.e. in config/development.rb? I ask because I turned your example into a test case last week and it passes if the GoB is turned on.

percyhanna:
Another example, using the `link_to` helper:

**`test.rb`**
```html
<%= link_to("/some_url") do %>
  <i class="fa-regular fa-bullseye"></i><span>Content</span>
<% end %>
```

Produces:
```html
<a href="/some_url">
    &lt;i class="fa-regular fa-bullseye"&gt;&lt;/i&gt;&lt;span&gt;Content&lt;/span&gt;
</a>
```

percyhanna:
I just wanted to follow up to confirm what I was seeing: the double-escaped HTML was coming from normals Rails views/partials, not a ViewComponent's views. I didn't try this out on a brand new Rails app without anything else, so it's entirely possible that some other part of our application is interfering with the rendering/escaping.

camertron:
Awesome, thanks @percyhanna! At first I couldn't reproduce, but then I tried using `link_to` _outside_ a component and was able to reproduce. I wouldn't have caught this without your help, so thank you!

I have since added fixed the problem and added a test. Can you confirm things are working in your app?

percyhanna:
Awesome! My regression test case passes, and the double-escaping issue has been resolved. Thanks! 🎉

PR #1308, opened 2022-03-14
Use `appraisal` gem in CI

boardfish:
Supersedes #1230

Uses `appraisal ~> 2.5` to run against an array of Rails minor versions from `6.1` onwards. Also moves us away from manually caching gem dependencies, as that can now be handled by the `setup-ruby` action.


Spone:
Looking good, thanks for wrapping this up! :tada: 

While re-reading #1230, I noticed there was one last thing to decide. Do we want to:
- merge this as it is (with the dependency on `excid3/appraisal` fork ([see this for background](https://github.com/github/view_component/pull/1230#discussion_r806892525))
- wait for https://github.com/thoughtbot/appraisal/pull/174 to be merged
- refactor this PR to not use appraisal directly in CI, but setting `BUNDLE_GEMFILE` via `matrix`, as suggested [here](https://github.com/github/view_component/pull/1230#discussion_r806926293)

Happy to hear what you think @joelhawksley @boardfish @BlakeWilliams

boardfish:
I'm leaning towards that first option. I think we should prioritize ensuring that running the tests works the same way in dev as it does in CI. It may be a good idea in the long run to mirror @excid3's fork to @github so we can ensure that version of the code remains publicly available until @excid3's PR is merged.

camertron (approved):
Nice this is great! I'm a big fan of appraisal :)

Spone:
@joelhawksley would you mind giving us your feedback on this? (See my latest comment above). I think we should try to merge this soon.

joelhawksley (changes requested):
@Spone IMO we need to be cautious of depending on a long-running upstream PR on Appraisal.

Let's take another approach, or work to get the upstream PR merged.

allmarkedup:
FWIW, I've just gone through the fun process of trying to get Lookbook CI working with Appraisal - I took a slightly different approach (and with a smaller matrix of ruby/rails versions for now) but I finally got it working. Just in case it's useful to have an example which uses the matrix approach but doesn't require the Appraisal fork: https://github.com/allmarkedup/lookbook/blob/main/.github/workflows/ci.yml

Spone:
There is [a new PR](https://github.com/thoughtbot/appraisal/pull/202) on Appraisal that could fix the issue and allow us to move forward with this approach.

If it does not, I'll take some time to mirror @allmarkedup's method.

nickcharlton:
Appraisal maintainer here! 👋 

I've just merged that PR mentioned above (https://github.com/thoughtbot/appraisal/pull/202), which I'm hoping can help you out here. In a few days I'll cut a new release.

We'd been a bit stuck with a good solution to that problem, but let us know over there if you're having problems.

Spone:
Awesome @nickcharlton! I'll try it out next week. Thanks for the heads up!

Spone:
Awesome @nickcharlton! I'll try it out next week. Thanks for the heads up!

Gemfile
```diff
@@ -17,6 +17,10 @@ group :test do
   gem "selenium-webdriver", "~> 4"
 end
 
+group :development, :test do
+  gem "appraisal", github: "thoughtbot/appraisal"
```
Spone, line 21: I'm targeting `main` for now, but we'll be able to use the [next version](https://github.com/thoughtbot/appraisal/issues/204) when it's released.

Spone:
I thought it was time to get this in shape for v3.0.0 :)

Do not forget this after merging: https://github.com/ViewComponent/view_component/pull/1308#discussion_r834585238

Thanks a lot @boardfish for your help on this!

joelhawksley:
@Spone I'd prefer we wait for the Appraisal release instead of depending on a branch.

joelhawksley:
@boardfish thoughts on testing this against https://github.com/thoughtbot/appraisal/issues/204#issuecomment-1635841844? Perhaps we don't need to wait for 3.0?

boardfish:
Yeah, that sounds good!

joelhawksley:
@boardfish cool, would you be up for bringing this PR up to date?

boardfish:
Sure! I can have a crack at that sometime in the next week.

boardfish:
Alright, I _think_ we're just about there. Personal TODO list of stuff I've recognized that's still left to do on this one:

- [ ] Check that the CC patch is still actually applied
- [ ] Figure out why simplecov's suggesting that `test_helpers.rb` isn't covered
- [ ] followup: see if we can use built-in bundler caching for other test runs instead of manually calling `actions/cache` to remove that layer of complexity from these workflows

.github/workflows/ci.yml
```diff
@@ -22,59 +26,32 @@ jobs:
...
       with:
         path: vendor/bundle
-        key: gems-build-rails-${{ matrix.rails_version }}-ruby-${{ matrix.ruby_version }}-${{ hashFiles('**/Gemfile.lock') }}
+        key: gems-build-rails-${{ matrix.rails_version }}-ruby-${{ matrix.ruby_version }}-${{ hashFiles('**/gemfiles/rails_${{ matrix.rails_version }}.gemfile.lock') }}
     - name: Build and test with Rake
       run: |
-        bundle config path vendor/bundle
-        bundle update
-        # Code-reloading isn't compatible with simplecov, so we need to run once
-        # to collect coverage, and again to test reloads.
```
boardfish, line 75: I'll restore this comment.

.github/workflows/ci.yml
```diff
@@ -22,59 +26,32 @@ jobs:
...
-            ruby_version: "3.1"
-            mode: "capture_patch_disabled"
-          - rails_version: "7.1.1"
-            ruby_version: "3.2"
-            mode: "capture_patch_disabled"
-          - rails_version: "main"
-            ruby_version: "3.2"
-            mode: "capture_patch_disabled"
+        rails_version: ["6.1", "7.0", "7.1", "main"]
+        ruby_version: ["3.0", "3.1", "3.2"]
```
boardfish, line 35: Is there a reason we were previously selective about which versions we'd run against? Maybe it was a holdover from when we needed specific version combos. That would still be possible [this way](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations).

test/test_helper.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "warning"
+
```
boardfish, line 4: This is a holdover from when I had to deal with warnings being raised, which I think might've followed from the rebase, but it's unnecessary now. I'll remove it.

```suggestion
```

.github/workflows/ci.yml
camertron: Yeah, certain versions of Rails require certain versions of Ruby.

camertron (commented):
LGTM! Can you merge main?

.github/workflows/ci.yml
boardfish: I was wondering if we'd still need to do this, but looking at [endoflife.date](https://endoflife.date/rails), the lists of versions we have here look good.

camertron:
Ugh sorry @boardfish, could you fix these merge conflicts? I'm not sure how to resolve the difference in the Gemfile 😓

gemfiles/rails_main.gemfile
```diff
@@ -5,13 +5,21 @@ source "https://rubygems.org"
...
 group :test do
-  gem "cuprite", "~> 0.8"
+  gem "cuprite", "~> 0.15"
   gem "puma", "~> 6"
+  gem "warning"
   gem "selenium-webdriver", "4.9.0"
 end
 
+group :development, :test do
+  gem "appraisal", github: "thoughtbot/appraisal"
```
joelhawksley, line 22: Can we use the rubygems source here?

lib/view_component/test_helpers.rb
```diff
@@ -252,15 +252,18 @@ def __vc_test_helpers_build_controller(klass)
...
         "#{described_class}Preview"
+        # :nocov:
       else
         self.class.name.gsub("Test", "Preview")
       end
       result = result.constantize
     rescue NameError
       raise NameError, "`render_preview` expected to find #{result}, but it does not exist."
     end
+    # :nocov:
```
joelhawksley, line 267: Why were these changes necessary?

.github/workflows/ci.yml
```diff
@@ -22,65 +26,50 @@ jobs:
...
       run: |
-        bundle config path vendor/bundle
-        bundle update
-        # Code-reloading isn't compatible with simplecov, so we need to run once
-        # to collect coverage, and again to test reloads.
-        MEASURE_COVERAGE=true bundle exec rake test spec
-        ENABLE_RELOADING=true bundle exec rake test spec
+        bundle exec appraisal rails-${{ matrix.rails_version }} bundle
+        MEASURE_COVERAGE=true bundle exec appraisal rails-${{ matrix.rails_version }} rake
+        ENABLE_RELOADING=true  bundle exec appraisal rails-${{ matrix.rails_version }} rake
```
camertron, line 72:
```suggestion
        ENABLE_RELOADING=true bundle exec appraisal rails-${{ matrix.rails_version }} rake
```

.github/workflows/ci.yml
```diff
@@ -22,65 +26,50 @@ jobs:
...
     - uses: actions/checkout@v4.1.1
     - name: Setup Ruby
       uses: ruby/setup-ruby@v1
       with:
         ruby-version: ${{ matrix.ruby_version }}
-    - uses: actions/cache@v3
-      with:
-        path: vendor/bundle
-        key: gems-build-rails-${{ matrix.rails_version }}-ruby-${{ matrix.ruby_version }}-${{ hashFiles('**/Gemfile.lock') }}
+        bundler-cache: true
```
camertron, line 65: 😎

camertron (changes requested):

lib/view_component/test_helpers.rb
boardfish: I'll try dropping them, but runs were failing due to coverage errors. I think this was the cause, though I'm not sure if anything in this PR in particular would've triggered it.

boardfish:
@joelhawksley Looks like we do still get [coverage-related failures](https://github.com/ViewComponent/view_component/actions/runs/7407617837/job/20154220996) without those nocov comments. Not sure why, though – maybe there's a discrepancy in the running environment between the runs on `main` vs. on this branch.

joelhawksley:
@boardfish I'm happy to merge and revisit given the complexity of this PR.

boardfish:
Awesome. I'll get rid of those merge conflicts and merge this in myself later, then 👍

PR #1316, opened 2022-03-23
Fix fragment caching in partials when global output buffer is enabled

camertron:
### Summary

I forgot to do one last run of the tests for github.com before merging https://github.com/github/view_component/pull/1307, and lo and behold there were some failures related to fragment caching. Unfortunately I was unable to reproduce the failure in view_component, but included several tests that exercise both branches of the `ActionView::Base#output_buffer=` method (see below for details).

#### Details

Rails' `CacheHelper` module calls `self.output_buffer=` in `#write_fragment_for`, which overwrites the global `OutputBufferStack` instance. `ViewComponent::Base` already contains a patch to accommodate this behavior, but since I also had to patch `ActionView::Base` to get helpers working in https://github.com/github/view_component/pull/1307, we now need a similar `self.output_buffer=` patch for `ActionView::Base`.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,10 +10,14 @@ title: Changelog
 * Add ADR for separate slot getter/setter API.
 
     *Blake Williams*
-    
+
 * Add the option to use a "global" output buffer so `form_for` and friends can be used with view components.
 
     *Cameron Dutro*, *Blake Williams*
+
+* Fix fragment caching in partials when global output buffer is enabled
+
+    *Cameron Dutro*
 
 ## 2.51.0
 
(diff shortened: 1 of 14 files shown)


docs/CHANGELOG.md
```diff
@@ -10,11 +10,15 @@ title: Changelog
 * Add ADR for separate slot getter/setter API.
 
     *Blake Williams*
-    
+
 * Add the option to use a "global" output buffer so `form_for` and friends can be used with view components.
 
     *Cameron Dutro*, *Blake Williams*
 
+* Fix fragment caching in partials when global output buffer is enabled
```
joelhawksley, line 18:
```suggestion
* Fix fragment caching in partials when global output buffer is enabled.
```

joelhawksley (approved):

PR #1318, opened 2022-03-25
Document how to use blocks with lambda slots

sampart:
### Summary

Adds documentation for how to render block content in a lambda slot.

It may just be that I'm a relative ruby newbie, but this felt like a gap in the documentation to me. It's also very different to how component slots do it, since the `content` method isn't available.

### Other Information

As requested in https://github.com/github/view_component/blob/main/docs/CONTRIBUTING.md#submitting-a-pull-request, I've added this change to the CHANGELOG and my avatar to the docs, even though this change is very minor. But I won't be offended if you deem neither of those changes necessary for such a small PR 😸


camertron (approved):

sampart:
Thanks for the approval, @camertron. Any thoughts on how we can get the [Lint check](https://github.com/github/view_component/runs/5734742706?check_suite_focus=true) passing?

docs/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ title: Changelog
 
 ## main
 
+* Document how to use blocks with lambda slots
```
```suggestion
* Document how to use blocks with lambda slots.
```

docs/guide/slots.md
```diff
@@ -175,6 +175,16 @@ class TableComponent < ViewComponent::Base
 end
 ```
 
+To provide content for a lambda slot via a block, you need to explicitly add a block parameter and render the content by calling its `call` method:
```
joelhawksley, line 178:
```suggestion
To provide content for a lambda slot via a block, add a block parameter and render the content by calling its `call` method:
```

docs/guide/slots.md
```diff
@@ -175,6 +175,16 @@ class TableComponent < ViewComponent::Base
...
+
+```ruby
+class BlogComponent < ViewComponent::Base
+  renders_one :header, -> (classes:, &block) do
+    content_tag :h1, block.call, class: classes
```
joelhawksley, line 183: Does this work?

```suggestion
    content_tag :h1, class: classes, &block
```

joelhawksley (changes requested):
@sampart would you be willing to add a test for this use case? I think it would be helpful to verify that the documentation we are adding is 1) correct and 2) remains correct ❤️

docs/guide/slots.md
sampart: yes it does, thanks!

docs/guide/slots.md
```diff
@@ -175,6 +175,16 @@ class TableComponent < ViewComponent::Base
 end
 ```
 
+To provide content for a lambda slot via a block, add a block parameter and render the content by calling its `call` method:
```
joelhawksley, line 178: I don't think this text matches the example code any more 😄

sampart: ooh good point! What about this?

```suggestion
To provide content for a lambda slot via a block, add a block parameter. You can then render the content by calling the block's `call` method, or pass the block directly to `content_tag`:
```

camertron:
@sampart sorry about that lint check, I fixed it in master and merged the change into this branch. Should be good to go now.

sampart:
@joelhawksley sorry for the delay; I've now added a test in https://github.com/github/view_component/pull/1318/commits/625c7d79f3eef905a56a7d7fbae6ea6450753919. I hope it's the kind of thing you were intending. Thanks

docs/guide/slots.md
```diff
@@ -175,6 +175,16 @@ class TableComponent < ViewComponent::Base
 end
 ```
 
+To provide content for a lambda slot via a block, add a block parameter. You can then render the content by calling the block's `call` method, or pass the block directly to `content_tag`:
```
joelhawksley, line 178:
```suggestion
To provide content for a lambda slot via a block, add a block parameter. Render the content by calling the block's `call` method, or by passing the block directly to `content_tag`:
```

test/sandbox/app/components/lambda_slot_component.html.erb
```diff
@@ -0,0 +1,5 @@
+<div>
+  <div class="lambda_slot">
+    <%= header %>
+  </div>
+</div>
```
joelhawksley, line 5:
```suggestion
<%= header %>
```

joelhawksley (approved):

test/view_component/slotable_v2_test.rb
```diff
@@ -508,4 +508,14 @@ def test_component_delegation_slots_work_with_helpers
...
   end
+
+  def test_lambda_slot_content_can_be_provided_via_a_block
+    render_inline LambdaSlotComponent.new do |c|
+      c.header(classes: "some-class") do
+        "This is a header!"
+      end
+    end
+
+    assert_selector("h1", text: "This is a header!")
```
joelhawksley, line 519:
```suggestion
    assert_selector("h1.some-class", text: "This is a header!")
```

sampart:
Thank you!

PR #1339, opened 2022-04-23
Introduce a new render_in_view_context test helper to render arbitrary logic in the view

camertron:
### Summary

I'm currently working on a forms framework at GitHub that uses view components. Forms are rendered like so:

```ruby
form_with(...) do |builder|
  render(FormClass.new(builder, ...))
end
```

The `render_inline` test helper works great for individual component instances, but doesn't let me also test the wrapping `form_with` call. I propose a modification to `render_inline` that evaluates the block in the view context using `instance_exec` so this is possible:

```ruby
render_inline do
  form_with(...) do |builder|
    render(FormClass.new(builder, ...))
  end
end

assert_selector ".whatever"
```


lib/view_component/test_helpers.rb
```diff
@@ -30,22 +30,37 @@ def refute_component_rendered
...
+    # Render a component or block inline. Internally sets `page` to be a `Capybara::Node::Simple`,
     # allowing for Capybara assertions to be used:
     #
     # ```ruby
     # render_inline(MyComponent.new)
     # assert_text("Hello, World!")
     # ```
     #
-    # @param component [ViewComponent::Base, ViewComponent::Collection] The instance of the component to be rendered.
+    # It is also possible to execute a block in the view context instead of rendering a component:
```
joelhawksley, line 41: Might we add something similar to the public docs?

camertron:
Consensus on the team is to define another method for this, since it's potentially surprising/confusing that the block is evaluated in the view context. Maybe call it `render_in_view_context` or `in_view_context`.

docs/api.md
```diff
@@ -221,6 +221,16 @@ render_inline(MyComponent.new)
 assert_text("Hello, World!")
 ```
 
+If `component` is `nil`, `render_inline` will execute the block in the view context instead:
```
joelhawksley, line 224:
```suggestion
If no `component` is provided, `render_inline` executes the block in the view context:
```

lib/view_component/test_helpers.rb
```diff
@@ -38,7 +38,6 @@ def refute_component_rendered
     # assert_text("Hello, World!")
     # ```
     #
-    # @param component [ViewComponent::Base, ViewComponent::Collection] The instance of the component to be rendered.
```
joelhawksley, line 41: I think we want to leave this line in ❤️

docs/api.md
```diff
@@ -221,6 +221,16 @@ render_inline(MyComponent.new)
...
+If no `component` is provided, `render_inline` executes the block in the view context:
+
+```ruby
+render_inline do
+  render(MyComponent.new)
+end
+
+assert_text("Hello, World!")
+```
+
```
joelhawksley, line 233: I think these changes should be removed ❤️ 

The API docs are auto-generated anyways.

lib/view_component/test_helpers.rb
camertron: Woops, thanks!

docs/api.md
camertron: Yeah... I thought the docs were generated and committed back to the repo, but nothing happened this time? Maybe I'm thinking of a different repo.

joelhawksley (approved):

PR #1343, opened 2022-04-26
Recompile inherited templates in development

patrickarnett:
### Summary

With class caching disabled, a view component will recompile after its template is updated.

However, in a case such as
```ruby
class ParentComponent < ViewComponent::Base; end
class ChildComponent < ParentComponent; end
```
where `ParentComponent` has a template and `ChildComponent` does not, updates to `parent_component.html.erb` will not trigger recompilation of `ChildComponent`.

With this change, compilation will check for templates first for the given component, then for each of its superclasses until a template is found (or the superclass is not a view component).


lib/view_component/compiler.rb
```diff
@@ -192,14 +192,23 @@ def templates
...
-          component_class._sidecar_files(extensions).each_with_object([]) do |path, memo|
-            pieces = File.basename(path).split(".")
-            memo << {
-              path: path,
-              variant: pieces.second.split("+").second&.to_sym,
-              handler: pieces.last
-            }
+          check_class = component_class
+
+          loop do
```
BlakeWilliams, line 197: Is it possible to avoid a `loop` here? I generally tend to avoid non-bounded loops since it makes it easier to introduce performance regressions or infinite loop errors.

Maybe we could look at the classes `ancestors` array instead?

lib/view_component/compiler.rb
```diff
@@ -192,14 +192,23 @@ def templates
...
+            files = check_class._sidecar_files(extensions)
+            if files.length > 0
+              break files.map do |path|
```
BlakeWilliams, line 200: Instead of using `break` to return values, could we combine the above suggestion with `find`/`detect` to return the first value found?

BlakeWilliams (commented):
Thanks for opening this PR and reporting this issue!

I'm curious if the root cause is due to us defining `call` explicitly on the subclass of the component instead of relying on inheritance to use the "correct" `call` method (on the parent component).

If that's the case, do you think we could rely on Ruby's inheritance behavior to solve the issue instead of looping over the component's ancestors?

lib/view_component/compiler.rb
```diff
@@ -31,6 +31,8 @@ def compile(raise_errors: false, force: false)
       return if compiled? && !force
       return if component_class == ViewComponent::Base
 
+      component_class.superclass.compile(raise_errors: raise_errors) if development? && templates.empty?
```
joelhawksley, line 34: Wouldn't we only want to do this if there is no `call` method either?

patrickarnett: @joelhawksley I've added a test that doesn't directly assert that we don't call `compile` on the superclass when `call` is defined on the inheriting component, but that we should expect to raise an error in the case when `compile` _is_ called on the superclass.

test/sandbox/app/components/inherited_from_uncompilable_component.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class InheritedFromUncompilableComponent < UncompilableComponent
+  def call(*)
```
camertron, line 4: I don't believe `call` ever accepts any arguments.

```suggestion
  def call
```

camertron (changes requested):

lib/view_component/compiler.rb
```diff
@@ -31,6 +31,8 @@ def compile(raise_errors: false, force: false)
       return if compiled? && !force
       return if component_class == ViewComponent::Base
 
+      component_class.superclass.compile(raise_errors: raise_errors) if should_compile_superclass?
```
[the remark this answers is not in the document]
patrickarnett: I could be wrong, but I believe the parent could already have been recompiled (e.g. the parent itself or another component that inherits from it has been rendered since its sidecar file changed), in which case another recompilation would be unnecessary.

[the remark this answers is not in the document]
patrickarnett: @camertron Am I incorrect?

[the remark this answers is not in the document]
camertron: Ah ok that makes sense, I get it 👍

camertron (approved):

PR #1347, opened 2022-05-04
add `render_preview` test helper

joelhawksley:
### Summary

I've documented this change thoroughly, see the diff ❤️


docs/guide/previews.md
```diff
@@ -6,11 +6,7 @@ parent: Guide
 
 # Previews
 
-`ViewComponent::Preview`, like `ActionMailer::Preview`, provides a quick way to preview components in isolation.
-
-_For a more interactive experience, consider using [ViewComponent::Storybook](https://github.com/jonspalmer/view_component_storybook) or [Lookbook](https://github.com/allmarkedup/lookbook)._
```
Spone, line 11: Maybe keep the mention of these 2 tools, but move it to the end of the page?

joelhawksley: Agreed! I re-added this 👍🏻

docs/CHANGELOG.md
```diff
@@ -9,13 +9,19 @@ title: Changelog
 
 ## main
 
+<<<<<<< HEAD
+* Introduce `render_preview` test helper.
+
+    *Joel Hawksley*
+=======
```
BlakeWilliams, line 16:
```suggestion
* Introduce `render_preview` test helper.

    *Joel Hawksley*
```

docs/CHANGELOG.md
```diff
@@ -9,13 +9,19 @@ title: Changelog
...
 * Add `with_*` slot API for defining slots. Note: we plan to deprecate the non `with_*` API for slots in an upcoming release.
 
     *Blake Williams*
 
 * Add QuickNode to list of companies that heavily rely on ViewComponent.
 
     *Luc Castera*
+>>>>>>> main
```
BlakeWilliams, line 24:
```suggestion
```

lib/view_component/test_helpers.rb
```diff
@@ -41,14 +41,49 @@ def refute_component_rendered
     # @param component [ViewComponent::Base, ViewComponent::Collection] The instance of the component to be rendered.
     # @return [Nokogiri::HTML]
     def render_inline(component, **args, &block)
-      @rendered_component =
+      @rendered_content =
```
BlakeWilliams, line 44: Super minor, but I wonder if this is being used directly by any libraries/apps and if renaming it would break anything.

app/controllers/concerns/view_component/preview_actions.rb
```diff
@@ -23,6 +23,8 @@ def index
     end
 
     def previews
+      find_preview
```
BlakeWilliams, line 26: Just curious, why was this change necessary?

BlakeWilliams (approved):
I like this a lot!

Similar to how we adopted i18n, slots v2 , and other features, what are your thoughts on extracting the test functionality into it's own namespace and not include it by default? That way we can make breaking changes as/if needed without needing a major release.

app/controllers/concerns/view_component/preview_actions.rb
joelhawksley: Good question. Since we're calling `previews` directly from `render_preview`, `before_action` isn't called. In fact, I think we could inline `find_preview` here!

lib/view_component/test_helpers.rb
joelhawksley: I do wonder that too. I'll add a note to the changelog.

cpjmcquillan: This indeed is used by lots of people! #1372 and #1373 offer two fixes.

I wonder if we should update the [documentation] to more explicitly direct people towards writing expectations that use `page` if they are using RSpec and Capybara?

```ruby
expect(page).to have_link("Home page", href: root_path)
```

[documentation]: https://viewcomponent.org/guide/testing.html#rspec-configuration

rmacklin: Related issue: https://github.com/github/view_component/issues/1375

PR #1353, opened 2022-05-10
Add render_parent helper method to avoid double-rendering hijinks

camertron:
### Summary

Although perhaps not a common pattern, it is possible to call `super` in a derived component to render the parent's template:

```html+erb
<%# base_component.html.erb %>
<div>Content from base component</div>

<%# derived_component.html.erb %>
<div>Content from derived component</div>
<% super %>
```

```ruby
# base_component.rb
class BaseComponent < ViewComponent::Base
end

# derived_component.rb
class DerivedComponent < BaseComponent
end
```

Rendering `DerivedComponent` results in the following, as expected:

```html
<div>Content from derived component</div>
<div>Content from base component</div>
```

Unfortunately however it is very tempting to echo the result of `super` out to the page, eg. `<%= super %>` instead of `<% super %>`, which leads to double rendering both components.

This PR introduces a new `render_parent` method that simply calls `super` and returns `nil`.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,8 +9,6 @@ title: Changelog
 
 ## main
 
-## 2.55.0
-
 * Add `render_parent` convenience method to avoid confusion between `<%= super %>` and `<% super %>` in template code.
 
     *Cameron Dutro*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,12 @@ title: Changelog
 
 ## main
 
+## 2.55.0
+
+* Add `render_parent` convenience method to avoid confusion between `<%= super %>` and `<% super %>` in template code.
+
+    *Cameron Dutro*
+
 ## 2.54.0
 
 * Add `with_*` slot API for defining slots. Note: we plan to deprecate the non `with_*` API for slots in an upcoming release.
--- a/docs/guide/templates.md
+++ b/docs/guide/templates.md
@@ -92,3 +92,14 @@ Component subclasses inherit the parent component's template if they don't defin
 class MyLinkComponent < LinkComponent
 end
 ```
+
+### Rendering the Parent's Template
+
+Subclasses that wish to render the parent's template may do so by calling the `render_parent` helper method.
+
+```ruby
+<%# my_link_component.html.erb %>
+<div class="base-component-template">
+  <% render_parent %>
+</div>
+```
--- a/docs/guide/templates.md
+++ b/docs/guide/templates.md
@@ -97,7 +97,7 @@ end
 
 To render a parent component's template from a subclass, call `render_parent`:
 
-```ruby
+```erb
 <%# my_link_component.html.erb %>
 <div class="base-component-template">
   <% render_parent %>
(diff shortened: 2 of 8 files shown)


docs/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ title: Changelog
 
 ## main
 
+## 2.55.0
+
```
```suggestion
```

These are added by the release script ❤️

docs/guide/templates.md
```diff
@@ -92,3 +92,14 @@ Component subclasses inherit the parent component's template if they don't defin
 class MyLinkComponent < LinkComponent
 end
 ```
+
+### Rendering the Parent's Template
```
joelhawksley, line 96:
```suggestion
### Rendering parent templates
```

docs/guide/templates.md
```diff
@@ -92,3 +92,14 @@ Component subclasses inherit the parent component's template if they don't defin
...
+
+Subclasses that wish to render the parent's template may do so by calling the `render_parent` helper method.
```
joelhawksley, line 98:
```suggestion
To render a parent component's template from a subclass, call `render_parent`:
```

docs/guide/templates.md
```diff
@@ -92,3 +92,14 @@ Component subclasses inherit the parent component's template if they don't defin
...
+
+```ruby
```
joelhawksley, line 100: This should be:

```suggestion
```erb
```

right?

lib/view_component/base.rb
```diff
@@ -117,6 +117,24 @@ def perform_render
       render_template_for(@__vc_variant).to_s + _output_postamble
     end
 
+    # Superclass components that call `super` inside their template code will cause a
```
joelhawksley, line 120:
```suggestion
    # Subclass components that call `super` inside their template code will cause a
```

lib/view_component/base.rb
```diff
@@ -117,6 +117,24 @@ def perform_render
...
+    # double render if they accidentally emit the result. In other words, this double
+    # renders:
```
joelhawksley, line 122:
```suggestion
    # double render if they accidentally emit the result:
```

lib/view_component/base.rb
```diff
@@ -117,6 +117,24 @@ def perform_render
...
+    #
+    # <%= super %>
```
joelhawksley, line 124:
```suggestion
    # <%= super %> # double-renders
```

lib/view_component/base.rb
```diff
@@ -117,6 +117,24 @@ def perform_render
...
+    #
+    # but this does not:
+    #
+    # <% super %>
+    #
```
joelhawksley, line 129:
```suggestion
    # <% super %> # does not double-render
    #
```

lib/view_component/base.rb
```diff
@@ -117,6 +117,24 @@ def perform_render
...
+    # The render_parent method simply calls `super` but makes sure to return nil to
+    # avoid rendering the result twice.
```
joelhawksley, line 131:
```suggestion
    # Calls `super`, returning nil to avoid rendering the result twice.
```

joelhawksley (approved):

docs/guide/templates.md
camertron: Woops, yes, thanks.

PR #1370, opened 2022-05-30
Test helper `#with_request_url` parses nested query parameters

richardmarbach:

Hey, when using the `#with_request_url` test helper, nested parameters are currently parsed as a string key:

**Current Behaviour**
```ruby
with_request_url "/?order[id]=asc" do
 assert_equal request.parameters["order[id]"], "asc"
end
```

The common behaviour in e.g. Rails is to parse the nested URL parameters into hashes. I ran into this issue today, since it's different to how the params are parsed in the actual request.

**New behaviour**
```ruby
with_request_url "/?order[id]=asc" do
  assert_equal request.parameters["order"], {"id" => "asc"}
end
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -25,6 +25,10 @@ title: Changelog
 
     *Nick Smith*
 
+* Fix `#with_request_url` test helper not parsing nested query parameters into nested hashes.
+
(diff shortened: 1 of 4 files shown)


joelhawksley:
Brilliant! Thank you for the merge-able contribution ❤️

PR #1377, opened 2022-06-02
Add missing require for translatable

Spone:

We got a `NameError: uninitialized constant ViewComponent::Translatable` in the `pvc` test suite https://github.com/github/view_component/runs/6714662961?check_suite_focus=true in #1291 and I noticed the `require` was missing, I guess it should have been part of  #1268 but was forgotten. It's apparently happening only in gems depending on `view_component` (such as `primer/view_components`), but not when using `view_component` directly in a Rails app?

Adding the `require` fixed the test suite.

@elia does this make sense?

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -9,6 +9,7 @@ require "view_component/polymorphic_slots"
 require "view_component/previewable"
 require "view_component/slotable"
 require "view_component/slotable_v2"
+require "view_component/translatable"
 require "view_component/with_content_helper"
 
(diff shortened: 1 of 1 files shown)


boardfish (approved):

BlakeWilliams (approved):

elia (approved):
Makes sense! If nothing else just to fix the error and for parity. That said I'm a bit confused on why the autoload directives in the `ViewComponent` module are not working in that case, but I didn't had time to find out more about that.

PR #1379, opened 2022-06-02
Strip trailing whitespace in templates

camertron:

Today a co-worker raised an issue with the `LinkComponent` in primer/view_components where it appeared the component was adding a newline to the end of the rendered output. It turned out there was a newline at the end of the corresponding template, which the ERB compiler faithfully turned into `@output_buffer.append("\n")`. Most of the time this isn't a problem, but it's common to include `LinkComponent` in a sentence. If it comes at the end of the sentence, there will be a space between the link text and the period.

IMHO this is pretty surprising behavior considering that the *nix standard is to leave a single trailing newline in all files. In almost all cases I don't think anyone will notice if view_component were to remove trailing whitespace, but I'm curious to hear from others.

I added an `.rstrip` call in the compiler that trims trailing whitespace from the template before compiling it.


Spone:
Wasn't this solved by https://github.com/github/view_component/issues/913 and https://github.com/rails/rails/pull/42279? :thinking:

Spone:
@camertron, makes sense, thanks for clarifying!

test/sandbox/test/layouts_test.rb
```diff
@@ -31,7 +31,7 @@ class LayoutsTest < ActionDispatch::IntegrationTest
     test "rendering with_content_for" do
       get "/layout_with_content_for"
       assert_response :success
-      assert_select 'body[data-layout="with_content_for"]', "Hello content for\n\n  Foo: bar"
+      assert_select 'body[data-layout="with_content_for"]', "Hello content for\n  Foo: bar"
```
Spone, line 34: I find this change a bit confusing. What about adding a dedicated test to be more explicit about what's the expected behavior?

lib/view_component/compiler.rb
```diff
@@ -237,7 +237,7 @@ def variants_from_inline_calls(calls)
 
     def compiled_template(file_path)
       handler = ActionView::Template.handler_for_extension(File.extname(file_path).gsub(".", ""))
-      template = File.read(file_path)
+      template = File.read(file_path).rstrip
```
BlakeWilliams, line 240: Super minor, but is there a case where you'd want to keep newlines? e.g. text based mailers

If-so, should we attempt to only strip a single newline off the end of the template?

BlakeWilliams, line 240:
```suggestion
      template = File.read(file_path)
      template.rstrip!
```

Also minor, but maybe this would result in less allocations?

BlakeWilliams (approved):
LGTM!

I lack context, but is there any risk of this breaking existing templates/behavior that would warrant a config option + deprecation warning?

camertron:
@BlakeWilliams hmm that's a really good call re: email templates. Maybe it makes sense to put the config option on `BaseComponent` so you can set it on a per-component basis.

lib/view_component/compiler.rb
camertron: Nice!

camertron: Hmm, perhaps. I think a per-component config option is a better way to go though.

joelhawksley:
> @BlakeWilliams hmm that's a really good call re: email templates. Maybe it makes sense to put the config option on `BaseComponent` so you can set it on a per-component basis.

Hmm. I'd love to avoid a config option long-term. Is there any way we can test this hypothesis?

joelhawksley (commented):
Needs docs beyond API

camertron:
We discussed the config option internally and came to consensus that a per-component class method is ok 👍  eg.

```ruby
class MyComponent
  strip_trailing_whitespace
end
```

docs/known_issues.md
```diff
@@ -38,3 +38,19 @@ Calls to form helpers such as `form_with` in ViewComponents [don't use the defau
   <%= f.text_field :name %>
 <% end %>
 ```
+
+## Unwanted whitespace
```
joelhawksley, line 42: 🤔 I wonder if this would be better off on the Templates docs page, as this is a feature!

test/sandbox/test/rendering_test.rb
```diff
@@ -2,7 +2,7 @@
 
 require "test_helper"
 
-class ViewComponentTest < ViewComponent::TestCase
+class RenderingTest < ViewComponent::TestCase
```
joelhawksley, line 5: 😅

docs/api.md
```diff
@@ -10,6 +10,18 @@ nav_order: 3
 
 ## Class methods
 
+### .strip_trailing_whitespace(value = true)
+
+Strips trailing whitespace from templates before compiling them.
+
+    class MyComponent < ViewComponent::Base
```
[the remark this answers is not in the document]
camertron: Ah ok I thought indentation was how you fenced, but I do see some `ruby` fencing in here too. Looks like we need to fix this in a bunch of places.

Spone (approved):

PR #1389, opened 2022-06-11
Add support for Capybara finders and other helpers

jacob-carlborg-apoex:
Extends the existing test helper with a simple Capybara session, used by the page and includes the Capybara DSL. With most of the Capybara helper and matchers available, it's now possible to do more fine grained testing by pinpoint various HTML elements, just as in browser based tests.

It includes support for finders (`Capybara::Node::Finders`), matchers (`Capybara::Node::Matchers`) and methods like `within` and `within_element`.


lib/view_component/test_helpers.rb
```diff
@@ -162,5 +173,58 @@ def with_request_url(path)
...
+    # This is a simpler version of {Capybara::Session}.
+    #
+    # It only includes {Capybara::Node::Finders}, {Capybara::Node::Matchers},
+    # {#within} and {#within_element}. It is useful in that it does not require a
+    # session, an application or a driver, but can still use Capybara's finders
+    # and matchers on any string that contains HTML.
+    #
+    # Most of this code is shamelessly stolen from the {Capybara::Session} class
+    # in the Capybara gem (https://github.com/teamcapybara/capybara/blob/e704d00879fb1d1e1a0cc01e04c101bcd8af4a68/lib/capybara/session.rb#L38).
+    class SimpleSession
```
jacob-carlborg-apoex, line 186: Should this class be located here or somewhere else?

jacob-carlborg-apoex:
Should I add a test that uses some of these new helpers?

lib/view_component/test_helpers.rb
```diff
@@ -162,5 +173,71 @@ def with_request_url(path)
...
+          scopes.push(new_scope)
+          yield if block_given?
+        ensure
+          scopes.pop
+        end
+      end
+
+      alias_method :within_element, :within
+
+      def native
```
jacob-carlborg-apoex, line 224: This method was needed by the PVC tests. But this method does not exist on the Capybara session of a regular browser based test.

docs/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Add support for Capybara finders and other helpers
```
joelhawksley, line 12: Might we include examples of the finders and helpers that we're adding support for?

lib/view_component/test_helpers.rb
```diff
@@ -52,6 +61,7 @@ def rendered_component
     # @param component [ViewComponent::Base, ViewComponent::Collection] The instance of the component to be rendered.
     # @return [Nokogiri::HTML]
     def render_inline(component, **args, &block)
+      @page = nil
```
joelhawksley, line 64: Can you explain why we need to reset `page` here?

lib/view_component/test_helpers.rb
joelhawksley: I'd vote that we put it in its own class.

joelhawksley (commented):
Looking good! I'd love to see this update include a change to the testing docs that fleshes out what capybara methods are supported. Might you be willing to write something up? Don't worry about it being perfect, I'll help you edit it ❤️

lib/view_component/test_helpers.rb
jacob-carlborg-apoex: `page` needs to be cached for `within` to work. If it's not reset here, then calling `render_inline` multiple times in the same test won't work. There are several existing tests that do this. Previously the result of `page` method was not cached.

docs/CHANGELOG.md
jacob-carlborg-apoex: Sure, I'll try and do that.

lib/view_component/test_helpers.rb
```diff
@@ -162,5 +173,71 @@ def with_request_url(path)
     def build_controller(klass)
       klass.new.tap { |c| c.request = request }.extend(Rails.application.routes.url_helpers)
     end
+
+    # This is a simpler version of {Capybara::Session}.
```
BlakeWilliams, line 177: I think this makes sense, but I'm not sure I completely understand the reason not to use the upstream class. Could you explain why we can't, or wouldn't want to utilize `Capybara::Session`?

BlakeWilliams (commented):
I had a few questions, but I really like having support for more capybara matchers!

lib/view_component/test_helpers.rb
```diff
@@ -5,11 +5,20 @@
...
   module TestHelpers
     begin
+      require "capybara/dsl"
       require "capybara/minitest"
+
+      include Capybara::DSL
       include Capybara::Minitest::Assertions
 
+      def self.included(mod)
+        Capybara::Node::Simple.send(:define_method, :to_capybara_node) do
```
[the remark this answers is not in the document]
jacob-carlborg-apoex: The reason for the `to_capybara_node` method is to be able to pass nodes, i.e. what's returned by `find`, to the `within` method. The implementation of `within` expects the argument to respond to `to_capybara_node`. In a full blown browser based test the nodes are of type `Capybara::Node::Base`, which implements `to_capybara_node`. In this implementation the nodes are of type `Capybara::Node::Simple`, which does **not** implement `to_capybara_node`. The `Capybara::Node::Simple` class is intended to be used for text when no browser is involved. To quote the documentation:

> It is useful in that it does not require a session, an application or a driver, but can still use Capybara's finders and matchers on any string that contains HTML.

The implementation of the `within` method which I've included here is copied from Capybara. An alternative would be to check if the argument passed to the `within` method is of type `Capybara::Node::Simple`, not sure if that's preferred. I don't know if there are any other types that can be passed to `within` which responds to `to_capybara_node`. Most of the code in this PR is copied from Capybara, I deliberately tried to keep it as close to the original code to make it easier to incorporate any future changes in Capybara.

lib/view_component/test_helpers.rb
```diff
@@ -162,5 +173,71 @@ def with_request_url(path)
...
+    #
+    # Most of this code is shamelessly stolen from the {Capybara::Session} class
+    # in the Capybara gem (https://github.com/teamcapybara/capybara/blob/e704d00879fb1d1e1a0cc01e04c101bcd8af4a68/lib/capybara/session.rb#L38).
+    class SimpleSession
+      # Stolen from: https://github.com/teamcapybara/capybara/blob/e704d00879fb1d1e1a0cc01e04c101bcd8af4a68/lib/capybara/session.rb#L767-L774.
+
+      Capybara::Session::NODE_METHODS.each do |method|
+        if RUBY_VERSION >= "2.7"
+          class_eval <<~METHOD, __FILE__, __LINE__ + 1
+            def #{method}(...)
```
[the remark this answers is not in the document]
jacob-carlborg-apoex: Honestly I haven't given it much thought. It's copied from Capybara, as the comment indicates.

lib/view_component/test_helpers.rb
```diff
@@ -162,5 +173,71 @@ def with_request_url(path)
...
+              @touched = true
+              current_scope.#{method}(...)
+            end
+          METHOD
+        else
+          define_method method do |*args, &block|
```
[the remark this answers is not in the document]
jacob-carlborg-apoex: I don't know. As I've mentioned elsewhere the code is copied from Capybara and I haven't given the edge cases much though. I would say if it's good enough for Capybara it good enough here.

lib/view_component/test_helpers.rb
```diff
@@ -162,5 +173,71 @@ def with_request_url(path)
...
+
+      # Initializes the receiver with the given string of HTML.
+      #
+      # @param html [String] the HTML to create the session out of
+      def initialize(html)
+        @document = Capybara::Node::Simple.new(html)
+      end
+
+      def within(*args, **kw_args)
+        new_scope = args.first.respond_to?(:to_capybara_node) ? args.first.to_capybara_node : find(*args, **kw_args)
```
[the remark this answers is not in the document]
jacob-carlborg-apoex: The coped is copied from Capybara, I would prefer to keep it as close to the original as possible, unless there are strong feelings against that.

lib/view_component/test_helpers.rb
jacob-carlborg-apoex: `Capybara::Session` expects a browser, or rather an application or a driver. The `Capybara::Session` class is quite big and complicated. It has quite a few methods that doesn't make sense in the context of testing a view component: window handling, executing JavaScript, handling alerts, navigation, response headers, status code and similar things.

BlakeWilliams: Thanks for the clarification.

I'm hesitant about having to copy the internals of Capybara classes to create our own class since it feels like we're relying on internal logic/behavior that could break any time Capybara decides to update the internals of their API's.

I'd really like to see if we could avoid that. I had two thoughts around avoiding including copy/pasted code:

* Could we write a wrapper around `Capybara::Session` that exposes only the public methods we need?
* Could we make a contribution to Capybara that would let us avoid having to copy/paste their code into ViewComponent?

For the wrapper idea, I was thinking something very roughly like:

```
class SimpleSession
  # only delegate the methods we want to expose to ViewComponent tests
  delegate :foo, :bar, :baz, to: :@original_session

  def initialize
    @original_session = Capybara::Session.new
  end
end
```

lib/view_component/test_helpers.rb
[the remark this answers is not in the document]
BlakeWilliams: > The implementation of the `within` method which I've included here is copied from Capybara. An alternative would be to check if the argument passed to the `within` method is of type `Capybara::Node::Simple`, not sure if that's preferred.

I think we should avoid modifying `Capybara::Node::Simple`. If capybara decides to implement that method or another library decides to copy this approach I believe we could end up with two gems that now have a compatibility issue due to modifying external dependencies, which we should strive to avoid.

lib/view_component/test_helpers.rb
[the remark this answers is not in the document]
BlakeWilliams: I see, but we will need to maintain this code if it becomes part of ViewComponent so I feel it's important we understand the code.

Looking at the [git blame](https://github.com/teamcapybara/capybara/blame/e704d00879fb1d1e1a0cc01e04c101bcd8af4a68/lib/capybara/session.rb#L783-L793) for that particular section of code, it looks like it was added to address 2.7 compatibility issues, which seems related to the comment below. Since 2.7 introduced `...` and the keyword arguments change, we can avoid having to call `ruby2_keywords` and can use the `...` method of forwarding.

lib/view_component/test_helpers.rb
jacob-carlborg-apoex: > I'm hesitant about having to copy the internals of Capybara classes to create our own class since it feels like we're relying on internal logic/behavior that could break any time Capybara decides to update the internals of their API's.

I understand your reasoning, but the code is not relying on the internals of Capybara. What the `SimpleSession` is doing is mirroring some of the public API of `Capybara::Session`. It's an implementation detail that I've copied some of the code from Capybara. I've looked through the code in `SimpleSession`, I don't see how Capybara could make non-breaking changes that would break the code in `SimpleSession`. If you have any concrete examples of that, please let me know.

jacob-carlborg-apoex: > Could we write a wrapper around Capybara::Session that exposes only the public methods we need?

No, not without relying on the internal code of `Capybara::Session`.

> Could we make a contribution to Capybara that would let us avoid having to copy/paste their code into ViewComponent?

I guess we could contribute `SimpleSession` to Capybara if they're interested. I'm actually already using the `SimpleSession` in regular view tests. But it would be a bit more work contributing to Capybara and try to deduplicate the code in the session classes.

lib/view_component/test_helpers.rb
jacob-carlborg-apoex: Fixed in 0a3c8409.

jacob-carlborg-apoex:
> Should I add a test that uses some of these new helpers?

This has been added.

docs/CHANGELOG.md
jacob-carlborg-apoex: Added: 4a859874.

jacob-carlborg-apoex:
> Looking good! I'd love to see this update include a change to the testing docs that fleshes out what capybara methods are supported. Might you be willing to write something up? Don't worry about it being perfect, I'll help you edit it ❤️

@joelhawksley added some documentation in eb710fad.

lib/view_component/test_helpers.rb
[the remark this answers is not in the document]
jacob-carlborg-apoex: Not needed, see https://github.com/github/view_component/pull/1389#discussion_r900428560.

test/sandbox/test/capybara_dsl_test.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class CapybaraDslTest < ViewComponent::TestCase
```
joelhawksley, line 5: For the sake of compatibility, IMO we should have a test for every method we're delegating ❤️

jacob-carlborg-apoex: Fixed: a9f88c25.

jacob-carlborg-apoex:
I'm not sure what's going on with the failing test: generating documentation for PVC. Does anyone have any insight?

Spone:
> I'm not sure what's going on with the failing test: generating documentation for PVC. Does anyone have any insight?

There a probably some changes to make to https://github.com/primer/view_components in order to fix this. @manuelpuyol would you mind having a look?

joelhawksley (approved):
Thanks for such a wonderful first time contribution! I look forward to your next one 😉 ❤️

Spone:
Let's add a version label to the docs when we release this. See #1374

jacob-carlborg:
Thanks 😃.

joelhawksley:
@jacob-carlborg-apoex @jacob-carlborg I've reverted this change due to the linked issues. I'd love to see this re-introduced if you have the time to look at the bugs it caused ❤️

jacob-carlborg:
I'll try to take a look at the issue.

PR #1394, opened 2022-06-15
Docs: add version info

### LINKED ISSUE #1313: Mention version info in docs
In the docs, it would be cool to mention the version of view_component certain features were introduced in.

Spone:

Improve the docs: add the versions various features were introduced in.

Close #1313 

To make the info stand out, I used the [label](https://just-the-docs.github.io/just-the-docs/docs/ui-components/labels/) component provided by the Jekyll theme.

image

Alternatively, we could also use a simple paragraph in italics.

- I'm not sure which version(s) to mention on the [Translations](https://viewcomponent.org/guide/translations.html) page
- Do you think we should add versions to the [API](https://viewcomponent.org/api.html) page?
- There's a [known display bug](https://github.com/just-the-docs/just-the-docs/issues/751) when multiple labels are used. I'll try to open a PR on the theme repo to fix it.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Improve the docs: add the versions various features were introduced in.
+
+    *Hans Lemuet*
+
 ## 2.57.0
 
 * Add missing `require` for `Translatable` module in `Base`.
--- a/docs/compatibility.md
+++ b/docs/compatibility.md
@@ -17,6 +17,9 @@ ViewComponent is tested against ERB, Haml, and Slim, but it should support most 
 
 ## Disabling the render monkey patch (Rails < 6.1)
 
+Added in 2.13.0
+{: .label }
+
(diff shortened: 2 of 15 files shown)


joelhawksley:
@Spone the design here looks good! I think `Since X` would be my preference.

I do wonder where we draw the line here though as to what gets a label. In theory, these notes should be accompanied by a good portion of our changes. I'm wondering if there might be a better way to solve this problem, perhaps even programmatically?

Spone:
> @Spone the design here looks good! I think `Since X` would be my preference.

I changed it.

> I do wonder where we draw the line here though as to what gets a label. In theory, these notes should be accompanied by a good portion of our changes. I'm wondering if there might be a better way to solve this problem, perhaps even programmatically?

I'd say that we add as many as we can now. Then, we can add a mention to the release docs: "if the release include new documented features, add a label in the docs". It can't be added as part of the original PR since at this point we don't know what the version of the next release will be.

There's probably a way to automate version tagging for each method on the [API](https://viewcomponent.org/api.html) page, using YARD. I suggest we tackle this in a separate PR.

PR #1401, opened 2022-06-16
Include polymorphic slots behavior by default

camertron:

The polymorphic slots feature has baked for a while and appears to be stable. Pending any concerns raised in [this discussion](https://github.com/github/view_component/discussions/1398), I think it's time to include it in `ViewComponent::Base` by default.

Simply `include`ing the module in `ViewComponent` base does the trick. There's no need to wait for v3 because the changes introduced by polymorphic slots are strictly additive.


joelhawksley (approved):
Feel free to merge this on June 22 if there are no objections in the discussion ❤️

PR #1406, opened 2022-06-24
Delegated slots

camertron:

We've known for a while that inheritance doesn't work very well with slots. Child components don't automatically inherit their parent's slots, meaning you have to write awkward code like this:

```ruby
class ParentComponent < ViewComponent::Base
  renders_one :header
end

  def header(*args, &block)
    if args.empty? && block.nil?
      # getter
      super
    else
      # setter
      super(*args, custom_arg, &block)
    end
  end
end
```

Because of stuff like this, I've always preferred composition over inheritance. However, composition suffers from the same problem, namely that defining a slot outside the slots API is awkward and requires special knowledge. I propose we introduce a canonical way to delegate slots to a child component.

I don't think it makes sense to allow delegated slots to return a different value than the original slot - you'd be better off defining your own slot in such a case. However, I do like the idea of being able to customize the parameters sent to the original slot. That's the use-case I've seen before in dotcom. To that end, this PR implements the following API:

```
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,5 +13,3 @@ if RUBY_VERSION >= "3.1"
   gem "net-pop", require: false
   gem "net-smtp", require: false
 end
-
-gem "pry-byebug"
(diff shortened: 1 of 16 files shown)


camertron:
@BlakeWilliams and I discussed other ways of delegating and since I don't have time to see this through right now, I'm going to close the PR and potentially pick it up later.

PR #1407, opened 2022-06-27
Consistently evaluate passed in block content

BlakeWilliams:
Currently slots aren't composable due to the lifecycle of components,
where we attempt to avoid evaluating the block passed to `render_in`
(aka `content` in the library). This creates an inconsistent experience
where delegation works only if `content` was already evaluated.

The following code looks correct, but the `c.title` code will never be
executed:

```ruby
class MyComponent
  delegate :title, to: :@sidebar

  def initialize
    @sidebar = SidebarComponent.new
  end

  def call
    content_tag :div do
      render @sidebar
    end
  end
end
```

```erb
<%= render MyComponent.new(:title) do |c| %>
  <%= c.title("My sidebar") %>
<% end %>
```

The block passed that includes the call to `c.title("My sidebar")` is
never evaluated. This is a side-effect of a performance optimization
where we attempt to lazily evaluate the block passed to `render_in`.
This optimization was introduced to avoid evaluating  the entire render 
chain for components return `false` when `render?` is called since it's
unnecessary work.

If the component were slightly different and called `content` for any
reason the above code would work as expected:

```ruby
class MyComponent
  delegate :title, to: :@sidebar

  def initialize
    @sidebar = SidebarComponent.new
  end

  def call
    # this causes `content` to be evaluated, resulting in
    # <%= c.title(title: "My sidebar") %> being called. Now this
    # component will work as expected.
    modified_content = content.replace('foo', 'bar')

    content_tag :div do
      render @sidebar
    end
  end
end
```

This is an example of the optimization causing inconsistent behavior in
the library. The change above causes `content` to be called before
`#call`, so any setup/mutations/side-effects are now guaranteed to be
reflected in the component output.

Here's what the change in the component lifecycle looks like:

```
#before_render -> #render? -> #call
(#content can be called anywhere in the above chain, or never called)
```

Now the lifecycle looks like:

```
#before_render -> #render? -> #content -> #call
(#content isn't used, it's just evaluated to ensure side-effects are performed)
```


test/sandbox/app/components/slots_v2_without_content_block_component.rb
```diff
@@ -4,7 +4,7 @@ class SlotsV2WithoutContentBlockComponent < ViewComponent::Base
   renders_one :title, "MyTitleComponent"
 
   class MyTitleComponent < ViewComponent::Base
-    def initialize(title:)
+    def initialize(title)
```
BlakeWilliams, line 7: This is a somewhat unrelated change, but having a kwarg called `title` for a method called `title` didn't seem great, but it also caused some issues with delegation.

BlakeWilliams:
This isn't a breaking change, but it does have the potential to expose bugs in existing components due to how it replaces the existing common rendering scenario (no evaluated content) with the less common one (evaluated content).

camertron (approved):
Makes sense to me 👍  but I wonder if we should ship it with v3.

BlakeWilliams:
@camertron I'd really like to ship it before 3.0, but adding a flag in the critical path would impact performance. I guess the question I have is, do we consider the inconsistent lifecycle a bug? I think that would help determine how we move forward.

adrianthedev:
A bit late at the party @joelhawksley and @BlakeWilliams , but I think this should have gone into v3 like @camertron suggested. It's quite a big breaking change.

For example, I'm using a wrapper component that hides the content if a passed variable is `nil`. Because the wrapper component hides that content, it never gets evaluated and doesn't raise an error when something inside breaks. The wrapper acts as a safe guard.
Now I have to change the logic in all components to take into account if the value exists or not. It kind of breaks the current way we compose components.

I'm not contesting if this change is better or worse, but just the way it was introduced in the release cycle and the impact it has on current work.

```ruby
class WrapperComponent
  def initialize(field: nil)
    @field = field
  end
end

class ChildComponent
  def initialize(field: nil)
    @field = field
  end
end
```

```erb
<!-- wrapper_component.html.erb -->
<% if @field.present? %>
  <%= content %>
<% else %>
  —
<% end %>

<!-- child_component.html.erb -->
<%= render WrapperComponent.new(field: @field) do %>
  <% @field.do_something %>
<% end %>
```

`do_something` will raise an error now because `@field` is `nil`. Pre `2.60` was not evaluated, but now (post `2.61`), it gets evaluated and crashes.

BlakeWilliams:
@adrianthedev Thanks for the report, and apologies for the change in behavior.

We just reverted this change since it's breaking code that our public API's support. This will be changed in the future though, so I'd recommend migrating any code that relies on `content` as described is migrated to rely on other mechanisms, such as slots or the `render?` callback method.

You should be able to update to `2.61.1`.

adrianthedev:
Hey @BlakeWilliams. No worries. All good!

Thank you for the swift response!

Have you given it any thought if this is going to land in v3, or should we follow the v2 changelog?

BlakeWilliams:
@adrianthedev If we can implement a flag check in a performant way, I'd like to land this in v2, otherwise this may end up being a breaking change in V3.

adrianthedev:
That sounds perfect 👌

Thanks! Let me know if I can help.

adrianthedev:
Hey @BlakeWilliams. Coming back to this one to plead for not removing the feature flag in V3 and going forward.
The reason why I think it's wrong to initially evaluate the block first is that it's counter-intuitive and different than how regular `erb`s work.

I'll post an example:

```ruby
# The helper
def wrapper(should_render: true, &block)
  # Conditionally stopping the execution of 
  return unless should_render

  render layout: "avo/partials/wrapper" do
    capture(&block)
  end
end
```

```erb
<!-- _wrapper.html.erb -->
<!-- We're just yielding for simplicity's sake -->
content -> <%= yield %>
```

```erb
<!-- We call the helper somewhere in the app. -->
<%= wrapper should_render: false do %>
  <%= raise 'error' %>
<% end %>
```

As you can see, this will not render any side effects and will follow the "regular" way of thinking about things. "I turned it off, so I don't have to worry about what's inside".

Yes, in this example, `should_render` is a basic use case, but think about us passing a more complex object and computing the visibility of the `body`.

Again, this is a plead towards, it's more natural to think about what happens in the wrapper first and then the content versus the other way around, and if you force changing that, it will introduce friction going forward.

BlakeWilliams:
@adrianthedev Thanks for the feedback! I think it's important to have these conversations, especially before we move forward with 3.0, so I appreciate you commenting here.

> The reason why I think it's wrong to initially evaluate the block first is that it's counter-intuitive and different than how regular `erb`s work.

That's a fair point, but it's also counter-intuitive to have your `slots` not be available in some circumstances because we haven't yet evaluated the passed in block. I think either way we go here, there's a compromise being made.

> As you can see, this will not render any side effects and will follow the "regular" way of thinking about things. "I turned it off, so I don't have to worry about what's inside".

I think that's reasonable, but do we have any real examples of side-effects that would cause problems with the new functionality? 

> Again, this is a plead towards, it's more natural to think about what happens in the wrapper first and then the content versus the other way around, and if you force changing that, it will introduce friction going forward.

I'd love to see examples of where this would cause friction. Without that, I think it's really difficult to weight the impacts of this change, which is one reason why this functionality is behind a flag.

In general, I think the core issue is that `content` isn't a slot. e.g.

```erb
<% render MyComponent do |c| %>
  <% c.do_something %>
  Test!
<% end %>
```

There's no way to evaluate `c.do_something` without also evaluating the content passed to the component. If `content` were a slot though, that changes:

```erb
<% render MyComponent do |c| %>
  <% c.do_something %>
  <% c.content do %>
    Test!
  <% end %>
<% end %>
```

Now we can safely evaluate the block without rendering the child content (best of both worlds at the cost of some boilerplate).

This might go a long way towards making component rendering more explicit, consistent, and predictable if we were to make that change. Unfortunately, it would likely introduce *a lot* of friction when upgrading from v2 -> v3.

I'd be curious to hear your thoughts about that in addition to the other questions above. cc @joelhawksley and @camertron since I'd love to hear y'alls feedback on the "content as a slot" idea.

adrianthedev:
> That's a fair point, but it's also counter-intuitive to have your slots not be available in some circumstances because we haven't yet evaluated the passed in block. I think either way we go here, there's a compromise being made.

IMHO the counter-intuitive way of doing things is to change the mental model of **everyone that writes erb** currently. They are used to this way of not evaluating the content first. It's the way it is now and they will apply the same workaround to view components as they do with regular `erb`.

This is not a compromise if we don't enforce changing the way of thinking about this. The "compromise" has already been settled by `erb`. Folks are alright with it IMHO.

> I think that's reasonable, but do we have any real examples of side-effects that would cause problems with the new functionality?

Yes. Definitely! Check out [my initial](https://github.com/ViewComponent/view_component/pull/1407#issuecomment-1191292165) post.

In the `content` slot, I actually use the `field` object. I run methods on it, I get properties, etc. but I rely on the wrapper to short-circuit that like so:

We check if the `field` object to see if it's `nil` or not. If it is, we stop rendering it (`content`) in the wrapper, thus the `content` is never executed.

That's one Avo field, but we have more. Others have built fields in the same way. They will be impacted too.

Again, this is one example that falls under the same way that `erb` currently works.

> I'd love to see examples of where this would cause friction. Without that, I think it's really difficult to weight the impacts of this change, which is one reason why this functionality is behind a flag.

This https://github.com/avo-hq/avo/issues/1074 is one example where this behavior popped up. It impacted a few of our users and they reported it right away. I made the revert and things went back to normal.

#### Your example

```ruby
<% render MyComponent do |c| %>
  <% c.do_something %>
  Test!
<% end %>
```

For your example, I would suggest to move the initialization from `c.do_something` to a different method or abstraction. It doesn't seem right to initialize "something" in a slot. 

#### Future API

```ruby
<% render MyComponent do |c| %>
  <% c.do_something %>
  <% c.content do %>
    Test!
  <% end %>
<% end %>
```

Regarding this syntax... I would vote against it. With all the dev-tools I built, I preferred to take the hit and do the heavy lifting behind the scenes vs. exposing a less-friendly API to the consumer.
It's up to you, the authors, but the difference is in the details.


```ruby
# simple API
render button do
  Save
end

# less-friendly API
render button do |c|
  c.content do
    Save
  end
end
```

IMHO, we have three options for this decision.

1. Keep things as they were and force the initialization of slots somewhere else (maybe in the `initialize` method).
2. Keep things as they are and allow the dev to choose the behavior type. This would be a compromise because it will mess with how people work with View Components across projects (projects may use different settings).
3. Change and enforce. That would be the most intrusive option TBH, because it will force some developers that are using VC this way to update their code (which sometimes is not an easy feat), and maybe never upgrading to 3.0 just because of that issue. This could create one of those `python 2/python 3` episode that divides everyone and keeps apps from being kept up to date (I may be a little dramatic here, but this scenario might happen).

I'd love to give more input on this.
Thank you for all the work. I really appreciate what I can accomplish using View Components.

BlakeWilliams:
> Yes. Definitely! Check out [my initial](https://github.com/ViewComponent/view_component/pull/1407#issuecomment-1191292165) post.
> 
> In the `content` slot, I actually use the `field` object. I run methods on it, I get properties, etc. but I rely on the wrapper to short-circuit that like so:
> 
> We check if the `field` object to see if it's `nil` or not. If it is, we stop rendering it (`content`) in the wrapper, thus the `content` is never executed.

Part of the problem here is that the code you posted *can* break under the right circumstances as-is, because there are times where `content` can be executed, like when slots are being used. That is part of the reason behind this PR, which is to make the evaluation model more consistent.

> Regarding this syntax... I would vote against it. With all the dev-tools I built, I preferred to take the hit and do the heavy lifting behind the scenes vs. exposing a less-friendly API to the consumer.

If you have suggestions on an API that will solve the problems being outlined, they're more than welcome. Unfortunately this is a difficult problem to solve and there's likely to be trade-offs that have to be made.

Personally, I don't think the API would be awful, since a large portion of components could do:

```
<%= MyComponent.new.with_content do %>
<% end %>
```

> * Keep things as they were and force the initialization of slots somewhere else (maybe in the `initialize` method).

ViewComponent doesn't control the `initialize` method, but I'm also not sure how this solves the problem. Do you mean move slots outside of the content block? I'm curious what you imagine this looking like in practice.

> * Keep things as they are and allow the dev to choose the behavior type. This would be a compromise because it will mess with how people work with View Components across projects (projects may use different settings).

Agreed, I think the reasons you outlined make this solution a no-go in my opinion, since the behavior is now even more unclear than it currently is.

> 3\. Change and enforce. That would be the most intrusive option TBH, because it will force some developers that are using VC this way to update their code (which sometimes is not an easy feat), and maybe never upgrading to 3.0 just because of that issue. This could create one of those `python 2/python 3` episode that divides everyone and keeps apps from being kept up to date (I may be a little dramatic here, but this scenario might happen).

I don't think this change is *that* extreme. Upgrades are tough, but there are ways to make it easier like through automatic codemods.

> I'd love to give more input on this.
Thank you for all the work. I really appreciate what I can accomplish using View Components.

Thank you, I appreciate the kind words. ❤️

adrianthedev:
Ok. This is quickly becoming a bit unproductive 😅

The point that I want to put across is that doing and enforcing this change (to evaluate the slots & the content first) is counter-intuitive to how `erb` works. Not sure about haml or phlex; maybe that's something worth evaluating too.

The vast majority of users that are coming to view components already have a mental model on the order in which the content is being evaluated, and this divide will cause friction in adoption and migration.

I'd love to hear @joelhawksley's opinion on this too.

Spone:
My 2 cents: you're assuming that people familiar with ERB layout / `yield` behavior will be confused by the component / slots behavior in ViewComponent. I'm not sure that's the case, because these are two quite distinct things in my opinion. I'd love to see more evidence of this confusion if there is.

I'll give a bit more thought to this and probably leave another comment later.

adrianthedev:
Thanks for your input.

It's not for the way VC evaluates slots that I'm worried about, but the way it evaluates the default `content` block. 

See [my original](https://github.com/ViewComponent/view_component/pull/1407#issuecomment-1277246524) example (simple non-slotted components).

PR #1408, opened 2022-06-28
Add support for RSpec to previews as test cases

thutterer:

I've noticed that the [(Experimental) Previews as test cases](https://viewcomponent.org/guide/testing.html#experimental-previews-as-test-cases) don't support RSpec yet, because of how it builds the class name in which it expects the preview:
https://github.com/github/view_component/blob/1a2e39d044f2be8c4dd18fe14d4aa5a3812b79f0/lib/view_component/render_preview_helper.rb#L13-L16

This PR adds support for RSpec by checking `self.respond_to?(:described_class)` and if so, using `"#{self.described_class}Preview"` as the preview class name.

- I haven't added a note about this to the docs. Not sure if that is needed in this case, because if you follow the naming conventions, it will just work.
- I haven't touched any tests. Not sure how RSpec compatibility is handled in the project.
- I haven't added myself to the `docs/index.md` as I hope #1383 gets merged before this one (or eventually).


docs/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Add support for RSpec to previews as test cases
```
```suggestion
* Add support for `render_preview` in RSpec tests.
```

joelhawksley (commented):
Wonderful! To little surprise, I'd love to see a test for this ❤️. IMO we should add example rspec tests to the example app. Would you mind doing so in this PR? Just a single test that uses `render_preview` is sufficient for now, but IMO we should expand what we have for the sake of example.

thutterer:
> IMO we should add example rspec tests to the example app. Would you mind doing so in this PR?

Good idea, I'll do that 👍🏼

thutterer:
Excuse the 👊🏼 -pushing but I've added the `rspec-rails` gem and made it so that `appraisal rake spec` runs all rspec tests.

For now there is only one rspec file covering the fix in this PR. I've added `:nocov:` comments around the code that is covered with RSpec.

@joelhawksley Should I go ahead and enable this also on CI by running `rake test spec` instead of `rake` in the test job?

joelhawksley:
> For now there is only one rspec file covering the fix in this PR. I've added :nocov: comments around the code that is covered with RSpec.

@thutterer yes, let's enable this on CI and remove the `:nocov:`. Simplecov should collate all of the builds to include the Rspec coverage.

thutterer:
@joelhawksley 
> let's enable this on CI and remove the :nocov:. Simplecov should collate all of the builds to include the Rspec coverage.

Done, test coverage back to 100% 💚 I had to move the `rspec-rails` gem into the top-level Gemfile for `rake spec` to work without appraisal.

lib/view_component/render_preview_helper.rb
```diff
@@ -11,15 +11,25 @@ module RenderPreviewHelper
     # ```
     #
     # Note: `#rendered_preview` expects a preview to be defined with the same class
-    # name as the calling test, but with `Test` replaced with `Preview`:
+    # name as the calling test, but with "Test" replaced with "Preview":
```
joelhawksley, line 14: IMO we should use backticks for constants.

```suggestion
    # name as the calling test, but with `Test` replaced with `Preview`:
```

lib/view_component/render_preview_helper.rb
```diff
@@ -15,11 +15,21 @@ module RenderPreviewHelper
     #
     # MyComponentTest -> MyComponentPreview etc.
     #
+    # With RSpec it uses `described_class` plus "Preview" as the class name.
```
joelhawksley, line 18:
```suggestion
    # In RSpec tests, `Preview` is appended to `described_class`.
```

joelhawksley:
Fantastic! Thanks so much for adding this support ❤️

PR #1415, opened 2022-07-05
Replace rubocop with StandardRB

joelhawksley:
In the interest of making sure ViewComponent doesn't feel
like it's strictly a GitHub project, this PR adopts
the base configuration of `standardrb` in place of `rubocop-github`.


boardfish:
Looking good! At first I saw `standardrb`'s choice of no spacing inside a hash with one pair as a little odd, but it's growing on me.

Think there's more to fix on [linter issues](https://github.com/github/view_component/runs/7205379900?check_suite_focus=true) before this is green, but it's well on the way 👍

elia (approved):
❤️

juanmanuelramallo (approved):

PR #1429, opened 2022-07-21
Pin CI to non-broken rails main

joelhawksley:
https://github.com/rails/rails/pull/45614 broke our Global Output Buffer, which we should just remove at this point. Pinning for now while we resolve other issues.


BlakeWilliams (approved):
LGTM

Should we open an issue to address it so we don't stick to that ref for too long?

PR #1432, opened 2022-07-21
Remove the global output buffer; compile on render_in

camertron:

A few months ago, I added an [experimental feature](https://github.com/github/view_component/pull/1307) called the global output buffer. We were attempting to build a forms framework at GitHub, and since it's a known issue that form helpers don't work very well with components, I was trying to work up a solution that would unblock the forms work. Unfortunately the final global output buffer solution monkeypatches actionview to an unsettling degree. It also turned out we didn't need it for the forms work after all. Finally, it appears it doesn't work with edge rails anymore. So let's get rid of it!

Because the global output buffer degraded performance a bit, I also worked up a [patch](https://github.com/github/view_component/pull/1302) to counteract it. Unfortunately the patch has led to a number of subtle bugs in development environments, so let's remove it too.


lib/view_component/base.rb
```diff
@@ -124,19 +126,18 @@ def render_in(view_context, &block)
       before_render
 
       if render?
-        perform_render
+        # preload content to support slot delegation and to consistently perform
+        # side-effects that may exist in the block
+        content
```
joelhawksley, line 131: I believe this change should not be here. We recently reverted this line.

Spone:
> It also turned out we didn't need it for the forms work after all.

@camertron If you're able to elaborate on this, I'd be interested in knowing a bit more!

lib/view_component/base.rb
BlakeWilliams, line 131:
```suggestion
```

Similar to @joelhawksley's comment (I can't see it in the files view?) this was part of a recent revert.

camertron:
@Spone good question! I was able to get around the form helper issues by creating what I've been calling "view_component lite" inside the [primer/rails_forms](https://github.com/primer/rails_forms) repo, which is where our new forms framework lives (for now). VC lite is really just a barebones VC implementation that doesn't inherit from `ActionView::Base`. The important bit is in the [`ActsAsComponent` module](https://github.com/primer/rails_forms/blob/main/lib/primer/rails_forms/acts_as_component.rb). It works by delegating everything to the view context, including the output buffer. I tried a similar approach in view_component but unfortunately the delegation adds quite a bit of overhead. Forms generally represent a fairly small area of the page so the performance hit was acceptable, but not reasonable for all view components at large.

joelhawksley (approved):

BlakeWilliams (approved):

stevegeek:
@camertron Understand the desire to remove nasty hacks etc, but just wondering if there is or will be any 'official' solution to the forms problem now? Was relying on the global output buffer to fix my form rendering woes! Could the solution in `rails_forms` be made into something that can be added more generally to other apps that dont use `rails_forms`? thanks for any suggestions!

tmaier:
Not for everyone looking [primer/rails_form](https://github.com/primer/rails_forms) has been archived. The work has been merged with primer/view_components. You can find the latest code at https://github.com/primer/view_components/pull/1238

camertron:
Woops @stevegeek, I totally missed your comment in August, sorry about that!

> Understand the desire to remove nasty hacks etc, but just wondering if there is or will be any 'official' solution to the forms problem now?

A semi-recent [change](https://github.com/rails/rails/pull/45731) to action view has removed the need to reassign `@output_buffer` when capturing, which was one of the big problems the global output buffer was designed to solve. Unfortunately our [action view compatibility tests](https://github.com/ViewComponent/view_component/blob/main/test/sandbox/test/action_view_compatibility_test.rb) still fail on Rails main, so the Rails change isn't enough. I'm planning on having a deeper look tomorrow.

> Could the solution in `rails_forms` be made into something that can be added more generally to other apps that don't use `rails_forms`?

Maybe? We could certainly extract the `ActsAsComponent` module into its own gem. We have tests in primer/view_components that demonstrate view components can be rendered inside `ActsAsComponent` templates. Theoretically you could write a component which includes `ActsAsComponent` (instead of inheriting from `ViewComponent::Base`) and call form helpers + render other view components in the template. I think that would work?

Honestly the long-term fix is probably to stop inheriting from `ActionView::Base`, but that's going to take a Herculean amount of effort.

stevegeek:
@camertron thanks for the reply! Ok understood. I will wait and see how it all evolves and look at ActsAsComponent. For now thanks for taking the time to elaborate on the situation!

camertron:
@stevegeek things move fast in the VC world! @BlakeWilliams has [this PR](https://github.com/ViewComponent/view_component/pull/1650) up for review right now which is another take on fixing our buffer issues. It's very likely going to get merged this week I'd say.

stevegeek:
@camertron amazing news!

PR #1437, opened 2022-07-25
Add predicate method to polymorphic slots

### LINKED ISSUE #1436: Polymorphic slots missing predicate method
### Steps to reproduce

1. Create a component with a polymorphic slot
2. Try to use the predicate method for the slot
3. See that the predicate is undefined

e.g.

```ruby
  renders_one icon, types: {
    svg: SvgComponent,
    font: IconFontComponent
  }

  def before_render
    svg("default.svg") unless icon?
  end
end
```

### Expected behavior

In the example `icon?` should be defined

### Actual behavior

`icon?` is not defined

### System configuration
**Rails version**: 7.0.3.1

**Ruby version**: ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]

**Gem version**: 2.59.0

TastyPi:

Add a `slot?` method for polymorphic slots. Fixes #1436.

Copied the implementation used for normal slots.

I haven't written tests because I haven't had time to work out how this project has set them up, but @Spone has said they will help.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -287,6 +287,9 @@ DEPENDENCIES
   haml (~> 5)
   jbuilder (~> 2)
   minitest (= 5.6.0)
+  net-imap
+  net-pop
+  net-smtp
   pry (~> 0.13)
   rails (~> 7.0.0)
   rake (~> 13.0)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -287,9 +287,6 @@ DEPENDENCIES
   haml (~> 5)
   jbuilder (~> 2)
   minitest (= 5.6.0)
-  net-imap
-  net-pop
-  net-smtp
   pry (~> 0.13)
   rails (~> 7.0.0)
   rake (~> 13.0)
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -25,6 +25,10 @@ title: Changelog
 
     *Josh Clayton*
 
+* Add predicate method support to polymorphic slots.
+
+    *Graham Rogers*
+
 ## 2.61.1
 
 * Revert `Expose Capybara DSL methods directly inside tests.` This change unintentionally broke other Capybara methods and thus introduced a regression. We aren't confident that we can fail forward so we have decided to revert this change.
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -282,7 +282,7 @@ Filling in the `visual` slot is done by calling the appropriate slot method:
 <% end %>
 ```
 
-To test whether a polymorphic slot has been passed to the component, use the provided `#{slot_name}?` method.
+To see whether a polymorphic slot has been passed to the component, use the `#{slot_name}?` method.
 
 ```erb
 <% if visual? %>
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -281,3 +281,13 @@ Filling in the `visual` slot is done by calling the appropriate slot method:
   <% end >
 <% end %>
 ```
+
+To test whether a polymorphic slot has been passed to the component, use the provided `#{slot_name}?` method.
+
+```erb
+<% if visual? %>
+  <%= visual %>
+<% else %>
+  <span class="visual-placeholder">N/A</span>
+<% end %>
+```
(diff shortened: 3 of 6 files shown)


joelhawksley (commented):
👋🏻 mind adding docs for this along with an entry to the changelog and an update to the docs index?

lib/view_component/polymorphic_slots.rb
```diff
@@ -42,6 +42,10 @@ def register_polymorphic_slot(slot_name, types, collection:)
           define_method(getter_name) do
             get_slot(slot_name)
           end
+
+          define_method("#{getter_name}?") do
+            get_slot(slot_name).present?
```
boardfish, line 47:
```suggestion
            public_send(getter_name).present?
```

boardfish: This way, if the implementation for the polymorphic slot ever changes internally (or is even overridden externally), we'll be calling that instead.

TastyPi:
I've added documentation and applied @boardfish's suggestion. I also figured out how to add a test, though if you think there's a better way to test it let me know.

lib/view_component/polymorphic_slots.rb
```diff
@@ -42,6 +42,10 @@ def register_polymorphic_slot(slot_name, types, collection:)
           define_method(getter_name) do
             get_slot(slot_name)
           end
+
+          define_method("#{getter_name}?") do
+            public_send(getter_name).present?
```
BlakeWilliams, line 47: I think we want to avoid `public_send` here, since it's likely slower than calling methods directly.

Could we use `get_slot` instead of `public_send`?

boardfish: Ah, I suppose that's important when it comes to ViewComponent. TIL!

TastyPi: Reverted back to using `get_slot`

docs/CHANGELOG.md
```diff
@@ -17,6 +17,10 @@ title: Changelog
 
     *Josh Clayton*
 
+* Implement predicate method for polymorphic slots
```
BlakeWilliams, line 20:
```suggestion
* Add predicate method support to polymorphic slots.
```

Super minor, just copying some previous changelog entries wording for consistency.

docs/guide/slots.md
```diff
@@ -281,3 +281,16 @@ Filling in the `visual` slot is done by calling the appropriate slot method:
...
+
+```erb
+<% if visual? %>
+  <div>
+    <div><%= content %></div>
+    <%= visual %>
+  </div>
+<% else %>
+  <div>content</div>
+<% end %>
```
BlakeWilliams, line 295:
```suggestion
<% if visual? %>
  <%= visual %>
<% else %>
  <span class="visual-placeholder">N/A</span>
<% end %>
```

Minor, but thoughts on simplifying this example to remove `content` from the equation (we also use `content` in two different ways in the current diff)?

BlakeWilliams (approved):
I left two more small pieces of feedback, but otherwise this looks great, thanks!

Gemfile.lock
```diff
@@ -287,6 +287,9 @@ DEPENDENCIES
   haml (~> 5)
   jbuilder (~> 2)
   minitest (= 5.6.0)
+  net-imap
+  net-pop
+  net-smtp
```
joelhawksley, line 292:
```suggestion
```

docs/guide/slots.md
```diff
@@ -281,3 +281,13 @@ Filling in the `visual` slot is done by calling the appropriate slot method:
   <% end >
 <% end %>
 ```
+
+To test whether a polymorphic slot has been passed to the component, use the provided `#{slot_name}?` method.
```
joelhawksley, line 285:
```suggestion
To see whether a polymorphic slot has been passed to the component, use the `#{slot_name}?` method.
```

test/sandbox/test/slotable_v2_test.rb
```diff
@@ -490,6 +490,15 @@ def test_polymorphic_slot_with_setters
     assert_selector("div .bar.custom-bar:last")
   end
 
+  def test_polymorphic_slot_question_mark
```
joelhawksley, line 493:
```suggestion
  def test_polymorphic_slot_predicate
```

joelhawksley (approved):

PR #1441, opened 2022-07-25
Add `raise_on_deprecated_slot_setter` macro to raise if deprecated no…

joelhawksley:
This PR adds a `raise_on_deprecated_slot_setter` macro to be used to preview the coming removal of the non-`with_*` methods for slots in ViewComponent v3.0.0 ❤️

### DIFF
--- a/lib/view_component/slotable_v2.rb
+++ b/lib/view_component/slotable_v2.rb
@@ -17,9 +17,19 @@ module ViewComponent
       # Hash of registered Slots
       class_attribute :registered_slots
       self.registered_slots = {}
+
+      class_attribute :_raise_on_deprecated_slot_setter
+      self._raise_on_deprecated_slot_setter = false
     end
 
     class_methods do
+      ##
+      # Enables deprecations coming to the Slots API in ViewComponent v3
+      #
+      def raise_on_deprecated_slot_setter
+        self._raise_on_deprecated_slot_setter = true
+      end
+
       ##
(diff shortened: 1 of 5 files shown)


lib/view_component/slotable_v2.rb
```diff
@@ -17,9 +17,19 @@ module SlotableV2
       # Hash of registered Slots
       class_attribute :registered_slots
       self.registered_slots = {}
+
+      class_attribute :_raise_on_deprecated_slot_setter
+      self._raise_on_deprecated_slot_setter = false
```
BlakeWilliams, line 22:
```suggestion
      class_attribute :_raise_on_deprecated_slot_setter, default: false
```

Can we rely on the `default` parameter `class_attribute` accepts?

BlakeWilliams (approved):
LGTM

It looks like polymorphic slots took a different approach to solving the same problem, [here](https://github.com/github/view_component/blob/main/lib/view_component/polymorphic_slots.rb#L60-L63). Can we consolidate the two approaches so we provide consistent behavior/warnings?

lib/view_component/slotable_v2.rb
joelhawksley: I don't believe so, at least with the Ruby/Rails combinations we support.

joelhawksley:
Closing for now, I'll bring this back in the future ❤️

PR #1442, opened 2022-07-26
Docs: add information about turbo-rails

### LINKED ISSUE #1440: Not updating turbo frames on POST when rendering from controllers
### Description

When rendering a `ViewComponent` from the controllers, it isn't updating the turbo frames tags specifically on POST. Using Rails partials it works fine. Using a Rails partial with a `ViewComponent` inside works too.

### Steps to reproduce

I made an example in this repository: https://github.com/MatheusPoliCamilo/view_component_1440

Basically a request to this `ExampleController#create` will updates a turbo frame (if the partial renders a turbo frame tag with the expected ID):
```ruby
class ExampleController < ApplicationController
  def create
    render partial: "example"
  end
end
```

But the same isn't occurring with the `ViewComponent`, this code doesn't update a turbo frame (considering that it will render a turbo frame tag with the expected ID):

```ruby
class ExampleController < ApplicationController
  def create
    render ExampleViewComponent.new
  end
end
```

### Expected behavior

Expected to the rendered `ViewComponent` by Controller updates the turbo frame when finishes the POST request, like when using Rails partials.

### Actual behavior

The rendered `ViewComponent` by Controller isn't updating the turbo frame tag.

### Workaround

Specify the `content_type: "text/html"` on the Controller render, since the response comes with the type `"text/vnd.turbo-stream.html"`.

```ruby
class ExampleController < ApplicationController
  def create
    render ExampleViewComponent.new, content_type: "text/html"
  end
end
```

### System configuration
**Rails version**: 7.0.3.1

**Ruby version**: 3.1.2

**Gem version**: 2.61.1

**Turbo rails version**: 1.1.1

MatheusPoliCamilo:

Added information about working with turbo-rails when rendering from controllers, on getting started guide.

Close #1440 

When using **turbo-rails** with **view_components**, and rendering a `ViewComponent` in the `Controller` on `#create`, the response has content type `text/vnd.turbo-stream.html`. This causes turbo frames to not update.

Just added information to use the `content_type: "text/html"` when working with **turbo-rails**.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,7 +9,7 @@ title: Changelog
 
 ## main
 
-* Improve the docs: add information about working with turbo-rails.
+* Add documentation about working with `turbo-rails`.
 
     *Matheus Poli Camilo*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Improve the docs: add information about working with turbo-rails.
+
+    *Matheus Poli Camilo*
+
 ## 2.62.0
 
 * Remove the experimental global output buffer feature.
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -99,6 +99,9 @@ end
 
 _Note: Content can't be passed to a component via a block in controllers. Instead, use `with_content`. In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout._
 
+_Note: If you are working with [turbo-rails](https://github.com/hotwired/turbo-rails), don't forget to set the
+`content_type` as `text/html`._
+
 ### Rendering ViewComponents to strings inside controller actions
 
 When rendering the same component multiple times for later reuse, use `render_in`:
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -99,7 +99,14 @@ end
 
 _Note: Content can't be passed to a component via a block in controllers. Instead, use `with_content`. In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout._
 
-_Note: When using [turbo-rails](https://github.com/hotwired/turbo-rails) be sure to set `content_type` as `text/html`._
+When using [turbo-rails](https://github.com/hotwired/turbo-rails) be sure to set `content_type` as `text/html`:
+
+```ruby
+# app/controllers/home_controller.rb
+def create
+  render(ExampleComponent.new, content_type: "text/html")
+end
+```
 
 ### Rendering ViewComponents to strings inside controller actions
 
--- a/docs/guide/getting-started.md
+++ b/docs/guide/getting-started.md
@@ -99,7 +99,7 @@ end
 
 _Note: Content can't be passed to a component via a block in controllers. Instead, use `with_content`. In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout._
 
-When using [turbo-rails](https://github.com/hotwired/turbo-rails) be sure to set `content_type` as `text/html`:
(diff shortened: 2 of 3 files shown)


keithamus (approved):

docs/guide/getting-started.md
```diff
@@ -99,6 +99,9 @@ end
 
 _Note: Content can't be passed to a component via a block in controllers. Instead, use `with_content`. In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout._
 
+_Note: If you are working with [turbo-rails](https://github.com/hotwired/turbo-rails), don't forget to set the
```
Spone, line 102: What about adding a code example just below this? It will make this clearer I think.

docs/guide/getting-started.md
```diff
@@ -99,6 +99,9 @@ end
...
+`content_type` as `text/html`._
```
joelhawksley, line 103:
```suggestion
_Note: If you are working with [turbo-rails](https://github.com/hotwired/turbo-rails), don't forget to set `content_type` as `text/html`._
```

docs/guide/getting-started.md
joelhawksley: Agreeed.

MatheusPoliCamilo: @Spone @joelhawksley Sure, I just added a code example, can you check again?

docs/guide/getting-started.md
```diff
@@ -99,6 +99,15 @@ end
 
 _Note: Content can't be passed to a component via a block in controllers. Instead, use `with_content`. In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout._
 
+When using [turbo-rails](https://github.com/hotwired/turbo-rails) be sure to set `content_type` as `text/html`:
+
+```ruby
+# app/controllers/home_controller.rb
+def create
+  render(ExampleComponent.new, content_type: "text/html")
```
Spone, line 107: Does this work?

```suggestion
  render html: ExampleComponent.new
```

jcoyne:
What if your component renders turbo stream directives? This would not be a solution in that case. 

I thinking it would be helpful if the docs explained your reasoning a bit more.

docs/guide/getting-started.md
MatheusPoliCamilo: @Spone this made the response content type "text/html", which is right, but the response will be a string with the `ViewComponent` instance, like:

- Controller:
```
class ExampleController < ApplicationController
  def create
    render html: ExampleComponent.new
  end
end
```

- Response:
```html
#<ExampleComponent:0x00007fa2f7bcf150>
```

MatheusPoliCamilo:
@jcoyne You're right, I'll change it to make it more specific that this is when we're only working with turbo frames.

docs/guide/getting-started.md
Spone: You're right! That a bit unexpected for users, I wonder if there's a way to improve that from ViewComponent side.

Spone (approved):

docs/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Improve the docs: add information about working with turbo-rails.
```
```suggestion
* Add documentation about working with `turbo-rails`.
```

docs/guide/getting-started.md
```diff
@@ -99,6 +99,15 @@ end
 
 _Note: Content can't be passed to a component via a block in controllers. Instead, use `with_content`. In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout._
 
+When using [turbo-rails](https://github.com/hotwired/turbo-rails), if you want to update a **turbo frame**, be sure to set `content_type` as `text/html`:
```
joelhawksley, line 102:
```suggestion
When using turbo frames with [turbo-rails](https://github.com/hotwired/turbo-rails), set `content_type` as `text/html`:
```

joelhawksley (approved):

Issue #1447, opened 2022-08-02
view_component + turbo + active_record callbacks can lead to a deadlock


shepmaster:
### Steps to reproduce

This is a tough one to reproduce as it deals with multithreading and occurs in a private application that I will need to spend a chunk of time to obfuscate / minimize. I can do that if useful, but I figured I'd open the issue to start discussion.

The general setup:

1. A Rails application 
1. A multithreaded web server 
1. Perform **parallel** web requests
1. One web request modifies data inside of an Active Record transaction
    - There is an `after_commit` callback that performs a websocket broadcast
    - The broadcast uses view_components to generate HTML
1. Another web request uses view_components to generate HTML  
    - It gathers data from Active Record

### Expected behavior

No deadlock occurs.

### Actual behavior

A deadlock occurs. 

### Notes / debugging

<details>
<summary>This stacktrace shows one thread acquires the Active Record connection lock, then attempts to get the view_component compiler lock</summary>

```
.../ruby-3.0.4/gems/rspec-core-3.11.0/exe/rspec:4:in `<top (required)>'
.../project/spec/features/my_work_spec.rb:163:in `block (3 levels) in <top (required)>'
.../ruby-3.0.4/gems/factory_bot-6.2.0/lib/factory_bot/evaluation.rb:18:in `create'
.../ruby-3.0.4/gems/activerecord-6.1.6.1/lib/active_record/transactions.rb:302:in `save!'
.../ruby-3.0.4/gems/activerecord-6.1.6.1/lib/active_record/connection_adapters/abstract/database_statements.rb:320:in `transaction'
.../ruby-3.0.4/gems/activerecord-6.1.6.1/lib/active_record/connection_adapters/abstract/transaction.rb:317:in `within_new_transaction'
.../ruby-3.0.4/gems/activesupport-6.1.6.1/lib/active_support/concurrency/load_interlock_aware_monitor.rb:21:in `synchronize'
.../ruby-3.0.4/gems/activerecord-6.1.6.1/lib/active_record/connection_adapters/abstract/transaction.rb:155:in `commit_records'
.../ruby-3.0.4/gems/activerecord-6.1.6.1/lib/active_record/transactions.rb:321:in `committed!'
.../ruby-3.0.4/gems/activesupport-6.1.6.1/lib/active_support/callbacks.rb:824:in `_run_commit_callbacks'
.../project/app/models/task.rb:46:in `block in <class:Task>'
[...]

PR #1448, opened 2022-08-02
Fix compiler deadlock

BlakeWilliams:
Currently there's a deadlock in the compiler due to rendering locking,
which can lead to edgecases where other locks prevent the completion of
rendering and the inability to complete rendering prevents other locks
from releasing as seen in https://github.com/github/view_component/issues/1447

This resolves the issue by replacing `Monitor` with a `ReadWriteLock`.
This allows the same component to render without requiring an exclusive
lock, resulting in concurrent renders.

In summary:

- Compilation will hold a write lock, blocking `#compile` and
  `render_template_for` calls.
- `render_template_for` calls will hold a read lock, blocking `#compile`
  but allowing other calls to `render_template_for` to check out another
  read lock.

See the documentation for `ReadWriteLock` for more details: http://ruby-concurrency.github.io/concurrent-ruby/master/Concurrent/ReadWriteLock.html

closes https://github.com/github/view_component/issues/1447

### DIFF
--- a/test/sandbox/test/rendering_test.rb
+++ b/test/sandbox/test/rendering_test.rb
@@ -1078,4 +1078,31 @@ class RenderingTest < ViewComponent::TestCase
       index += 1
     end
   end
+
+  def test_concurrency_deadlock
+    with_compiler_mode(ViewComponent::Compiler::DEVELOPMENT_MODE) do
+      with_new_cache do
+        mutex1 = Mutex.new
+        mutex2 = Mutex.new
+
+        t1 = Thread.new do
+          mutex1.synchronize do
+            sleep 0.02
+            render_inline(ContentEvalComponent.new)
+          end
+        end
+
+        t = Thread.new do
+          render_inline(ContentEvalComponent.new) do
+            mutex1.synchronize do
+              sleep 0.01
+            end
+          end
+        end
+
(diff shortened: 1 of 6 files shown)


test/sandbox/test/rendering_test.rb
```diff
@@ -1078,4 +1078,30 @@ def test_renders_nested_collection
...
+        t = Thread.new do
+          render_inline(ContentEvalComponent.new) do
+            mutex.synchronize do
+              sleep 0.01
+            end
+          end
+        end
+
+        t1.join
+        t.join
```
BlakeWilliams, line 1103: There's no assertion here, since the old code causes an error that looks like this:

```
Error:
RenderingTest#test_concurrency_deadlock:
fatal: No live threads left. Deadlock?
7 threads, 7 sleeps current:0x000000010697e260 main thread:0x00000001516046f0
* #<Thread:0x000000015205fae8 sleep_forever>
   rb_thread_t:0x00000001516046f0 native:0x00000001031ac580 int:0
   /Users/blake/Code/view_component/test/sandbox/test/rendering_test.rb:1103:in `join'
```

followed by a lot of stacktrace information 😄

joelhawksley (approved):

PR #1455, opened 2022-08-03
Raise `ArgumentError` when conflicting slots are defined.

joelhawksley:
Before this change it was possible to define Slots with conflicting names, for example:

```ruby
  renders_one :item
  renders_many :items
end
```

### DIFF
--- a/lib/view_component/slotable_v2.rb
+++ b/lib/view_component/slotable_v2.rb
@@ -80,6 +80,7 @@ module ViewComponent
       #   <% end %>
       def renders_one(slot_name, callable = nil)
         validate_singular_slot_name(slot_name)
+        validate_plural_slot_name(ActiveSupport::Inflector.pluralize(slot_name).to_sym)
 
         define_method :"with_#{slot_name}" do |*args, &block|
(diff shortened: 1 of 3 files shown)


lib/view_component/slotable_v2.rb
```diff
@@ -148,6 +149,7 @@ def renders_one(slot_name, callable = nil)
       #   <% end %>
       def renders_many(slot_name, callable = nil)
         validate_plural_slot_name(slot_name)
+        validate_singular_slot_name(ActiveSupport::Inflector.singularize(slot_name).to_sym)
 
         singular_name = ActiveSupport::Inflector.singularize(slot_name)
```
BlakeWilliams, line 154:
```suggestion
        singular_name = ActiveSupport::Inflector.singularize(slot_name)
        validate_plural_slot_name(slot_name)
        validate_singular_slot_name(ActiveSupport::Inflector.singularize(slot_name).to_sym)
```
Minor, but could we re-use `singular_name` right below?

BlakeWilliams (approved):
Nice catch!

PR #1463, opened 2022-08-10
Fix request.path inside with_request_url helper

franzliedke:

`#with_request_url` is great. It was started in #1058, but later #1221
introduced a tiny regression in the parsing of segments.

When using `#with_request_url` including a query string, the `#path`
helper would now contain the query string segment, when it should not.
This led to `#fullpath` having the query string twice.

I added regression tests to the existing scenarios and then fixed the parsing to make them pass. :)

I love this library. 😍

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,7 +9,7 @@ title: Changelog
 
 ## main
 
-* Fix `#with_request_url` (when used with query string) resulting in incorrect `request.path` and `request.fullpath`.
+* Fix bug where `#with_request_url`, when used with query string, set the incorrect `request.path` and `request.fullpath`.
 
     *Franz Liedke*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Fix `#with_request_url` (when used with query string) resulting in incorrect `request.path` and `request.fullpath`.
+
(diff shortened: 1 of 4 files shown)


franzliedke:
If I see things correctly, the failing build seems to also happen in `main` and originate from #1462.

lib/view_component/test_helpers.rb
```diff
@@ -190,10 +190,14 @@ def with_request_url(path)
       old_request_query_string = request.query_string
       old_controller = defined?(@controller) && @controller
 
+      # Started in https://github.com/github/view_component/pull/1058
+      # Broken in https://github.com/github/view_component/pull/1221
+
```
joelhawksley, line 195:
```suggestion
```

docs/CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ title: Changelog
 
 ## main
 
+* Fix `#with_request_url` (when used with query string) resulting in incorrect `request.path` and `request.fullpath`.
```
```suggestion
* Fix bug where `#with_request_url`, when used with query string, set the incorrect `request.path` and `request.fullpath`.
```

joelhawksley (approved):
Thank you for your wonderful contribution! ❤️

lib/view_component/test_helpers.rb
franzliedke: Whoops, forgot my good old TODO notes. 😇

PR #1472, opened 2022-08-15
Update gemspec

joelhawksley:

I'm trying to update our `gemspec` and license to better recognize that ViewComponent isn't just built by GitHub ❤️


camertron (approved):
Nice 🎉

PR #1476, opened 2022-08-17
Adjust stack trace in slot name deprecation warning output

mxriverlynn:

The current logging for slot name deprecation warnings displays the file in which the deprecation warning is called. To be more useful in finding and correcting deprecations, this PR adjusts the stack trace output to reflect the location of the named slot call.

The [core `ActiveSupport::Deprecation#warn` method implementation](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/deprecation/reporting.rb#L18-L29) includes a call to `caller_locations(2)`. This returns the location of the deprecation call, instead of the location where the slot is used. 

To correct the issue, `caller_locations(3)` is called from within the slotable deprecation check, adding the corrected stack trace to the `Deprecation#warn` call. 

I chose to add two new tests instead of adjusting the existing tests, to emphasize what each test is looking for. This does duplicate a small amount of the testing, but the readability and maintainability is worth the duplication.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,6 +13,11 @@ title: Changelog
 
     *Thomas Hutterer*
 
+* Corrects the deprecation warning for named slots to show the file and line
+  where the slot is called
+
+    *River Bailey*
+
 ## 2.68.0
 
 * Update `gemspec` author to be ViewComponent team.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -13,8 +13,7 @@ title: Changelog
 
     *Thomas Hutterer*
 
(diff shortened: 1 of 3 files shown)


mxriverlynn:
anyone have an idea why the PVC checks are failing? the build details for those steps are not terribly informative

docs/CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ title: Changelog
 
     *Thomas Hutterer*
 
+* Corrects the deprecation warning for named slots to show the file and line
+  where the slot is called
```
```suggestion
* Corrects the deprecation warning for named slots to show the file and line where the slot is called.
```

joelhawksley (approved):

PR #1486, opened 2022-08-23
clarify wording around testability of rails views

joelhawksley:
On Twitter, @LynxEyes and @chalmagean pointed out that our claims about testability of Rails views are not true: https://twitter.com/Lynx_Eyes/status/1561277373823307777

I think this updated language is more appropriate :heart:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -53,7 +53,7 @@ ViewComponent methods are implemented within the scope of the template, encapsul
 
 ### Testing
 
-ViewComponents are meant to be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
+ViewComponent was designed with the intention that all components should be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
 
 With ViewComponent, integration tests can be reserved for end-to-end assertions, with permutations covered at the unit level.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -53,7 +53,7 @@ ViewComponent methods are implemented within the scope of the template, encapsul
 
 ### Testing
 
-Unlike traditional Rails templates, ViewComponents can be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
+Unlike Rails templates, ViewComponents are meant to be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
 
 With ViewComponent, integration tests can be reserved for end-to-end assertions, with permutations covered at the unit level.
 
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -53,7 +53,7 @@ ViewComponent methods are implemented within the scope of the template, encapsul
 
 ### Testing
 
-Unlike traditional Rails templates, ViewComponents can be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
+Unlike Rails templates, ViewComponents are meant to be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
```
joelhawksley, line 56:
```suggestion
ViewComponents are meant to be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
```

docs/index.md
```diff
@@ -53,7 +53,7 @@ ViewComponent methods are implemented within the scope of the template, encapsul
 
 ### Testing
 
-Unlike traditional Rails templates, ViewComponents can be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
+ViewComponents are meant to be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
```
BlakeWilliams, line 56:
```suggestion
ViewComponent was designed with the intention that all components should be unit tested. In the GitHub codebase, ViewComponent unit tests are over 100x faster than similar controller tests.
```

I think something like this could make a stronger argument of "if you aren't testing your components, you're doing something wrong", but the existing wording is great too. 👍

BlakeWilliams (approved):

LynxEyes:
@joelhawksley I didn't said it is easy, and I didn't said ANY tradicional rails view can be tested without first obeying a few rules but they CAN be tested, so no, it is not a false claim.

That said, I NEVER said testing rails views is convenient or easy, just as I never said it is better than this. What I said, and keep saying, is that claiming that unit testing rails views can't be done is a false claim.

LynxEyes:
Personally, I'd word it like "unit testing rails views can be quite hard as ALL your views need to obey rules like only using local variables, requires all controller actions do do explicit renders, etc.. No one writes rails apps this way"

joelhawksley:
> Personally, I'd word it like "unit testing rails views can be quite hard as ALL your views need to obey rules like only using local variables, requires all controller actions do do explicit renders, etc.. No one writes rails apps this way"

I get what you're saying. I'm going to merge this for now and give it some more thought.

LynxEyes:
@joelhawksley I've re-read your statements on this PR and my response.. and I feel compelled to apologise, I read that in a rush and responded as a douche... I'm sorry, will avoid in any future occasion.

PR #1489, opened 2022-08-24
Lock compile cache while rendering

### LINKED ISSUE #1488: Compile deadlock with ActiveJob
I think the fix to this issue #1447 has created a regression because I am now experiencing deadlocks when using view_component and turbo streams.

After a lot of digging I found out that ViewComponent may reset the CompileCache while a request running in another thread is rendering some ViewComponents.

The line that refreshes the cache is this one and It seems to be triggered right after action cable streams a message (So probably after the BoradcastJob is performed).
https://github.com/github/view_component/blob/6397e87fcd9e606212493f31127943e13e3e336b/lib/view_component/engine.rb#L136

The fact that the cache is reset makes the rendering on the other thread to recompile the view component event though it is already compiled. Because of that it tries to get a write lock while still holding a read lock thus resulting in a deadlock.
(If it already holds a read lock that is because my components are nested)

To sum it up

```
In thread #1
Send turbo stream message
Finish BroadcastJob
Reset CompileCache

In thread #2
Render view component A <-- (Compile and is now cached)
  Render view component A <-- (Knows it is compiled so renders with read lock)
    -- Thread #1 has reset cache --
    Render view component A <-- (Thinks it needs to compile and try to get write lock)
      deadlock because we already have a read lock.
```

### Steps to reproduce
Unfortunately that issue is extremely hard to reproduce because of its nature

### Expected behavior
No deadlock

### Actual behavior
Deadlock :(

paulhenri-l:

This should fix #1488 by preventing any thread to clear the compile cache while another thread is rendering.

Given that the deadlock is caused by one thread emptying the cache while another one is still rendering the simplest approach seemed to be waiting for the rendering to be done. As this only happens in development it should not impact performance.

Given that I don't usually write concurrent code in ruby I'm not sure I tested that in the best possible way 😄

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -26,6 +26,10 @@ nav_order: 5
 
     *Marc Köhlbrugge*
 
+* Fix potential deadlock scenario in the compiler's development mode.
+
+  *Paul-Henri Leobon*
+
 ## 2.69.0
 
 * Add missing `require` to fix `pvc` build.
--- a/lib/view_component/compile_cache.rb
+++ b/lib/view_component/compile_cache.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require "view_component/compile_cache_lock"
+
 module ViewComponent
   # Keeps track of which templates have already been compiled
   # This isn't part of the public API
@@ -9,7 +11,7 @@ module ViewComponent
     end
 
     mattr_accessor :lock, instance_reader: false, instance_accessor: false do
-      Concurrent::ReadWriteLock.new
+      ViewComponent::Lock.new
     end
 
     module_function
--- a/lib/view_component/compile_cache.rb
+++ b/lib/view_component/compile_cache.rb
@@ -8,6 +8,10 @@ module ViewComponent
       Set.new
     end
 
+    mattr_accessor :lock, instance_reader: false, instance_accessor: false do
+      Concurrent::ReadWriteLock.new
+    end
+
     module_function
 
     def register(klass)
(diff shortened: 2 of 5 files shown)


paulhenri-l:
Okay so it appears that my solution still produces deadlocks as the read write lock will wait for readers to finish whenever a new writer tries to read.

paulhenri-l:
I’ll work on another solution. Making sure no one can clear the cache until there are readers.

paulhenri-l:
Well I have a solution that is now working on my side and the tests are still passing 😃 

As I am quite new to ruby multi threading and synchronization, I am not sure I went with the best solution but it seems to do the job.

I initially wanted to use the ReadWriteLock but as it prevents new reader to get the lock while there is a waiting writer the code was still deadlocking. 

That was because when rendering nested components we keep acquiring read locks and won't release them until we come back from all the nesting. While rendering another thread tried to acquire the write lock resulting in the lock preventing new readers from locking until the writer got the lock. If that makes sense :D

So I decided to create a new lock that is very similar to the ReadWriteLock with the only difference that readers are considered more important than writers. So a writer will only be able to write if no one is reading and it won't prevent new readers from acquiring the lock.

BlakeWilliams:
👋 Thanks for diving into this problem and opening this PR!

I'm a little hesitant about us maintaining our own concurrency primitive. I wonder if instead, we need to make the compile cache lock aware itself? e.g. In development each render should lock the cache to prevent it from writing until all renders are complete. I'm not sure how practical that would be though.
 
It's also possible our current implementation of having a separate compiler and separate compiler cache may not suit itself well to the concurrency challenges we have to solve, so re-imagining that implementation/approach may be on the table too.

PR #1497, opened 2022-08-25
[Do not merge] Minimal repro for issue testing lambda slots

just-joshing:

This PR is a minimal repro of a test issue and is not meant to be merged.

I'm writing a new component with a lambda slot that renders another component with its own slots.
My goal is to pass the block that was given to my component's slot to the slot of the component I'm rendering in my slot.

~The first oddity I encountered is that in my slot, I can't just instantiate the component it should render. I have to call `render` or else the slot doesn't output the component. However once I do then everything works the way I want visually.~
Edit: [This one is solved](https://github.com/pulls#discussion_r978981926).

But then I hit the second issue. I want to test that the content returned by the block given to my component's slot ends up rendered. But for some reason I cannot seem to figure out how to write my test to get it to output as expected.

This PR is a minimal repro to hopefully help diagnose the issue I'm facing.

Here's a screenshot of the expected rendered output of the minimal repro.
It behaves the way I want when rendered. It's just the test that isn't behaving.
screenshot of expected rendered output

One other oddity to mention is that if the block given to my slot returns a `String` instead of a `ViewComponent`, things seem to work. I feel like there's a number of little things at play here that are getting in the way. Like the fact that I had to call `render` in my slot.

Also, I want to call out that in my actual production code I'm rendering 3 different `ViewComponents`. It was just easier for the minimal repro to write one that "renders itself".

So in my prod example it'd look more like:
```ruby
<%= render Component1.new do |c| %>
  <% c.with_component_2 do %>
    <%= render Primer::Beta::Text.new do %>
      Content <%= render(Primer::LinkComponent.new(href: "?")) { "Link" } %>.
    <% end %>
  <% end %>
<% end %>
```


test/sandbox/app/components/just_joshing_repro_component.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class JustJoshingReproComponent < ViewComponent::Base
+  renders_one :slot1, -> (&block) do
+    # Won't render slot2 content unless I call 'render' here.
+    render JustJoshingReproComponent.new do |c|
+      c.with_slot2(&block)
+    end
```
just-joshing, line 8: To reviewers:

Callout here. Not sure why I have to call `render` to make this work.

just-joshing: Thanks @BlakeWilliams for helping me understand this issue.
I was able to change this code to
```suggestion
    JustJoshingReproComponent.tap do |c|
      c.with_slot2(&block)
    end
```

but I'm still having issues with the test failure.

BlakeWilliams: Can you verify if the HTML output looks correct?

It looks like it's failing on line 53, not line 66 as commented (assuming minitest is reporting it correctly)?

> expected to find css "> [data-test-selector='just-joshing-repro-component']" within #<Capybara::Node::Simple tag="div" path="/html/body/div/div"> 1 time but there were no matches

joelhawksley:
Closing as stale ❤️

Issue #1498, opened 2022-08-29
Improve error message when using `renders_one :content`


danieldiekmeier:
## Feature request

When naming a slot `content`, ViewComponent shows an argument error:

```rb
class MyComponent < ViewComponent::Base
  renders_one :content
end
```

```
ArgumentError: 
  AlertComponent declares a slot named content, which is a reserved word in the ViewComponent framework. 
  To fix this issue, choose a different name.
```

The argument validation happens here: https://github.com/github/view_component/blob/c5308ae62508925408625fffce3b420719093182/lib/view_component/slotable_v2.rb#L274-L280

I think this error message could be improved to tell the user that `content` already _is_ a slot, and they don't have to do anything.

### Motivation

While this error message is _correct_, I think it's not helpful enough. In our codebase, it lead to many developers creating a slot with another name, like `:text` or `:fill` or `:inside`, which has several problems:

- It' a worse name than `:content`
- It's completely unnecessary, if only we knew that `content` was an implicit slot that could always be used.

I think the ergonomics of this validation could be improved by explaining the special case of `content`.

### Suggestion

I could take a stab at solving this by introducing a special case for `:content`, which would say something like

```
ArgumentError:
  AlertComponent declares a slot named content, which is the name of the implicit slot. 
  To fix this issue, either use the implicit slot directly or choose a different name.
```

If this is acceptable for you, I'll gladly open a PR. 🥰

Spone:
Thanks for opening an issue to discuss this! I'm all for improving error messages to be more helpful and actionable.

In this case, I'm not sure that using the phrase "implicit slot" make things clearer. I don't think we refer to `content` this way in the docs (but maybe I'm mistaken?)

Could you suggest another phrasing for the error message?

danieldiekmeier:
@Spone You're right, that's a good point. In https://viewcomponent.org/guide/getting-started.html#implementation, I found this wording:

> Content passed to a ViewComponent as a block is captured and assigned to the `content` accessor.

My suggestion based on this would be something like:

```
ArgumentError:
  AlertComponent declares a slot named content, which is a reserved word in the ViewComponent framework.
  Content passed to a ViewComponent as a block is captured and assigned to the `content` accessor without having to create an explicit slot.
  To fix this issue, either use the `content` accessor directly or choose a different name.
```

But I'm afraid it's a little long. What do you think?

Spone:
[...]

PR #1506, opened 2022-08-31
Add special case for `renders_one :content`

danieldiekmeier:

As proposed in #1498, this PR adds a special error message when calling `renders_one :content`. 

The new error message looks like this:

```
ArgumentError: ExampleComponent declares a slot named content, which is a reserved word in the ViewComponent framework.

Content passed to a ViewComponent as a block is captured and assigned to the `content` accessor without having to create an explicit slot.

To fix this issue, either use the `content` accessor directly or choose a different name.
```

`content` is already very similar to a slot, so suggesting to the user that it may do what they want seems like a nice idea to help them improve their code.

I decided to add the special case above the general case in `validate_singular_slot_name`, instead of somehow generalizing it. I also didn't touch `RESERVED_NAMES[:singular]`, even though it is only used in this one spot and the `:content` check will now never happen. If you want me to clean this up, feel free to indicate that!

Not at the moment.

### DIFF
--- a/lib/view_component/slotable_v2.rb
+++ b/lib/view_component/slotable_v2.rb
@@ -272,6 +272,14 @@ module ViewComponent
       end
 
       def validate_singular_slot_name(slot_name)
+        if slot_name.to_sym == :content
+          raise ArgumentError.new(
+            "#{self} declares a slot named content, which is a reserved word in the ViewComponent framework.\n\n" \
+            "Content passed to a ViewComponent as a block is captured and assigned to the `content` accessor without having to create an explicit slot.\n\n" \
+            "To fix this issue, either use the `content` accessor directly or choose a different name."
+          )
+        end
+
         if RESERVED_NAMES[:singular].include?(slot_name.to_sym)
           raise ArgumentError.new(
             "#{self} declares a slot named #{slot_name}, which is a reserved word in the ViewComponent framework.\n\n" \
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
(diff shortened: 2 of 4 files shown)


Spone:
Hi @danieldiekmeier, sorry for the delay in reviewing this. Is there a reason why you closed the PR?

danieldiekmeier:
Sorry, my bad. I accidentally removed my fork after https://github.com/ViewComponent/view_component/pull/1511 was merged. I recreated it and I think it should be fine now.

lib/view_component/slotable_v2.rb
```diff
@@ -272,6 +272,14 @@ def validate_plural_slot_name(slot_name)
       end
 
       def validate_singular_slot_name(slot_name)
+        if slot_name.to_sym == :content
+          raise ArgumentError.new(
+            "#{self} declares a slot named content, which is a reserved word in the ViewComponent framework.\n\n" \
```
joelhawksley, line 277:
```suggestion
            "#{self} declares a slot named content, which is a reserved word in ViewComponent.\n\n" \
```

lib/view_component/slotable_v2.rb
```diff
@@ -272,6 +272,14 @@ def validate_plural_slot_name(slot_name)
...
+            "Content passed to a ViewComponent as a block is captured and assigned to the `content` accessor without having to create an explicit slot.\n\n" \
+            "To fix this issue, either use the `content` accessor directly or choose a different name."
```
joelhawksley, line 279:
```suggestion
            "To fix this issue, either use the `content` accessor directly or choose a different slot name."
```

joelhawksley (approved):

joelhawksley:
@danieldiekmeier mind resolving the merge conflict?

danieldiekmeier:
Recreated at #1514. (I'll also fix the messages really quick.)

PR #1516, opened 2022-09-14
Deprecate support for Ruby < 2.7

joelhawksley:

We are deprecating support for Ruby < 2.7 in the coming ViewComponent release: https://github.com/ViewComponent/view_component/issues/617

I added a deprecation warning to the component compiler as I thought it would be fairly visible here.


BlakeWilliams (approved):
LGTM

PR #1529, opened 2022-09-21
remove experimental _after_compile API

joelhawksley:

This PR remove removes the experimental `_after_compile` API.

I deleted it. As it's an experimental API, this is not breaking in SemVer.


BlakeWilliams (approved):
:shipit: 

I think we're using this a few places in dotcom, but otherwise LGTM

PR #1530, opened 2022-09-21
Rename _after_compile to after_compile., Remove experimental docs labels

joelhawksley:

This PR renames `_after_compile` to `after_compile`, making it a public API.

I renamed the method, leaving the original in place while we deprecate it. As this is an experimental API we do not need to make a breaking SemVer change.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Rename `_after_compile` to `after_compile`. `_after_compile` is deprecated and will be removed in `v3.0.0`.
+
+    *Joel Hawksley*
+
 * Remove experimental labels from docs.
 
     *Joel Hawksley*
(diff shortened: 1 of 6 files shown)


Spone (approved):

joelhawksley:
> > This PR renames _after_compile to after_compile, making it a public API.
> 
> I'm wondering if we actually want this to become a public API as-is. What are the current use-cases, are there alternatives, is this being used in the wild, etc.

@BlakeWilliams we use it in two places in PVC: https://github.com/primer/view_components/blob/4e87f1d2bdd8330570e587d8e09c120db67d20ba/app/components/primer/octicon_component.rb#L86 and https://github.com/primer/view_components/blob/4e87f1d2bdd8330570e587d8e09c120db67d20ba/app/components/primer/octicon_symbols_component.rb#L42.

Looking at those use-cases, I wonder if they could use an application initializer instead.

PR #1533, opened 2022-09-23
Remove _after_compile

joelhawksley:

Sorry for the back and forth, but I think we can remove this after all. The only usage we have anywhere at GitHub is in PVC and I think it's actually not a great use case. I've written a PR to move it to an initializer: https://github.com/primer/view_components/pull/1402.

Deletion. This is an experimental API with no deprecation or SemVer guarantees.


docs/guide/testing.md
```diff
@@ -31,9 +31,6 @@ _Note: `assert_selector` only matches on visible elements by default. To match o
 Since 2.56.0
 {: .label }
 
-Experimental
```
BlakeWilliams, line 34: Just double checking, we intentionally removed the experimental flags here?

BlakeWilliams (approved):

docs/guide/testing.md
joelhawksley: Yup 👍🏻

PR #1548, opened 2022-10-06
Allow render_parent to facilitate setting slots

camertron:

A few months ago we introduced the `render_parent` method as a bit of sugar around calling `super` in template code (it prevents double-rendering). However `render_parent` doesn't accept a block, making it feel different than `render`. It also means setting slots is sort of awkward:

```html+erb
<% with_some_slot(args) do |slot| %>
  <% capture do %>
    <%= render(AnotherComponent.new)) %>
  <% end %>
<% end %>

<%= render_parent %>
```

This PR modifies `render_parent` to `yield self` so this is possible:

```html+erb
<%= render_parent do |parent| %>
  <% parent.with_some_slot(args) do |slot| %>
    <%= render(AnotherComponent.new)) %>
  <% end %>
<% end %>
```

I simply added a `yield self` in `render_parent`. Seems to work 👍 

As @BlakeWilliams pointed out, it feels like `parent` in the example above is another object rather than `self`. I'm curious to hear what others think and reach a consensus.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -153,8 +153,17 @@ module ViewComponent
     # <% super %> # does not double-render
     # ```
     #
-    # Calls `super`, returning `nil` to avoid rendering the result twice.
+    # Calls `super`, returning `nil` to avoid rendering the result twice. Accepts a
+    # block that will be called with self, which can be used to define slots, eg:
+    #
+    # <%= render_parent do |c| %>
+    #   <% c.with_some_slot(...) %>
+    # <% end %>
+    #
     def render_parent
+      # yield the component here to make defining slots more ergonomic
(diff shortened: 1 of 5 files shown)


camertron:
Hey @dkniffin, good question. We decided to close this because it was confusing to yield `self` to the `render_parent` call, since the parent and `self` are the same object. Yielding makes them look like different objects.

To your specific question, it is possible to render additional content inside a parent's template by calling the `#with_content` method, although that call will have to happen in `#before_render` or perhaps `#render_in`. I tried hacking on a better solution last week but it proved to be too difficult. We rarely run into this problem at GitHub, so we decided the workarounds we already have (eg. calling `#with_content` in `#render_in`) are ok for now.

PR #1550, opened 2022-10-07
Remove locking mechanisms

camertron:

Over the last few months we've seen [several](https://github.com/ViewComponent/view_component/issues/869) [issues](https://github.com/ViewComponent/view_component/issues/1447) [filed](https://github.com/ViewComponent/view_component/issues/1488) regarding threadsafety in view_component. Our awesome community members have also submitted [several](https://github.com/ViewComponent/view_component/pull/1489) [PRs](https://github.com/ViewComponent/view_component/pull/1202) attempting to fix these issues.

Unfortunately despite our best efforts we have yet to come up with a viable solution that works in every case. In the process, we've made the compiler code more difficult to understand and reason about. This is absolutely nothing against the fine work already contributed to the framework - it's just an observation about the current state of the code.

My proposal with this PR is to remove our locking mechanisms in their entirety. The compiler lock was [originally introduced](https://github.com/ViewComponent/view_component/pull/1161) for two reasons:

1. to prevent race conditions where:
    1. Thread A checks for the existence of a particular compiled method. The method does not exist, so the thread compiles it and defines the method.
    1. Before thread A can add the component class to the compile cache, thread B enters the compile routine. Since the component class is not cached, thread B attempts to compile the method, and calls `undef_method` to avoid a Ruby method redefinition warning.
    1. Thread A attempts to call the method, which was just undefined by thread B.
    1. A `NoMethodError` is raised.
1. to avoid compiling template methods multiple times, ostensibly saving memory and execution time.

How does this PR address these issues?

1. Race conditions are only possible because of the non-atomic, 2-step operation of undefining and then redefining the method. If we redefine the method without first undefining it, there should be no way to achieve the same race condition.
1. I do not believe the performance impact of defining the same method multiple times justifies the complexity of the locking mechanisms we currently have in the compiler.
    1. The method will be compiled at most once per thread, meaning the cost should only affect apps during startup.
    1. Defining the method is idempotent, so there's no risk of defining a method with different behavior halfway through a request.

I completely removed the compiler lock, associated methods, and calls to `undef_method`. I also wrapped the method definition code in `silence_warnings` to suppress Ruby's method redefinition warnings.

/cc @shepmaster @horiaradu @paulhenri-l @stillheart @allmarkedup


BlakeWilliams:
Can we share this in https://github.com/ViewComponent/view_component/issues/1488 and ask folks to try this branch out and see if it resolves their issues?

lib/view_component/compiler.rb
```diff
@@ -38,46 +34,39 @@ def compile(raise_errors: false, force: false)
...
 
-          if component_class.instance_methods.include?(method_name.to_sym)
-            component_class.send(:undef_method, method_name.to_sym)
-          end
+      templates.each do |template|
+        # Remove existing compiled template methods,
+        # as Ruby warns when redefining a method.
+        method_name = call_method_name(template[:variant])
 
+        silence_warnings do
```
BlakeWilliams, line 69:
```suggestion
        method_name = call_method_name(template[:variant])
        component_class.silence_redefinition_of_method(method_name)
```

Thoughts on using the [ActiveSupport `silence_redefinition_of_method`](https://www.rubydoc.info/gems/activesupport/Module:silence_redefinition_of_method) method instead here? That would let us keep any non-redefinition warnings here, and maybe be a bit more intention revealing.

Spone: I'm all for being more specific!

camertron: Oh nice!

camertron: Unfortunately tests failed using the `silence_redefinition_of_method` approach because it's not threadsafe 🤦 It first checks to see if the method in question is defined, then cleverly uses a side-effect of `alias_method` to define the method. Unfortunately the method might not exist during the check but exist immediate after it, meaning the `class_eval` will still result in a warning. I pushed up a threadsafe version that passes all our tests, including the ones that use threads.

BlakeWilliams: I think the solution you has will work, but it's tracking per-thread which seems unnecessary?

It feels like a scenario where we'd want to actually use a lock, which I think would be safe since it could be tracked as internal compiler state?

I'll add a comment in the code too real quick.

lib/view_component/compiler.rb
```diff
@@ -38,46 +34,39 @@ def compile(raise_errors: false, force: false)
...
 
-          if component_class.instance_methods.include?(method_name.to_sym)
-            component_class.send(:undef_method, method_name.to_sym)
-          end
+      templates.each do |template|
+        # Remove existing compiled template methods,
+        # as Ruby warns when redefining a method.
+        method_name = call_method_name(template[:variant])
 
+        silence_redefinition_of_methods(method_name) do
```
BlakeWilliams, line 69: I know we generally want less locks, but I think this could solve our multi-threading issue with `silence_redefinition_of_method`:

```ruby
# somewhere above we do: @redefinition_lock = Thread::Mutex.new
@redefinition_lock.synchronize do
  silence_redefinition_of_method(method_name)
end
```

This should be safe since we don't hold the lock in a (potentially) recursively called method like we did with templates, and there's no state like `CompileCache` that can cause deadlocks.

Thoughts on going with this approach to avoid having to track/call `silence_redefinition_of_method` per-thread?

camertron: I was skeptical at first (because the whole point of this PR is to _remove_ locks 😅 ) but it appears to work 👍 It makes sense to me why it works too, since we're still not undefining any methods before defining them, thereby avoiding the race condition.

BlakeWilliams (approved):

camertron:
Thanks for the heads-up about this @DRBragg! We've opted to officially drop support for Rails versions <= 5.1, which is end-of-life. We don't even test 5.1 or older in our CI, which is why we didn't catch this.

DRBragg:
> Thanks for the heads-up about this @DRBragg! We've opted to officially drop support for Rails versions <= 5.1, which is end-of-life. We don't even test 5.1 or older in our CI, which is why we didn't catch this.

Yea, I kinda figured that was the case or at least happening in the near future. I opened #1553 to help make it clear.

camertron:
Alright, after some additional internal discussion, we've decided to continue support for >= 5.0. We consider dropping support for a Rails version as a breaking change, so until view_component v3, we'll continue to support those older versions.

PR #1552, opened 2022-10-14
Update lambda slot docs to mention lazy rendering

### LINKED ISSUE #1527: Delay running slot lambda until after the `render_in` block
## Feature request

It would be nice if slot lambdas were not called until after the block passed to `render_in` is called.

### Motivation

Sometimes I need to change the behaviour of a slot depending on the existence of another slot. A good example of this is a form field component that supports displaying an error message:

```ruby
class FieldComponent < ViewComponent::Base
  renders_one input, ->(type: "text") { text_field_tag name, value, class: ("error" if error?), type: }
  renders_one error
end
```

If this component is rendered like this:

```ruby
render FieldComponent.new do |field|
  field.with_error { "Error message" }
  field.with_input
end
```

then the `error` class is added as expected, however the following does not include the `error` class as `error?` returns `false`:

```ruby
render FieldComponent.new do |field|
  field.with_input
  field.with_error { "Error message" }
end
```

Unfortunately in this scenario the code reads better if the `with_input` comes first. If the lambda was called after the block finished this would not be an issue. I believe this would also be the expected behaviour for most people as you would expect the order of calling `with_` methods for `renders_one` slots to not matter.

TastyPi:
Documentation change discussed in #1527 

Fixes #1527


docs/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ nav_order: 5
 
     *Jonathan del Strother*
 
+* Mention lambda slots rendering returned values lazily in the guide
```
```suggestion
* Mention lambda slots rendering returned values lazily in the guide.
```

docs/guide/slots.md
```diff
@@ -217,6 +217,8 @@ class BlogComponent < ViewComponent::Base
 end
 ```
 
+Note that while the lambda is called when the `with_*` method is called, if an object that responds to `render_in` is returned (like a component) then it won't be fully rendered until it is first used.
```
joelhawksley, line 220:
```suggestion
_Note: While a lambda is called when the `with_*` method is called, a returned component is not rendered until first use._
```

joelhawksley (approved):

PR #1555, opened 2022-10-18
Support older Rails versions

camertron:

We consider it a breaking change to drop support for any particular Rails version, even if that version is technically end-of-life. We stopped testing against Rails 5.0 and 5.1 but never officially stated they were unsupported. This PR adds support back for these older Rails versions, which will be officially dropped in view_component v3.0.


lib/view_component/test_helpers.rb
```diff
@@ -200,7 +200,7 @@ def with_request_url(path)
 
       path, query = path.split("?", 2)
       request.path_info = path
-      request.path_parameters = Rails.application.routes.recognize_path_with_request(request, path, {})
+      request.path_parameters = Rails.application.routes.recognize_path(path, request.env)
```
joelhawksley, line 203: Can you add a TODO to switch this back in V3?

test/sandbox/app/components/collection_parameter_with_active_model_component.rb
```diff
@@ -1,9 +1,14 @@
 # frozen_string_literal: true
 
 class CollectionParameterWithActiveModelComponent < ViewComponent::Base
-  include ActiveModel::Attributes
+  # This file gets eager loaded, so we have to check the Rails version to avoid
```
joelhawksley, line 4: Add a TODO here as well?

test/sandbox/config/routes.rb
```diff
@@ -32,7 +32,7 @@
   get :inherited_sidecar, to: "integration_examples#inherited_sidecar"
   get :inherited_from_uncompilable_component, to: "integration_examples#inherited_from_uncompilable_component"
 
-  constraints(lambda { |request| request.env["warden"].authenticate! }) do
+  constraints(lambda { |request| Thread.current[:constrained_route_enabled] }) do
```
joelhawksley, line 35: I feel hesitant about this change. This does not feel like an equivalent test. Maybe we could use an integration test instead?

joelhawksley (commented):
Mind testing this against the GitHub.com test suite before we merge?

camertron:
*Sigh*. Ok after even more internal discussion, we've decided to focus our efforts towards releasing view_component v3, which will drop support for EOL Rails versions.

PR #1556, opened 2022-10-20
Enable (experimental) single file components

BlakeWilliams:
This change adds an experimental module,
`ViewComponent::InlineTemplate` that allows you to define your templates using a basic DSL directly within your component. This eliminates the need for sidecar templates.

This is what it looks like:

```ruby
class MyComponent < ApplicationComponent
  include ViewComponent::InlineTemplate

  template! <<~ERB
    <div>
      <h1>Hello, <%= @name %>!</h1>
    </div>
  ERB

  def initialize(name:)
    @name = name
  end
end
```

This results in component being "self-contained", in that a single file can represent the entire component. This makes it easier to understand a component and requires less jumping around between files. This also has benefits like enabling the following, which feels closely related to our multi-template efforts:

(this is not implemented, just an example of what this functionality can enable)

```ruby
class MyComponent < ApplicationComponent
  include ViewComponent::InlineTemplate

  template! <<~ERB
    <div>
      <h1>Hello, <%= @name %>!</h1>

      <% post.each do.%>
        <%= render_row(post) %>
      <% end %>
    </div>
  ERB

  fragment! :row, -> (post) {
    <<~ERB
    <tr>
      <td><%= post.title %></td>
      <td><%= post.body %></td>
    </tr>
    ERB
  }

  def initialize(name:)
    @name = name
  end

  def posts
    Post.all
  end
end
```


cheshire137:
> eliminates the need for sidecar templates

What do you mean by a sidecar template? Just "the .html.erb file with the same name as the component .rb file in the same directory"?

BlakeWilliams:
> What do you mean by a sidecar template? Just "the .html.erb file with the same name as the component .rb file in the same directory"?

Yep, exactly.

fsateler:
> @fsateler would this satisfy your requirements for multiple templates?

If `fragment` is implemented, yes. I will probably end up doing `fragment :foo, File.read(_dir_ + '/foo.html.erb')` anyway though.

joelhawksley:
> > @fsateler would this satisfy your requirements for multiple templates?
> 
> If `fragment` is implemented, with arguments, yes. I will probably end up doing `fragment :foo, File.read(_dir_ + '/foo.html.erb')` anyway though.
> 
> Although it seems to me this is punting on the difficult part, which is the arguments. The (not implemented) example would have a lambda taking an argument that returns an ERB string, how would that get turned into `def render_foo(post)`? Invoke the lambda passing `nil` as argument? what if the user tries to use the arguments directly in the template (as in string interpolation: `"#{post.name}"`). ?

@fsateler What if we used the arguments approach from your PR, including them in the heredoc?

BlakeWilliams:
I don't think the initial PR should support `fragment!`, but I think we have a few ways that we could potentially support arguments. Ideally I'd like to keep it really simple and use Ruby primitives, but if we can't achieve that we can always make some compromises to keep it simple but functional. e.g.

```ruby
  fragment! :row, accepts: [:post], <<~ERB
    <tr>
      <td><%= post.title %></td>
      <td><%= post.body %></td>
    </tr>
    ERB
```

We could take the `accepts` array and use that to build the `build_row` parameters definition e.g. `def render_row(post:)`. I was hoping we could use `Proc#parameters` in the original example, but it (sadly) omits default argument values.

test/sandbox/test/inline_erb_test.rb
```diff
@@ -0,0 +1,55 @@
...
+    template! <<~ERB
+      <h1>Hello, <%= name %>!</h1>
+    ERB
+
+    # fragment idea
+    # fragment :foo, -> (a, b, c) {
+    #   <<~ERB
+    #   ERB
+    # }
+
```
BlakeWilliams, line 20:
```suggestion
```

joelhawksley:
@fsateler @BlakeWilliams why not use the Strict Locals syntax here? It's a solved problem 😄

fsateler:
It has just occurred to me, will this give proper line numbers in stack traces?

BlakeWilliams:
> It has just occurred to me, will this give proper line numbers in stack traces?

As-is no, it's something I need to write a test for, but it's definitely doable.

test/sandbox/test/inline_erb_test.rb
```diff
@@ -0,0 +1,49 @@
...
+      @name = name
+    end
+  end
+
+  class InlineSlimComponent < ViewComponent::Base
+    include ViewComponent::InlineTemplate
+
+    attr_reader :name
+
+    template_language! :slim
```
joelhawksley, line 25: Why not make the language an argument of `template!`?

BlakeWilliams: I considered it, but I went with two different methods for a few reasons. The first is I wanted components to have a consistent templating language. e.g. if you have Slim and ERB in the same component that's going to be a bit difficult to read since you have to translate between the two.

The second reason is because I wanted `ApplicationComponent` to be able to set a default. That way all of your components can inherit the `template_language!` so it "just works™" instead of requiring callers to pass `slim` to each `template!` call.

BlakeWilliams:
I just marked this as ready for review. It's pretty minimal, but I think it covers the basics needed to start experimenting with inline templates.

We chatted a bit about fragments, but I want to keep that separate from this PR since that solve a separate problem and might require a lot more back-and-forth discussion. I'll tackle that in a follow-up PR assuming this PR gets a 👍.

Happy to hear thoughts and feedback on this approach and discuss any specific points folks make.

lib/view_component/inline_template.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+module ViewComponent # :nodoc:
+  module InlineTemplate
+    extend ActiveSupport::Concern
+    Template = Struct.new(:source, :language, :path, :lineno)
+
+    class_methods do
+      def template!(template)
+        caller = caller_locations(1..1)[0]
```
BlakeWilliams, line 10: This makes a small assumption, which is that all `template!` calls will be formatted like:

```ruby
template <<~ERB
ERB
```

and not

```ruby
template!(
  <<~ERB
  ERB
)
```

The newline is what's problematic here, since we assume that the template will start on the same line as the method call (`template!`). I think that's fine, but if it becomes a problem we can provide more "fine grained" options like allowing an offset to be passed

test/sandbox/test/inline_erb_test.rb
Spone: It would be awesome to be able to figure out the templating language based on the HEREDOC delimiter, but it does not seem possible, as a HEREDOC block is "just a string". I did not find a way to introspect the delimiter.

joelhawksley: @BlakeWilliams makes sense. Maybe we could set the language to be ERB by default?

BlakeWilliams: Sounds good, ERB is the default we fall back to if it's not already set: https://github.com/ViewComponent/view_component/pull/1556/files#diff-8678537e91c0b561c831072ad0aad8a3ed1208cf537591e96475a23959920b5dR27-R30

lib/view_component/inline_template.rb
fsateler: It also cannot be:

```ruby
template! "<div>Alert!"
tmplt = "asdf"
template! tmplt
def self.generate_template
  "a_template"
end
template! generate_template
template! File.read("some_file.html.erb")
```

BlakeWilliams:
> * if you plan on adding some docs already, or wait until it's no longer experimental

I was thinking about passing on documentation for the first round, then fill it in once it's been tried out for real in a few spots.

> * what's the rationale behind using bang methods for `template!` and `template_engine!`? preventing collision with existing user methods? just shouting at the component? 😆

Honestly, no real solid reason. I thought it would be nice to signal "this is happening only once!" but happy to change it if folks feel strongly.

> * if we should warn/raise when calling `template!` or `template_engine!` multiple times in a component

Good call. I think that will be easy to implement for `template!`, I can add that.

lib/view_component/inline_template.rb
joelhawksley: I'm not a fan of imposing this constraint. We should accept all valid Ruby.

joelhawksley: What if we were to wrap template errors with a message along the lines of "Error on line 3 of MyComponent inline template"

test/sandbox/test/inline_template_test.rb
```diff
@@ -0,0 +1,113 @@
...
+      @name = name
+    end
+  end
+
+  class InlineRaiseErbComponent < ViewComponent::Base
+    include ViewComponent::InlineTemplate
+
+    attr_reader :name
+
+    template! <<~ERB
```
joelhawksley, line 25: Could we register methods for each valid handler?

```suggestion
   erb_template! <<~ERB
```

BlakeWilliams: I'm hacking on this idea, but I have some concerns that some environments may run into load order problems. To do this in an efficient way, we'll likely need to do something like:

```ruby
module InlineTemplates
  class_methods do
    ActionView::Template.template_handler_extensions.each do |ext|
      define_method("#{ext}_template!") do 
      end
    end
end
```

In our office hours we discussed that this would be similar to the compiler, but I think it's slightly different because this has to happen when the *module* is first evaluated/included because these are class methods (or use the slower `method_missing` approach). When we compile, we explicitly call `compile` on all descendants after `ActiveSupport` is loaded, which gives us some control over what's loaded at the time vs it being when it's first required/evaluated.

I don't have time to setup an environment to try confirming that, but wanted to throw that out there as a potential gotcha.

BlakeWilliams: Okay, I worked around that with some clever use of `method_missing`. It feels more Ruby/Rails like since it's a bit more magical. 😄

test/sandbox/test/inline_template_test.rb
```diff
@@ -0,0 +1,114 @@
+# frozen_string_literal: true
```
joelhawksley, line 1: Mind adding a test showing these in use with slots?

BlakeWilliams: Happy to add more tests, but would you mind clarifying what you're looking for? Do you want the template to use slots, or did you want to see another component use an "inline template component" as a slot?

joelhawksley: I was thinking of the latter, but the former would be useful too. Just to make sure things play together happily.

BlakeWilliams: Added a test in https://github.com/ViewComponent/view_component/pull/1556/commits/ab0c15b5b779db487a0a4e417260324bfca3de4f

lib/view_component/inline_template.rb
BlakeWilliams: I have some ideas here on how we might be able to improve the error messages, but I'd prefer they weren't blocking. Any opposition to merging this as-is and exploring better error strategies later? This is marked as experimental, so a little bit of churn seems fine here imo.

joelhawksley (commented):
LGTM, mind adding docs? Happy to edit heavily if you can get something rough together.

joelhawksley (approved):

cheshire137:
Thanks for this! I think this will be really nice for tiny little components.

Bestra:
Oooh, I really like this! Thanks for keeping DX improvements like this coming!

PR #1560, opened 2022-10-24
`.with_collection` supports Components with splatted KeywordArgs

zspencer:

I'd like to be lazy and not define `initialize` on every view component when they inherit from a parent class that accepts keyword arguments in order for the component to work with `with_collection.

For example:

```ruby
  def initialize(**kwargs)
    super
    kwargs.each do |k, v|
      send(:"#{k}=", v)
    end
  end
end

class ProductComponent < ApplicationComponent
  attr_accessor :product
end

class OrderComponent < ApplicationComponent
  attr_accessor :order
end

```

However, ViewComponent doesn't recognize that a keyword args is an implicit `def initialize(product:)` or `def initialize(order:)`; which causes it to raise an `ArgumentError` for an unaccepted parameter.

I made it so Components with Keyword Args in their initialize pass the initialze parameter name check. This seemed like a small enough change?

### Anything you want to highlight for special attention from reviewers?=

- [ ] Are there side-effects of this I'm not aware of?
- [ ] Does this create subtle design pressure that encourages cruft?


zspencer:
Woops, I must have forgotten to include the empty `call` method definition when I pulled out the class. One moment...

docs/CHANGELOG.md
```diff
@@ -10,6 +10,9 @@ nav_order: 5
 
 ## main
 
+* `Component.with_collection` supports Components whose initialize takes Keyword Arguments
```
```suggestion
* `Component.with_collection` supports components that accept keyword arguments.

```

test/sandbox/test/collection_test.rb
```diff
@@ -17,5 +26,10 @@ def test_collection_has_a_size
     def test_is_a_collection_of_view_components
       assert_equal [ProductComponent], @collection.map(&:class).uniq
     end
+
+    def test_supports_components_with_keyword_args
+      collection = ComponentWithKeywordArgs.with_collection(@products)
+      render_inline(collection)
```
joelhawksley, line 32:
```suggestion
      render_inline(ComponentWithKeywordArgs.with_collection(@products))
```

Can you add an assertion that something is rendered correctly?

zspencer: Done!

docs/CHANGELOG.md
zspencer: Much tighter, thank you!

joelhawksley:
👋🏻 it looks like you do have a legit failure in CI ❤️

zspencer:
Woops! I ran them locally but must have misread the output. That's what I get for shoveling all my open-source stuff to end-of-business!

zspencer:
yes-chef

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* `Component.with_collection` supports components that accept keyword arguments.
```
jcoyne, line 13: Can this be changed to "accept splatted keyword arguments"?  I believe defined keyword args are not a problem.

lib/view_component/base.rb
```diff
@@ -591,7 +591,7 @@ def validate_collection_parameter!(validate_default: false)
         parameter = validate_default ? collection_parameter : provided_collection_parameter
 
         return unless parameter
-        return if initialize_parameter_names.include?(parameter)
+        return if initialize_parameter_names.include?(parameter) || initialize_parameter_names.include?(:kwargs)
```
jcoyne, line 594: Does this imply that the splatted argument has to have the name `kwargs`?   That is, would this work with `def initialize(**other_args)`?

zspencer: An excellent question to which I do not yet have a conclusive answer. I'll update the tests and report back.
i have no idea what im doing

docs/CHANGELOG.md
zspencer: That's much clearer!

lib/view_component/base.rb
zspencer: Excellent catch! I've updated to check for `:kwrest` in the flattened initialize parameters!

zspencer:
Once more into the fray! The lint failures appear to be unrelated to my changes; so I made a patch to resolve those as well: https://github.com/ViewComponent/view_component/pull/1576

lib/view_component/base.rb
jcoyne: Thanks for making that change.

zspencer:
OK look slike the 3.1 build is not happy will poke at it 🔜

zspencer:
OK, looks like we are back to green on CI; and I have resolved the test failure for ruby 3.1.

lib/view_component/base.rb
```diff
@@ -645,6 +645,12 @@ def iteration_argument_present?
         initialize_parameter_names.include?(collection_iteration_parameter)
       end
 
+      # @private
+      def splatted_keyword_argument_present?
```
joelhawksley, line 649: Are you sure this can't be `private`?

zspencer: Oh definitely. Let me move that.

zspencer: Fin.

joelhawksley:
@zspencer can you fix the merge conflicts? ❤️

zspencer:
@joelhawksley - I have several times at this point... They'll keep coming up when y'all merge other stuff, so for me to do it feels like a dog chasing its tail a bit.

zspencer:
@joelhawksley - You'll need to approve it in order for auto-merge to work, I think.

joelhawksley (approved):

joelhawksley:
That'll do it! Thanks @zspencer ❤️

zspencer:
autobots-roll-out

lib/view_component/base.rb
```diff
@@ -636,6 +636,11 @@ def iteration_argument_present?
 
       private
 
+      def splatted_keyword_argument_present?
+        initialize_parameters.flatten.include?(:keyrest) &&
+          !initialize_parameters.include?([:keyrest, :**]) # Un-named splatted keyword args don't count!
```
daniel-rikowski, line 641: Is there a specific reason for not accepting unnamed splatted arguments? This limitation creates some problems with initializers, which forward their arguments to helpers, e.g. dry-initialize, which uses `def initialize(...); __dry_initializer_initialize__(...); end` (`...` = actual Ruby code) or any kind of custom helper, like `def initialize(**); do_init(**); end;` (Sorry for "gravedigging", I'll open a separate issue, if that is more appropriate...)

zspencer: I can't remember my exact reasoning, but it was more a limitation-in-my-capacity and not necessarily a limitation-by-design.

reeganviljoen: @daniel-rikowski @zspencer can we please open up a new issue where this can be discussed further

joelhawksley: 👋 I'd also be interested in a PR with a failing test for the use case you want 😄

daniel-rikowski: See #2020

@zspencer It seems this was because of a Ruby 3.1 bug 😉

PR #1562, opened 2022-11-02
Remove consistent render change for 3.0

BlakeWilliams:
This removes the configuration option that allows ViewComponent users to
opt-in to consistent rendering which was planned to be the default for
3.0.

While there are some gaps in the existing API, it's a significant amount
of work to make this change for established ViewComponent users. While
not planned, this should be replaced by some other solution, such as
migrating the `content` of a component to become a slot.


adrianthedev:
Thanks for reconsidering this! I appreciate the thought. Keep up the excellent work!

PR #1570, opened 2022-11-05
Support variants with dashes in their names

javierm:

We're implementing multitenancy in our application and are allowing the possibility to use different variants for different tenants. We use the subdomain of the current request to get the current tenant.

While AFAIK it isn't documented anywhere and there aren't any test cases for it, Rails correctly selects and renders templates when the variant contains characters like dashes, dots or spaces. So adding a tenant with the subdomain `my-subdomain` and creating a view file ending with `.html+my-subdomain.erb` works fine.

However, ViewComponent defines methods dynamically with `def call_#{variant}`, which results in invalid Ruby code if the variant contains a dash.

While we could replace dashes with underscores, we'd prefer to keep the dashes and replace dots with underscores (for nested subdomains) instead.

Replacing `def` with `define_method` and using `send` allows using characters which can't be used when defining methods with `def`. As mentioned in, for instance, the [Rails test documentation](https://guides.rubyonrails.org/v7.0/testing.html#rails-meets-minitest):

> in Ruby technically any string may be a method name. This may require use of define_method and send calls to function properly, but formally there's little restriction on the name.

So, with these changes, not only dashes but also spaces and pretty much all characters except dots are allowed in variant names, but I'm only adding a test for dashes because, alongside the underscore, they're arguably the most commonly used word-separator in computing.

I'm using `send`, but I don't know whether `public_send` would be better in this case :thinking:. Any comments?

By the way, is this case worth documenting? As mentioned earlier, I don't see it anywhere in the Rails documentation.

And please let me know whether more tests should be added and where! Any other comments are also welcome :wink:.


javierm:
The failing CI check doesn't seem related to this pull request. Let me know if that's not the case! :pray:

lib/view_component/compiler.rb
```diff
@@ -92,7 +92,7 @@ def #{method_name}
 
     def define_render_template_for
       variant_elsifs = variants.compact.uniq.map do |variant|
-        "elsif variant.to_sym == :#{variant}\n    #{call_method_name(variant)}"
+        "elsif variant.to_sym == :'#{variant}'\n    send(:'#{call_method_name(variant)}')"
```
BlakeWilliams, line 95: Can we avoid `send` here somehow (it's slower than "default" method calls)? Maybe we could call `underscore` on `method_name` to make it a valid identifier?

javierm: @BlakeWilliams Thanks for your comment! I've updated the pull request.

lib/view_component/compiler.rb
```diff
@@ -234,7 +234,7 @@ def compiled_template(file_path)
 
     def call_method_name(variant)
       if variant.present? && variants.include?(variant)
-        "call_#{variant}"
+        "call_#{variant.to_s.underscore}"
```
BlakeWilliams, line 237: @camertron @joelhawksley any concerns against using underscore here (I suggested it, fwiw). It lets us avoid the performance penalty of using `send`, the only downside is that it increases the chance for conflicts between variants (although I'd say it's really unlikely).

BlakeWilliams (approved):

lib/view_component/compiler.rb
joelhawksley: @BlakeWilliams my gut says we should be collision-proof or detect collisions at a minimum.

camertron: Yeah, detecting collisions feels important, but otherwise I'm ok with `underscore`ing.

javierm: Thanks for the suggestion!

@BlakeWilliams I've updated the pull request so it detects collision. More suggestions are welcome, particularly regarding the error message.

By the way, just to confirm, I understand checking for collisions would have a smaller performance penalty than using `send` because collisions would be detected when pre-compling the templates while `send` would be executed at runtime. Is that correct?

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Support variants with dashes in their names
```
```suggestion
* Support variants with dashes in their names.
```

lib/view_component/compiler.rb
```diff
@@ -163,6 +163,18 @@ def template_errors
               "There can only be a template file or inline render method per variant."
           end
 
+          uniq_variants = variants.compact.uniq
```
joelhawksley, line 166: I happened to be poking around Bundler the other day and saw these rules:

```
- Any - characters in a host name are mapped to a triple dash (___) in the corresponding environment variable.
- Any . characters in a host name are mapped to a double dash (__) in the corresponding environment variable.
```

(Above https://bundler.io/man/bundle-config.1.html#CONFIGURE-BUNDLER-DIRECTORIES).

What do you think about using this naming approach to avoid collisions?

javierm: @joelhawksley I think that's a great idea! I'll update the pull request.

javierm: @joelhawksley I've updated the pull request so method names replace dashes with double underscores. Dots aren't replaced with triple underscores , though, because currently using dots wouldn't work because the code getting variants and handlers from filenames assumes there are only two dots in the filename. I intend to try to add support for dots in a few weeks, although I can't confirm for sure that I'll be able to do so.

In the meantime, more comments regarding this pull request are welcome :wink:.

joelhawksley (approved):

joelhawksley:
Fantastic! Thanks for your work on this @javierm ❤️

Issue #1565, opened 2022-11-03
TypeError: no implicit conversion of nil into String


krsyoung:
Thanks for all of your work on 2.75.0! Our gem seems to be less appreciative though...

We are hitting an issue with 2.75.0 causing `TypeError: no implicit conversion of nil into String` in our [matey gem](https://github.com/harled/matey). The problem does not occur in 2.74.1.

For some reason in `base.rb` `ViewComponent::Base.config.view_component_path` is nil which causes `Regex.quote` to throw an exception. Upon a little further inspection, the whole config object is empty:

```
(ruby) ViewComponent::Base.config
{}
```

If I add something like the following above `base.rb#509` it seems to work :-) 

```
ViewComponent::Base.config = ViewComponent::Config.new
```

I'm still digging around but wanted to post in case others are experiencing the same issue.

### Steps to reproduce
1. clone https://github.com/harled/matey
2. Switch to 2.75.0 in matey.gemspec
3. bin/setup 
4. rake spec  

### Expected behavior
Expect the test cases to pass as with 2.74.1

### Actual behavior
Test cases are not happy!

```
TypeError: no implicit conversion of nil into String
/usr/local/rvm/gems/default/gems/view_component-2.75.0/lib/view_component/base.rb:509:in `quote'
/usr/local/rvm/gems/default/gems/view_component-2.75.0/lib/view_component/base.rb:509:in `inherited'
/workspaces/matey/app/components/application_component.rb:4:in `<top (required)>'
/workspaces/matey/lib/matey.rb:5:in `require_relative'
/workspaces/matey/lib/matey.rb:5:in `<top (required)>'
/workspaces/matey/spec/sample/config/application.rb:15:in `<top (required)>'
```

### System configuration
**Rails version**: 7.0.4

**Ruby version**: ruby 3.1.2p20 

**Gem version**: 2.75.0

boardfish:
I think the change to `lib/view_component/engine.rb` and `base.rb` [between these versions](https://github.com/ViewComponent/view_component/compare/v2.74.1...v2.75.0#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9e) might be responsible. This was introduced in #1528.

I'm surprised that the tests that maintain compatibility between `ViewComponent::Base.config` and `Rails.application.config.view_component` allowed this - it seems to suggest those will need to be updated so that they fail for this instance.

[...]

PR #1571, opened 2022-11-06
Fix empty VC::Base.config during initialization

jdelStrother:

If the Railtie initialization hasn't yet run, and a third-party gem (eg [Matey](https://github.com/harled/matey/)) tries to subclass ViewComponent::Base, it would fail in Base.inherited due to config.view_component_path being nil:

```
TypeError: no implicit conversion of Hash into String
/Users/jon/Developer/vendor/view_component/lib/view_component/base.rb:510:in `quote'
/Users/jon/Developer/vendor/view_component/lib/view_component/base.rb:510:in `inherited'
/Users/jon/Developer/vendor/matey/app/components/application_component.rb:4:in `<top (required)>'
/Users/jon/Developer/vendor/matey/lib/matey.rb:5:in `require_relative'
/Users/jon/Developer/vendor/matey/lib/matey.rb:5:in `<top (required)>'
/Users/jon/Developer/vendor/matey/spec/sample/config/application.rb:15:in `<top (required)>'
/Users/jon/Developer/vendor/matey/spec/sample/Rakefile:4:in `require_relative'
/Users/jon/Developer/vendor/matey/spec/sample/Rakefile:4:in `<top (required)>'
/Users/jon/.gem/ruby/3.1.0/gems/railties-7.0.4/lib/rails/tasks/engine.rake:5:in `load'
/Users/jon/.gem/ruby/3.1.0/gems/railties-7.0.4/lib/rails/tasks/engine.rake:5:in `block (2 levels) in <top (required)>'
/Users/jon/.gem/ruby/3.1.0/gems/railties-7.0.4/lib/rails/tasks/engine.rake:4:in `block in <top (required)>'
/Users/jon/.gem/ruby/3.1.0/gems/railties-7.0.4/lib/rails/tasks/engine.rake:85:in `<top (required)>'
/Users/jon/Developer/vendor/matey/Rakefile:13:in `load'
/Users/jon/Developer/vendor/matey/Rakefile:13:in `<top (required)>'
```

This broke in 2.75 due to changes from https://github.com/ViewComponent/view_component/pull/1528

This ensures that if VC::Base.config is accessed during Rails initialization, it contains sane defaults. I think that's probably going to be fine for most cases and should fix the problems that gem authors are seeing in #1565.

It troubles me a bit that we're potentially initializing VC::Base.config once when it's first accessed by a third-party gem, who might try to set their own custom config on it, and that's going to be overwritten once initialization completes here - https://github.com/ViewComponent/view_component/blob/main/lib/view_component/engine.rb#L39 .

Possibly VC::Base.inherited could be tweaked to avoid accessing configuration by deferring the `virtual_path` calculation.

Another option would be to push the responsibility onto gem authors, who ideally wouldn't be forcing VC::Base to load as soon as their gem is required, since that then loads ActionView::Base, which is what we're trying to avoid in the first place in https://github.com/ViewComponent/view_component/pull/1528

Ugh, tests for this are a pain. In isolation, a one-off test like this might have caught the previously-empty configuration object:

```ruby
require "minitest/autorun"
require "rails"
require "action_view"
require "view_component"

class IsolatedTest < Minitest::Test
  def test_config_has_sane_defaults
    assert_equal "app/components", ViewComponent::Base.config.view_component_path
  end
end
```

but if that's loaded that into the same test process as everything else in test/sandbox, ViewComponent::Engine would already have run the initializers so the test wouldn't be testing what we're interested in.  Think you'd either have to change all the tests in test/sandbox to use, eg, ActiveSupport::TestCase::Isolation, or run two separate test processes - I'm not sure it's worth it. WDYT?

Would welcome input from @krsyoung / @jcoyne / @peteryates / anyone else affected by #1565


peteryates:
This resolves the issue for me, thank you @jdelStrother.

For what it's worth, your original fix worked fine for us in [our actual gem](https://github.com/DFE-Digital/govuk-components), it only caused problems when building the [statically generated documentation website](https://govuk-components.netlify.app/) - which we're definitely building in an _unintended_ fashion.

jcoyne:
This resolves this issue for me too.

BlakeWilliams:
> It troubles me a bit that we're potentially initializing VC::Base.config once when it's first accessed by a third-party gem, who might try to set their own custom config on it, and that's going to be overwritten once initialization completes here - https://github.com/ViewComponent/view_component/blob/main/lib/view_component/engine.rb#L39 .

I agree with this statement. I wonder if we need to change our approach to configuration to be component focused, encouraging folks to set configuration values on `ApplicationComponent` or some other root component when possible instead of relying on globals that may between gem/application.

BlakeWilliams:
Given the concerns about conflicts with configuration values, I'm thinking we might want to revert https://github.com/ViewComponent/view_component/pull/1528 to get back to a known good state at re-evaluate our configuration strategy. Maybe we could find some way to scope the configuration to a group of components instead of globally like mentioned in my comment?

/cc @boardfish since you have expertise here and might have thoughts on how best to proceed.

boardfish:
@BlakeWilliams That sounds good to me. I think there may be some questions worth discussing with the Matey crew around what they use the config for. Hopefully we can get some answers to whether the config changes made by gems like Matey should have precedence over, exist independently of, or override app settings. There may also be the chance that app-level config and gem-level config can be two different surfaces.

My original intent with making the config object shared between the Rails app and `ViewComponent::Base` was so that it behaved predictably. The central problem #1528 was trying to solve was that in order to mirror the config back to `ViewComponent::Base`, it'd need to load that class, and ActionView with it. Perhaps to solve both issues in the short term, there needs to be a "join table" between them – an object that's initialized that both `Rails.application.config.view_component` and `ViewComponent::Base` refer to.

In the long term, however, your idea sounds pretty solid 👍 I just wonder what `Rails.application.config.view_component` would refer to, if anything at all 🤔 Maybe the configuration *would* take place against some `ApplicationComponent`, as you say.

Laykou:
Does it make sense while these pvc tests would be failing due to this change?

jdelStrother:
> Does it make sense while these pvc tests would be failing due to this change?

Was that directed at me? I'm not sure what pvc is

Spone:
> Does it make sense while these pvc tests would be failing due to this change?

I ran the tests again and they passed. `pvc` tests can be pretty flaky :see_no_evil:

Laykou:
Can this be merged and released as new version, please?

boardfish:
@Laykou A decision remains to be made on this PR – as you might've seen from [this comment](https://github.com/ViewComponent/view_component/pull/1571#issuecomment-1317554595), we're contemplating reverting #1528 and re-evaluating our approach to configuration on ViewComponent. Thanks for your patience in the meantime.

@BlakeWilliams, it'd seem @Spone and I agree with your thoughts on where we should go from here. Feel free to continue ahead and revert that PR – I'll try and take some time in the new year to update our approach to config to reflect that new goal.

joelhawksley:
@boardfish how are you feeling about this PR? Are you still interested in revisiting our config approach?

boardfish:
@joelhawksley I'd still like to, but if it's something that might benefit from being done sooner, then I'm happy for someone else to pick it up – I've been a bit busier as of late! I'll be about for review, though 👍

joelhawksley:
Closing as stale ❤️

PR #1572, opened 2022-11-07
Adding a `from` argument to render_preview to specify the preview class rather than inferring it

jonrohan:

Currently the `render_preview` test helper infers the preview class by the name of the test calling it. In my test cases I needed to specify the class outside the context of the specific preview test. [Real world application](https://github.com/primer/view_components/pull/1574)

I extracted out the current code that's used to infer the test, and moved to a private method `preview_klass`. Then I added a new argument `klass` to `render_preview` with a fallback of inferring the class like before.

### DIFF
--- a/docs/guide/previews.md
+++ b/docs/guide/previews.md
@@ -76,7 +76,7 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
-By default, the name of the preview to be rendered is inferred from the name of the current test file. Use `from` to explicitly set the preview to be rendered:
+By default, the preview class is inferred from the name of the current test file. Use `from` to set it explicitly:
 
 ```ruby
 class ExampleTest < ViewComponent::TestCase
--- a/docs/guide/previews.md
+++ b/docs/guide/previews.md
@@ -76,12 +76,12 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
-By default, the name of the preview to be rendered is inferred from the name of the current test file. Use `klass` to explicitly set the preview to be rendered:
+By default, the name of the preview to be rendered is inferred from the name of the current test file. Use `from` to explicitly set the preview to be rendered:
 
 ```ruby
 class ExampleTest < ViewComponent::TestCase
   def test_render_preview
-    render_preview(:with_default_title, klass: ExampleComponentPreview)
+    render_preview(:with_default_title, from: ExampleComponentPreview)
 
     assert_text("Hello, world!")
   end
--- a/docs/guide/previews.md
+++ b/docs/guide/previews.md
@@ -76,6 +76,18 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+By default, the name of the preview to be rendered is inferred from the name of the current test file. Use `klass` to explicitly set the preview to be rendered:
+
+```ruby
+class ExampleTest < ViewComponent::TestCase
+  def test_render_preview
+    render_preview(:with_default_title, klass: ExampleComponentPreview)
+
+    assert_text("Hello, world!")
+  end
+end
+```
+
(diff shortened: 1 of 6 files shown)


docs/guide/previews.md
```diff
@@ -76,6 +76,18 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+By default, the name of the preview to be rendered is inferred from the name of the current test file. Use `klass` to explicitly set the preview to be rendered:
+
+```ruby
+class ExampleTest < ViewComponent::TestCase
+  def test_render_preview
+    render_preview(:with_default_title, klass: ExampleComponentPreview)
```
BlakeWilliams, line 84:
```suggestion
    render_preview(:with_default_title, preview_class: ExampleComponentPreview)
```

Minor API suggestion, thoughts on naming the argument `preview_class` since we should be able to use `class` with a prefix?

joelhawksley: @BlakeWilliams that's how @jonrohan implemented it originally, but I've been of the general belief that ruby param names should not be duplicative of the method's name. Since the method name already includes `preview`, I think `klass` is more idiomatic here. 

But I don't feel too strongly either way. I'd be happy to go with `preview_class` if you feel strongly about it ❤️

BlakeWilliams: > Since the method name already includes `preview`, I think `klass` is more idiomatic here.

Yeah, I generally agree about not "stuttering" as Go would call it (preview preview). I generally try to avoid `klass` though, since I've never found it to be descriptive enough and the spelling throws me off.

Is there an alternative name that would give us the best of both worlds? `component_class` maybe?

joelhawksley: Hmm. It isn't a component class though, it's a preview class.

Spone: Maybe something simpler such as `from` or `in`? This way it reads like a sentence: Render preview_method from/in PreviewClass.

joelhawksley: @Spone lovely. I'd 👍🏻 `from`.

mxriverlynn: do we have sorbet or any other strong typing in this gem? the naming is useful, but more specifically stating that the param must be a `ViewComponent::Preview` subclass would be 100x more useful

i also like the way this looks with `from`

```ruby
render_preview(:default, from: Primer::Beta::MyComponentPreview)
```

this reads really well, IMO 👍🏻

lib/view_component/test_helpers.rb
```diff
@@ -81,22 +81,10 @@ def render_inline(component, **args, &block)
...
-          "#{described_class}Preview"
-        else
-          self.class.name.gsub("Test", "Preview")
-        end
-        preview_klass = preview_klass.constantize
-      rescue NameError
-        raise NameError, "`render_preview` expected to find #{preview_klass}, but it does not exist."
-      end
-
+    def render_preview(name, klass: preview_klass, params: {})
```
mxriverlynn, line 87: does this need to be updated, based on the above conversation in the `previews.md` file?

joelhawksley (approved):

jonrohan:
Is there something I need to do to fix those `pvc` tests?

Spone (approved):

docs/guide/previews.md
```diff
@@ -76,6 +76,18 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+By default, the name of the preview to be rendered is inferred from the name of the current test file. Use `from` to explicitly set the preview to be rendered:
```
joelhawksley, line 79:
```suggestion
By default, the preview class is inferred from the name of the current test file. Use `from` to set it explicitly:
```

joelhawksley (approved):

PR #1576, opened 2022-11-11
Apply the new standardrb rules

zspencer:

CI is failing on lint for everyone, regardless of changes. This gets through that in a small, temporary way.

###

While working on [supporting splatted keyword
args](https://github.com/ViewComponent/view_component/pull/1560) I discovered that the linter was failing me on changes I didn't make.

When I ran standardrb locally, it was quite content, so I was a bit confused. Then I read [.github/workflows/lint.yml] and realized that view_component runs a `bundle update` prior to running standard; which bumps the standardrb version and the corresponding rubocop set.

This was introduced by @camertron in
[e7eec9869fc13e401401c687573a98f1b0ed57a8](https://github.com/ViewComponent/view_component/commit/e7eec9869fc13e401401c687573a98f1b0ed57a8) 7+ months ago to fix an error in CI.

It's unclear to me why `bundle install` was not working, so I decided that rather than swapping that back I would trust @camertron's instincts and roll-forward by updating the `Gemfile.lock` so folks' local matches CI (for now), and applying the changes it recommended.

I am not including a change to the Changelog because this seems internal and unnecessary to communicate to package consumers.

-[ ] I'd be down to just try `bundle install` rather than `bundle update` if that seems like a good idea to other folks with more context.


camertron:
> It's unclear to me why `bundle install` was not working

I changed it to `bundle update` exactly because of situations like this. Since we have a Gemfile.lock checked into version control, we miss testing against new dependency releases. Running `bundle update` in CI ensures we test against newer releases. The other option would be to stop version tracking Gemfile.lock, but that's considered good practice these days.

> I am not including a change to the Changelog because this seems internal and unnecessary to communicate to package consumers.

I get that, but we'd like every change to have an accompanying changelog entry.

camertron:
This PR should fix the PVC tests: https://github.com/ViewComponent/view_component/pull/1579

BlakeWilliams:
> Since we have a Gemfile.lock checked into version control, we miss testing against new dependency releases

If possible, we should omit tooling like standardrb from being updated like that, to prevent the case of it introducing "breaking changes".

PR #1579, opened 2022-11-12
Use npm instead of yarn to install PVC dependencies

camertron:

Our PVC test runs have been [failing recently](https://github.com/ViewComponent/view_component/actions/runs/3441193868/jobs/5740477527) in CI. This PR should fix things.

PVC [migrated](https://github.com/primer/view_components/pull/1388) away from Yarn in favor of npm a few months ago, but view_component's CI scripts were never updated. Don't ask me why, but using npm fixes the test issues.

The linting issues should be fixed by [this PR](https://github.com/ViewComponent/view_component/pull/1576).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,18 +79,18 @@ jobs:
         ruby-version: 2.7
     - uses: actions/setup-node@v2
       with:
-        node-version: '14'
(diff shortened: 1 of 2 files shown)


BlakeWilliams (approved):
Good find!

PR #1580, opened 2022-11-12
Fix YARD in docs:build 

Spone:

I noticed that the latest updates in the YARD comments (such as #1539) did not appear on the docs [API page](https://viewcomponent.org/api.html).

I tracked down the issue to an error when running `bundle exec rake docs:build`:

```
Building YARD documentation.
Files:          46
Modules:        43 (   35 undocumented)
Classes:        28 (   25 undocumented)
Constants:      15 (   14 undocumented)
Attributes:     26 (    4 undocumented)
Methods:       153 (   76 undocumented)
 41.89% documented
Converting YARD documentation to Markdown files.
rake aborted!
/home/hans/projects/vc/view_component/lib/view_component/base.rb:509:in `quote'
/home/hans/projects/vc/view_component/lib/view_component/base.rb:509:in `inherited'
/home/hans/projects/vc/view_component/lib/view_component/docs_builder_component.rb:4:in `<module:ViewComponent>'
/home/hans/projects/vc/view_component/lib/view_component/docs_builder_component.rb:3:in `<top (required)>'
/home/hans/projects/vc/view_component/Rakefile:86:in `require'
/home/hans/projects/vc/view_component/Rakefile:86:in `block (2 levels) in <top (required)>'
/home/hans/.asdf/installs/ruby/3.0.4/bin/bundle:23:in `load'
/home/hans/.asdf/installs/ruby/3.0.4/bin/bundle:23:in `<main>'
Tasks: TOP => docs:build
(See full trace by running task with --trace)
```

I chose to define `ViewComponent::Base.config.view_component_path` within the rake task, before requiring the docs builder component. I'm not sure if it's something that ought to be fixed in `Base`:

https://github.com/ViewComponent/view_component/blob/main/lib/view_component/base.rb#L509

Also, I think we may need to improve comments for `.config` and `.config=` class methods:

image


docs/api.md
```diff
@@ -10,6 +10,23 @@ nav_order: 3
...
+### .config=(value)
+
+Sets the attribute config
+
+### .sidecar_files(extensions)
+
+Find sidecar files for the given extensions.
+
+The provided array of extensions is expected to contain
+strings starting without the "dot", example: `["erb", "haml"]`.
```
boardfish, line 26: This could be fixed in [this file](https://github.com/ViewComponent/view_component/blob/22dbe8f648b44cc73d9663082a3e5024345b28ee/lib/view_component/base.rb#L416) – I don't think the word "dot" needs to be in quotes.

docs/api.md
```diff
@@ -10,6 +10,23 @@ nav_order: 3
 
 ## Class methods
 
+### .config
+
+
+
+### .config=(value)
+
+Sets the attribute config
```
boardfish, line 19: I agree that this could be changed. I figure it comes from `attr_accessor` as a default.

"Replaces the entire config. You shouldn't need to use this directly unless you're building a `ViewComponent::Config` elsewhere."

docs/api.md
```diff
@@ -10,6 +10,23 @@ nav_order: 3
 
 ## Class methods
 
+### .config
+
+
```
boardfish, line 15: "Returns the current config. This should be a `ViewComponent::Config` object."

Rakefile
```diff
@@ -83,6 +83,7 @@ namespace :docs do
     require "rails"
     require "action_controller"
     require "view_component"
+    ViewComponent::Base.config.view_component_path = "view_component"
```
Spone, line 86: @boardfish what do you think about this fix?

boardfish: It's isolated to this file, and this is just a build task, so I think it's alright. I do wonder if it's symptomatic of something else that's wrong with ViewComponent itself, though. It might be worth thinking about what we use `view_component_path` for a little more widely. I reckon there's probably a justifiable reason for it, but I wonder if we could infer necessary paths based on how folks use ViewComponent and leave the rest to Zeitwerk.

Spone: Yes, that's precisely what I was thinking. I think we should revisit the purpose and uses of this setting. But that's for another PR.

boardfish (commented):
The last Vale failure listed here probably requires a change to the docs generator to put the section title between backticks. Once that's done and the docs are regenerated, this should be good to go!

Spone:
> The last Vale failure listed here probably requires a change to the docs generator to put the section title between backticks. Once that's done and the docs are regenerated, this should be good to go!

Which error? I thought I fixed it already :thinking:

boardfish (approved):

PR #1589, opened 2022-11-24
Add documentation to test component with behaviors and POROs as input

nirebu:

Currently in the docs [there's a handy way to test component with behaviors](https://github.com/ViewComponent/view_component/blob/main/docs/guide/testing.md#previews-as-test-cases) leveraging the previews through system specs. This method works well for components without arguments or taking simple inputs such as strings and numbers.

If we want to test components taking ActiveRecord objects as input, we can pass the IDs in the URL and then find those records in the preview, but this causes tight coupling between the specs and the preview class, forcing developers to tweak multiple files at a time.

However, this method breaks if we want to pass in POROs, or other generic objects in these preview URLs.

Here's an example of how I implemented a rough solution to this problem using RSpec and Capybara:
```rb
class OrderSummaryComponent < ApplicationComponent
  def initialize(order:)
    @order = order
  end
end

class OrderSummaryComponentPreview < ApplicationPreview
  # The default preview is not needed to make this work, we can have a empty class as well
  def default
    render(OrderSummaryComponent.new(order: Order.complete.last))
  end
end

module Helpers
  module ViewComponentSystemSpecHelpers
    def with_spec_preview(&preview_proc)
      preview_class.define_method(:spec, &preview_proc)
    end
  end
end

RSpec.describe OrderSummaryComponent do
  it 'is expandable by clicking on the button' do
    order = create(:order)
    # not saving the record intentionally to prove we display non-persisted data
    order.shipment_total = 24
    with_spec_preview do
      render(described_class.new(order: order))
    end
    visit "/rails/view_components/#{described_class.to_s.underscore}/spec"

    expect(page).to have_text("Order Summary")
    expect(page).not_to have_text("Shipping: $24.00")

    click_button "Order Summary"

    expect(page).to have_text("Shipping: $24.00")
  end
end
```

For now I'm adding this as documentation. I'll try to submit other commits or a different PR to implement such a test helper if the team feels is useful.

### DIFF
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -233,3 +233,42 @@ To use component previews:
 # config/application.rb
 config.view_component.preview_paths << "#{Rails.root}/spec/components/previews"
 ```
+
+If you need to test ViewComponents with behaviors taking POROs or ActiveRecord objects (even with unpersisted attributes) as input:
+
+```ruby
+module Helpers
+  module ViewComponentSystemSpecHelpers
+    def with_spec_preview(&preview_proc)
+      preview_class.define_method(:test_preview, &preview_proc)
+    end
+  end
+end
+
+class Post
+  attr_reader :title, :body
+
+  def initialize(title:, body:)
+    @title = title
+    @body = body
+  end
+end
+
+RSpec.describe PostComponent do
+  it "expands the body when clicking the button" do
+    post = Post.new(title: "My post title", body: "A long form text here")
+    with_spec_preview do
+      render(PostComponent.new(post: post))
+    end
+    visit "/rails/view_components/#{described_class.to_s.underscore}/test_preview"
+
+    expect(page).to have_text("My post title")
+    expect(page).not_to have_text("A long form text here")
+
+    click_button "Show More"
+
+    expect(page).to have_text("My post title")
+    expect(page).to have_text("A long form text here")
+  end
+end
+```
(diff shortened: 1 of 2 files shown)


docs/guide/testing.md
```diff
@@ -233,3 +233,42 @@ To use component previews:
...
+  def initialize(title:, body:)
+    @title = title
+    @body = body
+  end
+end
+
+RSpec.describe PostComponent do
+  it "expands the body when clicking the button" do
+    post = Post.new(title: "My post title", body: "A long form text here")
+    with_spec_preview do
```
joelhawksley, line 260: Hmm, I'm hesitant to recommend this approach, as we're redefining `test_preview` each time this helper is used.

I think I'd prefer the approach @edwinthinks is working on in https://github.com/ViewComponent/view_component/pull/1061#pullrequestreview-1196421738. What do you think?

nirebu: Wow, thanks @joelhawksley for pointing that PR out. I totally missed it when looking for something solving the same problem. The solution proposed there @edwinthinks is indeed more thoughtful and robust, supporting a lot of customisations from the layout and slots standpoints.

> as we're redefining `test_preview` each time this helper is used.

As long as we don't have naming conflicts this shouldn't be an issue. I think it could be made better by defining a method with the example name, visiting the path and then undefining the method to avoid pollution or leaking.

Alas, seeing that work on #1061 is almost at the finish line, I'll keep this solution as a workaround in the application I'm working on until those helpers are released in a new version.

nirebu:
@joelhawksley closing this. Thanks for your feedback!

PR #1597, opened 2022-12-02
Remove deprecated slots setter methods. Use #with_SLOT_NAME instead.

joelhawksley:

I'm removing the deprecated slots setter methods. Next, I'll write a migration guide into the changelog that summarizes all of the deprecations.


Spone:
Any pointers on how to migrate a large project that still uses the former syntax to define slots?

joelhawksley:
@Spone yup, we provided guidance and tooling here: https://viewcomponent.org/CHANGELOG.html#2640

Spone:
@joelhawksley thanks! I totally forgot about this 😊

PR #1598, opened 2022-12-03
Add support for explicit preview_path when generating component

erinnachen:

When multiple preview_paths are configured, the preview generator will abort trying to generate the preview. This adds the option to pass in a path and configure it in the generate options.

There's probably a better name for this option other than `preview_path`, considering the deprecation in v3, but I couldn't come up with one.


lib/rails/generators/preview/component_generator.rb
```diff
@@ -4,15 +4,22 @@ module Preview
   module Generators
     class ComponentGenerator < ::Rails::Generators::NamedBase
       source_root File.expand_path("templates", __dir__)
+      class_option :preview_path, type: :string, desc: "Path for previews, required when multiple preview paths are configured", default: ViewComponent::Base.config.generate.preview_path
```
boardfish, line 7: Given `preview_path` is due for deprecation, maybe `preview_paths.first` might be a good option here.

```suggestion
      class_option :preview_path, type: :string, desc: "Path for previews, required when multiple preview paths are configured", default: ViewComponent::Base.config.generate.preview_paths.first
```

It might also be worth experimenting with using `preview_paths!.first` instead if the behavior feels right. Since this option appears to be for choosing from the existing `preview_paths`, I think it'd make sense in that context.

lib/rails/generators/preview/component_generator.rb
```diff
@@ -4,15 +4,22 @@ module Preview
...
       def create_preview_file
         preview_paths = ViewComponent::Base.config.preview_paths
-        return if preview_paths.count > 1
+        optional_path = options[:preview_path]
+        return if preview_paths.count > 1 && optional_path.blank?
+
+        path_prefix = if optional_path.present?
+          optional_path
+        else
+          preview_paths.one? ? preview_paths.first : "test/components/previews"
```
boardfish, line 20:
```suggestion
          preview_paths.any? ? preview_paths.first : "test/components/previews"
```

lib/rails/generators/preview/component_generator.rb
joelhawksley: @boardfish the deprecated `preview_path` option is for which previews VC looks up, not where previews are generated, right?

erinnachen: I hesitated to used preview_paths.first since its behavior is dependent on how your app is configured.

lib/view_component/config.rb
```diff
@@ -66,6 +66,17 @@ def defaults
...
       #
       #      config.view_component.generate.preview = true
+      #
+      # #### #preview_path
+      #
+      # Path to generate preview:
+      #
+      #      config.view_component.generate.preview_path = "test/components/previews"
+      #
+      # Required when there is more than one path defined in preview_paths.
```
joelhawksley, line 76:
```suggestion
      # Required when there is more than one path defined in config.view_component.preview_paths.
```

joelhawksley (approved):

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add ability to pass in explicit preview_path to preview generator.
```
```suggestion
* Add ability to pass explicit `preview_path` to preview generator.
```

joelhawksley (approved):

PR #1602, opened 2022-12-08
Use deprecation_warning for consistency

janklimo:

A lot of the deprecation messages are currently missing gem name which makes them hard to debug.

Using `deprecation_warning` ensures consistency in including gem name and deprecation horizon in every deprecation message.

### Some examples

Before:

```ruby
ViewComponent::Deprecation.warn(
  "Manually loading the engine is deprecated and will be removed in v3.0.0. " \
  "Remove `require \"view_component/engine\"`."
)
```
```
DEPRECATION WARNING: Manually loading the engine is deprecated and will be removed in v3.0.0. Remove `require "view_component/engine"`. (called from load at /Users/jan/.asdf/installs/ruby/3.0.3/bin/irb:23)
```

After:

```ruby
ViewComponent::Deprecation.deprecation_warning(
  "Manually loading the engine",
  "remove `require \"view_component/engine\"`"
)
```

```
DEPRECATION WARNING: Manually loading the engine is deprecated and will be removed from ViewComponent 3.0.0 (remove `require "view_component/engine"`) (called from load at /Users/jan/.asdf/installs/ruby/3.0.3/bin/irb:23)
```

When the second argument to `deprecation_warning` is a string, it gets included inside parentheses after the message.

Before

```ruby
ViewComponent::Deprecation.warn(
  "`#before_render_check` will be removed in v3.0.0.\n\n" \
  "To fix this issue, use `#before_render` instead."
)
```

```
DEPRECATION WARNING: `#before_render_check` will be removed in v3.0.0.

To fix this issue, use `#before_render` instead. (called from load at /Users/jan/.asdf/installs/ruby/3.0.3/bin/irb:23)
```

After
```ruby
ViewComponent::Deprecation.deprecation_warning(
  "`before_render_check`", :"`before_render`"
)
```

```
DEPRECATION WARNING: `before_render_check` is deprecated and will be removed from ViewComponent 3.0.0 (use `before_render` instead) (called from load at /Users/jan/.asdf/installs/ruby/3.0.3/bin/irb:23)
```

When the second argument to `deprecation_warning` is a symbol, it is used in the `(use ... instead)` template you can see [here](https://github.com/rails/rails/blob/90a272a1de6f13711943139c4292336dbff19c7c/activesupport/lib/active_support/deprecation/reporting.rb#LL116C5-L116C5).

This approach reduces the amount of code in quite a few places as you can see in the last example.


Spone:
Thanks! Can you add a changelog entry please?

Spone (approved):

janklimo:
@Spone updated 👍

PR #1606, opened 2022-12-13
Add documentation to utilize JS testing for RSpec

edwinthinks:

Adding setup instructions for utilizing the [newly added JS testing approach](https://github.com/ViewComponent/view_component/pull/1061) when using RSpec. Without these configurations, using the `with_rendered_component_path` would not work.

This was sparked by a comment left here - https://github.com/ViewComponent/view_component/pull/1061#issuecomment-1347567212

N/A

This is assuming Capybara is being used in RSpec setup, which is common. This may need some updates once we want to include instructions that don't rely on Capybara.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@ nav_order: 5
 
 ## main
 
-* Add missing documentation to use `with_rendered_component_path` with RSpec.
+* Add documentation to use `with_rendered_component_path` with RSpec.
 
     *Edwin Mak*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add missing documentation to use `with_rendered_component_path` with RSpec.
+
+    *Edwin Mak*
+
 ## 2.79.0
 
 * Add ability to pass explicit `preview_path` to preview generator.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@ nav_order: 5
 
 ## main
 
-* Add documentation to use `with_rendered_component_path` with RSpec.
+* Add documentation for using `with_rendered_component_path` with RSpec.
 
     *Edwin Mak*
 
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -197,7 +197,6 @@ RSpec.configure do |config|
   config.include ViewComponent::TestHelpers, type: :component
   config.include ViewComponent::SystemTestHelpers, type: :component
   config.include Capybara::RSpecMatchers, type: :component
-  config.include Capybara::DSL, type: :component
 end
 ```
 
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -195,7 +195,9 @@ require "capybara/rspec"
 
 RSpec.configure do |config|
   config.include ViewComponent::TestHelpers, type: :component
+  config.include ViewComponent::SystemTestHelpers, type: :component
   config.include Capybara::RSpecMatchers, type: :component
+  config.include Capybara::DSL, type: :component
 end
 ```
 


edwinthinks:
@joelhawksley a bit of an oversight, we didn't include documentation for using the `with_rendered_component_path` with RSpec. Let me know if this makes sense, I can confirm this works on my project that uses RSpec

developius:
@edwinthinks thanks for adding the docs! Even with this change, I'm still seeing something weird. Our setups must be a little different. The server must be running, as I'm getting a 404. What would be the most useful information to share with you?

<details><summary>Screenshot of error from headless browser</summary>
Screenshot 2022-12-13 at 3 45 13 PM
</details>

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add missing documentation to use `with_rendered_component_path` with RSpec.
```
```suggestion
* Add documentation to use `with_rendered_component_path` with RSpec.
```

edwinthinks:
@developius thanks for sharing that. Ummm can you confirm that `Rails.env.test?` is true? The route to `_system_test_entrypoint` is only open when its in Rails.env.test? It is likely it is already set that way.


Another option, which seems a bit strange... but let me know is to add change the `type: :system` and add the helpers to the system type:
```
  config.include ViewComponent::TestHelpers, type: :system
  config.include ViewComponent::SystemTestHelpers, type: :system
  config.include Capybara::DSL, type: :system
  config.include Capybara::RSpecMatchers, type: :system
```

You may also need to define (I noticed an issue if you don't define this when going this system test route), which you can add in the spec.
```
  def page
    Capybara.current_session
  end
```

Here is a sample of what I've got working on my local machine:
```
# frozen_string_literal: true

RSpec.describe "SidenavLinkComponent", type: :system do
  def page
    Capybara.current_session
  end

  it 'does render a link if no url is provided' do
    inline = render_inline(SidenavLinkComponent.new(label: 'Test', href: '#', options: {}))

    with_rendered_component_path(inline) do |path|
      visit path
      expect(page).to have_link('Test')
    end

  end
end
```

@developius lemme know if this works. Otherwise be interested to hear from others about why you might be encountering that issue. Might need to be a another issue we can add in the repo?

docs/guide/testing.md
```diff
@@ -195,7 +195,9 @@ require "capybara/rspec"
 
 RSpec.configure do |config|
   config.include ViewComponent::TestHelpers, type: :component
+  config.include ViewComponent::SystemTestHelpers, type: :component
   config.include Capybara::RSpecMatchers, type: :component
+  config.include Capybara::DSL, type: :component
```
edwinthinks, line 200:
```suggestion
```

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add documentation to use `with_rendered_component_path` with RSpec.
```
```suggestion
* Add documentation for using `with_rendered_component_path` with RSpec.
```

joelhawksley (approved):
LGTM, but let's see what @developius says.

developius:
Apologies for the red herring @edwinthinks, we're setting `config.view_component.preview_controller` to a custom controller which has some auth stuff because we use the previews to power Lookbook.
Commenting out that configuration allows everything to work as expected 💯 thanks for the help!

This could probably be moved to another PR/discussion/issue, but sharing anyway...
I'm not sure I would've expected the preview controller to be used for this since it's not using previews - it's only rendering a file. Is that something you're set on keeping? I would definitely vote for decoupling it completely from previews. I suspect others might have similar setups with a custom preview controller, which will make it difficult to use this awesome new approach out of the box. What do you think?

edwinthinks:
@developius no worries! Good find. I think you aren't going to be the only person caught with this issue.

My gut feeling is that it should be a different PR/discussion. And I think we could move it out of previews_controller which I do agree seems to be responsible for something else which we can discuss in that other discussion/issue.

@joelhawksley what do you think?

joelhawksley:
> I think we could move it out of previews_controller which I do agree seems to be responsible for something else which we can discuss in that other discussion/issue.

@edwinthinks I think we should fix the bug with custom preview controllers, yes. Would you be up for working on that?

edwinthinks:
> @edwinthinks I think we should fix the bug with custom preview controllers, yes. Would you be up for working on that?

Hey @joelhawksley yeah I think I can tackle that. Upfront do we know if we are open to moving the system endpoint to a different controller action?

joelhawksley:
> Upfront do we know if we are open to moving the system endpoint to a different controller action?

@edwinthinks yes, I think we should make a separate controller as this feature is not related to previews any more.

edwinthinks:
@joelhawksley sounds good! I think we can merge this in the meantime? The documentation is still neccessary for RSpec users.

Spone (approved):

PR #1557, opened 2022-10-20
chore: Bump activesupport version

DRBragg:
### What are you trying to accomplish?

Re-opening #1553 

Changes in #1550 and #1536 caused `view_component` to no longer be compatible with Rails < 5.2. This PR updates the `.gemspec` file to require `activesupport` version >= 5.2 so folks know about the compatibility issue when installing/updating.  I also updated the compatibility documentation to reflect this.

PR #1609, opened 2022-12-14
chore: Bump activesupport version

DRBragg:

Opening #1557 again to be included in v2.

The original PR should probably have targeted the `main` branch instead of `v3` since `view_component` hasn't been compatible with Rails < 5.2 since version 2.72.0 but you can still install 2.78.0 in an app with no bundler errors. You'll only know there'a an issue when you start rendering components and you sift through the backtraces...


joelhawksley (approved):

DRBragg:
@joelhawksley I just noticed [9543dc3](https://github.com/ViewComponent/view_component/pull/1609/commits/9543dc35f31cae2be609bf8019051fbfce2c93ba) is merging v3 stuff into this branch. I'm not sure if that can be picked out of if I need a whole new branch.

PR #1612, opened 2022-12-20
Move system test endpoint out of preview controller to avoid issues when previews have authentication

edwinthinks:

This PR aims to address an issue discovered in https://github.com/ViewComponent/view_component/pull/1606#issuecomment-1349640678 in which adding authentication to the previews controller causes the JS testing approach to not work. The solution we agreed upon in https://github.com/ViewComponent/view_component/pull/1606#issuecomment-1351728446 is to move the `_system_test_endpoint` into its own controller, since it isn't related to previews.

This approach was taken because it seemed the easiest and safest way to get the desired outcome.

N/A

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@ nav_order: 5
 
 ## main
 
-* Move system test endpoint out of the unrelated previews controller
+* Move system test endpoint out of the unrelated previews controller.
 
     *Edwin Mak*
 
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Move system test endpoint out of the unrelated previews controller
```
```suggestion
* Move system test endpoint out of the unrelated previews controller.
```

joelhawksley (approved):

developius:
Thank you @edwinthinks & @joelhawksley! This is great 🙏

PR #1616, opened 2022-12-21
Make render_in_view_context forward its args to the block

camertron:

Our `render_in_view_context` test helper uses `instance_exec` to evaluate the given block in the view context. Since `instance_exec` changes the value of `self`, instance variables, methods, etc defined in the caller's context are not available inside the block, which can be pretty confusing. For example, consider this test:

```ruby
class MyTest < Minitest::Test
  def setup
    @user = create(:user)
  end

  def test_my_component
    render_in_view_context do
      render(MyComponent.new(@user))
    end
  end
end
```

As a direct result of using `instance_exec`, `@user` will be `nil` inside the block.

Presumably to provide an escape hatch for this problem, `instance_exec` forwards its arguments when calling the block. This PR modifies `render_in_view_context` to forward _its_ arguments to `instance_exec`, so this is now possible:

```ruby
class MyTest < Minitest::Test
  def setup
    @user = create(:user)
  end

  def test_my_component
    render_in_view_context(@user) do |user|
      render(MyComponent.new(user))
    end
  end
end
```

### DIFF
--- a/.standard.yml
+++ b/.standard.yml
@@ -1,0 +1,1 @@
+ruby_version: 2.5
--- a/test/sandbox/test/rendering_test.rb
+++ b/test/sandbox/test/rendering_test.rb
@@ -15,6 +15,17 @@ class RenderingTest < ViewComponent::TestCase
     assert_selector("div", text: "hello,world!")
   end
 
+  def test_render_in_view_context_forwards_arguments
+    @foo = "foo"
+    @bar = "bar"
+
+    render_in_view_context(@foo, bar: @bar) do |foo, bar:|
+      render(MyComponent.new) { foo + bar }
+    end
+
+    assert_text "hello,world!\nfoobar"
+  end
+
   def test_render_inline_returns_nokogiri_fragment
     assert_includes render_inline(MyComponent.new).css("div").to_html, "hello,world!"
   end
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Modify the `render_in_view_context` test helper to forward its args to the block.
+
+    *Cameron Dutro*
+
 ## 2.80.0
 
 * Move system test endpoint out of the unrelated previews controller.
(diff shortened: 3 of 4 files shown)


.standard.yml
```diff
@@ -0,0 +1 @@
+ruby_version: 2.5
```
Spone, line 1: This change seems unrelated?

joelhawksley (commented):
Makes sense to me. Mind adding a regression test?

.standard.yml
camertron: Good call-out, but it actually is related. Standard wanted me to use argument forwarding, eg `render_in_view_context(...)` instead of `render_in_view_context(*args, &block)`, but arg forwarding wasn't introduced until Ruby 2.7. Since view_component supports Ruby versions all the way back to 2.5, I needed to tell standard to only apply rules that make sense for Ruby versions <= 2.5.

Spone: Makes sense, thanks for clarifying!

test/sandbox/test/rendering_test.rb
```diff
@@ -15,6 +15,17 @@ def test_render_in_view_context
     assert_selector("div", text: "hello,world!")
   end
 
+  def test_render_in_view_context_forwards_arguments
+    @foo = "foo"
+    @bar = "bar"
+
+    render_in_view_context(@foo, bar: @bar) do |foo, bar:|
+      render(MyComponent.new) { foo + bar }
```
joelhawksley, line 23: Not sure if it matters, but this test does not match the PR description, which had a reference to the ivar inside the `render` call 🤔

BlakeWilliams (approved):
Nice, I really like this solution. 👍

test/sandbox/test/rendering_test.rb
camertron: The example in the PR description that uses an ivar inside `render_in_view_context` is the one that doesn't work. There's nothing we can do about ivar access (since the value of `self` is changed by `instance_exec`), so there's no way to make the example work as written. The idea behind this PR is to be able to pass data _stored_ in ivars (and elsewhere) into the block as arguments (locals). It's an imperfect workaround, but I don't think there's much else we can do.

PR #1617, opened 2022-12-22
DO NOT MERGE Add support for relative translations in initializer

### LINKED ISSUE #1577: Relative translations don't work in initializer
This is similar to #74

If I have an initializer and I use the translation helper, then I get the error:
```
Cannot use t(".link_text") shortcut because path is not available
```
This happens even after calling `super`

If I move the translation into an instance method, it works as expected.

Using view_component-2.75.0, rails-7.0.4, ruby 3.1.0

joelhawksley:
In order for translations to work in an initializer, ActionView requires @virtual_path to be set. To set it, we allocate the component, set @virtual_path, then call the component's initializer. 

This is either terrible, brilliant, or both. What do you think?

Closes #1577

(I used the new argument forwarding syntax from Ruby 2.7, so we won't be able to land this before v3.0.0)

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -83,6 +83,18 @@ module ViewComponent
     _deprecated_generate_mattr_accessor :sidecar
     _deprecated_generate_mattr_accessor :stimulus_controller
 
+    # In order for translations to work in an initializer,
+    # ActionView requires @virtual_path to be set. To set it,
+    # we allocate the component, set @virtual_path, then
+    # call the component's initializer. This is either terrible,
+    # brilliant, or both.
+    def self.new(...)
+      precursor = self.allocate
+      precursor.instance_variable_set(:@virtual_path, virtual_path)
+      precursor.send(:initialize, ...)
+      precursor
+    end
+
     # Entrypoint for rendering components.
     #
     # - `view_context`: ActionView context from calling view
@@ -106,9 +118,6 @@ module ViewComponent
 
       # For content_for
       @view_flow ||= view_context.view_flow
-
-      # For i18n
-      @virtual_path ||= virtual_path
 
       # For template variants (+phone, +desktop, etc.)
       @__vc_variant ||= @lookup_context.variants.first
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -89,7 +89,7 @@ module ViewComponent
     # call the component's initializer. This is either terrible,
     # brilliant, or both.
     def self.new(...)
-      precursor = self.allocate
+      precursor = allocate
       precursor.instance_variable_set(:@virtual_path, virtual_path)
       precursor.send(:initialize, ...)
       precursor
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add support for relative translations in initializer.
+
+    *Joel Hawksley*
+
 * Move system test endpoint out of the unrelated previews controller.
 
     *Edwin Mak*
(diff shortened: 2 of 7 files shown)


lib/view_component/base.rb
```diff
@@ -83,6 +83,18 @@ def self._deprecated_generate_mattr_accessor(name)
     _deprecated_generate_mattr_accessor :sidecar
     _deprecated_generate_mattr_accessor :stimulus_controller
 
+    # In order for translations to work in an initializer,
+    # ActionView requires @virtual_path to be set. To set it,
+    # we allocate the component, set @virtual_path, then
+    # call the component's initializer. This is either terrible,
+    # brilliant, or both.
```
elia, line 90: > This is either terrible, brilliant, or both.

All of the above! 😅 

Another terrible/brilliant proposal is that we only redefine `allocate` and use `super` instead of `send`:

```rb
def self.allocate
  component = super
  component.instance_variable_set(:@virtual_path, virtual_path)
  component
end
```

This should also fix the syntax issues in ruby 2.5 and 2.6, assuming we'll keep supporting them.

---


This doesn't seem to affect performance, so it's just style and communication:

```
             vanilla   541.319k i/100ms
            allocate   589.440k i/100ms
                 new   588.357k i/100ms
             vanilla      5.403M (± 0.7%) i/s -     27.066M in   5.010067s
            allocate      5.891M (± 0.3%) i/s -     29.472M in   5.002889s
                 new      5.890M (± 0.5%) i/s -     30.006M in   5.094568s

Comparison:
            allocate:  5891035.6 i/s
                 new:  5889982.9 i/s - same-ish: difference falls within error
             vanilla:  5402556.1 i/s - 1.09x  (± 0.00) slower

```

```rb
require 'benchmark/ips'

class Vanilla
  def initialize(bar:)
    @foo = foo
    @bar = bar
  end
  
  def foo
    123
  end
end

class Allocate
  def allocate
    object = super
    object.instance_variable_set :@foo, foo
    object
  end
  
  def initialize(bar:)
    @bar = bar
  end
  
  def foo
    123
  end
end

class New
  def new(...)
    object = allocate
    object.instance_variable_set :@foo, foo
    object.initialize(...)
    object
  end
  
  def initialize(bar:)
    @bar = bar
  end
  
  def foo
    123
  end
end

Benchmark.ips do |x|
  x.report("vanilla") { Vanilla.new(bar: 234) }
  x.report("allocate") { Allocate.new(bar: 234) }
  x.report("new") { New.new(bar: 234) }
  x.compare!
end
```

elia (commented):
Looks terrible… and brilliant! 😅 

Anyway given this is library code I think we're allowed to mess a little bit with `new` and `allocate`. 😊 

I left a suggestion that should avoid the syntax issues and requires less code, it might be slightly less intentional, but in any case a comment on why we're touching `new`/`allocate` is due to our future selves reading the code.

lib/view_component/base.rb
```diff
@@ -83,6 +83,18 @@ def self._deprecated_generate_mattr_accessor(name)
...
+    def self.new(...)
+      precursor = allocate
+      precursor.instance_variable_set(:@virtual_path, virtual_path)
+      precursor.send(:initialize, ...)
```
BlakeWilliams, line 94: In general I don't think the indirection is worth it when it comes to `self.new`, but setting that aside I don't think we should move forward with this if this is only fixing translations in the initializer, since that might set the expectation that *all* behavior should be runnable in the initializer.

I'd love to see use-cases where translations are necessary in the initializer since I imagine a better pattern would be to push folks towards using some other hook (e.g. `before_render`) instead of relying on the initializer to setup view data. That would also help co-locate logic that needs to run before a component is rendered.

joelhawksley:
@BlakeWilliams as a counterpoint, I think the real awkwardness here is that Rails doesn't let us provide the virtual path explicitly, instead requiring it to be set as an instance variable. 

If Rails could check for an instance method on the current context, we could define that method to return the virtual path which we already know at compile time. Regardless, I think I'll write up a clearer exception error message for this case for now.

PR #1619, opened 2022-12-26
Fix layout nesting in system tests when using `with_rendered_component_path` with custom layouts

### LINKED ISSUE #1618: When using a custom layout, the `with_rendered_component_path` doesn't work as expected
When using a custom layout in tests, the `with_rendered_component_path` wraps the HTML output in the [default `Rails::ApplicationController` layout](https://github.com/rails/rails/blob/main/railties/lib/rails/templates/layouts/application.html.erb), with the component and custom layout nested in the `<body>` of that layout:

```rb
with_rendered_component_path(render_inline(MyComponent.new), layout: 'custom_layout') do |path|
  visit(path)
end
```

```html
<!-- `Rails::ApplicationController` layout -->
<!DOCTYPE html>
<html lang="en">
  <head></head>
  <body>
    <!-- custom layout -->
    <!DOCTYPE html>
    <html>
      <head></head>
      <body>
        <div>hello!</div>
      </body>
    </html>
  </body>
</html>
```

When a custom layout is used, that custom layout is applied before it reaches the controller via the [`with_rendered_component_path_helper`](https://github.com/ViewComponent/view_component/blob/main/lib/view_component/system_test_helpers.rb#L18). Then, since the [`ViewComponentSystemTestController`](https://github.com/ViewComponent/view_component/blob/main/app/controllers/view_components_system_test_controller.rb) inherits from [`Rails::ApplicationController`](https://github.com/rails/rails/blob/main/railties/lib/rails/application_controller.rb), which has a default layout defined, the default `Rails::ApplicationController` layout is also used. So essentially this ends up rendering the component with two nested layouts.

I'll open a PR with a fix.

### Steps to reproduce

Using `with_rendered_component_path(render_inline(MyComponent.new), layout: 'custom_layout')` with any layout file will result in the default `Rails::ApplicationController` layout wrapping the rendered HTML.

Given this component:

```rb
  def call
    tag.div('hello!')
  end
end
```

And this custom layout:

```erb
<!DOCTYPE html>
<html>
  <head>
    <title>Component Layout</title>
  </head>
  <body>
    <%= yield %>
  </body>
</html>
```

This spec fails, because the `<title>` tag from the custom layout ends up being wrapped in the body of the default Rails layout:

```rb
require "rails_helper"

RSpec.describe MyComponent, type: :feature do
  include ViewComponent::SystemTestHelpers

  let(:page) { Capybara.current_session }

  it 'does not include a <title> in the body when rendered with a layout' do
    with_rendered_component_path(render_inline(MyComponent.new), layout: 'custom_layout') do |path|
      visit(path)

      expect(page).not_to have_css('body > title', visible: false)
    end
  end
end
```

### Expected behavior

I'd expect the HTML generated by `with_rendered_component_path` in the example above to only include the content from the custom layout:

```html
<!DOCTYPE html>
<html>
  <head>
    <title>Component Layout</title>
  </head>
  <body>
    <div>hello!</div>
  </body>
</html>
``` 

### Actual behavior

The HTML generated by `with_rendered_component_path` in the example above wraps everything (including the custom layout) in the default `Rails::ApplicationController` layout:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <title></title>
    <style>
      /* ... */
    </style>
  </head>
  <body>
    <!-- content from custom layout is nested in the body of the default layout -->
    <!DOCTYPE html>
    <html>
      <head>
        <title>Component Layout</title>
      </head>
      <body>
        <div>hello!</div>
      </body>
    </html>
  </body>
</html>
```

### System configuration
**Rails version**: 7.0.4

**Ruby version**: 2.7.6

**Gem version**: 2.8.0

ihollander:

Fixes #1618

This PR fixes an issue where using the `with_rendered_component_path` helper with a custom layout results in that layout being nested in the body of the [default `Rails::ApplicationController` layout](https://github.com/rails/rails/blob/main/railties/lib/rails/templates/layouts/application.html.erb).

This also fixes another issue where layouts with script or stylesheet tags were being blocked by Content Secure Policy:

Screen Shot 2022-12-26 at 9 46 38 AM

...because `Rails::ApplicationController` includes a CSP header, which means that only layouts with inline scripts are allowed by CSP:

```rb
# https://github.com/rails/rails/blob/main/railties/lib/rails/application_controller.rb#L9-L12
  content_security_policy do |policy|
    policy.script_src :self, :unsafe_inline
    policy.style_src :self, :unsafe_inline
  end
```

The approach I took was to have the `ViewComponentsSystemTestController` inherit from `ActionController::Base`, which doesn't have a default layout defined so no layout is used unless one is explicitly passed to the helper. `ActionController::Base` also doesn't set any CSP header.

Another possible approach that I didn't pursue too far would be to not use a layout when rendering the component to a file [here](https://github.com/ViewComponent/view_component/blob/main/lib/view_component/system_test_helpers.rb#L18), and instead pass layout as a param to the `ViewComponentsSystemTestController`:

```rb
        file.write(controller.render_to_string(html: fragment.to_html.html_safe))
        file.rewind

        block.call("/_system_test_entrypoint?file=#{file.path.split("/").last}&layout=#{layout.to_s}")
```

This wouldn't fix the CSP issue, however.

Curious to get the reviewer's thoughts on what spec coverage is helpful here - I added a test case that checks if a `<title>` tag is rendered inside the `<body>` as a way to indirectly check for the nested layout issue, but I'm not sure how worthwhile that test is outside of fixing this specific issue.

### DIFF
--- a/app/controllers/view_components_system_test_controller.rb
+++ b/app/controllers/view_components_system_test_controller.rb
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
-require "rails/application_controller"
-
-class ViewComponentsSystemTestController < Rails::ApplicationController # :nodoc:
+class ViewComponentsSystemTestController < ActionController::Base # :nodoc:
   def system_test_entrypoint
     render file: "./tmp/view_components/#{params.permit(:file)[:file]}"
   end
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Fix tests using `with_rendered_component_path` with custom layouts.
+
+    *Ian Hollander*
+
 ## 2.80.0
 
 * Move system test endpoint out of the unrelated previews controller.
--- a/docs/index.md
+++ b/docs/index.md
@@ -205,6 +205,7 @@ ViewComponent is built by over a hundred members of the community, including:
 <img src="https://avatars.githubusercontent.com/yykamei?s=64" alt="yykamei" width="32" />
 <img src="https://avatars.githubusercontent.com/matheuspolicamilo?s=64" alt="matheuspolicamilo" width="32" />
 <img src="https://avatars.githubusercontent.com/erinnachen?s=64" alt="erinnachen" width="32" />
+<img src="https://avatars.githubusercontent.com/ihollander?s=64" alt="ihollander" width="32" />
 
 ## Who uses ViewComponent?
 
--- a/test/sandbox/app/views/layouts/application.html.erb
+++ b/test/sandbox/app/views/layouts/application.html.erb
@@ -6,15 +6,15 @@
 </head>
 <body data-layout="application">
   <%= yield %>
+
+  <script>
+    document.addEventListener("DOMContentLoaded", function(event) {
+      let button = document.querySelector('[data-button]')
+      let hiddenField = document.querySelector('[data-hidden-field]')
+      button.addEventListener('click', function() {
+        hiddenField.style.display = "block"
+      })
+    })
+  </script>
 </body>
 </html>
-
-<script>
-  document.addEventListener("DOMContentLoaded", function(event) {
-    let button = document.querySelector('[data-button]')
-    let hiddenField = document.querySelector('[data-hidden-field]')
-    button.addEventListener('click', function() {
-      hiddenField.style.display = "block"
-    })
-  })
-</script>
--- a/test/sandbox/test/view_component_system_test.rb
+++ b/test/sandbox/test/view_component_system_test.rb
@@ -44,4 +44,13 @@ class ViewComponentSystemTest < ViewComponent::SystemTestCase
       find(".title", text: "This is my title!")
     end
   end
+
+  def test_layout_is_not_nested
+    with_rendered_component_path(render_inline(MyComponent.new), layout: "application") do |path|
+      visit path
+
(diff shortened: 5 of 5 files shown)


app/controllers/view_components_system_test_controller.rb
```diff
@@ -1,8 +1,6 @@
 # frozen_string_literal: true
 
-require "rails/application_controller"
-
-class ViewComponentsSystemTestController < Rails::ApplicationController # :nodoc:
```
Spone, line 5: @edwinthinks I'm curious to know if there was a specific reason for using `Rails:: ApplicationController` in the first place?

joelhawksley (approved):
Are there any docs you think we should add for this change?

ihollander:
I think the docs [here](https://viewcomponent.org/guide/testing.html#component-system-tests) for the `with_rendered_component_path` with the `layout` option do a good job explaining how to use that method, and now with this fix in I think the `layout` option works the way I'd expect it to.

app/controllers/view_components_system_test_controller.rb
edwinthinks: Hey @Spone ; There is no specific reason for doing it with `ApplicationController`. Based on the issues this PR resolves, this does seem like a better option.

Thanks @ihollander for tackling this!

Spone: Thanks for your confirmation!

PR #1640, opened 2023-01-27
Add GetBlock to the list of users

JoGetBlock:
Adding GetBlock, an RPC provider to the list of users.


Spone:
Hi @JoGetBlock can you please keep the list ordered alphabetically and add a Changelog entry? Thanks!

PR #946, opened 2021-06-03
Adds documentation for usage with Rails forms

pinzonjulian:
Closes #908

### Summary

This PR adds a comprehensive guide to using `ViewComponent` with forms in `Ruby on Rails`. 

#### Notable changes
1. I decided to add a new section to the documentation called _Practical Examples_ as this guide is more of an explanation on how to use `ViewComponent` for a particular purpose rather than how to use the library itself.
2. I've removed the forms issue from the _Known Issues_ section and included it inside the guide for forms.

### Other Information
I wrote this guide with an opinionated tone as if these were a set of "good practices" when using `ViewComponent` in the context of Rails forms. The advice in this guide comes from having used this library in place of `ActionView` for the past year and a half and finding the best way possible to let each part of the framework do what they do best. However these are just the opinions from my experience so any feedback from other heavy `ViewComponent` users is super welcome!

#### Navigation screenshot with new section
image

#### Snippet of table of contents
For context when reviewing this PR

image

Issue #241, opened 2020-03-03
Strange behavior with code blocks using form helpers and Slim


javierav:
Hi!

I'm playing with this project for use in production but I encountered a possible bug related with passing a Slim/HTML block to Rails helper.

```ruby
# app/components/test_component.rb

class TestComponent < ActionView::Component::Base
  def initialize(form:)
    @form = form
  end

  attr_reader :form
end
```

```slim
/ app/components/test_component.html.slim

div
  = form.text_field :name
  = form.label :name do
    div.label__text
      | Nombre
```

```slim
/ app/views/tests/show.html.slim

= form_with model: @test do |form|
  /! Test Component Start
  = render TestComponent.new(form: form)
  /! Test Component End
```

Expected output:

```html
<!-- Test Component Start -->
<div>
  <input type="text" name="test[name]" id="test_name">
  <label for="test_name">
    <div class="label__text">Nombre</div>
  </label>
</div>
<!-- Test Component End -->
```

Current output:

```html
<!-- Test Component Start -->
<div>
  <input type="text" name="test[name]" id="test_name">
  <div class="label__text">Nombre</div>
  <label for="test_name">
    <div>
      <input type="text" name="test[name]" id="test_name">
      <div class="label__text">Nombre</div>
    </div><!-- Test Component End -->
  </label>
</div>
```

In normal Rails the block of code passed to the helper works perfectly. Any ideas? Thanks!

BlakeWilliams:
This looks really similar to the issue I ran into that caused me to open https://github.com/github/actionview-component/pull/240.

sanchez2210:
Just ran into this and for a simple case temporarily solved it by doing 
```
<%# 
  <% label_contents = '
    <span class="form-file-text">Cambiar imagen</span>
    <span class="form-file-button">Abrir</span>
  ' %>

  <%= @form.label(@attached_method, raw(label_contents), class: "form-file-label") %>
```

mmhan:
It seems that this error is happening with `link_to` code blocks and `tag.<tagname>` blocks too. 

Any reason why this is happening?

PR #1644, opened 2023-01-29
Update section on form_for compatibility

tmaier:

I think it helps people when they know what the compatibility issues between `form_for` and ViewComponents actually mean and what options they have.

#241 

I wanted to keep the section small, but mention all options known so far.
It does not seek to create a dedicated page on this particular topic, like #946 tried.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Improve documentation of known incompatibilities with Rails form helpers.
+
+    *Tobias L. Maier*
+
 * Add Mon Ami to list of companies using ViewComponent.
 
     *Ethan Lee-Tyson*
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -18,8 +18,8 @@ Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
-This means that it isn't supported to pass a form object to a view component.
-It won't render as expected. Some content will be rendered twice.
+Passing a form object reference (usually `f`) to a view component works for simple cases, like `f.text_field :name`.
+Content will be ill-ordered or duplicated in complex cases (such as passing blocks to form helpers or when nesting components).
 
 Consider following options:
 
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -14,9 +14,22 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
 
 Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_frame(id: dom_id(my_model))`.
 
-## form_for compatibility
+## `form_for` compatibility
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
+
+This means that it isn't supported to pass a form object to a view component.
+It won't render as expected. Some content will be rendered twice.
+
+Consider following options:
+
+- Render a form end-to-end within one view component
+- Render a [classical view partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the view component, which includes the form
+- Use [FormBuilder](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components.
+  Note: There are multiple approaches how to implement a FormBuilder:
+  - Classic approach (without ViewComponent)
+  - Using a lightweight implementation of ViewComponent (for example, [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) at [Primer ViewComponents](https://github.com/primer/view_components))
+  - Using ViewComponent (for example, [view_component-form gem](https://github.com/pantographe/view_component-form))
 
 ## Inconsistent controller rendering behavior between Rails versions
 
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -18,17 +18,17 @@ Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
-Passing a form object reference (usually `f`) to a view component works for simple cases, like `f.text_field :name`.
+Passing a form object reference (usually `f`) to a ViewComponent works for simple cases, like `f.text_field :name`.
 Content will be ill-ordered or duplicated in complex cases (such as passing blocks to form helpers or when nesting components).
 
 Consider following options:
 
-- Render a form end-to-end within one view component
-- Render a [classical view partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the view component, which includes the form
-- Use `[FormBuilder](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders)` to create reusable form components.
+- Render a form end-to-end within one ViewComponent
+- Render a [classical view partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the ViewComponent, which includes the form
+- Use a [custom `FormBuilder`](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components.
   Note: There are multiple approaches how to implement a FormBuilder:
   - Classic approach (without ViewComponent)
-  - Using ViewComponent (for example, [view_component-form gem](https://github.com/pantographe/view_component-form))
+  - Using a FormBuilder that overrides all field helpers to render a ViewComponent, so that you can customize each field individually (for example, [view_component-form gem](https://github.com/pantographe/view_component-form))
   - Using a lightweight implementation of ViewComponent. For example, [Primer ViewComponents](https://github.com/primer/view_components) implemented with [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) a lightweight version, which they use in the context of `FormBuilder`.
 
 ## Forms don't use the default `FormBuilder`
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -25,7 +25,7 @@ Some workarounds include:
 - Render an entire form within a single ViewComponent.
 - Render a [partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the ViewComponent which includes the form.
 - Use a [custom `FormBuilder`](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components:
-  - Without FormBuilder at all, [use Action View helpers](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html) to create the desired output
+  - Using FormBuilder the default way. This means to [use Action View helpers](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html) to create the desired output
   - Using a FormBuilder that overrides all field helpers to render a ViewComponent, so that each field can be customized individually (for example, [view_component-form](https://github.com/pantographe/view_component-form))
   - Using a lightweight implementation of ViewComponent. For example, [Primer ViewComponents](https://github.com/primer/view_components) implemented [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) which is used in the context of `FormBuilder`.
 
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -25,11 +25,11 @@ Consider following options:
 
 - Render a form end-to-end within one view component
 - Render a [classical view partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the view component, which includes the form
-- Use [FormBuilder](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components.
+- Use `[FormBuilder](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders)` to create reusable form components.
   Note: There are multiple approaches how to implement a FormBuilder:
   - Classic approach (without ViewComponent)
-  - Using a lightweight implementation of ViewComponent (for example, [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) at [Primer ViewComponents](https://github.com/primer/view_components))
   - Using ViewComponent (for example, [view_component-form gem](https://github.com/pantographe/view_component-form))
+  - Using a lightweight implementation of ViewComponent. For example, [Primer ViewComponents](https://github.com/primer/view_components) implemented with [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) a lightweight version, which they use in the context of `FormBuilder`.
 
 ## Inconsistent controller rendering behavior between Rails versions
 
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -14,7 +14,7 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
 
 Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_frame(id: dom_id(my_model))`.
 
-## `form_for` compatibility
+## Compatibility with Rails form helpers
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -18,16 +18,16 @@ Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
-Passing a form object reference (often `f`) to a ViewComponent works for simple cases, like `f.text_field :name`. Content may be ill-ordered or duplicated in complex cases, such as passing blocks to form helpers or when nesting components.
+Passing a form object (often `f`) to a ViewComponent works for simple cases like `f.text_field :name`. Content may be ill-ordered or duplicated in complex cases, such as passing blocks to form helpers or when nesting components.
 
 Some workarounds include:
 
 - Render an entire form within a single ViewComponent.
 - Render a [partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the ViewComponent which includes the form.
 - Use a [custom `FormBuilder`](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components:
-  - Using FormBuilder the default way. This means to [use Action View helpers](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html) to create the desired output
-  - Using a FormBuilder that overrides all field helpers to render a ViewComponent, so that each field can be customized individually (for example, [view_component-form](https://github.com/pantographe/view_component-form))
-  - Using a lightweight implementation of ViewComponent. For example, [Primer ViewComponents](https://github.com/primer/view_components) implemented [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) which is used in the context of `FormBuilder`.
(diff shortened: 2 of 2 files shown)


docs/known_issues.md
```diff
@@ -14,10 +14,23 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
 
 Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_frame(id: dom_id(my_model))`.
 
-## form_for compatibility
+## `form_for` compatibility
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
+This means that it isn't supported to pass a form object to a view component.
```
Spone, line 21: Actually it can work for some simple cases: you pass the form builder reference (usually `f`) to the component, as an argument, and use it in the component, ie. `f.text_field :name`. But for more complex cases (passing blocks to form helpers) you may have duplicate or wrongly ordered content.

Spone:
Thanks for taking a stab at this!

docs/known_issues.md
tmaier: You are right, @Spone. I am uncertain if we need to mention this details here... My focus is more the "way forward" with the options one has...

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Improve documentation of known error re. `form_for`
```
camertron, line 13:
```suggestion
* Improve documentation of known incompatibilities with Rails form helpers.
```

docs/known_issues.md
```diff
@@ -14,10 +14,23 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
 
 Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_frame(id: dom_id(my_model))`.
 
-## form_for compatibility
+## `form_for` compatibility
```
camertron, line 17: There are actually two form helpers affected, `form_for` and `form_with`. I think it makes sense to refer to them both together as "form helpers."

```suggestion
## Compatibility with Rails form helpers
```

docs/known_issues.md
```diff
@@ -14,10 +14,23 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
+This means that it isn't supported to pass a form object to a view component.
+It won't render as expected. Some content will be rendered twice.
+
+Consider following options:
+
+- Render a form end-to-end within one view component
+- Render a [classical view partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the view component, which includes the form
+- Use [FormBuilder](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components.
```
camertron, line 28: I'm not sure a form builder will help across components. The builder is actually a big part of the reason form helpers aren't compatible with view components, namely because they retain a reference to the view context in which they were called. I was trying to address that [here](https://github.com/ViewComponent/view_component/pull/1642). I suppose you could write your own form helper that does something similar, but there would be no way to know which view context is the "current" one without some additional help from the VC library itself (i.e. what I was trying to do with `@render_stack`).

Or maybe you mean a form builder in the generic sense, i.e. not Rails form builders?

docs/known_issues.md
```diff
@@ -14,10 +14,23 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+  Note: There are multiple approaches how to implement a FormBuilder:
+  - Classic approach (without ViewComponent)
+  - Using a lightweight implementation of ViewComponent (for example, [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) at [Primer ViewComponents](https://github.com/primer/view_components))
```
camertron, line 31: I'm not sure `ActsAsComponent` is something we want to mention in the official VC docs. I think of it as a private class that's only designed to be used internally. We could expose it as its own gem, but it lacks many standard VC features like slots.

camertron (changes requested):

docs/known_issues.md
tmaier: It is just an example. One could do something else. I for one like the last option for example...

We can revise this section, when there is a better solution. But think about someone from the developer/user side. He/she is looking for a solution. So it makes sense to mention all known options

docs/known_issues.md
tmaier: I also mean the rails FormBuilder. I know this is an issue.

When I did my research on this topic, I saw multiple repos, where people tried to use view components to style HTML form elements. E.g. to style a textarea field. However, in this case, FormBuilder can be considered a view component implementation of some sort. And maybe this is enough for them and their use case at this moment.

docs/known_issues.md
tmaier: If there is demand, I can of course elaborate on this. @Spone and @ianterrell, what do you think?

tmaier: @ianterrell, thanks for giving some insights. What is your argument to not put the label, description and error component into a form builder? For example, viewcomponent-form gem would allow you to combine both worlds.

docs/known_issues.md
tmaier: I tried to re-word it. Please see, if you like it.

docs/known_issues.md
ianterrell: Do you mean this [view_component-form](https://github.com/pantographe/view_component-form) gem? For large projects I try to run as few dependencies as possible. We don't even use Devise. I wonder if we should be using ViewComponent 😅.

I don't have a strong argument not to put the label etc into a form builder, other than it just seemed natural to use view components directly, and we have some inertia in that direction — we use them almost everywhere else, after all. I didn't think about using form builders at first, and now we have a lot of forms built this way. 

I was surprised when in some cases ViewComponents didn't work. In fact, in some cases they did work and then stopped working, when the global buffer changes were being tried out. 

All that said... since they play so poorly together — or at least, poorly together _sometimes_ — we probably should transition to using our own form builder.

Spone: Overall, I think if we're trying to document how things currently are, we should aim at being as specific as possible, so that people can make an informed choice between the available options.

I would draw the line at describing specifics about external libraries, since I prefer people get their info first-hand from their docs (and we don't have to keep it updated on our side).

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
 Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_frame(id: dom_id(my_model))`.
 
-## form_for compatibility
+## Compatibility with Rails form helpers
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
-## Inconsistent controller rendering behavior between Rails versions
+Passing a form object reference (usually `f`) to a view component works for simple cases, like `f.text_field :name`.
+Content will be ill-ordered or duplicated in complex cases (such as passing blocks to form helpers or when nesting components).
```
ianterrell, line 22: This may be unhelpful, but — is this limited to form helpers? I feel like I remember seeing issues with other block capturing helpers, and `link_to` is similarly mentioned in one of the issues: https://github.com/ViewComponent/view_component/issues/241#issuecomment-905233222

Does that help clarify the difference between the "simple" and "complex" cases?

Spone: Turbo helpers as well.

docs/known_issues.md
camertron: Ok I'll take another look. I understand it's an example, I just don't want to support someone using it outside the context it was designed for 😅

docs/known_issues.md
camertron: Yes, the compatibility issues affect numerous Rails helpers. Another one that hasn't been mentioned yet is `content_for`, and I think `content_tag` too.

docs/known_issues.md
camertron: Ahh ok I understand. That's an interesting point.

camertron (commented):
Alright I'm cool with this change, although I wonder if it's obviated by @BlakeWilliams' [timely PR](https://github.com/ViewComponent/view_component/pull/1650). At the very least I think we should include the new `CaptureCompatibility` module in the list.

docs/known_issues.md
tmaier: Shall we mention the other affected helpers as well? Maybe within an additional section?

```markdown
## Other Rails helpers with compatibility issues

Similar to the form helpers (see above), the following helpers have compatibility issues:

- `content_for` --> Set directly within a Rails view
- `link_to`(?) --> ???
- `content_tag` --> Use xzy instead

To keep this PR small, I would like to do this in another PR

File: docs/known_issues.md [hunk 7]
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
 
 Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_frame(id: dom_id(my_model))`.
 
-## form_for compatibility
+## Compatibility with Rails form helpers
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
-## Inconsistent controller rendering behavior between Rails versions
+Passing a form object reference (usually `f`) to a view component works for simple cases, like `f.text_field :name`.
```
Spone, line 21:
```suggestion
Passing a form object reference (usually `f`) to a ViewComponent works for simple cases, like `f.text_field :name`.
```

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+Content will be ill-ordered or duplicated in complex cases (such as passing blocks to form helpers or when nesting components).
 
-In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout.
+Consider following options:
 
-## Forms don't use the default form builder
+- Render a form end-to-end within one view component
```
Spone, line 26:
```suggestion
- Render a form end-to-end within one ViewComponent
```

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+- Render a [classical view partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the view component, which includes the form
```
Spone, line 27:
```suggestion
- Render a [classical view partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the ViewComponent, which includes the form
```

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+- Use `[FormBuilder](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders)` to create reusable form components.
```
Spone, line 28:
```suggestion
- Use a custom `[FormBuilder](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders)` to create reusable form components.
```

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+  Note: There are multiple approaches how to implement a FormBuilder:
+  - Classic approach (without ViewComponent)
```
Spone, line 30: Can you elaborate on what "classic approach" means? You mean not using ViewComponents at all to generate the form, right?

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+  - Using ViewComponent (for example, [view_component-form gem](https://github.com/pantographe/view_component-form))
```
Spone, line 31: I may be contradicting [my own comment](https://github.com/ViewComponent/view_component/pull/1644#discussion_r1092464568) here, but I now think we need to describe the overall approach.

```suggestion
  - Using a FormBuilder that overrides all field helpers to render a ViewComponent, so that you can customize each field individually (for example, [view_component-form gem](https://github.com/pantographe/view_component-form))
```

docs/known_issues.md
tmaier: in the sense of how to customize the form builder. I mean, whatever you find when you google for "custom form builder rails"

docs/known_issues.md
camertron: Yeah another PR sounds like a good idea.

For both `link_to` and `content_tag`, the best suggestion is probably to just write the HTML by hand 😢

Spone (approved):
👌

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
 Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_frame(id: dom_id(my_model))`.
 
-## form_for compatibility
+## Compatibility with Rails form helpers
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
-## Inconsistent controller rendering behavior between Rails versions
+Passing a form object reference (usually `f`) to a ViewComponent works for simple cases, like `f.text_field :name`.
+Content will be ill-ordered or duplicated in complex cases (such as passing blocks to form helpers or when nesting components).
```
joelhawksley, line 22:
```suggestion
Passing a form object reference (often `f`) to a ViewComponent works for simple cases, like `f.text_field :name`. Content may be ill-ordered or duplicated in complex cases, such as passing blocks to form helpers or when nesting components.
```

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
 
-In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout.
+Consider following options:
```
joelhawksley, line 24:
```suggestion
Some workarounds include:
```

docs/known_issues.md
```diff
@@ -14,15 +14,24 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+- Render a form end-to-end within one ViewComponent
+- Render a [classical view partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the ViewComponent, which includes the form
+- Use a [custom `FormBuilder`](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components.
+  Note: There are multiple approaches how to implement a FormBuilder:
+  - Without FormBuilder at all, where one would [use Action View helpers](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html) to create the desired output
+  - Using a FormBuilder that overrides all field helpers to render a ViewComponent, so that you can customize each field individually (for example, [view_component-form gem](https://github.com/pantographe/view_component-form))
+  - Using a lightweight implementation of ViewComponent. For example, [Primer ViewComponents](https://github.com/primer/view_components) implemented with [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) a lightweight version, which they use in the context of `FormBuilder`.
```
joelhawksley, line 32:
```suggestion
- Render an entire form within a single ViewComponent.
- Render a [partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the ViewComponent which includes the form.
- Use a [custom `FormBuilder`](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components:
  - Without FormBuilder at all, [use Action View helpers](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html) to create the desired output
  - Using a FormBuilder that overrides all field helpers to render a ViewComponent, so that each field can be customized individually (for example, [view_component-form](https://github.com/pantographe/view_component-form))
  - Using a lightweight implementation of ViewComponent. For example, [Primer ViewComponents](https://github.com/primer/view_components) implemented [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) which is used in the context of `FormBuilder`.
```

joelhawksley (commented):
👋🏻 Just some grammar and voicing edits to keep the docs cohesive. Happy to re-edit if material changes are requested.

docs/known_issues.md
Spone: > Render a [partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) which includes the form, within the ViewComponent.

What about reordering this? Because it's the partial that includes the form, not the ViewComponent.

Spone: > Without FormBuilder at all

It feels weird that this option is introduced as a sub-item of "Use a custom `FormBuilder`"

tmaier: > Without FormBuilder at all

This statement is just wrong. I need to fix it.

camertron (approved):

camertron:
Hey @tmaier, could you resolve that merge conflict? As soon as that's done, I'll merge this in 😎

tmaier:
@camertron  done

camertron:
@tmaier sorry for the churn here. Looks like there's another conflict now 🤦

docs/known_issues.md
```diff
@@ -14,15 +14,22 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
 
 Note: For the same functionality as `turbo_frame_tag(my_model)`, use `tag.turbo_frame(id: dom_id(my_model))`.
 
-## form_for compatibility
+## Compatibility with Rails form helpers
 
 ViewComponent [isn't compatible](https://github.com/viewcomponent/view_component/issues/241) with `form_for` helpers by default.
 
-## Inconsistent controller rendering behavior between Rails versions
+Passing a form object reference (often `f`) to a ViewComponent works for simple cases, like `f.text_field :name`. Content may be ill-ordered or duplicated in complex cases, such as passing blocks to form helpers or when nesting components.
```
joelhawksley, line 21:
```suggestion
Passing a form object (often `f`) to a ViewComponent works for simple cases like `f.text_field :name`. Content may be ill-ordered or duplicated in complex cases, such as passing blocks to form helpers or when nesting components.
```

docs/known_issues.md
```diff
@@ -14,15 +14,22 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
 
-In versions of Rails < 6.1, rendering a ViewComponent from a controller doesn't include the layout.
+Some workarounds include:
 
-## Forms don't use the default form builder
+- Render an entire form within a single ViewComponent.
+- Render a [partial](https://guides.rubyonrails.org/layouts_and_rendering.html#using-partials) within the ViewComponent which includes the form.
+- Use a [custom `FormBuilder`](https://guides.rubyonrails.org/form_helpers.html#customizing-form-builders) to create reusable form components:
+  - Using FormBuilder the default way. This means to [use Action View helpers](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html) to create the desired output
```
joelhawksley, line 28:
```suggestion
  - Using FormBuilder with [Action View helpers](https://api.rubyonrails.org/classes/ActionView/Helpers/FormBuilder.html).
```

docs/known_issues.md
```diff
@@ -14,15 +14,22 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+  - Using a FormBuilder that overrides all field helpers to render a ViewComponent, so that each field can be customized individually (for example, [view_component-form](https://github.com/pantographe/view_component-form))
```
joelhawksley, line 29:
```suggestion
  - Using a FormBuilder overriding all field helpers to render a ViewComponent so each field can be customized individually (for example, [view_component-form](https://github.com/pantographe/view_component-form)).
```

docs/known_issues.md
```diff
@@ -14,15 +14,22 @@ As a workaround, use `tag.turbo_frame` instead of `turbo_frame_tag`.
...
+  - Using a lightweight implementation of ViewComponent. For example, [Primer ViewComponents](https://github.com/primer/view_components) implemented [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) which is used in the context of `FormBuilder`.
```
joelhawksley, line 30:
```suggestion
  - Using a lightweight re-implementation of ViewComponent. For example, [Primer ViewComponents](https://github.com/primer/view_components) implemented [`ActsAsComponent`](https://github.com/primer/view_components/blob/main/lib/primer/forms/acts_as_component.rb) which is used in the context of `FormBuilder`.
```

joelhawksley (approved):

PR #1646, opened 2023-02-01
Fix #1641: Conflict with private API in TestHelpers

### LINKED ISSUE #1641: Conflict with private API in TestHelpers
### Steps to reproduce

I have this RSpec test:

```ruby
RSpec.describe FooComponent do
  def controller
    FooController
  end
  
  class FooController < ::ApplicationController
    def action_name
      "index"
    end
  end

  it "renders a Tailwind page" do
    with_controller_class(controller) do
      render_inline(FooComponent.new)
      expect(page).to have_content("foo")
    end
  end
end
```

### Expected behavior
I expect the test to pass.

### Actual behavior
Running this test results in the following unexpected error:

```
NoMethodError:
 undefined method `view_context' for FooController:Class
 Did you mean?  view_context_class
# /Users/jacobcarlborg/.rvm/gems/ruby-2.7.2@auti/gems/view_component-2.82.0/lib/view_component/test_helpers.rb:57:in `render_inline'
```

The problem is that `TestHelpers` contains some private API:

https://github.com/ViewComponent/view_component/blob/9c5a9c7727b6186ece8b86fb832e7f6fd969580a/lib/view_component/test_helpers.rb#L112

And since `TestHelpers` is included in the test context, the private API is now suddenly available to tests. I unknowingly created a method, `controller`, that overwrote one of the methods part of the private API thereby breaking `TestHelpers`.

This is probably not avoidable with they way `include` works in Ruby. But perhaps the risk of causing a conflict can be reduced by prefixing all symbols (methods, instance variables, constants, etc.) with an underscore or something similar. Since this would technically be breaking change it would be nice if some form of solution could be included in ViewComponent 3.

### System configuration
**Rails version**: 6.1.5.1
**Ruby version**: 2.7.2
**Gem version**: 2.82.0

jacob-carlborg-apoex:

Reduce the risk of causing conflicts with the private API in ViewComponent tests, see #1641.

I chose to hide all existing private API behind one single method, to reduce the surface of the private API. Due to the way `include` in Ruby works, I don't think it's possible to completely fix this issue. Only reduce the risk of it occurring.

Several of the existing tests use the private API. I've only made sure the tests continue to pass without trying to remove the usage of the private API. My changes clearly highlight where the private API is used.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Fixed conflict with private API in TestHelpers ([#1646](https://github.com/ViewComponent/view_component/pull/1646))
+
+    *Jacob Carlborg*
+
 * Added example of a custom preview controller.
 
     *Graham Rogers*
--- a/lib/view_component/system_test_helpers.rb
+++ b/lib/view_component/system_test_helpers.rb
@@ -15,7 +15,7 @@ module ViewComponent
 
       file = Tempfile.new(["rendered_#{fragment.class.name}", ".html"], "tmp/view_components/")
       begin
-        file.write(controller.render_to_string(html: fragment.to_html.html_safe, layout: layout))
+        file.write(_view_component_private.controller.render_to_string(html: fragment.to_html.html_safe, layout: layout))
         file.rewind
 
         block.call("/_system_test_entrypoint?file=#{file.path.split("/").last}")
--- a/lib/view_component/test_helpers.rb
+++ b/lib/view_component/test_helpers.rb
@@ -8,7 +8,7 @@ module ViewComponent
       include Capybara::Minitest::Assertions
 
       def page
-        @page ||= Capybara::Node::Simple.new(rendered_content)
+        @page ||= Capybara::Node::Simple.new(_view_component_private.rendered_content)
       end
 
       def refute_component_rendered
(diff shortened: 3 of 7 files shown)


jacob-carlborg-apoex:
Primer is using the private API, or even worse, an instance variable: https://github.com/primer/view_components/blob/c8223c5ee56e4cac88b680cf0578fe11beaf42cc/test/components/beta/link_test.rb#L24

BlakeWilliams:
I'm not positive that we want to hide all of this state. There are use-cases where `controller` is overridden for an entire test via `def controller` which is a nice feature of the current API.

jacob-carlborg-apoex:
> I'm not positive that we want to hide all of this state

That's what I asked in the issue: https://github.com/ViewComponent/view_component/issues/1641#issuecomment-1410152534.

> There are use-cases where controller is overridden for an entire test via def controller which is a nice feature of the current API.

For that particular use case there's `with_controller_class` [1]. I think we either should hide the state or make it public (including adding documentation). Now it's too easy to break or misuse.

[1] https://viewcomponent.org/guide/testing.html#configuring-the-controller-used-in-tests.

BlakeWilliams:
> For that particular use case there's `with_controller_class` [1]. I think we either should hide the state or make it public (including adding documentation). Now it's too easy to break or misuse.

True, but that requires each render to be wrapped explicitly, which I don't think is necessarily a bad thing.

Either way this would be a breaking change, and I'm not positive that the benefits here are worth breaking the public API.

jacob-carlborg-apoex:
> Either way this would be a breaking change, and I'm not positive that the benefits here are worth breaking the public API.

That's why I wanted to get this in before v3. But @joelhawksley don't think it counts as public API: https://github.com/ViewComponent/view_component/issues/1641#issuecomment-1406845885.

joelhawksley:
Closed by #1661

PR #1650, opened 2023-02-01
Add optional compatibility module for capture/form/turbo compatibility fixes

### LINKED ISSUE #241: Strange behavior with code blocks using form helpers and Slim

### LINKED ISSUE #1099: turbo_frame_tag inside a component duplicates it
Ok, so I'm not sure if the title makes sense. Hopefully the code will.

When rendering a component which has a `turbo_frame_tag` inside it, ViewComponent (or Turbo?) will render the contents twice.
Once inside a `turbo_frame_tag` as expected, and again, with whatever's inside the `turbo_frame_tag`

Note: The content of the component html doesn't matter from my testing. I initially thought it might be another turbo frame tag or component conflicting, since my VC is quite complex.

```
- achievements.each do |achievement|
  = turbo_frame_tag(achievement.event) do
    = render(Provider::TrophyComponent.new(trophy: achievement))
```
```
= turbo_frame_tag(@trophy) do

  .cell-auto.card.--with-image data-controller='provider--trophy-component--trophy-component' id=dom_id(@trophy)
    .row.items-stretch
      .cell-auto.card class="#{ @trophy.cover.attached? ? '' : 'animate-pulse' } #{ @trophy.is_a?(XBOX::Trophy) ? 'w-60' : 'w-38' }"
        = helpers.lazy_image_tag(@trophy.cover, class: 'w-full h-full rounded-md object-cover') if @trophy.cover.attached?
```
Screenshot 2021-10-19 at 13 27 20

To fix this, I have to extract the component inside a rails partial and wrap it inside a turbo_frame_tag in there, instead of the VC.

```
- achievements.each do |achievement|

  = turbo_frame_tag(achievement.event) do
    = render(partial: 'fix', locals: { achievement: achievement }, cache: false)
```
```
= turbo_frame_tag(achievement.event) do
  = render(Provider::TrophyComponent.new(trophy: achievement))
```
### Expected behavior
The ViewComponent shouldn't render the contents twice.

### System configuration
**Rails version**:
ruby -Rails 6.1.4.1
**Ruby version**:
ruby 2.7.2p137 (2020-10-01 revision 5445e04352) [x86_64-darwin20]
**Gem version**:
  view_component (2.40.0)
  turbo-rails (7.1.0)

### LINKED ISSUE #1137: `turbo_stream` helpers not capturing correctly

### LINKED ISSUE #1384: &block rendered twice or at the wrong position from helper
I have a custom form builder but since I want to be able to use the default one as well I created helper method for it:
```
# form_helper.rb

module FormHelper
  def my_form_with(**args, &block)
    form_with(builder: MyFormBuilder, **args, &block)
  end
end
```

And when I try to use it inside any component I get two different results but neither of them is what I expect it to be. One renders the block twice - before and inside the `form` tag. The other one renders the block only before the `form` tag.

Note that when I use the `form_with` directly in the component's template it works so this has something to do with it being called from the helper.

### Steps to reproduce
Create helper that receives `&block`, pass that block to any rails helper like `form_with`, `link_to`, etc. Use the created helper in the component.

### Expected behavior
It should render the block inside the `form` tag and only once. The same way it works when used outside the component

### Actual behavior
Example 1:
```
# component.html.erb

<%= helpers.my_form_with do %>
  Example string
<% end>
```

Result:
```
Example string

<form action="..." method="post">
    Example string
</form>
```

Example 2:
```
# component.html.erb

<%= helpers.my_form_with(model: Model) do |form| %>
  <% form.text_field(:title) %>
<% end>
```

Result:
```
<input id="model_title" type="text" name="model[title]">

<form action="..." method="post">
</form>
```

### System configuration
**Rails version**: 6.1

**Ruby version**: 2.7.6

**Gem version**: 2.56.2

### LINKED ISSUE #1485: helpers.link_to with a block causes broken HTML output
`helpers.link_to` does not work with a block. Removing `helpers` prefix fixes the behavior. 

However, it is very common to use `link_to` in view helpers. A simple usage is marking the links as active or inactive based on the current URL:

```ruby
module LinksHelper
  def link_to_active(name = nil, options = {}, html_options = {}, &block)
    # For simplicity logic omitted

    link_to(name, options, html_options, &block)
  end
end
```

Usage of `link_to_active` with a block just breaks `view_component`.

### Steps to reproduce

Inside a component, render:

```
<%= helpers.link_to "/", class: "flex" do %>
  <%= helpers.heroicon :check %>
  <span class="ml-2">Link</span>
<% end %>
```

### Expected behavior

```HTML
<a class="flex" href="/">
  <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" class="h-5 w-5">
    <path fill-rule="evenodd" d="M16.707 5.293a1 1 0 010 1.414l-8 8a1 1 0 01-1.414 0l-4-4a1 1 0 011.414-1.414L8 12.586l7.293-7.293a1 1 0 011.414 0z" clip-rule="evenodd"></path>
  </svg>
  <span class="ml-2">Link</span>
</a>
```

### Actual behavior

It renders incorrect HTML:

```HTML
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" class="h-5 w-5">
  <path fill-rule="evenodd" d="M16.707 5.293a1 1 0 010 1.414l-8 8a1 1 0 01-1.414 0l-4-4a1 1 0 011.414-1.414L8 12.586l7.293-7.293a1 1 0 011.414 0z" clip-rule="evenodd"></path>
</svg>
<span class="ml-2">Link</span>
<a class="flex" href="/">
  &lt;span class="ml-2"&gt;Link&lt;/span&gt;
</a>
```

### System configuration

**Rails version**: Rails 7.0.3.1
**Ruby version**: ruby 3.1.2p20
**Gem version**: 2.66.0

### LINKED ISSUE #1607: Double Output When Rendering HAML Partials Containing Form Helpers
When using view component (ERB, HAML, whatever) to render a form,
and then rendering a partial from that component,
if that component is both written in HAML and contains a tag like `fields_for`,
the output will be doubled.

This does not happen if the partial doesn't contain such a tag, or if the partial is written in ERB, or if the partial is rendered from a normal view without VC. So, all three conditions are required.

### Steps to reproduce
Here is a link to a repository demonstrating the behavior: https://github.com/Aesthetikx/view_component_haml_partial_double

### Expected behavior
plain_form

### Actual behavior
haml_partial

### System configuration
Rails 7.0.4
Ruby 3.1.2
ViewComponent 2.79.0
HAML 6.1.1

### LINKED ISSUE #1651: duplicate elements when combining helpers and <%=
### Steps to reproduce
this component:
```erb
<div>
  <%= helpers.tag.div do %>
    content of helper and =
  <% end %>

  <% helpers.tag.div do %>
    content of helper and no =
  <% end %>

  <% tag.div do %>
    content of no helper and no =
  <% end %>

  <%= tag.div do %>
    content of no helper and =
  <% end %>
</div>
```

### Expected behavior
```html
<div>
  <div>
    content of helper and =
  </div>

  <div>
    content of no helper and =
  </div>
</div>
```

### Actual behavior
```html
<div>
  content of helper and =
  <div>
    content of helper and =
  </div>
  content of helper and no =
  <div>
    content of no helper and =
  </div>
</div>
```
its as if when we are using the `helpers` keyword, it gets rendered even if we use `<%` instead of `<%=`. And when we do user `<%=`, then it renders it in double.

i'm 99% sure its the root cause of issue https://github.com/ViewComponent/view_component/issues/1099 and https://github.com/viewcomponent/view_component/issues/241, and might be related to some other issues like https://github.com/viewcomponent/view_component/issues/241

### System configuration
**Rails version**:
Rails 7.0.3.1
**Ruby version**:
ruby 3.1.2p20
**Gem version**:
2.62.0

BlakeWilliams:
This adds an optional module, `CaptureCompatibility` which is meant to
monkey patch `ActionView::Base` so that forms, turbo frames, and
content_for, and other common Rails code works as-expected.

Close #1651, close #1384, close #1485, close #1099, close #241, close #1607, close #1137

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,6 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
+        capture_patch_enabled: [true, false]
         include:
           - rails_version: "5.2.6"
             ruby_version: "2.7"
@@ -56,6 +57,7 @@ jobs:
         RAISE_ON_WARNING: 1
         MEASURE_COVERAGE: true
         RAILS_VERSION: ${{ matrix.rails_version }}
+        CAPTURE_PATCH_ENABLED: ${{ matrix.output_buffer == 'capture_patch_enabled' && 'true' || 'false' }}
     - name: Upload coverage results
       uses: actions/upload-artifact@master
       if: always()
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,16 +26,34 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        capture_patch_enabled: [true, false]
         include:
           - rails_version: "5.2.6"
             ruby_version: "2.7"
+            capture_patch_enabled: true
           - rails_version: "6.0.4.4"
             ruby_version: "2.7"
+            capture_patch_enabled: true
           - rails_version: "6.1.4.4"
             ruby_version: "2.7"
+            capture_patch_enabled: true
           - rails_version: "7.0.2.3"
             ruby_version: "3.0"
+            capture_patch_enabled: true
+          - rails_version: "main"
+            ruby_version: "3.1"
+            capture_patch_enabled: true
+          - rails_version: "5.2.6"
+            ruby_version: "2.7"
+            capture_patch_enabled: false
+          - rails_version: "6.0.4.4"
+            ruby_version: "2.7"
+            capture_patch_enabled: false
+          - rails_version: "6.1.4.4"
+            ruby_version: "2.7"
+            capture_patch_enabled: false
+          - rails_version: "7.0.2.3"
+            ruby_version: "3.0"
+            capture_patch_enabled: false
           - rails_version: "main"
             ruby_version: "3.1"
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,33 +29,34 @@ jobs:
         include:
           - rails_version: "5.2.6"
             ruby_version: "2.7"
-            capture_patch_enabled: true
+            mode: "capture_patch_enabled"
           - rails_version: "6.0.4.4"
             ruby_version: "2.7"
-            capture_patch_enabled: true
+            mode: "capture_patch_enabled"
           - rails_version: "6.1.4.4"
             ruby_version: "2.7"
-            capture_patch_enabled: true
+            mode: "capture_patch_enabled"
           - rails_version: "7.0.2.3"
             ruby_version: "3.0"
-            capture_patch_enabled: true
+            mode: "capture_patch_enabled"
           - rails_version: "main"
             ruby_version: "3.1"
-            capture_patch_enabled: true
+            mode: "capture_patch_enabled"
           - rails_version: "5.2.6"
             ruby_version: "2.7"
-            capture_patch_enabled: false
+            mode: "capture_patch_disabled"
           - rails_version: "6.0.4.4"
             ruby_version: "2.7"
-            capture_patch_enabled: false
+            mode: "capture_patch_disabled"
           - rails_version: "6.1.4.4"
             ruby_version: "2.7"
-            capture_patch_enabled: false
+            mode: "capture_patch_disabled"
           - rails_version: "7.0.2.3"
             ruby_version: "3.0"
-            capture_patch_enabled: false
+            mode: "capture_patch_disabled"
           - rails_version: "main"
             ruby_version: "3.1"
+            mode: "capture_patch_disabled"
     steps:
     - uses: actions/checkout@master
     - name: Setup Ruby
@@ -75,7 +76,7 @@ jobs:
         RAISE_ON_WARNING: 1
         MEASURE_COVERAGE: true
         RAILS_VERSION: ${{ matrix.rails_version }}
-        CAPTURE_PATCH_ENABLED: ${{ matrix.output_buffer == 'capture_patch_enabled' && 'true' || 'false' }}
+        CAPTURE_PATCH_ENABLED: ${{ matrix.mode == 'capture_patch_enabled' && 'true' || 'false' }}
     - name: Upload coverage results
       uses: actions/upload-artifact@master
       if: always()
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add `capture_compatibility_patch_enabled` configuration option to enable an experimental patch that resolves rendering issues related to forms, capture, turbo frames, and more.
+
+    *Blake Williams*
+
 * Added example of a custom preview controller.
 
     *Graham Rogers*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,10 @@ nav_order: 5
 
     *Blake Williams*
 
+* Add `#content?` method that indicates if content has been passed to component.
+
+    *Joel Hawksley*
+
 * Added example of a custom preview controller.
 
     *Graham Rogers*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@ nav_order: 5
 
 ## main
 
-* Add `capture_compatibility_patch_enabled` configuration option to enable an experimental patch that resolves rendering issues related to forms, capture, turbo frames, and more.
+* Add experimental `config.view_component.capture_compatibility_patch_enabled` option resolving rendering issues related to forms, capture, turbo frames, etc.
 
     *Blake Williams*
 
--- a/lib/view_component/capture_compatibility.rb
+++ b/lib/view_component/capture_compatibility.rb
@@ -26,9 +26,12 @@ module ViewComponent
 
     module InstanceMethods
       def capture(*args, &block)
+        # Handle blocks that originate from C code and raise, such as `&:method`
+        return original_capture(*args, &block) if block.source_location.nil?
+
         block_context = block.binding.receiver
 
-        if block_context.class < ActionView::Base && block_context != self
+        if block_context != self && block_context.class < ActionView::Base
           block_context.original_capture(*args, &block)
         else
           original_capture(*args, &block)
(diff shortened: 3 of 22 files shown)


.github/workflows/ci.yml
```diff
@@ -56,6 +75,7 @@ jobs:
         RAISE_ON_WARNING: 1
         MEASURE_COVERAGE: true
         RAILS_VERSION: ${{ matrix.rails_version }}
+        CAPTURE_PATCH_ENABLED: ${{ matrix.output_buffer == 'capture_patch_enabled' && 'true' || 'false' }}
```
camertron, line 78: Huh this looks... weird to me. Do we want `matrix.capture_patch_enabled` instead?

lib/view_component/capture_compatibility.rb
```diff
@@ -0,0 +1,39 @@
...
+  module CaptureCompatibility
+    def self.included(base)
+      base.class_eval do
+        alias_method :original_capture, :capture
+      end
+
+      base.prepend(Methods)
+    end
+
+    module Methods
```
camertron, line 27: Can we call this module `InstanceMethods`? I think that's a little more idiomatic.

camertron (approved):

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add `capture_compatibility_patch_enabled` configuration option to enable an experimental patch that resolves rendering issues related to forms, capture, turbo frames, and more.
```
```suggestion
* Add experimental `config.view_component.capture_compatibility_patch_enabled` option resolving rendering issues related to forms, capture, turbo frames, etc.
```

.github/workflows/ci.yml
rmacklin: Yeah. With this logic, `CAPTURE_PATCH_ENABLED` is always getting set to `false`, even in the `capture_patch_enabled: true` variants of the CI matrix:
https://github.com/ViewComponent/view_component/actions/runs/4068065964/jobs/7006166117#step:5:10
which means those builds are actually skipping the tests they meant to run:
https://github.com/ViewComponent/view_component/actions/runs/4068065964/jobs/7006166117#step:5:568

With
```suggestion
        CAPTURE_PATCH_ENABLED: ${{ matrix.capture_patch_enabled }}
```
it runs as desired:
https://github.com/ViewComponent/view_component/actions/runs/4071219661/jobs/7012819664#step:5:10
https://github.com/ViewComponent/view_component/actions/runs/4071219661/jobs/7012819664#step:5:568

BlakeWilliams: Yep! Just missed it on the migration over. I wasn't a fan of the build saying `true` or `false` so I'm changing up the approach to use a string value instead.

lib/view_component/capture_compatibility.rb
BlakeWilliams: Done in https://github.com/ViewComponent/view_component/pull/1650/commits/66ec2eab0c363b7f55c0dadb5edbce0bcad2f62f

BlakeWilliams:
Alright, got the matrix build working and validated it's running the compatibility tests because the `capture_patch_enabled` builds have no skipped tests.

camertron (approved):

Spone:
@BlakeWilliams I updated the description to mention all issues (I could find) that would be solved by this PR.

BlakeWilliams:
@Spone great idea, thanks for updating it!

joelhawksley (approved):
🎊 I'll be sure to mention asking for feedback on this when I publish the next RC changelog ❤️

lib/view_component/config.rb
```diff
@@ -137,6 +138,13 @@ def defaults
       # A custom default layout used for the previews index page and individual
       # previews.
       # Defaults to `nil`. If this is falsy, `"component_preview"` is used.
+      #
+      # @!attribute capture_compatibility_patch_enabled
+      # @return [String]
```
rmacklin, line 143:
```suggestion
      # @return [Boolean]
```

BlakeWilliams: Thanks!

BlakeWilliams:
Alright, I tested this module on GitHub locally, clicked around and it all worked as expected. I also ran the tests with the patch enabled and all is 🟢. That being said, this is still experimental so please validate that it works as expected in your own applications and report any bugs you run into (with reproducible examples, please).

sdhull:
Thank you for this!! Just blundered into this particular bug and was super stoked to find all I had to do was update the gem and enable the experimental mode 🎉

marcelolx:
Thank you for this, everyone who was involved. You all rock!

PR #1652, opened 2023-02-03
Add prefix option to polymorphic slots

### LINKED ISSUE #1526: Allow more flexibility in polymorphic slot names
## Feature request

It would be nice if there was more customisation in how polymorphic slots were named. Specifically I would like to have the polymorphic slot name be a suffix instead of a prefix, but I could also imagine use cases where the polymorphic slot name isn't included in the individual slot names at all.

Something like this for example:

```ruby
class MyFormComponent < ViewComponent::Base
  renders_many :fields, suffix: true, types: {
    text: TextFieldComponent,
    email: EmailFieldComponent,
    ...
  }
end
```

### Motivation

I'm creating a form component where naming slots like `text_field` would be more consistent with Rails instead of `field_text` that is currently forced.

Spone:
Close #1526

Add support for a `prefix: true` option when declaring a polymorphic slot, that flips the order of the slot accessor methods names.

This is a first draft, in order to move the discussion in #1526 forward.

This PR introduces a breaking change to the current default behavior (put the slot name first and the type second).

Now the slot name is second, which makes more sense in most cases: `with_standard_header`, `with_text_field` etc.

Adding `prefix: true` to the slot definitions reverts to the previous behavior.

### DIFF
--- a/lib/view_component/polymorphic_slots.rb
+++ b/lib/view_component/polymorphic_slots.rb
@@ -57,7 +57,8 @@ module ViewComponent
 
           setter_name =
             if collection
-              prefix ? "#{ActiveSupport::Inflector.singularize(slot_name)}_#{poly_type}" : "#{poly_type}_#{ActiveSupport::Inflector.singularize(slot_name)}"
+              singular_slot_name = ActiveSupport::Inflector.singularize(slot_name)
+              prefix ? "#{singular_slot_name}_#{poly_type}" : "#{poly_type}_#{singular_slot_name}"
             else
               poly_slot_name
             end
--- a/lib/view_component/polymorphic_slots.rb
+++ b/lib/view_component/polymorphic_slots.rb
@@ -25,17 +25,17 @@ module ViewComponent
         return super unless callable.is_a?(Hash) && callable.key?(:types)
 
         validate_singular_slot_name(slot_name)
-        register_polymorphic_slot(slot_name, callable[:types], collection: false)
+        register_polymorphic_slot(slot_name, callable[:types], collection: false, prefix: callable[:prefix] || false)
       end
 
       def renders_many(slot_name, callable = nil)
         return super unless callable.is_a?(Hash) && callable.key?(:types)
 
         validate_plural_slot_name(slot_name)
-        register_polymorphic_slot(slot_name, callable[:types], collection: true)
+        register_polymorphic_slot(slot_name, callable[:types], collection: true, prefix: callable[:prefix] || false)
       end
 
-      def register_polymorphic_slot(slot_name, types, collection:)
+      def register_polymorphic_slot(slot_name, types, collection:, prefix:)
         unless types.empty?
           getter_name = slot_name
 
@@ -49,15 +49,17 @@ module ViewComponent
         end
 
         renderable_hash = types.each_with_object({}) do |(poly_type, poly_callable), memo|
+          poly_slot_name = prefix ? "#{slot_name}_#{poly_type}" : "#{poly_type}_#{slot_name}"
+
           memo[poly_type] = define_slot(
-            "#{slot_name}_#{poly_type}", collection: collection, callable: poly_callable
+            poly_slot_name, collection: collection, callable: poly_callable
           )
 
           setter_name =
             if collection
-              "#{ActiveSupport::Inflector.singularize(slot_name)}_#{poly_type}"
+              prefix ? "#{ActiveSupport::Inflector.singularize(slot_name)}_#{poly_type}" : "#{poly_type}_#{ActiveSupport::Inflector.singularize(slot_name)}"
             else
-              "#{slot_name}_#{poly_type}"
+              poly_slot_name
             end
 
           define_method("with_#{setter_name}") do |*args, &block|
--- a/test/sandbox/app/components/polymorphic_content_slot_component.rb
+++ b/test/sandbox/app/components/polymorphic_content_slot_component.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class PolymorphicContentSlotComponent < ViewComponent::Base
-  renders_one :leading_visual, types: {
+  renders_one :leading_visual, prefix: true, types: {
     icon: lambda { |title|
       content_tag("h1") do
         title
--- a/test/sandbox/app/components/polymorphic_slot_component.rb
+++ b/test/sandbox/app/components/polymorphic_slot_component.rb
@@ -6,7 +6,7 @@ class PolymorphicSlotComponent < ViewComponent::Base
     special: lambda { |&block| content_tag(:div, class: "special", &block) }
   }
 
-  renders_many :items, types: {
+  renders_many :items, prefix: true, types: {
     foo: "FooItem",
     bar: lambda { |class_names: "", **_system_arguments|
       classes = (class_names.split(" ") + ["bar"]).join(" ")
--- a/test/sandbox/test/slotable_v2_test.rb
+++ b/test/sandbox/test/slotable_v2_test.rb
@@ -492,7 +492,7 @@ class SlotsV2sTest < ViewComponent::TestCase
 
   def test_polymorphic_slot_with_setters
     render_inline(PolymorphicSlotComponent.new) do |component|
-      component.with_header_standard { "standard" }
+      component.with_standard_header { "standard" }
       component.with_item_foo(class_names: "custom-foo")
       component.with_item_bar(class_names: "custom-bar")
     end
@@ -538,7 +538,7 @@ class SlotsV2sTest < ViewComponent::TestCase
 
   def test_polymorphic_slot
     render_inline(PolymorphicSlotComponent.new) do |component|
-      component.with_header_standard { "standard" }
+      component.with_standard_header { "standard" }
       component.with_item_foo(class_names: "custom-foo")
       component.with_item_bar(class_names: "custom-bar")
     end
@@ -559,8 +559,8 @@ class SlotsV2sTest < ViewComponent::TestCase
   def test_singular_polymorphic_slot_raises_on_redefinition
     error = assert_raises ArgumentError do
       render_inline(PolymorphicSlotComponent.new) do |component|
-        component.with_header_standard { "standard" }
-        component.with_header_special { "special" }
+        component.with_standard_header { "standard" }
+        component.with_special_header { "special" }
       end
     end
 


camertron:
Awesome, thanks for this @Spone!

> This PR introduces a breaking change to the current default behavior (put the slot name first and the type second).

I'm hesitant about this. Can we keep the default behavior and change the parameter to `suffix: true`?

camertron:
> What if we keep the current behavior as the default and introduce a config option to redefine the default value application-wide?

Hmm... yeah I'm ok with that 👍

joelhawksley:
> What if we keep the current behavior as the default and introduce a config option to redefine the default value application-wide?

I'm not so sure. How would this work in codebases that mix application components with those from a gem? I could see that being confusing.

Spone:
Yeah I gave it more thought and I'm finally convinced adding a config option would add too much complexity.

But I'm not sure the `prefix: true` option has any interest if it's not the default. 🤔

pfeiffer:
I think this is a great addition and much needed for better readability. In our specific app case, the method names generated currently make little sense and seem 'off'.

> Yeah I gave it more thought and I'm finally convinced adding a config option would add too much complexity.
> 
> But I'm not sure the `prefix` option has any interest if `false` is not the default. 🤔

How about adding a config option that, when set, would generate _both_ the prefixed and the suffixed versions? This would allow first-party components to use the new suffixed slot method names, while keeping compatibility with gems that use the prefixed method names.

It would also provide a good deprecation path if needed in a new major release, as deprecation warnings could be added without breaking functionality.

Another approach could be adding the option as a class option on the component, similar to [strip_trailing_whitespace](https://viewcomponent.org/api.html#strip_trailing_whitespacevalue--true), ie:

```ruby
class MyComponent < ..
  prefix_polymorphic_slot_names false
end
```

.. which would of course be inheritable, so setting it for your app-wide `ApplicationComponent` (or whatever) would apply it to all inheriting components per default. Combined with the `prefix: false` option for the individual slot definitions would allow to mix and match even in the same component to build the method names that makes most sense.

Spone:
Alright, I'm aligned with you @BlakeWilliams!

Any suggestions for tooling? I also wanted to investigate creating a codemod for easing migration to v3.0.0's `#with_[slot]` syntax.

I'm a bit more inclined to use something focused on modifying code, not just linting (so maybe not rubocop).

I've identified [codeshift](https://github.com/rajasegar/codeshift) / [cybertron](https://github.com/rajasegar/cybertron), but if the rules are relatively simple it may also be a plain Ruby script, or based on Thor?

BlakeWilliams:
> I'm a bit more inclined to use something focused on modifying code, not just linting (so maybe not rubocop).
> 
> I've identified [codeshift](https://github.com/rajasegar/codeshift) / [cybertron](https://github.com/rajasegar/cybertron), but if the rules are relatively simple it may also be a plain Ruby script, or based on Thor?

I haven't used either of those, so I can't provide any thoughts there. I was thinking Rubocop since it's got great parsing/rewriting functionality built-in and most codebases already have it installed. I have no strong feelings either way though.

joelhawksley:
@Spone thanks for opening this PR to prompt further discussion!

Giving this some more thought, I think we need to go all-in on this change if we're going to do it. I'd vote we go for a similar approach to the `with_*` API migration, adding something like a `warn_on_deprecated_polymorphic_slot_setter` macro and change the API to be the suggested prefix instead of suffix. 

I think we could do this with a two-week cycle from deprecation to removal as Polymorphic Slots are pretty niche feature.

What do y'all think?

joelhawksley:
In our office hours today, @camertron encouraged us to consider using an `as:` keyword to configure the name for each polymorphic slot type.

I think that would be a good way forward without making breaking changes. What do you think @Spone?

Spone:
> Giving this some more thought, I think we need to go all-in on this change if we're going to do it. I'd vote we go for a similar approach to the `with_*` API migration, adding something like a `warn_on_deprecated_polymorphic_slot_setter` macro and change the API to be the suggested prefix instead of suffix.
> 
> I think we could do this with a two-week cycle from deprecation to removal as Polymorphic Slots are pretty niche feature.

I agree that we should aim at replacing the current default, but I think it should be considered as a breaking change, even if Polymorphic Slots can be viewed as a niche feature. So maybe deprecate now and remove for v4.0?

> In our office hours today, @camertron encouraged us to consider using an `as:` keyword to configure the name for each polymorphic slot type.
> 
> I think that would be a good way forward without making breaking changes. What do you think @Spone?

Can you elaborate on the syntax? Where would you add the `as` keyword in this example?

```rb
renders_many :fields, types: {
  text: TextFieldComponent,
  email: EmailFieldComponent,
}
```

joelhawksley:
@camertron what was your proposed `as` API?

camertron:
I had two proposals. The first was to turn each value in the `types` hash into hashes themselves:

```ruby
renders_many :fields, types: {
  text: { renderable: TextFieldComponent, as: :text_field },
  email: { renderable: EmailFieldComponent, as: email_field },
}
```

However that's duplicating a bit of info, so I also suggested an array-based approach:

```ruby
renders_many :fields, types: [
  { renderable: TextFieldComponent, as: :text_field },
  { renderable: EmailFieldComponent, as: :email_field },
]
```

Spone:
Okay it's similar to what we discussed in the issue (https://github.com/ViewComponent/view_component/issues/1526#issuecomment-1380821018) but I like the array-based approach with the `as` keyword better!
I prefer the explicit way of naming the setter, instead of having to reason about prefixes / suffixes.

My concerns:
- we would have to support both syntaxes (`types` as Hash + as array) until the former one is removed
- it doesn't globally solve the issue of the current setter naming which is a bit backwards: you'd still have to update each component class + all occurrences of slot setters

or maybe I'm mistaken?

camertron:
> I prefer the explicit way of naming the setter, instead of having to reason about prefixes / suffixes.

Yeah me too :)

> we would have to support both syntaxes (`types` as Hash + as array) until the former one is removed

Right, which is not _great_ but also not a huge deal imo. We could also simply leave in the hash functionality indefinitely.

> it doesn't globally solve the issue of the current setter naming which is a bit backwards: you'd still have to update each component class + all occurrences of slot setters

Correct, slot setters would still be prefixed with `with_`, meaning a bunch of tedious updates in app code. I don't think that's something we can avoid though.

Spone:
> Right, which is not _great_ but also not a huge deal imo. We could also simply leave in the hash functionality indefinitely.

Agreed. It will be easy to differentiate between the two.

> Correct, slot setters would still be prefixed with `with_`, meaning a bunch of tedious updates in app code. I don't think that's something we can avoid though.

Yeah, at least that's something that you can do gradually if both syntaxes coexist.

Just to make sure I got it right, this slot:

```rb
renders_many :fields, types: [
  { renderable: TextFieldComponent, as: :text_input },
  { renderable: EmailFieldComponent, as: :email_input },
]
```

would be used like this:

```rb
render(MyComponent.new) do [c[
  c.with_text_input { ... }
  c.with_email_input { ... }
end
```

so you could potentially use `text_input` as a polymorphic slot name, that's totally unrelated to the `fields` slot name.

We will need to check the polymorphic names against other existing slots to prevent collisions.

allmarkedup:
Thanks for your work on this all :-) Just a really minor thing to bring up but I feel like `renders` is potentially a bit nicer than `renderable`? It makes it feel more declarative and reads better, IMO:

```
renders_many :fields, types: [
  { renders: TextFieldComponent, as: :text_input },
  { renders: EmailFieldComponent, as: :email_input },
]
```

Just my 2 cents. Feel free to ignore if you disagree!

Spone:
@allmarkedup I think the idea is to refer to [`ActionView::Template::Renderable`](https://github.com/rails/rails/blob/e78ed07e008676752b2ed2cff97e74b31ffacaf5/actionview/lib/action_view/template/renderable.rb), but I tend to agree with you that `renders`, or maybe `component` reads better.

camertron:
> so you could potentially use `text_input` as a polymorphic slot name, that's totally unrelated to the `fields` slot name.

Yes, that's right. It sort of obviates the `fields` slot name, which I'm not super happy about, but I think that's unavoidable.

> We will need to check the polymorphic names against other existing slots to prevent collisions.

Oh yeah that's a good idea 👍 

> Just a really minor thing to bring up but I feel like `renders` is potentially a bit nicer than `renderable`?

As @Spone pointed out, I was referencing `ActionView::Template::Renderable`, but I suppose a "renderable" isn't mentioned anywhere in the ViewComponent API, so `renders` is better 👍

Spone:
Okay I think I've got enough info to rework the PR to implement the "array of Hashes" approach, if everybody's on board with this.

camertron:
@Spone I say go for it! I've actually needed this feature twice in the last week, so I'm looking forward to seeing it land!

joelhawksley:
> Yes, that's right. It sort of obviates the fields slot name, which I'm not super happy about, but I think that's unavoidable.

@camertron wouldn't `fields` still be used to name the accessor?

camertron:
@joelhawksley oh that's true, I hadn't thought of that. Nice!

camertron:
Hey everyone 👋  I've decided to take a crack at implementing this after our discussion during office hours yesterday. After exploring the problem a bit more, I actually think the hash-of-hashes approach is better, eg:

```ruby
renders_many :fields, types: {
  text: { renders: TextFieldComponent, as: :text_field },
  email: { renders: EmailFieldComponent, as: :email_field },
}
```

The hash-of-hashes approach has the following benefits:

1. It's closer to the existing API and will cause IMHO less confusion for consumers. If we go with the array approach, we'll have to document and support both. I'm worried about people's potential confusion around which data structure to use.
2. It allows the framework to fall back to the original setter name if no `as:` argument is provided. The array approach does not offer this because the "poly name," i.e. "text" and "email" in the example above, are not available.
3. It allows consumers to override setters for individual polymorphic types. There are scenarios where the original setter name is perfectly acceptable and scenarios where it makes no linguistic sense. The array approach forces consumers to specify a setter name for every polymorphic type, even if the original (computed) name is ok.
4. It maintains the original syntax for use-cases the original setter names are well-suited for. Even though we would support both an array-of-hashes and a hash-of-hashes for the foreseeable future, I expect we'd converge on one approach and remove the other in the next major release of the framework. Using a hash instead of an array is more backwards-compatible and prevents this sort of fracturing.
5. It's simpler to implement. This is the least important reason, but I found the code was easier to understand for the hash approach because `types` is always a hash.

Here's the draft PR: https://github.com/ViewComponent/view_component/pull/1740

Spone:
Closing in favor of #1740

PR #1656, opened 2023-02-07
Expose translations on the component class

elia:

Ref: https://github.com/ViewComponent/view_component/discussions/1648#discussioncomment-4843103

We want to expose translations to the outside world with class methods:

```rb
MyComponent.compile
MyComponent.translate ".my_translation"
```

Did not go for feature parity between the instance translation methods and the class ones as there is a number of fallbacks to `super` on the instance that couldn't work in the same way on the class.

Expecting the class methods won't be used within view I left the support for html_safe only in the instance methods.

I moved one method to be private, doesn't look like a concern for backwards compatibility and semver to me, but open to extracting that to a different PR if that's the case.


lib/view_component/translatable.rb
```diff
@@ -32,6 +32,25 @@ def build_i18n_backend
...
+
+      def i18n_key(key, scope = nil)
+        scope = scope.join(".") if scope.is_a? Array
+        key = key&.to_s unless key.is_a?(String)
+        key = "#{scope}.#{key}" if scope
+        key = "#{i18n_scope}#{key}" if key.start_with?(".")
+        key
+      end
+
+      def translate(key = nil, **options)
```
joelhawksley, line 44: What do you think about having this method compile the component if it hasn't been compiled? That way we wouldn't have to expect folks to call `compile` first.

PR #1657, opened 2023-02-10
Remove `evironment` dependency of rake task `view_component:statsetup`

svetlins:

Running `rake stats` in a Rails app that's using `view_component` may fail if the app is not bootable in the current environment (e.g. CI). That differs from the behavior of `rake stats` if `view_component` is not installed

The reason is that the gem hooks its own setup task as a dependency of the `stats` task and prepended task itself depends on `environment`, thus introducing `environment` as a transitive dependency of `stats`.

What's more it doesn't seem there's a reason for adding this dependency (at least I can't see one).

My goal is to restore the behavior of stats being runnable even if the app is not bootable.

The specific problem we hit arising from this is that we run `rake stats` as CI step which started breaking once we introduced `view_component` because we have initializers that depend on Rails credentials for which we don't setup a key on the CI. This is also a problem of our code base in itself as depending on credentials in initializers is wrong but that's a separate issue :)

I think this should be fixed as I suspect other people run into it and the errors arising from this are completely dependent on the actual details of the app and can be quite mysterious. There's also a (very small) performance overhead.

I simply changed the `view_component:statssetup` task to not depend on `environment`. I added a test case even though I'm not sure adding is the correct trade-off as it has some side effects and doesn't test a whole lot.

* I would kindly ask if anybody has knowledge of a reason for `view_component:statssetup` depending on `environment` that I have missed.
* I would like feedback on the viability of the added test case - it calls `.load_tasks` and invokes the task itself which modifies a top level constant both of which have side effects and might have unintended consequences in the future.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@ nav_order: 5
 
 ## main
 
-* Remove dependency on environment task from view_component:statsetup
+* Remove dependency on environment task from `view_component:statsetup`.
 
     *Svetlin Simonyan*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,7 @@ nav_order: 5
 # Changelog
 
 ## main
+
 * Remove dependency on environment task from view_component:statsetup
 
     *Svetlin Simonyan*
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -9,6 +9,9 @@ nav_order: 5
 # Changelog
 
 ## main
+* Remove dependency on environment task from view_component:statsetup
+
+    *Svetlin Simonyan*
 
 * Add experimental `config.view_component.capture_compatibility_patch_enabled` option resolving rendering issues related to forms, capture, turbo frames, etc.
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -206,6 +206,7 @@ ViewComponent is built by over a hundred members of the community, including:
 <img src="https://avatars.githubusercontent.com/matheuspolicamilo?s=64" alt="matheuspolicamilo" width="32" />
 <img src="https://avatars.githubusercontent.com/erinnachen?s=64" alt="erinnachen" width="32" />
 <img src="https://avatars.githubusercontent.com/ihollander?s=64" alt="ihollander" width="32" />
+<img src="https://avatars.githubusercontent.com/svetlins?s=64" alt="svetlins" width="32" />
 
 ## Who uses ViewComponent?
 
--- a/lib/view_component/rails/tasks/view_component.rake
+++ b/lib/view_component/rails/tasks/view_component.rake
@@ -3,7 +3,7 @@
 task stats: "view_component:statsetup"
 
 namespace :view_component do
-  task statsetup: :environment do
+  task :statsetup do
     require "rails/code_statistics"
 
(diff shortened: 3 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Remove dependency on environment task from view_component:statsetup
```
```suggestion
* Remove dependency on environment task from `view_component:statsetup`.
```

joelhawksley (approved):

eneagoe:
This changeset seems to have broken `rake stats`:

```
eugens-MacBook-Pro-3:stats-tester (main)$ rake stats
rake aborted!
<internal:dir>:15:in `open'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/railties-7.0.5/lib/rails/code_statistics.rb:46:in `foreach'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/railties-7.0.5/lib/rails/code_statistics.rb:46:in `calculate_directory_statistics'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/railties-7.0.5/lib/rails/code_statistics.rb:40:in `block in calculate_statistics'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/railties-7.0.5/lib/rails/code_statistics.rb:40:in `map'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/railties-7.0.5/lib/rails/code_statistics.rb:40:in `calculate_statistics'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/railties-7.0.5/lib/rails/code_statistics.rb:21:in `initialize'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/railties-7.0.5/lib/rails/tasks/statistics.rake:36:in `new'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/railties-7.0.5/lib/rails/tasks/statistics.rake:36:in `block in <main>'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/gems/rake-13.0.6/exe/rake:27:in `<top (required)>'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/bin/ruby_executable_hooks:22:in `eval'
/Users/eneagoe/.rvm/gems/ruby-3.0.4/bin/ruby_executable_hooks:22:in `<main>'
Tasks: TOP => stats
(See full trace by running task with --trace)
```

also mentioned at https://github.com/ViewComponent/view_component/commit/5b8fb263e2a6e41d008011cbee651cf6a3a9293f#r116718974

svetlins:
@eneagoe I ran a bisect between `main` and this commit, and the first offending commit is 01510f331fc976288569ab9d9c1644f08f035558. Specifically [this change](https://github.com/ViewComponent/view_component/commit/01510f331fc976288569ab9d9c1644f08f035558#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9eR24) seems to affect it.

I don't understand the change introduced by the commit well enough to fix that though. @camertron maybe you can take a look?

As for the change introduced here - the idea is that the usual `stats` task in Rails does not execute `environment` and the goal was to preserve this. If that proves to be infeasible I think this can be reverted - it won't be a huge deal.

eneagoe:
Thank you for looking into it - I'm not familiar enough with the codebase to figure out why that commit broke `rake stats`. It's probably a very low priority, but it would be nice if it'd be fixed at some point.

PR #1661, opened 2023-02-17
Rename private TestHelpers methods

### LINKED ISSUE #1641: Conflict with private API in TestHelpers

joelhawksley:
In #1641, @jacob-carlborg-apoex provided a test case showing how our private #controller method can create confusion when writing tests. This PR renames our private methods to avoid such conflicts. I've opted for a narrower solution than the one proposed in #1646.

Closes #1641, #1646

### DIFF
--- a/test/sandbox/test/integration_test.rb
+++ b/test/sandbox/test/integration_test.rb
@@ -214,7 +214,7 @@ class IntegrationTest < ActionDispatch::IntegrationTest
     assert_response :success
     assert_includes response.body, "Rendered"
 
-    cookies[:shown] = true
+    cookies[:hide] = true
 
     get "/render_check"
     assert_response :success
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,6 +18,10 @@ nav_order: 5
 
     *Joel Hawksley*
 
+* BREAKING: Rename private TestHelpers#controller, #build_controller, #request, and #preview_class to avoid conflicts. Note: While these methods were undocumented and marked as private, they was easily accessible in tests. As such, we're cautiously considering this to be a breaking change.
+
+    *Joel Hawksley*
+
 * Avoid loading ActionView::Base during Rails initialization. Originally submitted in #1528.
 
     *Jonathan del Strother*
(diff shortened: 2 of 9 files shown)


test/sandbox/test/integration_test.rb
```diff
@@ -214,7 +214,7 @@ def test_optional_rendering_component_depending_on_request_context
     assert_response :success
     assert_includes response.body, "Rendered"
 
-    cookies[:shown] = true
```
camertron, line 217: Wait, isn't "shown" the opposite of "hide"? It would make sense if you were changing `shown == false` to `hide == true`, but...

camertron (commented):
Seems ok to me, just wondering about the tests you changed.

test/sandbox/test/integration_test.rb
joelhawksley: In this case, it's not the opposite, which I too found confusing. It's because it's the _past_ tense of `show`. That's why I changed this to `hide` for clarity.

PR #1664, opened 2023-02-20
Protect against Arbitrary File Read in System Test Controller

nickmalcolm:

Resolves https://github.com/ViewComponent/view_component/issues/1663. When a file path is passed to `ViewComponentsSystemTestController` it will ensure that the path exists and is within `/tmp/view_components/`. It will also raise an error if run outside of a test environment.

One thing I can't quickly figure out is how to get the test to pass properly 😅 It raises the error as expected, but I think the Rails App is catching the error instead of passing it to Minitest. It says nothing was raised ... then says the test failed because an error was raised.

```
Failure:
ViewComponentSystemTest#test_path_traversal_raises_error:
ArgumentError expected but nothing was raised.

Error:
ViewComponentSystemTest#test_path_traversal_raises_error:
ArgumentError: Invalid file path
```

See https://github.com/ViewComponent/view_component/issues/1663 for discussion.

Assistance with getting the test to pass, please.

### DIFF
--- a/app/controllers/view_components_system_test_controller.rb
+++ b/app/controllers/view_components_system_test_controller.rb
@@ -19,7 +19,7 @@ class ViewComponentsSystemTestController < ActionController::Base # :nodoc:
   # Ensure that the file path is valid and doesn't target files outside
   # the expected directory (e.g. via a path traversal or symlink attack)
   def validate_file_path
-    base_path = ::File.realpath(TEMP_DIR).freeze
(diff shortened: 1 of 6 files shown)


nickmalcolm:
@joelhawksley can you assist with this MR, please? 🙇

BlakeWilliams (approved):
LGTM, thanks for fixing up those concerns!

PR #1666, opened 2023-02-22
Raise error if translations are used in initializer

### LINKED ISSUE #1577: Relative translations don't work in initializer

joelhawksley:
After exploring a potential solution in https://github.com/ViewComponent/view_component/pull/1617, this PR raises a clear error that guides folks towards our recommended approach.

Closes #1577

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -31,13 +31,13 @@ module ViewComponent
       attr_writer :config
     end
 
+    ViewContextCalledBeforeRenderError = Class.new(StandardError)
+
     include ViewComponent::ContentAreas
     include ViewComponent::PolymorphicSlots
     include ViewComponent::SlotableV2
     include ViewComponent::Translatable
     include ViewComponent::WithContentHelper
-
-    ViewContextCalledBeforeRenderError = Class.new(StandardError)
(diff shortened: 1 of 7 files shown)


lib/view_component/base.rb
```diff
@@ -28,12 +28,12 @@ def config
       attr_writer :config
     end
 
+    ViewContextCalledBeforeRenderError = Class.new(StandardError)
```
camertron, line 31: Why not just

```ruby
class ViewContextCalledBeforeRenderError < StandardError; end
```

camertron (approved):
Makes sense to me 👍 Same thing we do when you try to call other helpers.

lib/view_component/base.rb
joelhawksley:
```suggestion
    class ViewContextCalledBeforeRenderError < StandardError; end
```

PR #1668, opened 2023-02-26
Fix bug where content? and with_content did not work reliably for slots

dkniffin:
_I've taken over this PR - @joelhawksley_

This PR fixes the behavior of `content?` and `with_content` to work with slots as reported by @dkniffin.


joelhawksley:
Thanks for capturing this behavior @dkniffin ❤️ 

@BlakeWilliams as our resident Slots expert, what are your thoughts on this failure mode?

lib/view_component/slot.rb
```diff
@@ -12,6 +12,24 @@ def initialize(parent)
       @parent = parent
     end
 
+    def content?
+      if (defined?(@__vc_content) && @__vc_content.present?) ||
+          (defined?(@__vc_content_set_by_with_content) && @__vc_content_set_by_with_content.present?) ||
+          (defined?(@__vc_content_block) && @__vc_content_block.present?)
+        return true
+      end
```
BlakeWilliams, line 20:
```suggestion
      return true if defined?(@__vc_content) && @__vc_content.present?
      return true if defined?(@__vc_content_set_by_with_content) && @__vc_content_set_by_with_content.present?
      return true if defined?(@__vc_content_block) && @__vc_content_block.present?
```

What are your thoughts on converting these conditionals into a guard clause to help separate logic and highlight the intention?

lib/view_component/slot.rb
```diff
@@ -12,6 +12,24 @@ def initialize(parent)
...
+
+      return @__vc_component_instance.content? if __vc_component_instance?
```
BlakeWilliams, line 22:
```suggestion
      return false if !__vc_component_instance?

      @__vc_component_instance.content?
```

Can we make sure this always returns a `true` or `false` value and not `nil`?

I think with the above guard clause change, this would have a pretty clear pattern of success/failure cases.

BlakeWilliams (approved):
Left some minor feedback, but overall LGTM

PR #1669, opened 2023-02-28
V3SlotSetters linter/codemod draft

Spone:

I'd like to provide a linter / codemod to ease the upgrade to ViewComponent v3.0.0, especially the new slot setters syntax which is in my opinion the main hassle when upgrading.

For now, this draft only outputs suggestions of changes. If we're confident enough we don't have false positives, we can change it to actually modify the files (or allow running it as a linter or as a codemod).

I first tried to experiment with Ruby AST / Parser-based solutions, but given that we have to access the slot names in ViewComponent classes **and** parse the content of various templates (erb, slim, haml), it did not appear as a viable solution.

Then, I implemented a two-steps approach:

1. Load all ViewComponents and read the name of their slots;
1. Try to find **exact matches**: scan all components and view files to find `render`s of the components, called with blocks, and methods matching the name of their slots' setter methods. These suggestions are prefixed with `probably` in the output ;
1. In some cases (such as when using helpers to render components instead of `render(MyComponent.new)`), this was not enough to find all occurrences so I added another "**uncertain matches**" step: try to match the name of the slots in all files. These suggestions are prefixed with `maybe` in the output.

I tested this on several projects, here is the output:
- [lookbook v1.4.5](https://github.com/ViewComponent/lookbook/tree/v1.4.5) which is the latest version without ViewComponent v3 support: [view output](https://gist.github.com/Spone/eb5e951b4cde53419527684d83ddc524#file-lookbook-output-txt) and compare with the changes made manually:
  - https://github.com/ViewComponent/lookbook/commit/f3cffecd9a317efba0e54bf5c10b27cba3fa7757
  - https://github.com/ViewComponent/lookbook/pull/333/commits/2254901cb38c96bff0506eeef056c7b68f113459
- [polaris_view_components](https://github.com/baoagency/polaris_view_components) which does not support ViewComponent v3 yet: [view output](https://gist.github.com/Spone/eb5e951b4cde53419527684d83ddc524#file-polaris-output-txt)
- [avo](https://github.com/avo-hq/avo) which does not support ViewComponent v3 yet: [view output](https://gist.github.com/Spone/eb5e951b4cde53419527684d83ddc524#file-avo-output-txt)
- a few internal projects I'm working on (I can't share the output infortunately)

I packaged this as a class that could be included to the gem.
The exact way people can run this remains to be defined.

We also have to figure a way to test this :)

### What's left to do

- [ ] fix [`render_match`](https://github.com/ViewComponent/view_component/pull/1669#discussion_r1180938848)
- [ ] [autodetect view paths](https://github.com/ViewComponent/view_component/pull/1669#discussion_r1132338978)
- [ ] improve uncertain matches by targeting full slot names instead of partial. This should reduce false positive results
- [x] add a rake task that will run codemod
- [x] make it accessible in rails console too
- [ ] add a "--fix" mode that modifies the files (similar to `standardrb --fix` or `rubocop -A`)
- [ ] tests
- [ ] docs
- [ ] changelog


camertron:
Very cool, thanks for taking this on! Before I really dive in I have a few questions.

1. Have you had a look at [erb-lint](https://github.com/Shopify/erb-lint)? It can parse ERB and I think uses the Parser gem under the hood. Might be a viable way to parse and manipulate ERB code.
2. I know we probably want this codemod to work statically, but I'll bet we could identify slot getters/setters really accurately by setting a tracepoint and rendering all the previews + running all the tests. What do you think?

Spone:
>     1. Have you had a look at [erb-lint](https://github.com/Shopify/erb-lint)? It can parse ERB and I think uses the Parser gem under the hood. Might be a viable way to parse and manipulate ERB code.

I considered it but I wanted something that would work for any templating language (or at least the most popular ones: ERB, Slim, Haml), since I use Slim in most projects.
 
>     2. I know we probably want this codemod to work statically, but I'll bet we could identify slot getters/setters really accurately by setting a tracepoint and rendering all the previews + running all the tests. What do you think?

That would be an interesting approach, but it would only work on thoroughly previewed / tested components, which is sadly not always the case :upside_down_face: 
I'd love to see what this would look like though! Maybe we can add this as another scanning step.

allmarkedup:
@Spone this is fantastic! Really great work mate. I have been putting off the big upgrade job at work but this looks like it'll get me a lot closer. Very exciting, I'm sure that the slot name changes are a blocker for quite a few people so this looks like a great solution, even if it only ended up getting 95% of the way there.

I'm off for a couple of days but will give it a test on our codebase when I'm back. I think I'll probably still try to combine it with some sort of monkey-patched ViewComponent gem to hopefully log anywhere that old-style slot names are still being used after the codemod has run, just to be sure.

boardfish:
I'm also planning to test this on our migration to ViewComponent 3 once I get around to it. Thanks for putting it together!

I really like providing codemods as a pattern to help folks along their migration path. We could tuck this inside a generator that helps folks to handle migrations, similar to Rails' own `app:update` task.

joelhawksley:
@Spone I like it! Are you hoping to land this in a Release Candidate ahead of 3.0.0 landing?

Spone:
@joelhawksley I'd love to, if I get sufficient feedback from people who tried it!

I ran this on more codebases, and noticed a few false-positives that I'd like to fix as well.

lib/view_component/codemods/v3_slot_setters.rb
```diff
@@ -0,0 +1,183 @@
+# Usage (in rails console):
+#
+# Make sure all your ViewComponents are loaded before running the codemod.
+#
+#     Zeitwerk::Loader.eager_load_all
```
joelhawksley, line 5: Would it make sense to add this to the script itself?

lib/view_component/codemods/v3_slot_setters.rb
```diff
@@ -0,0 +1,183 @@
...
+#
+# Run the codemod:
+#
+#     ViewComponent::Codemods::V3SlotSetters.new.call
+#
+# If your app uses custom paths for views, you can pass them in:
```
joelhawksley, line 11: Could we load these from the rails config instead of requiring them to be passed in?

lib/view_component/codemods/v3_slot_setters.rb
```diff
@@ -0,0 +1,183 @@
...
+      def initialize(view_component_path: [], view_path: [])
+        @view_component_path = view_component_path
+        @view_path = view_path
+      end
+
+      def call
+        puts "Using ViewComponent path: #{view_component_paths.join(", ")}"
+        puts "Using Views path: #{view_paths.join(", ")}"
+        puts "#{view_components.size} ViewComponents found"
+        puts "#{slottable_components.size} ViewComponents using slots found"
```
joelhawksley, line 34:
```suggestion
        puts "#{slottable_components.size} ViewComponents using Slots found"
```

lib/view_component/codemods/v3_slot_setters.rb
```diff
@@ -0,0 +1,183 @@
...
+        puts "#{view_component_files.size} view component files found"
```
joelhawksley, line 35:
```suggestion
        puts "#{view_component_files.size} ViewComponent templates found"
```

joelhawksley (approved):
I think others will find this tool helpful. Would it be worth writing a simple test for it? We don't need to be too thorough.

joelhawksley:
@Spone I'm hoping to wrap up the v3 release cycle in the next week or two, so perhaps you could decide whether to land this in the next week?

boardfish:
Ran this against our codebase @fac. I'd already gone through and made all the necessary changes previously on another branch.

When I ran this before I had updated the slot syntax, it came up with a lot of `maybe` cases. Enough of these were false positives that I can say this was quite noisy for my use case. In particular, it caught a lot of method names that begin with the names of slots that exist on components. For example, it would catch `total_somethings` where there was a component with a slot named `total`. It'd be good to rework that side of things and perhaps also make it easier to filter the output.

When I ran this after I had updated the slot syntax, it didn't turn up any cases at all, which I'm hoping suggests that I had found everything that needed to be fixed.

Ideally, I think it would be really good to use `erb-lint` and `rubocop` to do this, with a view to making it possible to autocorrect these with confidence. That being said, there are definitely "maybe" cases, especially in cases where folks have components aliased to helpers.

Spone:
@joelhawksley Sorry I missed your previous comment. I'd like to continue improving this but I cannot commit to a specific time frame. I feel like this can be part of an upcoming v3.x minor version release. What do you think?

Spone:
I'd be open to some pairing to wrap this up!

lib/view_component/codemods/v3_slot_setters.rb
```diff
@@ -0,0 +1,180 @@
...
+      end
+
+      private
+
+      def scan_exact_matches(file)
+        [].tap do |suggestions|
+          rendered_components = []
+          content = File.read(file)
+
+          if (render_match = content.match(RENDER_REGEX))
```
owaiswiz, line 65: Is this assuming only a single render_match per file?

If I render multiple different components in the same file, would this catch it?

kirillplatonov:
@Spone what's left here in order to get this ready for merge? Some suggestions I have are:

- Improve uncertain matches by targeting full slot names instead of partial. This should reduce false positive results
- Add a rake task that will run codemod
- Make it accessible in rails console too

Spone:
@kirillplatonov I added a "What's left to do" section to the PR description, including your suggestions. Would you like to take care of some of these items?

kirillplatonov:
@Spone sure! Should I open a PR with my changes to `codemods/v3-slots` branch?

Spone:
Yes please do!

joelhawksley (approved):
Might we add something to the docs for this?

Spone:
> Might we add something to the docs for this?

Yes, but not sure where. We don't yet have a "Upgrading" page.

Also, we need to fix the changelog, the entry has to be under the `main` heading.

Twistedben:
Hey guys! 

I started working on a similar solution last week and had no idea! At eXp-Realty we use a lot of View Components and have 10+ production apps with them in them. I built a Ruby script to automatically update these slot calls with the `with_` change. It's amazing how similar our solutions are though I think yours is better. I just wanted to add to the discussion and my script file if it can help at all. In one of our apps that I used it, it correctly updated over 327 slots. 

https://github.com/Twistedben/view_component_3_slot_upgrade

camertron:
Hey @Spone, how's this PR coming along? Is it ready to merge?

Spone:
Haven't had time to look at this again before holidays, I'll get back to it when I'm back to work. Thanks everyone for your patience 🙏

JWShuff:
> Hey guys!
> 
> I started working on a similar solution last week and had no idea! At eXp-Realty we use a lot of View Components and have 10+ production apps with them in them. I built a Ruby script to automatically update these slot calls with the `with_` change. It's amazing how similar our solutions are though I think yours is better. I just wanted to add to the discussion and my script file if it can help at all. In one of our apps that I used it, it correctly updated over 327 slots.
> 
> https://github.com/Twistedben/view_component_3_slot_upgrade

Just used that script to great effect, TY! Noted a small issue with encoding and have opened a PR against your repo if you'd like.

joelhawksley:
@Spone how do you think we should proceed with this PR?

Twistedben:
> > Hey guys!
> > I started working on a similar solution last week and had no idea! At eXp-Realty we use a lot of View Components and have 10+ production apps with them in them. I built a Ruby script to automatically update these slot calls with the `with_` change. It's amazing how similar our solutions are though I think yours is better. I just wanted to add to the discussion and my script file if it can help at all. In one of our apps that I used it, it correctly updated over 327 slots.
> > https://github.com/Twistedben/view_component_3_slot_upgrade
> 
> Just used that script to great effect, TY! Noted a small issue with encoding and have opened a PR against your repo if you'd like.

Thanks @JWShuff ! I'm glad it worked well for you. I merged your PR on my script

reeganviljoen:
@Spone I would really love to see this go through, if you need my help in getting it over the line please ask me as I would love to help if its needed

joelhawksley:
Closing as stale, especially given our path towards a `v4` release: https://github.com/ViewComponent/view_component/pull/2085

Issue #1670, opened 2023-02-28
Using Components with Multipart Emails (multiple formats)


jaredmoody:
I was trying to convert a partial that's used in my mailers to a ViewComponent today, and because the partial is used in a multipart email, it has a plaintext template (_thing.text.erb) as well as an html template (_thing.html.erb). I was surprised I could not define a component with two different template formats, I get the following error:

`ViewComponent::TemplateError: More than one template found for Mailers::ThingComponent. There can only be one default template file per component.`

Maybe I could workaround by using a variant?

It sure would be nice if this just worked out of the box the way that Rails does with different formats.

joelhawksley:
See our work in progress here: https://github.com/ViewComponent/view_component/pull/1681.

At this point, it looks like it will be too much work to land this any time soon. I'm tagging this issue as `help wanted` in case someone would like to step in and give it shot.

reeganviljoen:
@jaredmoody @joelhawksley I am working on adding support for this by checking using `request.format` to determine the format used in rendering the template

joelhawksley:
@reeganviljoen sounds good, let me know what you think of our previous attempt 😄

reeganviljoen:
@joelhawksley I looked at the previous attempt and it seemed solid, the only place where I differ is I believe the format should be assumed from the request

joelhawksley:
Closed by https://github.com/ViewComponent/view_component/pull/2079

PR #1681, opened 2023-03-14
Failing test for request formats

jaredmoody:
An initial stab at a failing test for #1670 - multipart emails are one part of this, multiple templates per request format may be a more generic case? If not, multiple formats is also a compelling feature for me.


jaredmoody:
That's ok, thanks for looking at it and for the pair! Maybe someday...

joelhawksley:
Closing in favor of https://github.com/ViewComponent/view_component/pull/1694

PR #1683, opened 2023-03-15
Protect against SystemStackError if CaptureCompatibility module is included more than once

camertron:

I tried setting `config.view_component.capture_compatibility_patch_enabled = true` in GitHub's monolith today and ran into a `SystemStackError` indicating infinite recursion. After digging in I discovered the `CaptureCompatibility` module is getting `include`d more than once, which causes `alias_method :original_capture, :capture` to be evaluated multiple times. The second time around, the alias makes both `#capture` and `#original_capture` point to the same method (`CaptureCompatibility::InstanceMethods#capture`). This means that when `#capture` calls `#original_capture`, it ends up calling itself 🤦 

I added a single-line `return` statement to bail out if `InstanceMethods` has already been prepended to `ActionView::Base`.

I think the above covers it.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Protect against `SystemStackError` if `CaptureCompatibility` module is included more than once.
+
+    *Cameron Dutro*
+
 ## v3.0.0.rc4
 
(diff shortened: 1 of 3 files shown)


BlakeWilliams (approved):
Good catch, :shipit:!

PR #1686, opened 2023-03-16
Update testing slots example

matthewlehner:

It looks like the API for slots has changed since the test examples were written. This updates the calls to `component` to use the `with_{slot_name}` method.

We're using RSpec for testing, so I'm not 100% certain that the api for slots is the same in MiniTest, but I assume these examples are just using the component API.


Spone:
Good catch! 👌

Spone:
I merged without it. For very simple docs changes like this I believe it's not mandatory.

Thanks a lot for your contribution!

PR #1687, opened 2023-03-16
Fix inline template error in development

kenyonj:

When using inline templates in development we try to compile the superclass of the component when there are no sidecar templates available. When your component inherits from `ViewComponent::Base` this is not a problem. However, when you have another base component that inherits from `ViewComponent::Base` (e.g. `class ApplicationComponent < ViewComponent::Base`) we fail to compile the superclass, `ApplicationComponent`.

However, in the case of `inline_calls` or inline `erb_template` usage, we don't actually want to try and compile the super class.

This PR:
  - Adds a failing test to exercise this problem
  - Modifies the `should_compile_superclass?` method to not only check for sidecar erb files, but also inline `erb_template` usage.

n/a

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,8 +10,7 @@ nav_order: 5
 
 ## main
 
-* Check for `inline_calls` and inline `erb_template` calls when deciding whether
-or not to compile a component's superclass.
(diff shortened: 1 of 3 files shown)


BlakeWilliams (approved):
Thanks for the fix!

PR #1688, opened 2023-03-16
Fix bug where `mkdir_p` failed due to incorrect permissions.

joelhawksley:


Gemfile.lock
```diff
@@ -303,9 +303,6 @@ DEPENDENCIES
   jbuilder (~> 2)
   m (~> 1)
   minitest (= 5.6.0)
-  net-imap
```
BlakeWilliams, line 306: Why did this change? 🤔

BlakeWilliams (approved):

Gemfile.lock
Spone: Seems related to https://github.com/rails/rails/pull/44083

PR #1695, opened 2023-03-28
Two small cleanups

joelhawksley:
1) Remove unused block arguments in `Slotable`
2) Rename `pvc` build to clarify why the build exists, as folks ask about it on occaision.

### DIFF
--- a/lib/view_component/slotable.rb
+++ b/lib/view_component/slotable.rb
@@ -84,7 +84,6 @@ module ViewComponent
           define_method slot_name do
             get_slot(slot_name)
           end
-          ruby2_keywords(slot_name.to_sym) if respond_to?(:ruby2_keywords, true)
 
           define_method "#{slot_name}?" do
             get_slot(slot_name).present?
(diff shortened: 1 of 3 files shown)


BlakeWilliams (approved):

lib/view_component/slotable.rb
```diff
@@ -81,10 +81,9 @@ def renders_one(slot_name, callable = nil)
           end
           ruby2_keywords(:"with_#{slot_name}") if respond_to?(:ruby2_keywords, true)
 
-          define_method slot_name do |*args, &block|
```
boardfish, line 84: I'll have a play with this at some point to see if it answers my own question, but would this break code like this in a component's template?

```html.erb
<% if help_link? %>
  <%= icon icon: :lifebuoy %>
<% else %>
  <%= icon icon: :external_link %>
<% end %>
```

`icon` in this case is a lambda slot that returns a component instance.

I guess this setup might need to be done internally to the lambda now, but I'm not sure if it also gave us unintended bonus behaviour of the arguments to `icon` becoming "default" arguments to the lambda in these instances.

joelhawksley: @boardfish wouldn't the setter be handling the arguments for the lambda slot, not the getter? So this example would be `with_icon...`

PR #1697, opened 2023-03-31
Add documentation for error

joelhawksley:

I'm looking to come up with an approach to solving https://github.com/ViewComponent/view_component/issues/896, adding our errors to the docs.

I created a unique error class for the specific case where `controller` is called before `render` and added it to the docs by extending our existing docs builder.

This will work automatically for any error class ending in `Error` and defining a `MESSAGE` constant. I plan to flesh out all of our errors once we can agree on an approach.


docs/api.md
```diff
@@ -128,7 +128,7 @@ so helpers, etc work as expected.
 
 ## Configuration
 
-### `.capture_compatibility_patch_enabled` → [String]
```
BlakeWilliams, line 131: It looks like we lost our return types?

joelhawksley: That was intentional. We meant to exclude them for config docs, but the exclusion code was broken 😄

BlakeWilliams (approved):

PR #1700, opened 2023-03-31
Fix broken links in documentation

llenk:

Fixing the broken links in the documentation. 

One link (ActionView::Components in Rails 6.1, Drifting Ruby - paid content) points to content that is behind a paywall. I'm unsure if that was the case when the link was originally added to the list. I added a note to that listing to be up-front about the availability of that content.


joelhawksley (approved):
Fantastic, thank you!

PR #1701, opened 2023-04-03
Use custom error classes instead of generic errors

joelhawksley:

The goal of this PR is to add all of our error messages to our docs. In order to do so, I moved all of our errors to an `Errors` module and had YARD look for classes ending in `Error` that have a `MESSAGE` constant.

I also cleaned up a few errors that should have been removed previously. I'll follow up with a review of our error messages themselves.

What do y'all think of the names for these errors?

### DIFF
--- a/app/controllers/view_components_system_test_controller.rb
+++ b/app/controllers/view_components_system_test_controller.rb
@@ -15,7 +15,7 @@ class ViewComponentsSystemTestController < ActionController::Base # :nodoc:
   private
 
   def validate_test_env
-    raise "ViewComponentsSystemTestController must only be called in a test environment" unless Rails.env.test?
+    raise ViewComponent::SystemTestControllerOnlyAllowedInTestError unless Rails.env.test?
   end
 
   # Ensure that the file path is valid and doesn't target files outside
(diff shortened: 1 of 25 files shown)


app/controllers/view_components_system_test_controller.rb
```diff
@@ -24,7 +24,7 @@ def validate_file_path
     base_path = ::File.realpath(self.class.temp_dir)
     @path = ::File.realpath(params.permit(:file)[:file], base_path)
     unless @path.start_with?(base_path)
-      raise ArgumentError, "Invalid file path"
+      raise ViewComponent::SystemTestControllerNefariousPathError
```
camertron, line 27: Lol "nefarious"

lib/view_component/errors.rb
```diff
@@ -1,16 +1,213 @@
...
+    def initialize
+      super(self.class::MESSAGE)
+    end
+  end
+
+  class DuplicateSlotContentError < StandardError
+    MESSAGE =
+      "It looks like a block was provided after calling `with_content` on COMPONENT, " \
+      "which means that ViewComponent doesn't know which content to use.\n\n" \
+      "To fix this issue, use either `with_content` or a block."
```
camertron, line 12: Would these look better with squiggly heredocs?

```suggestion
    MESSAGE = <<~MESSAGE
      It looks like a block was provided after calling `with_content` on COMPONENT,
      which means that ViewComponent doesn't know which content to use.

      To fix this issue, use either `with_content` or a block.
    MESSAGE
```

camertron (approved):
Oh wow this is awesome! 🎉

lib/view_component/errors.rb
joelhawksley: I'm not sure these lines are equivalent, though? I'm trying to have the first sentence of the output be a single line of text.

joelhawksley: Confirmed in the console: `=> "It looks like a block was provided after calling `with_content` on COMPONENT,\n`

PR #1702, opened 2023-04-03
Allow ActiveRecord objects to renders_many slots

leighhalliday:

There was an issue (https://github.com/ViewComponent/view_component/issues/1547) which wanted to allow passing ActiveRecord objects to `renders_many` slots.

Normal example:
```ruby
    render_inline(SlotsComponent.new(classes: "mt-4")) do |component|
      component.with_items([{highlighted: true}, {highlighted: false}])
    end
```

With ActiveRecord objects, causes an error `TypeError: no implicit conversion of Post into Hash`:
```ruby
    render_inline(SlotsComponent.new(classes: "mt-4")) do |component|
      component.with_posts([Post.new(title: "Title A"), Post.new(title: "Title B")])
    end
```

The double splat `**` operator is meant specifically for hashes (or objects that respond to `to_hash` method). I attempted to solve this by checking if the args respond to `to_hash?`... if that is the case we can stick with the double splat, otherwise we can use the single splat operator to pass along arguments to the slot being rendered.

```ruby
{}.respond_to?(:to_hash) # true
```


docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Allow ActiveRecord objects to be passed to `renders_many`
```
```suggestion
* Allow ActiveRecord objects to be passed to `renders_many`.
```

joelhawksley (approved):
LGTM! I'm going to check this out locally and see if I we can support ActiveRecord collections too.

PR #1712, opened 2023-04-23
Added support for CSP nonces with `javascript_tag` 

reeganviljoen:

This allows `javascript_tag` to work with the nonce parameter without throwing this error `undefined local variable or method `content_security_policy_nonce' for #<NonceComponent:0x00007fdee2bc2f50 @_routes=nil, @view_context=#<ActionView::Base:0x0000000000ea60>`

I used an aproach similar to adding the CSRF token helper to components to make sure the added code fits the style of the codebase. The primary trade of with this aproach is that it couples the requests global state into the component but I believe this is an acceptable use case as their would be no way to pull in a nonce value without coupling global state somewhat 

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -38,6 +38,9 @@ module ViewComponent
     # For CSRF authenticity tokens in forms
     delegate :form_authenticity_token, :protect_against_forgery?, :config, to: :helpers
 
+    #For Content Security Policy nonces
+    delegate :content_security_policy_nonce, to: :helpers
+
     # Config option that strips trailing whitespace in templates before compiling them.
     class_attribute :__vc_strip_trailing_whitespace, instance_accessor: false, instance_predicate: false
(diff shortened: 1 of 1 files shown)


docs/CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ nav_order: 5
 
      *Mike Munroe*
 
+* Add support for csp nonces inside of components
```
Spone, line 21:
```suggestion
* Add support for CSP nonces inside of components.
```

Spone (commented):
Thanks for your contribution. Do you think you can add a test for this feature?

reeganviljoen:
> Thanks for your contribution. Do you think you can add a test for this feature?

I will get working on that tonight thank you

docs/CHANGELOG.md
reeganviljoen: I will fix that when I write the tests tonight thank you

reeganviljoen:
Hi @Spone I added tests like you asked me , thank you

joelhawksley (approved):

PR #1738, opened 2023-05-02

donapieppo:
When view-component is required in a Rails::Engine the variable
`ViewComponent::Base.config.view_component_path` is null and `Regexp.quote(ViewComponent::Base.config.view_component_path)` 
raises `TypeError: no implicit conversion of nil into String`.

Checking `defined?(Rails) && Rails.application` seems to solve the problem.

May be related to [#1565](https://github.com/ViewComponent/view_component/issues/1565)
but in 3.0.0 version.


donapieppo:
@joelhawksley I can try but I have no experience on how to test a gem against another gem :-). 
I mean that the problem comes when you create an engine and use wiew_component in that engine.
So I gess I have to create a test engine and verify that the `dummy rails app` inside the engine does not raise the error.

Does it make sense?

joelhawksley:
@donapieppo sounds like it might not be worth the trouble. I'll merge as-is!

donapieppo:
Thanx!  
I actually gave another look into the problems in my engine and in #1565. I add notes here just for reference.

Create the standard example blorgh engine as in rails documentation:
```bash
rails plugin new blorgh --mountable
```
and add `spec.add_dependency "view_component"` in blorgh.gemspec and `require "view_component"` in `lib/blorgh.rb`.

Than create a first component (for example `app/components/blorgh/first_component.rb` and `app/components/blorgh/first_component.html.erb`) and run the test. Everything is ok.

If in `lib/blorgh.rb` you add 
`require_relative "../app/components/blorgh/first_component"`
or if you put the components in another path and require them you get the `no implicit conversion of nil into String` error. This is what is done in matey gem (#1565).

With this pull request it works also with the require of the component in `lib/blorgh.rb`.

PR #1746, opened 2023-05-08
Codemod v3 slots improvements

kirillplatonov:
This PR includes improvements for #1669:

- Adds rake task to detect legacy slots `bin/rails view_component:detect_legacy_slots`
- Adds rake to migrate legacy slots automatically `bin/rails view_component:migrate_legacy_slots`
- Adds component previews to codemods view paths
- Autodetection for Rails view paths
- Improves uncertain matches detection
- Updated changelog

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,7 +18,7 @@ nav_order: 5
 
    *Joseph Carpenter*
 
-* Add codemod to detect and migrate deprecated Slots setters to new `with_*` prefix.
+* Add codemod to detect and migrate deprecated Slots setters to new `with_*` prefix introduced in v3.0.0.
 
     ```bash
     bin/rails view_component:detect_legacy_slots
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,6 +18,15 @@ nav_order: 5
 
    *Joseph Carpenter*
 
+* Add codemod to detect and migrate deprecated Slots setters to new `with_*` prefix.
+
+    ```bash
+    bin/rails view_component:detect_legacy_slots
+    bin/rails view_component:migrate_legacy_slots
+    ```
+
+    *Hans Lemuet, Kirill Platonov*
+
 ### v3.0.0
 
 1,000+ days and 100+ releases later, the 200+ contributors to ViewComponent are proud to ship v3.0.0!
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,7 +18,7 @@ nav_order: 5
 
    *Joseph Carpenter*
 
-* Add codemod to detect and migrate deprecated Slots setters to new `with_*` prefix introduced in v3.0.0.
+* Add codemod to detect and migrate deprecated Slots setters to new `with_*` prefix introduced in v3.x. Note: This codemod is non-deterministic and works on a best-effort basis.
 
     ```bash
     bin/rails view_component:detect_legacy_slots
--- a/lib/view_component/codemods/v3_slot_setters.rb
+++ b/lib/view_component/codemods/v3_slot_setters.rb
@@ -93,7 +93,7 @@ module ViewComponent
         [].tap do |suggestions|
           File.open(file) do |f|
             f.each_line do |line|
-              if (matches = line.scan(/(?<!\s)\.(?<slot>#{Regexp.union(all_registered_slot_names)})/))
+              if (matches = line.scan(/(?<!\s)\.(?<slot>#{Regexp.union(all_registered_slot_names)})\b/))
                 next if matches.size == 0
 
                 matches.flatten.each do |match|
--- a/lib/view_component/codemods/v3_slot_setters.rb
+++ b/lib/view_component/codemods/v3_slot_setters.rb
@@ -159,9 +159,15 @@ module ViewComponent
         "{#{view_component_paths.join(",")}}"
       end
 
+      def rails_view_paths
+        ActionController::Base.view_paths.select do |path|
+          path.to_s.include?(Rails.root.to_s)
+        end.map(&:to_s)
+      end
+
       def view_paths
         @view_paths ||= [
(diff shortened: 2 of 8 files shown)


lib/view_component/codemods/v3_slot_setters.rb
```diff
@@ -7,7 +7,7 @@
 # If your app uses custom paths for views, you can pass them in:
 #
 #     ViewComponent::Codemods::V3SlotSetters.new(
-#       view_path: "../app/views",
+#       view_path: Rails.root.join("../app/views"),
```
Spone, line 10: If we join to Rails' root, I guess we don't need the `..\`

```suggestion
#       view_path: Rails.root.join("app/views"),
```

docs/CHANGELOG.md
```diff
@@ -18,6 +18,15 @@ nav_order: 5
 
    *Joseph Carpenter*
 
+* Add codemod to detect and migrate deprecated Slots setters to new `with_*` prefix.
```
Spone, line 21:
```suggestion
* Add codemod to detect and migrate deprecated Slots setters to new `with_*` prefix introduced in v3.0.0.
```

lib/view_component/codemods/v3_slot_setters.rb
```diff
@@ -1,28 +1,31 @@
...
     class V3SlotSetters
       TEMPLATE_LANGUAGES = %w[erb slim haml].join(",").freeze
       RENDER_REGEX = /render[( ](?<component>\w+(?:::\w+)*)\.new[) ]+(do|\{) \|(?<arg>\w+)\b/ # standard:disable Lint/MixedRegexpCaptureTypes
 
       Suggestion = Struct.new(:file, :line, :message)
 
-      def initialize(view_component_path: [], view_path: [])
-        Zeitwerk::Loader.eager_load_all
+      def initialize(view_component_path: [], view_path: [], migrate: false)
+        Rails.application.eager_load!
```
Spone, line 24: If I remember correctly, I used `Zeitwerk` directly so the codemod works in gems that do no require Rails (this needs to be verified).

lib/view_component/codemods/v3_slot_setters.rb
```diff
@@ -159,9 +191,16 @@ def view_component_path_glob
...
+        ActionController::Base.view_paths.select do |path|
+          path.to_s.include?(Rails.root.to_s)
+        end.map(&:to_s)
+      end
+
       def view_paths
         @view_paths ||= [
-          "app/views",
+          rails_view_paths,
+          Rails.application.config.view_component.preview_paths,
```
Spone, line 203: Does this break when running the codemod in a gem that does not require Rails?

docs/CHANGELOG.md
```diff
@@ -18,6 +18,15 @@ nav_order: 5
...
+
+    ```bash
+    bin/rails view_component:detect_legacy_slots
+    bin/rails view_component:migrate_legacy_slots
```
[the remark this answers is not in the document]
Spone: Also, maybe we can pass custom paths to the task, similar to rubocop:

```
rubocop          # default
rubocop app lib  # run on specific paths
```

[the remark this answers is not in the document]
kirillplatonov: In rake tasks we can use `--fix` option only with the options separator:
```bash
rails view_component:detect_legacy_slots --- --fix
```
Not sure if it's cleaner than a separate rake task for migrations.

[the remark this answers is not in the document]
kirillplatonov: > Also, maybe we can pass custom paths to the task, similar to rubocop:
Added 👍

lib/view_component/codemods/v3_slot_setters.rb
kirillplatonov: This will cause issues with apps that don't use `Zeitwerk`. Eg some versions of Rails in the `view_component` test suite will fail. The library is Rails-only so I don't think it should be a problem.

lib/view_component/codemods/v3_slot_setters.rb
kirillplatonov: In gems app rake tasks require dummy app and at least development rails dependency:
```
# Rakefile
APP_RAKEFILE = File.expand_path("test/dummy/Rakefile", __dir__)
load "rails/tasks/engine.rake"
```
And can be executed with the prefix:
```
bin/rails app:view_component:detect_legacy_slots
```
Considering that, I don't think Rails view paths call will be a problem.

kirillplatonov:
@Spone can we merge it at the current state? I included this class in Polaris ViewComponents v1 release and tested upgrade on a couple production apps. It works great. 

Even if the current implementation is not perfect it will be beneficial to have it shipped with ViewComponent now when people are doing upgrades to v3. In a couple of months this will be less useful I suppose.

Spone:
I agree. What do you think @joelhawksley?

joelhawksley (commented):
I like the idea of getting this into `main` even if it isn't perfect ❤️

docs/CHANGELOG.md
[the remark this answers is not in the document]
kirillplatonov: Added a note to changelog. Not sure where to put info about this codemod in docs. It seems like there's no upgrade section at the moment.

Gemfile
```diff
@@ -14,7 +14,7 @@ group :test do
   gem "cuprite", "~> 0.8"
   gem "puma", "~> 5"
 
-  gem "selenium-webdriver", "~> 4"
+  gem "selenium-webdriver", "~> 4", "< 4.9.1"
```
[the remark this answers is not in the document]
kirillplatonov: Removed it

kirillplatonov:
@joelhawksley please note that this PR is a contribution to #1669. We'll need to make a final review there once merged.

Spone (approved):

Spone:
Let's wrap this up in #1669!

PR #1752, opened 2023-05-16
Fix duplication in configuration docs

thomascchen:

This fixes some incorrect duplication in the configuration docs for `.capture_compatibility_patch_enabled` and `.default_preview_layout`. The descriptions for both config options were included under both headings. This seems to have been introduced in this commit: https://github.com/ViewComponent/view_component/commit/72830c8cf4aa034570fb2cc3167dedbd69cf7f4b.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Fix duplication in configuration docs.
+
+    *Tom Chen*
+
 * Fix helpers not reloading in development.
 
     *Jonathan del Strother*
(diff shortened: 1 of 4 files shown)


Spone:
Hi @thomascchen the `docs/api.md` is generated with YARD. There may be something to fix in the source code comments, to make the fix permanent (else it may be overwritten next time we generate the docs). Can you have a look?

thomascchen:
@Spone that makes sense, thanks! I found an extra `#` in the comments separating the documentation for these two methods and removed it. Then ran `rake docs:build` which updated the return value for the `.config` class method as well.

Spone (approved):
Thanks for the fix!

PR #1753, opened 2023-05-20
Add Iterators for slots

reeganviljoen:
closes https://github.com/ViewComponent/view_component/issues/1620
Add a parameter similar to the iteration context provided when [rendering collections](https://viewcomponent.org/guide/collections.html#collection-iteration-context), ViewComponent could provide contextual information, such as #size, #index, #first?, and #last?, when rendering slots that have been defined with renders_many,
additionally provide a collection counter to slots defined with renders_many.

I chose to use ActionView::PartialIteration to get the iteration count and the iteration context because that is what was used in the collection iteration counter and iteration context helpers 


lib/view_component/slotable.rb
```diff
@@ -387,6 +392,16 @@ def set_slot(slot_name, slot_definition = nil, *args, &block)
 
       slot
     end
+
+    def set_slot_itterator(options, component, slot_name)
```
Spone, line 396:
```suggestion
    def set_slot_iterator(options, component, slot_name)
```

reeganviljoen: fixed thank you

reeganviljoen:
@Spone I fixed that spelling mistake a re-review would be much appreciated

docs/CHANGELOG.md
```diff
@@ -30,6 +30,14 @@ nav_order: 5
 
     *Joseph Carpenter*
 
+* Add support for CSP nonces inside of components.
+
+     *Reegan Viljoen*
```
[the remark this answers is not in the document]
reeganviljoen: When I added that fix in another PR I accidentally put it under a release to I moved it to main

test/sandbox/app/components/numbered_tab_component.html .erb
```diff
@@ -0,0 +1,7 @@
+<% if @tab_iteration.first? %>
+  <p><%= @tab_number%>. first <%=@title%></p>
+<% else %>
+  <p><%= @tab_number%>. not first <%=@title%></p>
+<% end %>
+
+
```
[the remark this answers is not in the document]
reeganviljoen: fixed thank you @joelhawksley

test/sandbox/app/components/slots_iteration_component.rb
```diff
@@ -0,0 +1,21 @@
...
+class SlotsIterationComponent < ViewComponent::Base
+  renders_many :numbered_tabs, NumberedTabComponent
+  renders_many :numbered_items, "NumberedItemComponent"
+
+  def initialize(classes: "")
+    @classes = classes
+  end
+
+  class NumberedItemComponent < ViewComponent::Base
+    delegate :first?, :last?, to: :item_iteration
```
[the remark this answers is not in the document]
reeganviljoen: @joelhawksley I stripped out the use of `delegate` in my example code

test/sandbox/app/components/numbered_tab_component.html.erb
```diff
@@ -0,0 +1,7 @@
+<% if @tab_iteration.first? %>
+  <p><%= @tab_number%>. first <%=@title%></p>
+<% else %>
+  <p><%= @tab_number%>. not first <%=@title%></p>
+<% end %>
+
+
```
joelhawksley, line 7:
```suggestion
```

joelhawksley (commented):
Looks good! Mind adding some docs? I'll happily edit them for you ❤️

reeganviljoen:
@joelhawksley I added some docs. I leveraged a lot of the content from the collection docs. A  second set of eyes would be much appreciated

lib/view_component/slotable.rb
```diff
@@ -356,9 +356,14 @@ def set_slot(slot_name, slot_definition = nil, *args, &block)
 
       # If class
       if slot_definition[:renderable]
+        item_options = set_slot_iterator(args, slot_definition[:renderable], slot_name)
+        args.last.merge!(item_options) if item_options.present?
```
corasaurus-hex, line 360: I think it might be better to avoid mutation here? If I'm understanding the code correctly this would mutate a user's hash that they pass in, which could lead to surprising behavior for the user.

Maybe something like this?

```ruby
args = args.dup.tap{|a| a[-1] = a[-1].merge(item_options) } if item_options.present?
```

lib/view_component/slotable.rb
```diff
@@ -356,9 +356,14 @@ def set_slot(slot_name, slot_definition = nil, *args, &block)
...
+
         slot.__vc_component_instance = slot_definition[:renderable].new(*args)
       # If class name as a string
       elsif slot_definition[:renderable_class_name]
+        item_options = set_slot_iterator(args, self.class.const_get(slot_definition[:renderable_class_name]), slot_name)
+        args.last.merge!(item_options) if item_options.present?
```
corasaurus-hex, line 366: Ditto here!

docs/guide/slots.md
```diff
@@ -282,6 +282,108 @@ Slot content can also be set using `#with_content`:
...
+  <%= child %>
+<% end %>
+```
+
+```ruby
+# app/components/child_component.rb
+class ChildComponent < ViewComponent::Base
+  def initialize(first_name:, child_iteration:)
+    @first_name = first_name
+    @counter = child_counter
```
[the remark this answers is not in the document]
reeganviljoen: Thank you, I will fix it later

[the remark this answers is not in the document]
reeganviljoen: @corasaurus-hex I have fixed it, thank you.

lib/view_component/slotable.rb
```diff
@@ -395,6 +400,16 @@ def set_slot(slot_name, slot_definition = nil, *args, &block)
...
       slot
     end
+
+    def set_slot_iterator(options, component, slot_name)
+      item_options = {}
+      @collection_iterator ||= ActionView::PartialIteration.new(send(slot_name))
+      item_options[component.collection_counter_parameter] = @collection_iterator.index if component.counter_argument_present?
+      item_options[component.collection_iteration_parameter] = @collection_iterator.dup if component.iteration_argument_present?
+      @collection_iterator.iterate!
+      options.last.merge!(item_options) if item_options.present?
```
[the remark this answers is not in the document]
reeganviljoen: @corasaurus-hex I have treied to implement your changes but they seem to break all the other test  no matter what I do, so I will keep the implementation as it is unless you can provide a working example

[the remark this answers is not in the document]
reeganviljoen: @corasaurus-hex I have removed all the merges and I have implemented the pieces of your code that I could get to work

reeganviljoen:
@joelhawksley a re-review of this pr at some point would be much appreciated

joelhawksley:
@reeganviljoen sorry for the delay, I lost track of this one!

@corasaurus-hex would you be up for re-reviewing first?

joelhawksley:
@reeganviljoen I've pushed a few minor refactors while I review this ❤️

reeganviljoen:
Thank you @joelhawksley , I see the tests are failing I will see if I can fix that

joelhawksley:
@corasaurus-hex I'd still be interested in your review ❤️

lib/view_component/slotable.rb
```diff
@@ -356,9 +356,11 @@ def set_slot(slot_name, slot_definition = nil, *args, &block)
 
       # If class
       if slot_definition[:renderable]
+        args = add_slot_iterator_options(args, slot_definition[:renderable], slot_name) if args.any?
         slot.__vc_component_instance = slot_definition[:renderable].new(*args)
       # If class name as a string
       elsif slot_definition[:renderable_class_name]
+        args = add_slot_iterator_options(args, self.class.const_get(slot_definition[:renderable_class_name]), slot_name) if args.any?
```
corasaurus-hex, line 363: What if instead of having `if args.any?` at the call sites we had a guard at the top of the `add_slot_iterator_options` method that returned args if args is empty?

lib/view_component/slotable.rb
```diff
@@ -395,6 +397,31 @@ def set_slot(slot_name, slot_definition = nil, *args, &block)
...
+
+        if component.iteration_argument_present?
+          options[-1][component.collection_iteration_parameter] = @collection_iterator.dup
+        end
+
+        @collection_iterator.iterate!
+      end
+
+      options
+    end
```
corasaurus-hex, line 423: I like the changes that have been made to this method!

I do think it's a litle confusing, though. We're mutating the options argument on lines 412 and 416 and then returning the mutated arguments. Usually if the method mutates the objects you're passing in it doesn't then return them since you still have the original mutated references. And even if it did mutate and return, it would be for chaining convenience and not for how we're using it at the call site.

So the method use is telling me that it's not mutating the arguments but the body does mutate the arguments, which is a little mixed up, I think.

I think it would be better to avoid mutating the objects on 412 and 416 by adding on 410:

```ruby
options = options.dup.tap{|o| o[-1] = o[-1].dup }
```

I can go into a few reasons I think it's worth avoiding mutation if you'd like, but it's probably good enough to say that if the user passes in frozen objects this will raise an error.

lib/view_component/slotable.rb
```diff
@@ -395,6 +397,31 @@ def set_slot(slot_name, slot_definition = nil, *args, &block)
 
       slot
     end
+
+    def add_slot_iterator_options(options, component, slot_name)
```
corasaurus-hex, line 401: Can we maybe name the first argument `args` to match the call sites? It feels to me like the last element in `args` is the `options` hash that we're adding values to, is it the same for you?

corasaurus-hex (commented):
I think this has shaped up a lot overall but there are a couple places that could use a little polish before we ship this, I think.

docs/guide/slots.md
```diff
@@ -282,6 +282,92 @@ Slot content can also be set using `#with_content`:
...
+```ruby
+#app/components/parent_component.rb
+class ParentComponent < ViewComponent::Base
+  renders_many :children, ChildComponent
+end
+```
+
+```erb
+#app/components/parent_component.html.erb
+<% children.each do |child| %>
```
BlakeWilliams, line 358: What happens if I call `sort`, or `reverse`? Will the iterator methods still work as-intended, or will they return misleading or incorrect values?

lib/view_component/slotable.rb
reeganviljoen: @corasaurus-hex I renamed all the options methods to args

lib/view_component/slotable.rb
reeganviljoen: @corasaurus-hex I have moved the guard clause to the method

lib/view_component/slotable.rb
reeganviljoen: @corasaurus-hex I agree that their is a danger of mutating an array that is passed to you because it may be frozen, so I tried to implement the duped array aswell as the nested hash but the nested hash when duped does not work with the splat operator for soem reason, here teh error I get when I splat the hash
```
ArgumentError: wrong number of arguments (given 1, expected 0; required keywords: title, numbered_tab_counter, numbered_tab_iteration)

Comment (reeganviljoen) [reply lib/view_component/slotable.rb:424]: on another note the args is not passed as an array to this method are not passed by the user bu rather the individual args of the with_slot_name accesor so it I do not think it is reasonable to expect the args to be in a immutable state from being frozen, etc

File: docs/guide/slots.md (hunk 13 above)
Comment (reeganviljoen) [reply docs/guide/slots.md:358]: @BlakeWilliams the order of the iterator is dependent on the order on which they are registered not the order in which they are sorted into inside the component, and while this may seem misleading I believe it may make more sense for the iteration context to be consistent with the order in which the slot setter is invoked

File: lib/view_component/slotable.rb (hunk 11 above)
Comment (corasaurus-hex) [reply lib/view_component/slotable.rb:426]: How about this?

```diff
diff --git a/lib/view_component/slotable.rb b/lib/view_component/slotable.rb
index 8aa36a8b..ec637128 100644
--- a/lib/view_component/slotable.rb
+++ b/lib/view_component/slotable.rb
@@ -357,13 +357,13 @@ module ViewComponent

       # If class
       if slot_definition[:renderable]
-        args = add_slot_iterator_args(args, slot_definition[:renderable], slot_name)
-        slot.__vc_component_instance = slot_definition[:renderable].new(*args)
+        *args, opts = add_slot_iterator_options(args, slot_definition[:renderable], slot_name)
+        slot.__vc_component_instance = slot_definition[:renderable].new(*args, **opts)
       # If class name as a string
       elsif slot_definition[:renderable_class_name]
-        args = add_slot_iterator_args(args, self.class.const_get(slot_definition[:renderable_class_name]), slot_name)
+        *args, opts = add_slot_iterator_options(args, self.class.const_get(slot_definition[:renderable_class_name]), slot_name)
         slot.__vc_component_instance =
-          self.class.const_get(slot_definition[:renderable_class_name]).new(*args)
+          self.class.const_get(slot_definition[:renderable_class_name]).new(*args, **opts)
       # If passed a lambda
       elsif slot_definition[:renderable_function]
         # Use `bind(self)` to ensure lambda is executed in the context of the
@@ -399,9 +399,15 @@ module ViewComponent
       slot
     end

-    def add_slot_iterator_args(args, component, slot_name)
-      args = args.deep_dup
-      return args unless args.any?
+    def add_slot_iterator_options(args, component, slot_name)
+      args = args.dup
+
+      if args.last.is_a?(Hash)
+        args[-1] = args[-1].dup
+      else
+        args << {}
+      end
+
       if component.counter_argument_present? || component.iteration_argument_present?
         @__vc_set_slots ||= {}
         @collection_iterator ||=
```

docs/guide/slots.md
BlakeWilliams: I think the big thing for me here is that we can't reliably know _what_ the order is here. We're declaring it as first defined, but I imagine it would be easy for folks to assume that it would be based on rendering order.

e.g. any of the following rendering scenarios could lead to some unfortunate, and subtle bugs:

```ruby
children.each { |child| child.to_s }
children.reverse.each { |child| child.to_s }
children.sort.each { |child| child.to_s }
```

Since it's Ruby, it's also possible to do things like:

```ruby
def children
  super.reverse
end
```

In any of those cases the iterator methods we provide would return an incorrect value because of the lack of abstractions around rendering a collection of slots, and I don't believe that order of registration is a good substitute for that scenario.

So my general concern is that there's a good few ways to have drift between the reality of the iterator methods (first, last, etc) and what users expect to get back from the API. View Component leans heavily into "it's just Ruby" and I while I appreciate all the effort that went into this PR, I think that may be what we want to encourage folks to do here instead of providing an API on-top of iterators based on registration order for them to utilize.

corasaurus-hex: This is a really great point and you've changed my mind on this PR. Generally I care about the rendering order and not the order slots were added, at least in cases like I describe in the https://github.com/ViewComponent/view_component/issues/1737.

reeganviljoen:
@joelhawksley @corasaurus-hex @Spone @BlakeWilliams after much thought I have decided to abonden this pr because of the very valid points braught forth

PR #1771, opened 2023-06-13
Allow Rails to provide timing data in browser dev-tools via Server-Timing header

tgaff:

Per this discussion:
https://github.com/ViewComponent/view_component/issues/1490#issuecomment-1588377941

This may resolve in part https://github.com/ViewComponent/view_component/issues/1490

This alters the key used for ActiveSupport Instrumentation so that Rails automatically picks it up and includes a sum total in the browser headers sent to the client ([feature added to Rails here](https://github.com/rails/rails/pull/36289)).  By default the `Server-Timing` header is only enabled in development but can be turned on via `config.server_timing = true`.  Rails excludes any notification key that starts with `!` when setting the Server-Timing header.

All desktop browsers I've checked (FF, Chrome, Safari, Edge, Opera) will automatically add this data to the client-side timing data on the dev-tools Network tab. This makes for a quick "low-detail" look at what happened in a particular request.

Screen Shot 2023-06-13 at 12 19 39 PM
Above is what you see in Firefox.

* marks the old key as deprecated
  * and avoids reporting this on every render
* adds a config setting to choose the deprecated key (default) or the new one

~I've gone with the simplest approach of just renaming the key _for now_ as that's what was initially discussed in the issue.  Please see discussion below.~

Based on feedback, this now introduces a new config to avoid the breaking change. 

**the below "breaking change" issue has been resolved with an additional config option per @joelhawksley instruction**
<s>
As I [mentioned](https://github.com/ViewComponent/view_component/issues/1490#issuecomment-1588480194) on the issue, this could be considered a _breaking change_ for anyone currently subscribed to the current notification name.  I proposed 3 ways of handling that, but could use guidance on what maintainers think is best.

1. wait for a major release to include this change
2. publish both events: `!render.view_component` and `render.view_component`.  The overhead is low assuming no one [subscribes to both events](https://github.com/rails/rails/blob/e88857bbb9d4e1dd64555c34541301870de4a45b/activesupport/lib/active_support/notifications.rb#L204-L210) (in which case its double the current). 
3. another config flag to set which name to use + maybe a deprecation for the old `!` name.

I have a slight preference towards 2 since it gets this in quickly, but not opposed to any of these.
</s>
Also I haven't proven it out yet, but I think there's a chance that nested components could also suffer from this bug that I reported over on the Rails repo: https://github.com/rails/rails/issues/48375

### DIFF
--- a/docs/guide/instrumentation.md
+++ b/docs/guide/instrumentation.md
@@ -30,7 +30,7 @@ ActiveSupport::Notifications.subscribe("render.view_component") do |*args| # or 
 end
 ```
 
-### Viewing instrumentation sums in the browser developer tools
+## Viewing instrumentation sums in the browser developer tools
 
 When using `render.view_component` with `config.server_timing = true` (default in development) in Rails 7, the browser developer tools display the sum total timing information in Network > Timing under the key `render.view_component`.
 


joelhawksley:
> another config flag to set which name to use + maybe a deprecation for the old ! name.

Let's do that! Let's default the config to false.

boardfish (commented):
Looks good to me!

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Allow instrumentation to be automatically included in Server-Timing headers generated by Rails. To enable this set the config `config.instrumentation_use_deprecated_name = false`.  The old key `!render.view_component` is deprecated, please update your ActiveSupport::Notification subscriptions to `render.view_component`.
```
```suggestion
* Allow instrumentation to be automatically included in Server-Timing headers generated by Rails. To enable this set the config `config.use_deprecated_instrumentation_name = false`.  The old key `!render.view_component` is deprecated: update ActiveSupport::Notification subscriptions to `render.view_component`.
```

docs/guide/instrumentation.md
```diff
@@ -15,14 +15,19 @@ To enable ActiveSupport notifications, use the `instrumentation_enabled` option:
 # config/application.rb
 # Enable ActiveSupport notifications for all ViewComponents
 config.view_component.instrumentation_enabled = true
+config.view_component.instrumentation_use_deprecated_name = false
```
joelhawksley, line 18:
```suggestion
config.view_component.use_deprecated_instrumentation_name = false
```

docs/guide/instrumentation.md
```diff
@@ -15,14 +15,19 @@ To enable ActiveSupport notifications, use the `instrumentation_enabled` option:
...
 ```
 
-Then subscribe to the `!render.view_component` event:
+Setting `instrumentation_use_deprecated_name` configures the event name used. If `false` the name `"render.view_component"` is used. If `true` (default) it will use the deprecated name `"!render.view_component"`.
```
joelhawksley, line 21:
```suggestion
Setting `use_deprecated_instrumentation_name` configures the event name used. If `false` the name `"render.view_component"` is used. If `true` (default) the deprecated `"!render.view_component"` will be used.
```

docs/guide/instrumentation.md
```diff
@@ -15,14 +15,19 @@ To enable ActiveSupport notifications, use the `instrumentation_enabled` option:
...
+
+You can then subscribe to the event:
```
joelhawksley, line 23:
```suggestion
Subscribe to the event:
```

docs/guide/instrumentation.md
```diff
@@ -15,14 +15,19 @@ To enable ActiveSupport notifications, use the `instrumentation_enabled` option:
...
-ActiveSupport::Notifications.subscribe("!render.view_component") do |*args|
+ActiveSupport::Notifications.subscribe("render.view_component") do |*args| # or !render.view_component
   event = ActiveSupport::Notifications::Event.new(*args)
-  event.name    # => "!render.view_component"
+  event.name    # => "render.view_component"
   event.payload # => { name: "MyComponent", identifier: "/Users/mona/project/app/components/my_component.rb" }
 end
 ```
+
+If you're using the non-deprecated key & have `config.server_timing = true` (default in development) set in Rails 7, you can see sum total timing information in your browser's developer-tools. Find the request on the Network tab and view it's "timing" sub-tab. On most browsers server results are shown beneath the client-side data. Look for the key `render.view_component`.
```
joelhawksley, line 33:
```suggestion
When using `render.view_component` with `config.server_timing = true` (default in development) in Rails 7, the Chrome developer tools display the sum total timing information in Network > Timing under the key `render.view_component`.
```

lib/view_component/config.rb
```diff
@@ -17,6 +17,7 @@ def defaults
           preview_route: "/rails/view_components",
           show_previews_source: false,
           instrumentation_enabled: false,
+          instrumentation_use_deprecated_name: true,
```
joelhawksley, line 20:
```suggestion
          use_deprecated_instrumentation_name: true,
```

lib/view_component/config.rb
```diff
@@ -99,6 +100,13 @@ def defaults
       # Whether ActiveSupport notifications are enabled.
       # Defaults to `false`.
 
+      # @!attribute instrumentation_use_deprecated_name
```
joelhawksley, line 103:
```suggestion
      # @!attribute use_deprecated_instrumentation_name
```

lib/view_component/instrumentation.rb
```diff
@@ -19,5 +19,13 @@ def render_in(view_context, &block)
         super(view_context, &block)
       end
     end
+
+    private
+
+    def notification_name
+      return "!render.view_component" if ViewComponent::Base.config.instrumentation_use_deprecated_name
```
joelhawksley, line 26:
```suggestion
      return "!render.view_component" if ViewComponent::Base.config.use_deprecated_instrumentation_name
```

test/sandbox/test/config_test.rb
```diff
@@ -14,6 +14,7 @@ def test_defaults_are_correct
       assert_equal @config.preview_route, "/rails/view_components"
       assert_equal @config.show_previews_source, false
       assert_equal @config.instrumentation_enabled, false
+      assert_equal @config.instrumentation_use_deprecated_name, true
```
joelhawksley, line 17:
```suggestion
      assert_equal @config.use_deprecated_instrumentation_name, true
```

test/sandbox/test/instrumentation_test.rb
```diff
@@ -4,16 +4,29 @@
 
 class InstrumentationTest < ViewComponent::TestCase
   def test_instrumentation
+    with_config_option(:instrumentation_use_deprecated_name, false) do
```
joelhawksley, line 7:
```suggestion
    with_config_option(:use_deprecated_instrumentation_name, false) do
```

joelhawksley (approved):
👏🏻

lib/view_component/config.rb
```diff
@@ -17,6 +17,7 @@ def defaults
           preview_route: "/rails/view_components",
           show_previews_source: false,
           instrumentation_enabled: false,
+          use_deprecated_instrumentation_name: true,
```
tgaff, line 20: I agree this name reads better. I had used the other name so that it would alphabetically show up next to "instrumentation_enabled" in any sort of auto-gened docs.

I just noticed that `instrumentation_enabled` shows in the API.md doc file.  Should I add this key there as well?

docs/guide/instrumentation.md
```diff
@@ -15,14 +15,19 @@ To enable ActiveSupport notifications, use the `instrumentation_enabled` option:
...
-ActiveSupport::Notifications.subscribe("!render.view_component") do |*args|
+ActiveSupport::Notifications.subscribe("render.view_component") do |*args| # or !render.view_component
   event = ActiveSupport::Notifications::Event.new(*args)
-  event.name    # => "!render.view_component"
+  event.name    # => "render.view_component"
   event.payload # => { name: "MyComponent", identifier: "/Users/mona/project/app/components/my_component.rb" }
 end
 ```
+
+When using `render.view_component` with `config.server_timing = true` (default in development) in Rails 7, the Chrome developer tools display the sum total timing information in Network > Timing under the key `render.view_component`.
```
tgaff, line 33: Not important but: specifying Chrome here is perhaps a little "chrome-centric".  The feature works nearly the same in FF, Safari, Edge and Opera.  The screenshot in the PR is from FF 113 IIRC.

reeganviljoen: Is there anyway you can add Firefox and safari instructions

lib/view_component/config.rb
Spone: `api.md` is auto-generated by YARD using the comments from the source code. I think the comments you added are enough.

You can regenerate the docs locally with `bundle exec rake docs:build`.

tgaff: Thanks Spone! I did not know that.

docs/guide/instrumentation.md
tgaff: They are nearly identical aside from coloration, so I don't think it's necessary.  

1. Network tab (refresh if its empty)
2. Click the request you're interested in.
3. on the right-side OR the bottom-side (depends on size of the dev-tools area) you'll see several tabs for that request; choose "Timing" **or "Timings" (in FF)**
4. Look for server timings at the bottom beneath the client-side data.

Biggest caveats with Server-Timings in general:
* It's not provided for every request.  Normal MVC rails controllers in development mode will provide it (it's a middleware) - but assets won't.  In production you won't see it without enabling it in production.rb.
* It does not give you the level of detail (or cross-request metrics) you can get from a product like Skylight or New Relic.  I'd suggest using it for quick-checks mainly, e.g.: "huh active record is crazy on this one, maybe I should take a look at my queries."
* Does not exist prior to Rails 7.

timings
From top-left going clockwise: Chrome, Firefox, Safari.  You can see its pretty much the same.

lib/view_component/config.rb
joelhawksley: > I just noticed that instrumentation_enabled shows in the API.md doc file. Should I add this key there as well?

API.md is auto-generated from YARD comments when we deploy the docs, so that should take care of itself 😄

docs/guide/instrumentation.md
joelhawksley, line 33:
```suggestion
When using `render.view_component` with `config.server_timing = true` (default in development) in Rails 7, the browser developer tools display the sum total timing information in Network > Timing under the key `render.view_component`.
```

joelhawksley: How about we just say "browser here"? Perhaps we should add this screenshot to the docs?

joelhawksley:
@tgaff it looks like we might have some missing test coverage here. Care to investigate?

tgaff:
I've been trying to get coverage on that deprecation warning and have run into a bit of a wall.  Maybe someone has advice.

In `engine_test.rb` something like this works for testing it:
```rb
  def test_deprecation_notice_when_use_deprecated_instrumentation_name
    with_config_option(:instrumentation_enabled, true) do
      with_config_option(:use_deprecated_instrumentation_name, true) do
        assert_deprecated(/!render.view_component/, ViewComponent::Deprecation) do
          ActiveSupport.run_load_hooks(:view_component, self)
        end
      end
    end
  end

  def test_no_deprecation_notice_when_not_using_deprecated_name
    with_config_option(:instrumentation_enabled, true) do
      with_config_option(:use_deprecated_instrumentation_name, false) do
        assert_not_deprecated(ViewComponent::Deprecation) do
          ActiveSupport.run_load_hooks(:view_component, self)
        end
      end
    end
  end
```

However it's got a pretty big down-side of causing the other instrumentation tests to fail in an order-dependent way and I suspect having all the initializer hooks (in engine.rb) run twice could also cause unexpected issues elsewhere in the test suite too (someday).  I've been trying to find a way to bind those initializers to another context but I haven't been able to come up with anything that actually works.  
I've even tried running just those two hooks, bound to a fresh Rails app but it still poisons the suite.  

Any ideas?

joelhawksley:
> I've been trying to find a way to bind those initializers to another context to test but I haven't been able to come up with anything that actually works. I've even tried running just the two config and instrumentation hooks, bound to a fresh Rails app but it still poisons the suite.

😅 sounds like we should `:nodoc:` this section then. Thanks for digging in so thoroughly. Perhaps you could include a summary comment of what you said here with the nodoc?

tgaff:
> 😅 sounds like we should `:nodoc:` this section then. Thanks for digging in so thoroughly. Perhaps you could include a summary comment of what you said here with the nodoc?

I couldn't find a better alternative. 🙁   `nocov` added around that section - with comment.

docs/api.md
```diff
@@ -342,14 +349,6 @@ with_request_url("/users/42") do
 end
 ```
 
-To use a specific host, pass the host param:
-
-```ruby
-with_request_url("/users/42", host: "app.example.com") do
-  render_inline(MyComponent.new)
-end
-```
```
rmacklin, line 351: It looks like this may have been inadvertently deleted by regenerating the API docs, which happened because it seems this part was manually added directly in `api.md` in https://github.com/ViewComponent/view_component/pull/1773, rather than in the corresponding YARD comment.

rmacklin: I've opened a PR here to restore this example here: https://github.com/ViewComponent/view_component/pull/1794

oliverguenther:
> Also I haven't proven it out yet, but I think there's a chance that nested components could also suffer from this bug that I reported over on the Rails repo: https://github.com/rails/rails/issues/48375

I can confirm that this happens. The logged render statements for an exemplary stacked ViewComponent (lots of nested component calls) is 

`:I, [2023-09-14T11:04:59.370006 #44280]  INFO -- :   Rendered Meetings::ShowComponent within layouts/base (Duration: 226.1ms | Allocations: 215335)`

The reported timing is ~6s, which is the sum of all component instrumentation calls combined. But as they are nested, inner calls should be excluded.

Bildschirmfoto 2023-09-14 um 11 14 27

PR #1773, opened 2023-06-15
Add host option to `with_request_url` test helper

pbstriker38:
- This allows testing components that use path helpers that are constrained by a subdomain in routes.rb.

We have routes that are constrained by a subdomain in our routes file. If the requests host does not match the constraint then a `ActionController::RoutingError: No route matches "..."` is raised.

`config/routes.rb`
```ruby
Rails.application.routes.draw do
  constraints subdomain: 'app' do
    resources: :products
  end
end
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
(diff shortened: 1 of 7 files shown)


joelhawksley (approved):
👏🏻 thanks for the clean PR!

Issue #1741, opened 2023-05-05
ViewComponent::Base.config.view_component_path nil error


geekdreamzz:
Hi - I recently updated to 3.0.0 and suddenly started getting this error on all my components:

ActionView::Template::Error (no implicit conversion of nil into String)

When I go into the stack trace I see it errors out at https://github.com/ViewComponent/view_component/blob/main/lib/view_component/base.rb#L452 
`Regexp.quote(ViewComponent::Base.config.view_component_path)`

I'm not sure what ViewComponent::Base.config.view_component_path is for? but it's nil for me and this Regexp.quote method throws the error. 

To get back up and running I just set ViewComponent::Base.config.view_component_path = '' and I'm good, but it's clearly a workaround. 

TBH, I'm not sure if I'm using the gem the intended way but I essentially create my own classes that inherit from ViewComponent::Base. I could send a PR to account for nil ViewComponent::Base.config.view_component_path but it would just mask the issue. I'm a bit out of my wheelhouse trying to debug the underlying issue, I'm not able to follow what the comments here imply:

`# Removes the first part of the path and the extension.
        child.virtual_path = child.source_location.gsub(
          /(.*#{Regexp.quote(ViewComponent::Base.config.view_component_path)})|(\.rb)/, ""
        )`

I'd be happy to contribute to the gem if you want to give some feedback on what ViewComponent::Base.config.view_component_path is for and what child.virtual_path is designed for.

My base class that inherits from ViewComponent::Base also has a config so it's prob something dumb on my end. You could also just disregard this but figured I share the error in case it happens to anyone else.

reeganviljoen:
so `ViewComponent::Base.config.view_component_path` is used to specify where to loacte compkents if teh default app/componets isnt used

it Defaults to `nil`. If this is falsy, `app/components` is used.

reeganviljoen:
If you want to use a parent component generate a component with this `bin/rails generate component Example title content --parent MyBaseComponent`

GyozaGuy:
I'm seeing this error as well as of today after updating my gems.

@reeganviljoen your explanation makes sense, but the app won't even run without manually setting the config.

Spone:
@boardfish mind looking into this?

reeganviljoen:
@thank you @GyozaGuy I see

reeganviljoen:
@GyozaGuy @Spone @boardfish this pr seems to have caused the issue https://github.com/ViewComponent/view_component/pull/1659

[...]

PR #1774, opened 2023-06-18
Initialize ViewComponent::Config with defaults before framework load

boardfish:
This PR introduces `ViewComponent::Config.current`, which should not be used directly by consumers of ViewComponent, but is used as the common object from which `ViewComponent::Base` and `Rails.application.config.view_component` will load their config once each is ready. It'll make config available from the very start with the intent of not relying upon the framework to find these options - we can manually set anything reliant on Action View, for example, within the app or engine config.

#1659 made the config object empty before ViewComponent and Rails were initialised. This would mean that code within ViewComponent that relied on configured defaults would break, resulting in issues like #1741. Since ViewComponent::Config can be loaded independently of Action View, this should hopefully no longer be the case.

If you were affected by #1741, please test against this branch to see if it fixes your issue. Thanks for your patience!


camertron (commented):
Hey @boardfish thanks for working on this. I don't understand how `current` get set - there doesn't appear to be any call to it?

Spone:
> Hey @boardfish thanks for working on this. I don't understand how `current` get set - there doesn't appear to be any call to it?

I think the `defaults` method is called here:

```rb
class_attribute :current, default: defaults, instance_predicate: false
```

boardfish:
Hey folks, is there anything else you'd like me to do with this so we can get it shipped? I'm aware there are now multiple issues that this will fix.

lib/view_component/preview.rb
```diff
@@ -4,7 +4,7 @@
 
 module ViewComponent # :nodoc:
   class Preview
-    include Rails.application.routes.url_helpers if defined?(Rails.application.routes.url_helpers)
+    include Rails.application.routes.url_helpers if Rails && Rails.application.present?
```
BlakeWilliams, line 7: Do we still need the `defined?` check to support non-Rails environments?

BlakeWilliams (approved):

lib/view_component/preview.rb
boardfish: Yeah, I think this must've fallen through in a rebate 🧐 I'll make sure to fix it before it gets merged.

camertron:
@Spone looks like there's a coverage issue as well 😦

Spone:
Do you know what's happening with the coverage @boardfish?

boardfish:
I think it wants `ViewComponent::Base#config=` to be tested, but I think that method might even be able to be removed.

ryansdwilson:
Thank you all for your work on this! I ran into this issue today and went down a rabbit hole trying to figure it out. I really appreciate all the work you're doing to maintain VC.

camertron (approved):

boardfish:
Thanks for wrapping this up @camertron, and sorry I haven't been able to dedicate the time to it! 😅 🙏

PR #1777, opened 2023-06-20
detect closest match preview path when namespaced previews exist

reedrolemodel:

When using components without the `Component` suffix there can be ambiguous preview paths. For instance, if there is a [`milestone`](https://github.com/reedrolemodel/view_component-bug-replica/blob/master/app/components/milestone.rb) and [`milestone/card`](https://github.com/reedrolemodel/view_component-bug-replica/blob/master/app/components/milestone/card.rb), both are candidates when rendering a [`:with_default_title`](https://github.com/reedrolemodel/view_component-bug-replica/blob/master/test/components/previews/milestone/card_preview.rb) preview. [app/controllers/concerns/view_component/preview_actions.rb#62](https://github.com/ViewComponent/view_component/blob/72b4f7de53d0834f5b9bf1ee83bee5ff79e2a09b/app/controllers/concerns/view_component/preview_actions.rb#L62) always detects the first candidate, `milestone`, which is the wrong one when testing `milestone/card`.

Reverse sorting the candidate paths by length results in the longest match which means we get the most specific one.

Depending on the project, this bug can result in `NameError` if the preview method doesn't exist in the other preview class. In [this reproduction](https://github.com/reedrolemodel/view_component-bug-replica/) the test renders the wrong preview text:

```
Failure:
Milestone::CardTest#test_render_preview [/home/reed/src/view_component-bug-replica/test/components/milestone/card_test.rb:7]:
expected to find text "Milestone card component default" in "\n  \n    ViewComponentBugReplica\n    \n    \n    \n\n    \n  \n\n  \n        Milestone component default\n\n\n\n\n  \n"
```

### DIFF
--- a/app/controllers/concerns/view_component/preview_actions.rb
+++ b/app/controllers/concerns/view_component/preview_actions.rb
@@ -59,7 +59,7 @@ module ViewComponent
     def find_preview
       candidates = []
       params[:path].to_s.scan(%r{/|$}) { candidates << Regexp.last_match.pre_match }
-      preview = candidates.sort_by { |candidate| -candidate.length }.detect { |candidate| ViewComponent::Preview.exists?(candidate) }
+      preview = candidates.sort_by(&:length).reverse_each.detect { |candidate| ViewComponent::Preview.exists?(candidate) }
 
       if preview
         @preview = ViewComponent::Preview.find(preview)
--- a/app/controllers/concerns/view_component/preview_actions.rb
+++ b/app/controllers/concerns/view_component/preview_actions.rb
@@ -59,7 +59,7 @@ module ViewComponent
     def find_preview
       candidates = []
       params[:path].to_s.scan(%r{/|$}) { candidates << Regexp.last_match.pre_match }
-      preview = candidates.detect { |candidate| ViewComponent::Preview.exists?(candidate) }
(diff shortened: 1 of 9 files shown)


app/controllers/concerns/view_component/preview_actions.rb
```diff
@@ -59,7 +59,7 @@ def show_previews?
     def find_preview
       candidates = []
       params[:path].to_s.scan(%r{/|$}) { candidates << Regexp.last_match.pre_match }
-      preview = candidates.detect { |candidate| ViewComponent::Preview.exists?(candidate) }
+      preview = candidates.sort_by { |candidate| -candidate.length }.detect { |candidate| ViewComponent::Preview.exists?(candidate) }
```
Spone, line 62:
```suggestion
      preview = candidates.sort_by(&:length).reverse_each.detect { |candidate| ViewComponent::Preview.exists?(candidate) }
```

May be a bit clearer?

Spone (commented):
Thanks for your contribution! Can you please add a test and a README entry for this?

reedrolemodel:
I added a test that fails on `main` with
```
RenderPreviewTest#test_render_preview_unsuffixed:
NameError: undefined method `other' for class `UnsuffixedPreview'
```

Spone (approved):

PR #1781, opened 2023-06-23
use Inline templates in docs

joelhawksley:
This change updates our docs to lean on inline templates as they greatly simplify our example code <3

I've intentionally omitted a changelog.


Spone:
Do you think we should add an option for this to the generator? (Given that the `--inline` flag is already used.)

Maybe `--inline-template`?

joelhawksley:
@Spone filed! https://github.com/ViewComponent/view_component/issues/1792

PR #1794, opened 2023-07-07
Restore API documentation example for with_request_url's `host` param

rmacklin:

Addresses https://github.com/ViewComponent/view_component/pull/1771#pullrequestreview-1519743265

It looks like this may have been inadvertently deleted by regenerating the API docs. That happened because this example was manually added directly in `api.md` in https://github.com/ViewComponent/view_component/pull/1773, rather than in the corresponding YARD comment. To fix this, we've added the example to the YARD comment and regenerated the API docs.

I did not include a changelog entry since this is just basically just reverting an inadvertent change from the most recent PR.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -349,6 +349,14 @@ with_request_url("/users/42") do
 end
 ```
 
+To use a specific host, pass the host param:
+
+```ruby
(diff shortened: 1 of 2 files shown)


joelhawksley:
Thanks Richard! hope you are well ❤️

PR #1805, opened 2023-07-18
Allow setting method when using the  test helper

aduth:

We have components whose logic varies depending based on the request method, which is not easy to test with the current `with_request_url` helper.

This adds a new `method` option to `with_request_url`. This follows the pattern previously established with the `host` option which exists to similarly specify the `host` associated with the test request.

### DIFF
--- a/lib/view_component/test_helpers.rb
+++ b/lib/view_component/test_helpers.rb
@@ -163,10 +163,20 @@ module ViewComponent
     # end
     # ```
     #
+    # To specify a request method, pass the method param:
+    #
+    # ```ruby
+    # with_request_url("/users/42", method: "POST") do
+    #   render_inline(MyComponent.new)
+    # end
+    # ```
+    #
     # @param path [String] The path to set for the current request.
     # @param host [String] The host to set for the current request.
-    def with_request_url(full_path, host: nil)
+    # @param method [String] The request method to set for the current request.
+    def with_request_url(full_path, host: nil, method: nil)
       old_request_host = vc_test_request.host
+      old_request_method = vc_test_request.request_method
       old_request_path_info = vc_test_request.path_info
       old_request_path_parameters = vc_test_request.path_parameters
       old_request_query_parameters = vc_test_request.query_parameters
@@ -177,6 +187,7 @@ module ViewComponent
       vc_test_request.instance_variable_set(:@fullpath, full_path)
       vc_test_request.instance_variable_set(:@original_fullpath, full_path)
       vc_test_request.host = host if host
+      vc_test_request.request_method = method if method
       vc_test_request.path_info = path
       vc_test_request.path_parameters = Rails.application.routes.recognize_path_with_request(vc_test_request, path, {})
       vc_test_request.set_header("action_dispatch.request.query_parameters",
@@ -185,6 +196,7 @@ module ViewComponent
       yield
     ensure
       vc_test_request.host = old_request_host
+      vc_test_request.request_method = old_request_method
       vc_test_request.path_info = old_request_path_info
       vc_test_request.path_parameters = old_request_path_parameters
       vc_test_request.set_header("action_dispatch.request.query_parameters", old_request_query_parameters)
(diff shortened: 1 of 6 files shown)


lib/view_component/test_helpers.rb
```diff
@@ -175,6 +184,7 @@ def with_request_url(path, host: nil)
 
       path, query = path.split("?", 2)
       vc_test_request.host = host if host
+      vc_test_request.request_method = method if method
```
Spone, line 187: I think the previous value should be reverted in the `ensure` block below.

BlakeWilliams:
> We have components whose logic varies depending based on the request method, which is not easy to test with the current `with_request_url` helper.

Can you share an example? I'd like to understand the use case better before we move forward here.

lib/view_component/test_helpers.rb
aduth: > I think the previous value should be reverted in the `ensure` block below.

Good call! Updated in 3f5e2df56751d6162c95b66764e89b36d1636b21.

aduth:
> > We have components whose logic varies depending based on the request method, which is not easy to test with the current `with_request_url` helper.
> 
> Can you share a real world example/use case? I'd like to understand the use case better before we move forward here.

Sure thing. We have a component which renders a set of navigation links and highlights the current page. It's implemented by comparing the current request's handler (`request[:controller]` and `request[:action]`) against what would be the handler for a given link, using [`recognize_path`](https://apidock.com/rails/ActionDispatch/Routing/RouteSet/recognize_path).

There's some challenge in forms with both a `new` and `create` action, using the same path and `GET` and `POST` methods respectively. We want to `render :new` for an invalid form. To support this, the component needs to consider the request method when calling `recognize_path` to be able to compare against the current request method. It's currently difficult to simulate this in tests for the component.

Maybe easier to share the actual code, effect, and our current workaround:

- https://github.com/18F/identity-idp/pull/8800
- https://github.com/18F/identity-idp/pull/8800#discussion_r1267194842

lib/view_component/test_helpers.rb
```diff
@@ -163,10 +163,20 @@ def with_controller_class(klass)
...
+    # with_request_url("/users/42", method: "POST") do
+    #   render_inline(MyComponent.new)
+    # end
+    # ```
+    #
     # @param path [String] The path to set for the current request.
     # @param host [String] The host to set for the current request.
-    def with_request_url(path, host: nil)
+    # @param method [String] The request method to set for the current request.
+    def with_request_url(path, host: nil, method: nil)
```
reeganviljoen, line 177: maybe rename method to `http_method` for clarlrty

Spone: It's indeed a clearer name but I'm a bit reluctant to rename the keyword since it would break people's code.

Spone: I'm sorry it's my first review since I got back from vacation and I guess I'm not 100% ready to work yet 🫣
I did not realize this was a new option introduced by this PR. 😅
So yeah, I'm okay with renaming it if others think it's clearer too.

reeganviljoen: @Spone @aduth considering your responses I feel sufficient context is provided

reeganviljoen:
@Spone @aduth any updates on this pr

aduth:
@reeganviljoen From my perspective, I'm expecting this to be ready for review as-is. Based on the discussion at https://github.com/ViewComponent/view_component/pull/1805#discussion_r1307013023, I don't have a clear sense that there's a suggested rename, and I'm reasonably happy with the current naming absent any objections.

I rebased the branch to resolve merge conflicts that existed due to drift.

Happy to make any revisions if suggestions are made!

reeganviljoen:
@Spone any insights ?

Spone:
Looks good to me. Maybe @BlakeWilliams can review this too?

@aduth can you please resolve the conflicts?

Spone (approved):

camertron (approved):

PR #1827, opened 2023-08-24
Fix API documentation and revert unnecessary change in `preview.rb`

rmacklin:
Addresses https://github.com/ViewComponent/view_component/pull/1774#pullrequestreview-1593279882

### DIFF
--- a/lib/view_component/preview.rb
+++ b/lib/view_component/preview.rb
@@ -4,7 +4,7 @@ require "active_support/descendants_tracker"
 
 module ViewComponent # :nodoc:
   class Preview
-    include Rails.application.routes.url_helpers if defined?(Rails) && Rails.application.present?
+    include Rails.application.routes.url_helpers if defined?(Rails.application.routes.url_helpers)
     include ActionView::Helpers::TagHelper
     include ActionView::Helpers::AssetTagHelper
     extend ActiveSupport::DescendantsTracker
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Fix API documentation and revert unnecessary change in `preview.rb`.
+
+    *Richard Macklin*
+
 * Initialize ViewComponent::Config with defaults before framework load.
 
     *Simon Fish*
--- a/docs/api.md
+++ b/docs/api.md
@@ -14,11 +14,6 @@ nav_order: 3
 
 Returns the current config.
 
-### `.config=(value)`
-
-Replaces the entire config. You shouldn't need to use this directly
-unless you're building a `ViewComponent::Config` elsewhere.
-
 ### `.sidecar_files(extensions)`
 
 Find sidecar files for the given extensions.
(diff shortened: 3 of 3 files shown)


rmacklin:
cc @boardfish @camertron

lib/view_component/preview.rb
```diff
@@ -4,7 +4,7 @@
 
 module ViewComponent # :nodoc:
   class Preview
-    include Rails.application.routes.url_helpers if defined?(Rails) && Rails.application.present?
+    include Rails.application.routes.url_helpers if defined?(Rails.application.routes.url_helpers)
```
camertron, line 7: What is the intent behind the guard here? I thought it was to protect against including URL helpers in non-Rails setups (which already doesn't make sense to me since view_component is very, very dependent on Rails). If that's the case, then how does `defined?(Rails.application.routes.url_helpers)` help? Won't a `NameError` be raised because the `Rails` constant doesn't exist?

rmacklin: > What is the intent behind the guard here? I thought it was to protect against including URL helpers in non-Rails setups (which already doesn't make sense to me since view_component is very, very dependent on Rails).

The intent behind the guard here is to avoid an exception when `preview.rb` is required in an environment that does not (or not yet) have a Rails application. I think the main use case is inside a gem, where the test environment may not go through the lengths to set up a complete dummy application. (Though even in view_component's own `docs:build` rake task there is no `Rails.application`, which is how I personally encountered the exception.)

> If that's the case, then how does `defined?(Rails.application.routes.url_helpers)` help? Won't a `NameError` be raised because the `Rails` constant doesn't exist?

No, `defined?(Rails.application.routes.url_helpers)` will return `nil`. Here's a clean `irb` session:
```
$ irb
irb(main):001:0> defined?(Rails.application.routes.url_helpers)
=> nil
irb(main):002:0> Rails
(irb):2:in `<main>': uninitialized constant Rails (NameError)
        from /home/rmacklin/.asdf/installs/ruby/3.0.2/lib/ruby/gems/3.0.0/gems/irb-1.3.5/exe/irb:11:in `<top (required)>'
        from /home/rmacklin/.asdf/installs/ruby/3.0.2/bin/irb:23:in `load'
        from /home/rmacklin/.asdf/installs/ruby/3.0.2/bin/irb:23:in `<main>'
```

The main point here is that the change being reverted in daf395c98d7605660db5a17d42977ab9120fa2be was unnecessary due to an improper git rebase conflict resolution, and it is technically less safe since it no longer handles the case where `Rails.application` is present (perhaps stubbed in the test environment of a gem) but `Rails.application.routes.url_helpers` is not. As mentioned in https://github.com/ViewComponent/view_component/pull/1774#discussion_r1304706189, I do agree that is probably a rare scenario, but nonetheless it doesn't seem worth changing this line to begin with given that the previous guard offered broader/more accurate protection from potential exceptions.

camertron: Great explanation, thanks @rmacklin!

camertron (approved):

Issue #1825, opened 2023-08-21
bug: Slots with names that start with the characters `call` raise an exception


wenley:
### Steps to reproduce

1. Define a ViewComponent with a `renders_one` slot
2. Name the slot with the `call` prefix (e.g. `callout_action`)
3. Use the new component in a template (i.e. `render(NewComponent.new(...))`). It does not matter whether the slot is used or not.
4. Load the page that uses the new component

### Expected behavior
The page should render.

### Actual behavior
A `NameError` "undefined method" exception is thrown, causing the page to fail to load.

Here is a PR with a failing test: https://github.com/ViewComponent/view_component/pull/1821

**Backtrace** (from [this PR CI run](https://github.com/ViewComponent/view_component/actions/runs/5872234737/job/15923396689?pr=1821)):
```
/home/runner/work/view_component/view_component/lib/view_component/compiler.rb:99:in `instance_method': undefined method `call_callout_title' for class `EmptySlotComponent' (NameError)
	from /home/runner/work/view_component/view_component/lib/view_component/compiler.rb:99:in `block in define_render_template_for'
	from /home/runner/work/view_component/view_component/lib/view_component/compiler.rb:97:in `map'
	from /home/runner/work/view_component/view_component/lib/view_component/compiler.rb:97:in `define_render_template_for'
	from /home/runner/work/view_component/view_component/lib/view_component/compiler.rb:84:in `compile'
	from /home/runner/work/view_component/view_component/lib/view_component/base.rb:512:in `compile'
	from /home/runner/work/view_component/view_component/lib/view_component/engine.rb:75:in `each'
	from /home/runner/work/view_component/view_component/lib/view_component/engine.rb:75:in `block (2 levels) in <class:Engine>'
	from /home/runner/work/view_component/view_component/vendor/bundle/ruby/2.7.0/gems/activesupport-5.2.6/lib/active_support/lazy_load_hooks.rb:71:in `instance_eval'
	from /home/runner/work/view_component/view_component/vendor/bundle/ruby/2.7.0/gems/activesupport-5.2.6/lib/active_support/lazy_load_hooks.rb:71:in `block in execute_hook'
	from /home/runner/work/view_component/view_component/vendor/bundle/ruby/2.7.0/gems/activesupport-5.2.6/lib/active_support/lazy_load_hooks.rb:62:in `with_execution_control'
[...]

PR #1828, opened 2023-08-26
Fix slot names that start with call

BlakeWilliams:
When a component is registered that starts with `call` the compiler
detects it as a template method which is not correct. This changes the
compiler to look for `call(_|$)` instead of `call` to avoid this
issue.

This means that slots can't (and couldn't) start with `call_` since we
still rely on the `call_` naming convention to generate the
template methods. To make this more dev friendly, this adds a check
that raises an error if a slot name starts with `call_`.

A more long-term fix would be to use some kind of template method
container instead of relying on the `call_` naming convention.

Fixes https://github.com/ViewComponent/view_component/issues/1825


reeganviljoen:
@BlakeWilliams maybe adding the fact that slot names cant start with `call` should be added to known issues in the docs so users can better troubkle shoot this issue going forward

test/sandbox/app/components/header_component.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
```
camertron, line 1: Is this used anywhere?

camertron (approved):
LGTM! To echo @reeganviljoen's point tho, would this also be a good thing to _call_ out in the docs (see what I did there?!)

davidgumberg:
Minor nit, could add a test that the #1825 case is fixed:

```ruby
def test_slot_names_can_start_with_call
  assert_nothing_raised do
    Class.new(ViewComponent::Base) do
      renders_one :callhome_et
    end
  end
end
```

Otherwise LGTM 👍

reeganviljoen:
@BlakeWilliams any progress on this ?

camertron:
@reeganviljoen I can take this one over and get it across the finish line 😄

BlakeWilliams:
I haven't had a ton of time to dedicate to this recently and documentation is always the hardest part. 🙂 

Thanks for the ping @reeganviljoen and @camertron for getting this shipped! 🚀

Issue #1836, opened 2023-08-30
Rename `helpers` to `globals`


reeganviljoen:
Hey ViewCompoent community  👋

# The problem 
storing the global view_context in a method called helpers is unintuitive for beginners, and even for experienced developers a method should describe what it does, which helpers barely does and not well at it. I was thinking a name like globals would accurately describe what it does and maybe even make it more of a thought process before coupling global state into a component which is generally a bad idea and should be used as a lasts resort.

I know that this would break a lot of existing code so I propose that this change be looked at for version 4

reeganviljoen:
@Spone @boardfish @BlakeWilliams @camertron any thoughts , all opinions welcome

camertron:
Hey @reeganviljoen, thanks for bringing this up! I believe it's called `#helpers` is to mimic Rails, which [introduced the `#helpers` method in controllers in Rails 5](https://www.bigbinary.com/blog/rails-add-helpers-method-to-ease-usage-of-helper-modules-in-controllers#new-way-to-call-helper-methods-in-rails-5). In fact, the [original implementation](https://github.com/rails/rails/pull/24866/files#diff-6b77af243a23efcde71c16accfd0b56c643a129fa1a725e5ca1b30e0ee06b994R119) of `#helpers` in Rails simply returned `@view_context`, just like we're doing in ViewComponent.

I'm not sure `#globals` is the right name, since the idea behind `#helpers` is to make it obvious how to call helper methods in component classes. I agree there is more functionality available on the view context than just helpers, but if we'd like to signal that functionality to framework consumers, I would propose adding another method that also returns the view context instead of changing the existing name.

reeganviljoen:
@camertron I get that this is something we are doing to be consistent with rails but I think the context is very different. In controllers and views it makes sense because all state is global state to so the naming makes sense.In ViewComponent however we have local and global state and the name helpers doesn't tell me hey if you need global state look here which as led me to personally experienced people people being confused about the purpose of helpers because the naming is unclear, so many people just dont use and pass in objects as a params , i.e (`current_user` in devise) when its already present in `helpers`

reeganviljoen:
I would however happy keeping `helpers` and maybe adding `globals` as a method that wraps around `helpers` and add a note that `globals` is exactly that but is their for clarity

boardfish:
> so many people just dont use and pass in objects as a params , i.e (`current_user` in devise) when its already present in `helpers`

[...]

PR #1841, opened 2023-09-02
Refer to helpers in `NameError` response in development/test

boardfish:

We identified in #1836 that some things might influence whether the developer is aware of methods provided by the helper layer, particularly in the case where they may have used the help of libraries like Devise and Kaminari to get going. In order to make things easier when you're getting started like this, this PR adds extra context to `NameError`s raised by method calls that mirror something the view context already responds to.

`method_missing` seemed like a natural fit here, but as suggested by @camertron, I've conditionally defined this behaviour for development and test environments only.

Additionally, the source location is removed from the backtrace – this was causing another test to fail.

Worth noting that I needed to `define_instance_method` in order to stub in the test, since calling MiniTest's `stub` would raise an error because view contexts don't naturally define `current_user`.

I wonder if there's anything more we can add here? A link to some docs, perhaps?

Also, it's worth noting that for older versions of Ruby, `DidYouMean` appends its suggestion a couple of times (notice `Did you mean?   current_page?` showing up twice in the below logs):

```rb
[1] pry(#<ViewComponent::Base::UnitTest>)> exception.message
=> "undefined local variable or method `current_user' for #<ReferencesMethodOnHelpersComponent:0x0000000151a98560>\nDid you mean?  current_page?\n\nYou may be trying to call a method provided as a view helper. Did you mean `helpers.current_user'?\nDid you mean?  current_page?"
```

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Refer to `helpers` in `NameError` message in development and test environments.
+
+    *Simon Fish*
+
 * Fix API documentation and revert unnecessary change in `preview.rb`.
 
     *Richard Macklin*
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -219,6 +219,21 @@ module ViewComponent
       @__vc_helpers ||= __vc_original_view_context || controller.view_context
     end
 
+    if Rails.env.development? || Rails.env.test?
+      def method_missing(method_name, *args) # rubocop:disable Style/MissingRespondToMissing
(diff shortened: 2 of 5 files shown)


reeganviljoen:
```
I wonder if there's anything more we can add here? A link to some docs, perhaps?
```
@boardfish I think that would be unnecessary as this is something I feel documents itself pretty well

reeganviljoen (approved):

Spone (approved):
This is awesome Simon, I'm a big fan of errors that point you in the right direction!

BlakeWilliams (approved):
Love it! :shipit:

PR #1849, opened 2023-09-11
Support devise using ViewComponent::TestCase

fugufish:

This updates ViewComponent::TestHelpers to better detect and support Devise. It resolves the issue described in this discussion https://github.com/ViewComponent/view_component/discussions/371.

Detect whether or not `Devise::Test::ControllerHelpers`, and if so, set up the `@request` instance variable and include `Devise::Test::ControllerHelpers`. The `@request` variable MUST be present before Devise's own setup hook is called, the only way to ensure this is to create setup initializing the controller and setting the `@request` instance variable before inclusion.

I'd like not to have to add janky `included` callbacks, but this is the only way that seems to work with Devise.

### DIFF
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -164,17 +164,6 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
-## Testing with Devise
-
-Testing with Devise requires you include `Devise::Test::ControllerHelpers` in the `ViewComponent::TestCase` class. This
-is typically done by adding the following to `test/test_helper.rb`:
-
-```ruby
-class ViewComponent::TestCase
-  include Devise::Test::ControllerHelpers
-end
-```
-
 ## Setting `request.path_parameters`
 
 Since 2.31.0
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -164,6 +164,17 @@ class ExampleComponentTest < ViewComponent::TestCase
 end
 ```
 
+## Testing with Devise
+
+Testing with Devise reqiures you include `Devise::Test::ControllerHelpers` in the `ViewComponent::TestCase` class. This
+is typically done by adding the following to `test/test_helper.rb`:
+
+```ruby
+class ViewComponent::TestCase
+  include Devise::Test::ControllerHelpers
+end
+```
+
 ## Setting `request.path_parameters`
 
 Since 2.31.0
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -166,7 +166,7 @@ end
 
 ## Testing with Devise
 
-Testing with Devise reqiures you include `Devise::Test::ControllerHelpers` in the `ViewComponent::TestCase` class. This
+Testing with Devise requires you include `Devise::Test::ControllerHelpers` in the `ViewComponent::TestCase` class. This
 is typically done by adding the following to `test/test_helper.rb`:
 
 ```ruby
--- a/lib/view_component/test_helpers.rb
+++ b/lib/view_component/test_helpers.rb
@@ -230,6 +230,11 @@ module ViewComponent
       super
     end
 
+    def teardown
+      super
+      @vc_render_preview_controller = nil
+    end
+
     # Note: We prefix private methods here to prevent collisions in consumer's tests.
     private
 
--- a/lib/view_component/test_helpers.rb
+++ b/lib/view_component/test_helpers.rb
@@ -103,9 +103,9 @@ module ViewComponent
     #
     # assert_text("Hello, World!")
     # ```
-    def render_in_view_context(*, &block)
+    def render_in_view_context(*args, &block)
       @page = nil
-      @rendered_content = vc_test_controller.view_context.instance_exec(*, &block)
+      @rendered_content = vc_test_controller.view_context.instance_exec(*args, &block)
       Nokogiri::HTML.fragment(@rendered_content)
     end
     ruby2_keywords(:render_in_view_context) if respond_to?(:ruby2_keywords, true)
--- a/lib/view_component/test_helpers.rb
+++ b/lib/view_component/test_helpers.rb
@@ -21,11 +21,25 @@ module ViewComponent
       if ENV["DEBUG"]
         warn(
           "WARNING in `ViewComponent::TestHelpers`: Add `capybara` " \
-          "to Gemfile to use Capybara assertions."
+            "to Gemfile to use Capybara assertions."
         )
       end
 
       # :nocov:
+    end
+
+    def self.included(base)
+      if defined?(Devise::Test::ControllerHelpers)
+        # the `@request` instance variable is used by Devise::Test::ControllerHelpers and must
+        # be in a `setup` block before including `Devise::Test::ControllerHelpers`
+        base.instance_eval do
+          setup do
+            @request = __vc_render_preview_controller.request
+          end
+        end
+
+        base.include(Devise::Test::ControllerHelpers)
+      end
     end
 
     # Returns the result of a render_inline call.
@@ -75,7 +89,7 @@ module ViewComponent
     # @param params [Hash] Parameters to be passed to the preview.
     # @return [Nokogiri::HTML]
     def render_preview(name, from: __vc_test_helpers_preview_class, params: {})
-      previews_controller = __vc_test_helpers_build_controller(Rails.application.config.view_component.preview_controller.constantize)
+      previews_controller = __vc_render_preview_controller
 
       # From what I can tell, it's not possible to overwrite all request parameters
       # at once, so we set them individually here.
@@ -108,6 +122,7 @@ module ViewComponent
       @rendered_content = vc_test_controller.view_context.instance_exec(*args, &block)
       Nokogiri::HTML.fragment(@rendered_content)
     end
+
     ruby2_keywords(:render_in_view_context) if respond_to?(:ruby2_keywords, true)
 
     # Set the Action Pack request variant for the given block:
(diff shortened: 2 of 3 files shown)


lib/view_component/test_helpers.rb
```diff
@@ -21,13 +21,27 @@ def refute_component_rendered
...
+      if defined?(Devise::Test::ControllerHelpers)
+        # the `@request` instance variable is used by Devise::Test::ControllerHelpers and must
+        # be in a `setup` block before including `Devise::Test::ControllerHelpers`
+        base.instance_eval do
+          setup do
+            @request = __vc_render_preview_controller.request
+          end
+        end
+
+        base.include(Devise::Test::ControllerHelpers)
```
BlakeWilliams, line 41: I don't think we want to include third party modules into the ViewComponent test helper by default.

Can we detect if devise has already included these helpers or not?

lib/view_component/test_helpers.rb
```diff
@@ -103,11 +117,12 @@ def render_preview(name, from: __vc_test_helpers_preview_class, params: {})
     #
     # assert_text("Hello, World!")
     # ```
-    def render_in_view_context(*args, &block)
+    def render_in_view_context(*, &block)
       @page = nil
-      @rendered_content = vc_test_controller.view_context.instance_exec(*args, &block)
+      @rendered_content = vc_test_controller.view_context.instance_exec(*, &block)
```
BlakeWilliams, line 122: Why did this change?

lib/view_component/test_helpers.rb
fugufish: probably

lib/view_component/test_helpers.rb
fugufish: pretty certain Standard did that, in any case I can revert (may have been a config on my end).

lib/view_component/test_helpers.rb
fugufish: before I look for an alternative, I'd contend we do exactly that on line 8. Doing it this way would certainly make things less difficult on the end user.

BlakeWilliams: I'm not certain we need to include this manually at all. I looked at the devise source a bit and I think we could get away with something similar to what you have here, but maybe modified a bit. Thoughts?

```ruby
def before_setup
  super
  _vc_reset!
end

def _vc_reset!
  @request = ActionDispatch::TestRequest.create
  @request.session = ActionController::TestSession.new
  @request
end
```

fugufish: As it turns out, Devise and Warden hook rather deeply into the controller, and `ActionDispatch::TestRequest` wont provide the necessary context. I simplified my implementation to:

```ruby
 def before_setup
      @request = __vc_render_preview_controller.request
      super
 end
 ```

The request must be set up before `super` is called, otherwise a hook in the devise test helpers will fail.

File: docs/guide/testing.md [hunk 3]
```diff
@@ -164,6 +164,17 @@ class ExampleComponentTest < ViewComponent::TestCase
...
+
+Testing with Devise requires you include `Devise::Test::ControllerHelpers` in the `ViewComponent::TestCase` class. This
+is typically done by adding the following to `test/test_helper.rb`:
+
+```ruby
+class ViewComponent::TestCase
+  include Devise::Test::ControllerHelpers
+end
+```
+
```
BlakeWilliams, line 177:
```suggestion
```

I don't think we should be prescriptive here. I think the Devise docs are clear enough on how to include the test helpers where you need them.

lib/view_component/test_helpers.rb
```diff
@@ -244,5 +249,9 @@ def __vc_test_helpers_preview_class
     rescue NameError
       raise NameError, "`render_preview` expected to find #{result}, but it does not exist."
     end
+
+    def __vc_render_preview_controller
+      @vc_render_preview_controller ||= __vc_test_helpers_build_controller(Rails.application.config.view_component.preview_controller.constantize)
```
BlakeWilliams, line 254: This is slightly changing behavior since we're now memoizing this value. Is this okay to memoize or do we need to reset it each render? I _think_ we want to keep the old behavior and have it re-evaluate per-call to `render_preview` so each call has a fresh controller instance?

lib/view_component/test_helpers.rb
```diff
@@ -225,6 +225,11 @@ def vc_test_request
         end
     end
 
+    def before_setup
+      @request = __vc_render_preview_controller.request
```
BlakeWilliams, line 229: This isn't correct since this is creating the request from the preview controller, not using the same request the `request` method uses. Can we update this to use the proper request?

BlakeWilliams (commented):
Thanks for making those changes, I think this is getting close but there's a few small things I think we want to address first.

docs/guide/testing.md
fugufish: @BlakeWilliams I merged those changes, however, I think it might be a mistake to not add the documentation. ViewComponent "strays" off of the Rails golden path. While Devise does describe how to setup tests, it doesn't specify how to set them up for `ViewComponent::TestCase`, which again is not a typical implementation.

While most engineers who google around would be able to find the actual solution, many will not, leading to potentially erroneous bug reports. By adding this do the documentation, it will help people self solve.

BlakeWilliams: > However, I think it might be a mistake to not add the documentation. ViewComponent "strays" off of the Rails golden path

I don't think we do, but I also don't feel super strongly about removing the documentation although it does feel a bit out-of-place.

lib/view_component/test_helpers.rb
fugufish: I'm not sure I follow. On line 78  we have `previews_controller = __vc_render_preview_controller`, `#request` tracks to   `ViewComponent::Base#request` which memoizes the request.

If you debug the call you can see that the request objects are indeed the same.

`@request ->  #<ActionDispatch::TestRequest:0x000000010c675838>`
`preview_controller.request -> #<ActionDispatch::TestRequest:0x000000010c675838>`

These are identical. Am I missing something?

lib/view_component/test_helpers.rb
fugufish: Added a `#teardown` method that clears the variable.

docs/guide/testing.md
fugufish: Happy to let it lie, and it can be re added if it becomes an annoyance. Looking deeper at the low number of people complaining it is probably a non issue.

fugufish:
how can I retrigger test suite, looks like a flakey test

lib/view_component/test_helpers.rb
BlakeWilliams:
```suggestion
      @request = vc_test_request
```

That wasn't completely clear in hindsight, sorry about that. 

We shouldn't use `__vc_render_preview_controller.request` here because we don't need to couple the `@request` to the preview controller or how it's construction. Also not every test requires a preview controller to be present either, so we should skip instantiating one except where necessary.

lib/view_component/test_helpers.rb
BlakeWilliams: I mentioned it above, but not every test needs a preview controller. Can we revert this back to the previous, lazy loaded implementation to avoid extra work in test suites?

BlakeWilliams:
Not relevant to this PR but we really should address that flaky test. I can re-run it real quick.

lib/view_component/test_helpers.rb
fugufish: Actually we do. The Warden context is not correctly set up in `ActionDispatch::TestRequest.create`. I tested this myself with one  of your earlier suggestions.

https://github.com/ViewComponent/view_component/pull/1849#discussion_r1323312240

lib/view_component/test_helpers.rb
fugufish: Unfortunately to support devise I don't see this being possible. We can't predict when `render_preview` is going to be used. The entire problem is that to support devise, setup needs to happen before `render_preview` is ever called. If we don't memoize, then the test may still work in most cases, but keep in mind it wont be accurate, as you're warden request environ would be a different instance than your render_preview request environ.

One possible alternative that might meet your requirements is a new concern, which sets up a test environment for devise and `render_preview`. That way a user is actively opting in essentially stating "I am using this pattern", and then update documentation to the concern's existence.

Honestly if it weren't for the fact that this project already has a bunch of minitests built before I started working on it, I'd just move the whole thing over to rspec and be done with it :D

fugufish: Something like

```ruby
module ViewComponent::TestHelpers::DevisePreviewHelper
   def self.included(base)
      base.instance_eval do
        setup do
          @request = __vc_render_preview_controller.request
        end
         
        teardown do
          @request = nil
          @vc_render_preview_controller = nil
        end
          
      # ....
   end
    
    def __vc_render_preview_controller
      @vc_render_preview_controller ||= super
    end
      
    
end
```

lib/view_component/test_helpers.rb
BlakeWilliams: We don't always use the preview controller instance to render since users can pass their own controller class to us and `render_inline` uses its own controller.

What does Devise need the controller instance for?

lib/view_component/test_helpers.rb
BlakeWilliams: To help clarify things, are devise helpers only broken for previews or for previews + render_inline?

fugufish: Just checked, it works for `render_inline`. I believe it is because `render_preview` uses an actual controller instance, causing it to pick up Warden's hooks.

lib/view_component/test_helpers.rb
fugufish: https://github.com/heartcombo/devise/blob/f6e73e5b5c8f519f4be29ac9069c6ed8a2343ce4/lib/devise/test/controller_helpers.rb#L44

fugufish: @BlakeWilliams to be clear, this path is ONLY triggered if you call `current_user` in any of your components. Which is kindof edge case ish right now. I actually refactored so that current_user is only ever passed into components when they need it.

That said, I read somewhere that the long term goal of ViewComponent is that it would be used for page rendering as well, which will make this a lot less of an edge case.

fugufish:
@BlakeWilliams are we considering this PR dead on arrival?

joelhawksley:
@fugufish sorry for the lack of response on this PR. Would you be interested in bringing it up to date and adding a test case?

joelhawksley:
Closing as stale.

PR #1850, opened 2023-09-12
Fix flaky translation test

BlakeWilliams:
This replaces the call to a stubbed method with a call to the original
method. This should be more reliable and less likely to break during
CI runs.


camertron (approved):
OMG I can't believe you fixed this 😍

BlakeWilliams:
I *think* I fixed it but I'm feeling good about it 😅

Issue #614, opened 2021-02-10
Allow slots to be rendered by default


manuelpuyol:
## Feature request

It'd be helpful to allow _some_ slots to be rendered even when they aren't explicitly called in the template.
e.g:

rb:
```rb
class MyComponent < ViewComponent::Base
  renders_one :slot, required: true, "RequiredSlot"
  renders_one :other_slot, "NotRequiredSlot"

  class RequiredSlot < ViewComponent::Base
  end

  class NotRequiredSlot < ViewComponent::Base
  end
end
``` 

erb:
```erb
<div class="slot">
  <%= slot %>
</div>
<%= other_slot %>
```

call:
```erb
<%= render MyComponent.new %>
```

This should render `slot` but not `other_slot`

### Motivation

Some components **always** have to render a slot, so the burden goes to the user to make sure to always call the required slot in the call block. Since most of the times they will use the default arguments for the slot, the explicit call could be removed.

### Current workaround

We can use `before_render` to reach this behavior:

```rb
def before_render
  slot(args...) unless slot
end
```

hjhart:
Does anyone else have any other strategies for this? I suppose we could create an abstraction that does the same thing that @manuelpuyol, but this seems like a common enough use-case that we could have some documentation surrounding it in the Slots docs.

cc @joelhawksley

boardfish:
I think it's also possible to roll with `<%= slot || render(ExampleComponent.new) %>` in the template. This makes more sense to me because you're also declaring the default content to go in that place inside of the template.

joelhawksley:
@manuelpuyol what do you think of @boardfish's idea?

manuelpuyol:
I think @boardfish's idea is a good workaround, but I still feel like this should be native to slots to avoid duplication.
It's quite common to set parameters in the `renders_one/many` lambda call. Something like: 

https://github.com/primer/view_components/blob/482c7cfd70584fd3ede8be6963cf4a433b1ac5c2/app/components/primer/navigation/tab_component.rb#L20-L39

Using`<%= slot || render(ExampleComponent.new(arg1: val1, arg2: val2, ...) %>` makes it harder to sync all the desired parameters

Spone:
Would something like that solve the issue?

```rb
class MyComponent < ViewComponent::Base
  renders_one :slot, "RequiredSlot"
  renders_one :other_slot, "NotRequiredSlot"

  def slot
    super || render(RequiredSlot.new)
  end

[...]

PR #1861, opened 2023-09-25
Support `required: true` for renders_one slots

jalyna:
Hi everyone, first of all thanks for the amazing work! I really love working with view_component!

I'm very new to the project and hope I can contribute to this project. Looking forward to any feedback.

Trying to solve the issue described in #614

Add `required` option to `renders_one` to indicate that there is a default slot to be rendered, which automatically applies lambda options. This only works with lambda or Component slots (and not for passthrough slots). It can be used with polymorphic slots but exactly one specific option can be required, otherwise an error is shown.

This is just my first attempt and probably a lot of fine tuning is needed as this project is used by many people.

I added a new method `__vc_before_render` to the `Base` class in order to register a method called from the `Slotable` module. This way I was able to add defaults to the slots before `before_render` is called. I'm not sure if this is the best approach yet, but I'm happy to get any feedback.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add `required` option to `renders_one` to indicate that there is a default slot to be rendered, which automatically applies lambda options. This only works with lambda or Component slots (and not for passthrough slots). It can be used with polymorphic slots but exactly one specific option can be required, otherwise an error is shown.
+
+    *Jalyna Schröder*
+
 * Add Simundia to list of companies using ViewComponent.
 
     *Alexandre Ignjatovic*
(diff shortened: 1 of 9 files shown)


BlakeWilliams:
@jalyna Sorry for the delay here, and thanks for the PR!

I like the direction this is going, but I think there's a few tweaks we want to make before moving forward here. I think we should omit `required` as a keyword since that and providing a default value have different meanings. `required` should mean that a consumer _has_ to provide the slot value.

Instead, I'd propose that we have a conventional/automagical field like `default_line_item_icon` that when present, is called and used as the slots default value if the consumer didn't provide one.

e.g.

```ruby
class MyComponent
  renders_one :header

  # If a `with_header` is not called, this method will be called and the result will be used as the slots value
  def default_header
    MyHeaderComponent.new(title: "Hello world")
  end
end
```

This has a few advantages:

- The existing public slot API stays the same with no changes
- It doesn't require components to implement default arguments
- The method allows you to pass your own arguments to the default slot and can use component local state to make decisions on what those arguments should be.
- We can avoid having to add the internal `__vc_before_render` hook, and instead hook directly into the accessors generated by the slotable module.

Having default slot values is something we've wanted to implement for a while, so thanks for pushing on this!

joelhawksley:
@jalyna what do you think of @BlakeWilliams's suggestion?

PR #1864, opened 2023-10-02
Document the capture compatibility patch on the Known issues page

### LINKED ISSUE #1838: Capture compatibility patch isn't documented at Known Issues page
https://viewcomponent.org/known_issues.html doesn't mention enabling the capture compatibility patch as a solution to the first two issues. I'm unsure as to why, but it'd be good to add it.

I also think it'd be good to have some explanation there as to why the patch is something that needs to be enabled. (My memory of how it works is hazy, but I assume it's patching Action View.)

boardfish:
Closes #1838

References the patch as a solution to the issues it's known to fix, and includes the explanation we have in `CaptureCompatibility`.

### DIFF
--- a/docs/known_issues.md
+++ b/docs/known_issues.md
@@ -6,7 +6,21 @@ nav_order: 9
 
 # Known issues
 
-## turbo_frame_tag double rendering or scrambled HTML structure
+## Issues resolved by the optional capture compatibility patch
+
+[Set `config.view_component.capture_compatibility_patch_enabled` to `true`](https://viewcomponent.org/api.html#capture_compatibility_patch_enabled) to resolve these issues.
+
+These issues arise because the related features/methods keep a reference to the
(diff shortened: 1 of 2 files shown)


docs/known_issues.md
```diff
@@ -6,15 +6,29 @@ nav_order: 9
 
 # Known issues
 
-## turbo_frame_tag double rendering or scrambled HTML structure
+## Issues resolved by the optional capture compatibility patch
+
```
Spone, line 10: Maybe we should first define what "these issues" are, in terms of symptoms? (Duplicate content, wrongly-ordered HTML tags...)

boardfish:
```suggestion

If you're experiencing issues with duplicated content or malformed HTML output, the capture compatibility patch may resolve these.

```

Spone (approved):

PR #1868, opened 2023-10-10
fix: reduce UnboundMethod objects by memoizing initialize_parameters

rainerborene:
I have just discovered a simple optimization on `ViewComponent::Base` that can reduce `UnboundMethod` objects allocation. I have removed other objects from this memory allocation report comparison to keep things simple and clear. These are the results in one of our main Rails endpoints.

Before:
```
allocated memory by class
-------------------------
4080  UnboundMethod
```

After:
```
allocated memory by class
-------------------------
1280  UnboundMethod
```

If you want to test this patch yourself, just do this on your `ApplicationComponent`:

```ruby
  private
    def self.initialize_parameters
      @initialize_parameters ||= super
    end
end
```

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -553,6 +553,7 @@ module ViewComponent
       # @param parameter [Symbol] The parameter name used when rendering elements of a collection.
       def with_collection_parameter(parameter)
         @provided_collection_parameter = parameter
+        @initialize_parameters = nil
       end
 
       # Strips trailing whitespace from templates before compiling them.
@@ -648,7 +649,7 @@ module ViewComponent
       end
 
       def initialize_parameters
-        instance_method(:initialize).parameters
+        @initialize_parameters ||= instance_method(:initialize).parameters
       end
 
       def provided_collection_parameter
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,6 +18,10 @@ nav_order: 5
 
     *Alexandre Ignjatovic*
 
+* Reduce UnboundMethod objects by memoizing initialize_parameters
+
+    Rainer Borene
+
 ## 3.6.0
 
 * Refer to `helpers` in `NameError` message in development and test environments.


reeganviljoen:
@rainerborene can you please add your change to the changelog while waiting on a review

Spone:
I'm a bit unsure if adding memorization here can have unwanted effects. Mind chiming in @camertron?

camertron:
Cool, thanks for looking into this @rainerborene :)

The only issue I can think of would be if the `#initialize` method gets redefined via `class_eval` or something. It's possible in such a scenario that the memoized list of parameters would be out-of-date by the time `Base` tries to find the collection parameter. However I can't imagine that sort of thing is very common, so I think this change is ok 👍

camertron:
I'm on leave right now @Spone, would you or another maintainer be able to look into the Vale CI failure and merge this?

Spone:
Yes I will, thanks for your input!

lib/view_component/base.rb
```diff
@@ -648,7 +648,7 @@ def initialize_parameter_names
       end
 
       def initialize_parameters
-        instance_method(:initialize).parameters
+        @initialize_parameters ||= instance_method(:initialize).parameters
```
camertron, line 651: Please set `@initialize_parameters = nil` in `.with_collection_parameter` above to cover the case where the collection parameter is changed after memoization.

camertron (changes requested):

lib/view_component/base.rb
rainerborene: @camertron `initialize_parameters` is not inherited. I don't think there is a need to set it to `nil`. Even if `.with_collection_parameter` is called the `initialize` method was already defined in the first place. Can you plz give me an example to reproduce the issue you're talking about?

camertron: Consider this (admittedly contrived) example:

```ruby
module Primer
  class ProductComponent < Primer::Component
    def initialize(product: nil)
      @product = product
    end
  end
end

# ensure parameters are memoized
Primer::ProductComponent.send(:initialize_parameters)

Primer::ProductComponent.class_eval do

  def initialize(item: nil)
    @item = item
  end
end
```

Trying to render this component as a collection will raise a `MissingCollectionArgumentError`.

rainerborene: @camertron Got it. It's ready to merge now.

camertron (approved):

PR #1870, opened 2023-10-14
[Docs]Update document for changing view_component_path

hachi8833:

I'm trying to update the document so that users can change the `view_component_path`.

I got stuck when I tried to introduce ViewComponent to Rails 7.1 app because changing just `config.view_component.view_component_path` did not work.

Finally I found that `config.eager_load_paths` should also be changed with the same path to get this change work.

I checked the fact with the newly generated Rails 7.1/7.0/6.1 apps.

### DIFF
--- a/docs/guide/generators.md
+++ b/docs/guide/generators.md
@@ -38,7 +38,13 @@ bin/rails generate component Sections::Example title content
 
 You can specify options when running the generator. To alter the default values project-wide, define the configuration settings described in [API docs](/api.html#configuration).
 
-Generated ViewComponents are added to `app/components` by default. Set `config.view_component.view_component_path` to use a different path.
+Generated ViewComponents are added to `app/components` by default. Set `config.view_component.view_component_path` to use a different path. Note that you need to add the same path to `config.eager_load_paths` as well.
+
+```ruby
+# config/application.rb
+    config.view_component.view_component_path = "app/views/components"
+    config.eager_load_paths << Rails.root.join("app/views/components")
+```
 
 ### Override template engine
 
--- a/docs/guide/generators.md
+++ b/docs/guide/generators.md
@@ -42,8 +42,8 @@ Generated ViewComponents are added to `app/components` by default. Set `config.v
 
 ```ruby
 # config/application.rb
-    config.view_component.view_component_path = "app/views/components"
-    config.eager_load_paths << Rails.root.join("app/views/components")
+config.view_component.view_component_path = "app/views/components"
+config.eager_load_paths << Rails.root.join("app/views/components")
 ```
 
 ### Override template engine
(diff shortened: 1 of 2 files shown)


Spone:
Thanks for reporting this! I'm wondering if we can inject the `view_component_path` into the `eager_load_paths` when needed, so the users don't need to do it themselves.

hachi8833:
Sounds a better idea!
If rhe auto-injection is supported, the PR can be closed instead.

Spone:
Would you like to try it and open a PR for the changes?

hachi8833:
After that I change my mind: just adding the documents might be safer because the eager-loading config is explicit and we don't need to change the current code.

I'm a bit worried about changing the code might affect some patches to ViewComponent such as following:

https://github.com/palkan/view_component-contrib

reeganviljoen:
Maybe we can get the view_component_contrib maintaners to join this discussion, they may be keen to add this to their lib too

reeganviljoen:
@spone @camertron Since this could be a breaking change I suggest that maybe we add this to v4 checklist

camertron:
What specifically would be the danger in automatically adding the `view_component_path` to the list of eager load paths? Seems like a pretty safe thing to do, unless I'm really missing something?

reeganviljoen:
From what I understand their is a bit of concern that existing community patches may break but on further thought we can't gurantee the stability of other libraries that patch vc

camertron:
@reeganviljoen right yeah, I'm curious about the specific issues community patches would have if ViewComponent were to automatically append to the set of Rails eager load paths. It's possible we need to do some investigation here.

docs/guide/generators.md
```diff
@@ -38,7 +38,13 @@ bin/rails generate component Sections::Example title content
 
 You can specify options when running the generator. To alter the default values project-wide, define the configuration settings described in [API docs](/api.html#configuration).
 
-Generated ViewComponents are added to `app/components` by default. Set `config.view_component.view_component_path` to use a different path.
+Generated ViewComponents are added to `app/components` by default. Set `config.view_component.view_component_path` to use a different path. Note that you need to add the same path to `config.eager_load_paths` as well.
+
+```ruby
+# config/application.rb
+    config.view_component.view_component_path = "app/views/components"
+    config.eager_load_paths << Rails.root.join("app/views/components")
```
camertron, line 46:
```suggestion
config.view_component.view_component_path = "app/views/components"
config.eager_load_paths << Rails.root.join("app/views/components")
```

camertron (changes requested):

camertron (approved):

camertron:
Going to merge this for now since it's just a documentation change 👍

PR #1874, opened 2023-10-17
Add docs about inline templates interpolation

### LINKED ISSUE #1869: String Interpolation Issue with Slim Inline Template
### Steps to reproduce

1 - Create a component with an inline slim template
2 - Use any helper or instance method inside a string interpolated bracket

Here is a dummy component:
```rb
class DummyComponent < ViewComponent::Base
  slim_template <<~SLIM
    span = foo
    span #{bar}
  SLIM

  def foo
    "foo"
  end

  def bar
    "bar"
  end
end
```

Screenshot 2023-10-14 at 10 13 50

### Expected behavior

Inline Template should behave like a Sibling Template.
Code inside `#{}` should be run in the Compoment instance context

### Actual behavior

Code inside `#{}` is run in the Compoment class context

Quick example:
```rb
class DummyComponent < ViewComponent::Base
  slim_template <<~SLIM
    div = self
    br
    div #{self}
    br
  SLIM
end
```
is printing:

Screenshot 2023-10-14 at 10 21 49

**Backtrace**:

### System configuration
**Rails version**: 7.0.8

**Ruby version**: 3.2.1

**Gem version**: 3.6.0

**Slim version**: 5.1.1

Spone:

Clarify interpolation behavior in inline templates

Close #1869

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,6 +18,10 @@ nav_order: 5
 
     *Alexandre Ignjatovic*
 
+* Improve docs about inline templates interpolation
+
+    *Hans Lemuet*
+
 ## 3.6.0
 
 * Refer to `helpers` in `NameError` message in development and test environments.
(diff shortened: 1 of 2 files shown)


boardfish (approved):

reeganviljoen (approved):
looks good @Spone

PR #1885, opened 2023-10-26
Failing test case using image_tag and tag slot

hchtlz:
I'm attempting to address an unexpected test failure that occurred while using the renders_one :tag feature in my component. The issue seems to be connected to the use of the image_tag method within the same component. The specific error message I encountered is:

````
Error:
SlotableTest#test_slotable_component:
ArgumentError: wrong number of arguments (given 2, expected 0)
    /view_component/lib/view_component/slotable.rb:91:in `block in renders_one'
    /.rbenv/versions/3.1.4/lib/ruby/gems/3.1.0/gems/actionview-7.0.8/lib/action_view/helpers/asset_tag_helper.rb:413:in `image_tag'
````
I suspect that the issue may be related to the naming conflict between renders_one :tag and the image_tag method within the same component. When I changed the name "tag" to something different, the test case ran without any issues.

### CODE REVIEW

**Comment by Spone** [comment]
Thanks @hchtlz for opening this!

I think I understand what happens here:
- We create a slot named `tag`, so the `tag` methods is defined in the component.
- When we call `image_tag`, the [`tag` helper](https://github.com/rails/rails/blob/main/actionview/lib/action_view/helpers/tag_helper.rb#L467) from Rails is called [here](https://github.com/rails/rails/blob/23938052acd773fa24068debe56cd892cbf8d868/actionview/lib/action_view/helpers/asset_tag_helper.rb#L437C28-L437C28)... with 2 arguments, whereas the slot `tag` method doesn't expect any.

Now... I'm not sure how we want to address this:
- raise an error when you try to define a slot that has the same name as a Rails' helper? (but this would be *a lot* of reserved names)
- only add `tag` to the list of reserved names (it's a pretty generic name so there are more chances of collision with this helper than, say, `image_tag`)
- figure out a way to prevent the collision (maybe related to the idea of allow-listing the helpers, see #1860)

**Comment by Spone** [comment]
What would be your preferred approach for handling this @camertron @BlakeWilliams @boardfish?

**Comment by camertron** [comment]
@Spone what if we just raised an error if you tried to define a slot with the same name as an already-defined method? We actually already do this for [polymorphic slot setters](https://github.com/ViewComponent/view_component/blob/2bc5763a39c67e0d7da335cfa4ee403afa363d75/lib/view_component/slotable.rb#L239).

**Comment by Spone** [comment]
Sounds good. I'll work with @hchtlz on this.

**Comment by Spone** [comment]
Let's see if we need to do that for `renders_many` as well.

File: docs/CHANGELOG.md [hunk 1]
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add the error `RedefinedExistingMethodError` in case of conflict after naming a slot with the name of an existing method.
```
```suggestion
* Raise the error `RedefinedExistingMethodError` when declaring a slot which conflicts with an existing method.
```

test/sandbox/test/slotable_test.rb
```diff
@@ -719,7 +719,7 @@ def test_slot_with_content_shorthand
 
     assert component.title.content?
   end
-
+
```
Spone, line 722:
```suggestion

```

test/sandbox/test/slotable_test.rb
```diff
@@ -741,4 +741,25 @@ def test_slot_names_can_start_with_call
...
+  def test_raises_error_on_conflicting_slot_names_with_content_for_renders_many
+    exception = assert_raises ViewComponent::RedefinedExistingMethodError do
+      Class.new(ViewComponent::Base) do
+        renders_many :tags
+      end
+    end
+  
+    assert_includes exception.message, "declares a slot named tag"
+  end
+
```
Spone, line 764:
```suggestion
  end
```

test/sandbox/test/slotable_test.rb
```diff
@@ -741,4 +741,25 @@ def test_slot_names_can_start_with_call
       end
     end
   end
+
+  def test_raises_error_on_conflicting_slot_names_with_content_for_renders_one
```
Spone, line 745:
```suggestion
  def test_raises_error_on_slot_name_conflicting_with_existing_method_for_renders_one
```

test/sandbox/test/slotable_test.rb
```diff
@@ -741,4 +741,25 @@ def test_slot_names_can_start_with_call
...
+    exception = assert_raises ViewComponent::RedefinedExistingMethodError do
+      Class.new(ViewComponent::Base) do
+        renders_one :tag
+      end
+    end
+  
+    assert_includes exception.message, "declares a slot named tag"
+  end
+  
+  def test_raises_error_on_conflicting_slot_names_with_content_for_renders_many
```
Spone, line 755:
```suggestion
  def test_raises_error_on_slot_name_conflicting_with_existing_method_for_renders_many
```

Spone:
The failure in the Primer ViewComponents check comes from this component in Lookbook: https://github.com/ViewComponent/lookbook/blob/main/app/components/lookbook/icon_button/component.rb#L3C5-L3C51

We're calling `renders_many :icons`, which raises an exception since in this PR, we no longer allow slot names to overwrite existing method names.

But actually, I'm not sure whether we need this check for `renders_many` slots, since the singular slot name is never defined as a method by ViewComponent, right? So it would be legal to call both `renders_many :icons` and `attr_reader :icon` as you do in this component.

@allmarkedup mind chiming in?

Spone:
Currently, `validate_singular_slot_name` is called on both `renders_one` and `renders_many`slots.

https://github.com/ViewComponent/view_component/blob/b11ba1f197e71cab1cc0758373a58c9b2e364755/lib/view_component/slotable.rb#L153

The check for plural slot name conflicting with singular slot name has been introduced in #1455, but I'm not sure if I understand why that would be an issue (aside from being very confusing when you read the component).

joelhawksley:
> The check for plural slot name conflicting with singular slot name has been introduced in https://github.com/ViewComponent/view_component/pull/1455, but I'm not sure if I understand why that would be an issue

@Spone I believe this is because plural slots still have singular setters, i.e. `renders_many :items` will generate the `item` setter.

Closing as stale.

PR #1893, opened 2023-11-01
Set request full path with request url

nachiket87:

Fixing https://github.com/ViewComponent/view_component/issues/1867

Issue may be a Rails issue here: https://github.com/rails/rails/blob/8e73e620b1072c831b81604449385dfbb60bdb63/actionpack/lib/action_dispatch/http/request.rb#L262

Where `@fullpath` is cached from previous request. To get around this, I have manually set the values of `@fullpath` and `@original_fullpath` in the test helper.

A similar issue was fixed here: https://github.com/ViewComponent/view_component/commit/4241a14be92298852d80b12308bf3da9390d4db5

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Fix bug where `#with_request_url`, set the incorrect `request.fullpath`.
+
+    *Nachiket Pusalkar*
+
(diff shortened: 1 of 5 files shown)


camertron (approved):
Awesome, thank you @nachiket87 😄 I'm fine merging this, but I would heartily recommend submitting a patch upstream as well.

Issue #1620, opened 2023-01-02
Slots iteration helpers


Spone:
## Feature request

Similar to the iteration context provided when [rendering collections](https://viewcomponent.org/guide/collections.html#collection-iteration-context), ViewComponent could provide contextual information, such as `#size`, `#index`, `#first?`, and `#last?`, when rendering slots that have been defined with `renders_many`.

### Motivation

I personally encountered this need at least a dozen times in various projects. This also has been discussed in #545, #982 and #986.

reeganviljoen:
@Spone I am trying to get more involved in view component is this possibly an issue I could help with

reeganviljoen:
here is a pr where I have some failing tests https://github.com/ViewComponent/view_component/pull/1753

reeganviljoen:
@Spone the slot itteration counter helper has been added , I will be adding the iteration context soon

reeganviljoen:
@spone I have fully implemented this feature in https://github.com/ViewComponent/view_component/pull/1753
can I ask for you to code review it since you requested this feature

joelhawksley:
Closed per debate in #1753

PR #1895, opened 2023-11-03
Add slot unwrapping API

camertron:
Fixes https://github.com/ViewComponent/view_component/issues/1737

This PR introduces an API for retrieving the underlying component instance inside `ViewComponent::Slot`s. At the moment, getting a reference to the underlying component necessitates using `instance_variable_get`, eg:

```ruby
  renders_one :header, HeaderComponent
  renders_many :items, ItemComponent
end

render(MyComponent.new) do |component|
  # returns an instance of ViewComponent::Slot
  component.header

  # returns an instance of HeaderComponent
  component.header.instance_variable_get(:@__vc_component_instance)
end
```

This is awkward. The `@__vc_component_instance` is private and not designed to be accessed outside the `Slot` class.

I propose the following API:

```ruby
render(MyComponent.new) do |component|
  # returns an instance of HeaderComponent
  component.header_instance

  # returns an array of ItemComponent instances
  component.item_instances
end
```

This works for any component slot. For lambda slots, the `*_instance` methods are not defined.

Is the `_instance` suffix the best naming? Maybe `_component` would be better? Open to suggestions 😄

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add API for "unwrapping" slots, i.e. return the component instance inside `ViewComponent::Slot`.
+
+    *Cameron Dutro*
+
 * Don't add ActionDispatch::Static middleware unless `public_file_server.enabled`.
 
     *Daniel Gonzalez*
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -219,6 +219,25 @@ end
 
 _Note: While a lambda is called when the `with_*` method is called, a returned component isn't rendered until first use._
 
+## Retrieving slot component instances
+
+Since 3.8.0
+{: .label }
+
+Component slots are wrapped in an instance of `ViewComponent::Slot` which transparently proxies method calls to the original instance via Ruby's `#method_missing`. To retrieve the underlying component instance(s), call the corresponding `*_instance` or `*_instances` method.
+
+```ruby
+blog = BlogComponent.new { ... }
+
+# renders_one slot
+blog.header  # returns an instance of ViewComponent::Slot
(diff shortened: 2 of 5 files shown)


Spone:
Re: the suffix I think `_component` is clearer than `_instance` (... of what?)

Anyway I think both are more explicit than the initial `unwrap` idea.

BlakeWilliams:
Just writing this down real quick since I don't have the time to write up a longer response, but I'm still on the fence here. I get the intention behind this change, but I also think we should try to avoid having another API for accessing slots in addition to the APIs we already have.

Spone:
Yeah my comment was only about the naming.

I've never had a use case for this in the projects I'm working on, so I'm not especially for this addition.

I reckon it adds another layer of complexity to the Slots which are already the most complex part of ViewComponent (in terms of usage I mean).

camertron:
Thanks for your feedback @Spone and @BlakeWilliams! I agree it adds a bit more complexity to an already complex feature, and I also agree there are limited occasions for use. I have personally run into a similar issue to the one the OP describes in https://github.com/ViewComponent/view_component/issues/1737 (comparing two wrapped component instances). I've also had occasion to unwrap slots when attempting to forward them to other components, and while debugging. In other words, I think the framework needs some way of unwrapping slots. If this isn't it, that's fine, but I would like to figure out the right approach.

reeganviljoen:
I would like to agree with @camertron, I have also needed access to the base component in a slot via unwrapping when I was attempting to add #1620 , this may even unblock it and give us the chance at re-attempting to add it, however due to the uncertainty may I suggest we add this as an optional experimental feature similar to how inline templates where added until we can see if it really is useful in the real world

camertron:
We talked about this internally and have decided removing the `Slot` wrapper is the better way forward 👍

PR #1896, opened 2023-11-04
Fix invalid rubocop errors

BlakeWilliams:
When running ViewComponent in VSCode I started getting a ton of errors
from Rubocop despite the project using standardrb and requiring it in
the rubocop config.

This fixes the issue by using the `inherit_gem` option in the rubocop
config.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -329,9 +329,6 @@ DEPENDENCIES
   jbuilder (~> 2)
   m (~> 1)
   minitest (~> 5.18)
-  net-imap
-  net-pop
-  net-smtp
   pry (~> 0.13)
   puma (~> 6)
   rails (~> 7.0.0)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -329,9 +329,6 @@ DEPENDENCIES
   jbuilder (~> 2)
   m (~> 1)
   minitest (~> 5.18)
-  net-imap
-  net-pop
-  net-smtp
```
reeganviljoen, line 334: are't these required for ruby 3.2 +

Spone:
TIL about `inherit_gem` :)

Spone (approved):

Gemfile.lock
BlakeWilliams: Maybe, but I think that can be fixed with a `bundle` call? Either way we'll see churn when running with different Ruby versions so I'll keep this in for now.

PR #1898, opened 2023-11-04
Remove support for unsupported versions of ruby and rails

### LINKED ISSUE #1897: Deprecate the usage of unsupported versions of Ruby and Rails
# Deprecate the usage of unsupported versions of Ruby and Rails

Currently view_component supports the usage of unsupported versions of ruby (i.e 2,7) as-well as Rails

This will likely cause issues in the  future where developers use view_component on unsupported versions of Ruby or Rails where we will not be able to fully support them as their issues are coming from unsupported versions of Ruby or Rails

## Ruby Life cycle
image
## Rails Life cycle
image

It is therefore my suggestion that starting with view_component 4 we adopt a ruby and Rails support life cycle aligned with their support dates

reeganviljoen:
Closes #1897 

Remove support for unsupported versions of ruby and rails


docs/compatibility.md
```diff
@@ -8,9 +8,9 @@ nav_order: 6
 
 ## Ruby & Rails
 
-ViewComponent is [supported natively](https://edgeguides.rubyonrails.org/layouts_and_rendering.html#rendering-objects) in Rails 6.1, and compatible with Rails 5.2+ via an included [monkey patch](https://github.com/viewcomponent/view_component/blob/main/lib/view_component/render_monkey_patch.rb).
+View Component supports all actively supported versions of Ruby (3.0+) and Ruby on Rails(6.1+) and is tested against a combination of these versions of ruby and rails
 
-ViewComponent is tested for compatibility [with combinations of](https://github.com/viewcomponent/view_component/blob/22e3d4ccce70d8f32c7375e5a5ccc3f70b22a703/.github/workflows/ruby_on_rails.yml#L10-L11) Ruby v2.5+ and Rails v5.2+. Ruby 2.4 is likely compatible, but is no longer tested.
+Rails >= 6.0 and Ruby >= 2.7 may still work but is not actively supported and any issues submitted against them will be immediately closed
```
Spone, line 13: Could this be rewritten with a friendlier tone? I'm not sure we need to warn that issues will be immediately closed (I guess we would rather discuss the issue first).

reeganviljoen: thank you for your feedback @Spone, I have updated it, will you let me know what you think ?

docs/compatibility.md
```diff
@@ -8,9 +8,9 @@ nav_order: 6
 
 ## Ruby & Rails
 
-ViewComponent is [supported natively](https://edgeguides.rubyonrails.org/layouts_and_rendering.html#rendering-objects) in Rails 6.1, and compatible with Rails 5.2+ via an included [monkey patch](https://github.com/viewcomponent/view_component/blob/main/lib/view_component/render_monkey_patch.rb).
+View Component supports all actively supported versions of Ruby (3.0+) and Ruby on Rails(6.1+) and is tested against a combination of these versions of ruby and rails
```
BlakeWilliams, line 11:
```suggestion
ViewComponent supports all actively supported versions of Ruby (3.0+) and Ruby on Rails(6.1+) and is tested against a combination of these versions of Ruby on Rails.
```

docs/compatibility.md
```diff
@@ -8,9 +8,9 @@ nav_order: 6
...
 
-ViewComponent is tested for compatibility [with combinations of](https://github.com/viewcomponent/view_component/blob/22e3d4ccce70d8f32c7375e5a5ccc3f70b22a703/.github/workflows/ruby_on_rails.yml#L10-L11) Ruby v2.5+ and Rails v5.2+. Ruby 2.4 is likely compatible, but is no longer tested.
+Rails >= 6.0 and Ruby >= 2.7 may still work with view_component, but this configuration is not actively supported and any issues submitted against unsupported versions of Ruby or Rails have no guarantee of being fixed as they may eb arising from issues from these unsupported versions
```
BlakeWilliams, line 13:
```suggestion
While EOL (end-of-life) versions of Ruby (>= 2.7) and Ruby on Rails (> 6.0) may still work with ViewComponent, they are not actively supported and we do not active test the gem against them. We will still accept patches on a case-by-case basis to support older Ruby and Ruby on Rails versions based on the complexity and maintenance burden. Please open an issue before submitting a Pull Request that addresses issues with EOL Ruby/Ruby on Rails versions.
```

Thoughts on this? This is how I was imagining we'd handle this situation since it adds some reasonable flexibility without keeping us stuck maintaining EOL dependencies.

reeganviljoen: I like this @BlakeWilliams, I struggle with the wording sometimes, but this sums it up just right

docs/compatibility.md
```diff
@@ -8,9 +8,9 @@ nav_order: 6
...
+ViewComponent supports all actively supported versions of Ruby (3.0+) and Ruby on Rails(6.1+) and is tested against a combination of these versions of Ruby on Rails.
 
-ViewComponent is tested for compatibility [with combinations of](https://github.com/viewcomponent/view_component/blob/22e3d4ccce70d8f32c7375e5a5ccc3f70b22a703/.github/workflows/ruby_on_rails.yml#L10-L11) Ruby v2.5+ and Rails v5.2+. Ruby 2.4 is likely compatible, but is no longer tested.
+While EOL (end-of-life) versions of Ruby (>= 2.7) and Ruby on Rails (> 6.0) may still work with ViewComponent, they are not actively supported and we do not active test the gem against them. We will still accept patches on a case-by-case basis to support older Ruby and Ruby on Rails versions based on the complexity and maintenance burden. Please open an issue before submitting a Pull Request that addresses issues with EOL Ruby/Ruby on Rails versions.
```
Spone, line 13:
```suggestion
While EOL (end-of-life) versions of Ruby and Ruby on Rails may still work with ViewComponent, they are not actively supported and no longer tested. We will still accept patches on a case-by-case basis to support older Ruby & Rails versions based on the complexity and maintenance burden. Please open an issue before submitting such a Pull Request.
```

I'm a bit reluctant to mention the specific EOL versions here since we'll need to keep that up-to-date.

reeganviljoen: @Spone I agree with you, but we will still have to remove and add versions to CI when versions get released and EOL'd

docs/compatibility.md
```diff
@@ -8,9 +8,9 @@ nav_order: 6
 
 ## Ruby & Rails
 
-ViewComponent is [supported natively](https://edgeguides.rubyonrails.org/layouts_and_rendering.html#rendering-objects) in Rails 6.1, and compatible with Rails 5.2+ via an included [monkey patch](https://github.com/viewcomponent/view_component/blob/main/lib/view_component/render_monkey_patch.rb).
+ViewComponent supports all actively supported versions of Ruby (3.0+) and Ruby on Rails(6.1+) and is tested against a combination of these versions of Ruby on Rails.
```
Spone, line 11:
```suggestion
ViewComponent supports all actively supported versions of Ruby (3.0+) and Ruby on Rails (6.1+) and is tested against a combination of these versions of Ruby on Rails.
```

Spone (approved):

Spone:
FYI I updated the branch protection rules to change which checks must pass in order to be able to merge, and re-ran the checks.

reeganviljoen:
> Code coverage check fails because we're no longer running tests for older versions of Rails, and there a still conditionals such as this one:
> 
> https://github.com/reeganviljoen/view_component/blob/rv-remove-unsupported-rails-ruby/app/helpers/preview_helper.rb#L16-L45
> 
> Not sure how to handle that. If we remove the conditionals to keep only the most recent versions, we intentionally break support (that would require us to wait for v4).

We can start raising deprecation warnings however

reeganviljoen:
@Spone I have added deprecation warnings for all the EOL Ruby and Rails stuff in #1902

Spone:
I'm just thinking of it now, but could we use `:nocov:` comments to wrap the code that support older versions of Rails?

See https://github.com/simplecov-ruby/simplecov#ignoringskipping-code

reeganviljoen:
@Spone I think that would be a great short term fix but we should then add depraction notices and possibly remove that code in v4

Spone (approved):

PR #1903, opened 2023-11-07
Error related to naming a slot with an uncountable name.

hchtlz:

I am submitting this PR to report an error related to naming a slot with an uncountable name in my component. I attempted to create a slot named "series," but it resulted in an error. This error is due to the fact that the name "series" is uncountable and does not follow Rails convention.

With reference to the uncountable words defined in [Rails' inflections list](https://github.com/rails/rails/blob/632bcb391840bacc77556eee6707178ddf293b99/activesupport/lib/active_support/inflections.rb#L70), I expected to use the name "series" for my slot. However, Rails expects a pluralized slot name, leading to this error.

Here is the specific error message I encountered:

````
Error:
SlotableTest#test_slot_with_unplurialized_name:
NoMethodError: undefined method `with_serie' for #<NamingSlotTestComponent:0x000000010fd97c30 @_routes=nil>
Did you mean?  with_series
````

I suggest creating a more explicit error message or exploring alternative ways to resolve this issue.

### DIFF
--- a/test/sandbox/app/components/naming_slot_test_component.rb
+++ b/test/sandbox/app/components/naming_slot_test_component.rb
@@ -2,6 +2,8 @@ class NamingSlotTestComponent < ViewComponent::Base
   renders_many :series, "Serie"
 
   class Serie < ViewComponent::Base
+    attr_reader :name
+
     def initialize(name:)
       @name = name
     end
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: test/sandbox/app/components/naming_slot_test_component.rb [hunk 1]
```diff
@@ -0,0 +1,9 @@
+class NamingSlotTestComponent < ViewComponent::Base
+  renders_many :series, "Serie"
+
+  class Serie < ViewComponent::Base
```
Spone, line 4:
```suggestion
  class Serie < ViewComponent::Base
    attr_reader :name

```

reeganviljoen:
@BlakeWilliams we had a similar error when using call in slot names I am string to think we should add a general error for conflicting slot names

PR #1904, opened 2023-11-08
Raise error uncountable slot names in plural slots

reeganviljoen:
closes #1903 
This Attempts to fix the [bellow bug](https://github.com/ViewComponent/view_component/pull/1903) by raising an error when a plural slot is used with an uncountable name
> ### What are you trying to accomplish?
> I am submitting this PR to report an error related to naming a slot with an uncountable name in my component. I attempted to create a slot named "series," but it resulted in an error. This error is due to the fact that the name "series" is uncountable and does not follow Rails convention.
> 
> With reference to the uncountable words defined in [Rails' inflections list](https://github.com/rails/rails/blob/632bcb391840bacc77556eee6707178ddf293b99/activesupport/lib/active_support/inflections.rb#L70), I expected to use the name "series" for my slot. However, Rails expects a pluralized slot name, leading to this error.
> 
> Here is the specific error message I encountered:
> 
> ```
> Error:
> SlotableTest#test_slot_with_unplurialized_name:
> NoMethodError: undefined method `with_serie' for #<NamingSlotTestComponent:0x000000010fd97c30 @_routes=nil>
> Did you mean?  with_series
> ```
> 
> ### What approach did you choose and why?
> I suggest creating a more explicit error message or exploring alternative ways to resolve this issue.


reeganviljoen:
@hchtlz I credited you with me in the change log because you discovered this issue and gave us a failing test to work with

hchtlz:
> @hchtlz I credited you with me in the change log because you discovered this issue and gave us a failing test to work with
@reeganviljoen Thank you for the credit in the changelog. I was delighted to contribute to solving this issue

BlakeWilliams (approved):
Great fix, thanks!

BlakeWilliams:
Actually, I think this would have been problematic in a previous API. Is that still the case for this iteration of slots?

Would we expect `has_many :series` to result in:

```
Component#series # gets the slots
Components#with_series # sets the slot
```

Which seems fine, unless I'm missing something here.

In the [example failure](https://github.com/ViewComponent/view_component/pull/1903/files#diff-81155907656ee8532caa24b032d5816353ae79d28c3b3fc7bf3e8054fe000beaR725) we shouldn't be calling `with_serie` since that's not a valid word. So I guess, do we need this check?

BlakeWilliams:
Ahhhh, we do still need this because of the extra methods we define: https://github.com/ViewComponent/view_component/blob/122604c9130820d7ee915aa80b8b6bb710dc57f0/lib/view_component/slotable.rb#L169-L185

PR #1906, opened 2023-11-09
Add `config.include_application_routes` to allow skipping inclusion of application routes

elia:

Work with components for Rails engines that provide route helpers that should stay separated from the main app routes.

Instead of trying to automatically detect and include the engine's routes seems better to just opt out of application routes and manually `include MyGem::Engine.routes.url_helpers`.

Would be even better to opt out of application routes individually for children of `ViewComponent::Base` so that the engine components can have their base class while `ApplicationComponent` still gets the app routes by default.


joelhawksley:
@elia do you still think this change is a good idea?

elia:
@joelhawksley absolutely, will give much more freedom when it comes to using VC within engines.

joelhawksley:
@elia cool! I'm wondering if there are any other options besides blanket disabling of routes. Can you think of another approach?

elia:
@joelhawksley in this case rather than disabling all the routes just puts them behind their "mounted_helpers" so the main app routes are still reachable via `main_app.posts_path`. This allows engines to put their routes as the default or to stick to explicit naming (`my_comments_engine.comments_path`). 

Would you prefer a solution always having a set of routes included? Be it the engine's or the main app's?

joelhawksley:
> Would you prefer a solution always having a set of routes included? Be it the engine's or the main app's?

My preference is to avoid a breaking change. What do you think?

joelhawksley:
Closing as stale. I'd still love to make an improvement here if you're up for it @elia!

PR #1911, opened 2023-11-16
Avoid allocating new string when output_postamble is blank

mitchellhenke:

This pull request is strictly to improve performance by reducing allocations

The change is exclusively focused on the `output_postamble` method and the default empty string. The concatenation calls https://github.com/rails/rails/blob/v7.0.8/activesupport/lib/active_support/core_ext/string/output_safety.rb#L247-249 which includes a duplication of the original string and concatenation of an empty one.

Avoiding the concatenation saves an allocation of the rendered template, which can be quite large.

This was found while working on #1884.

The code changes are quite small, but a potentially significant thing I'm not sure of is whether there is a good way to add tests.


reeganviljoen:
@mitchellhenke sorry about the late reply, I can't seem to find your issue, @joelhawksley are you able to help

mitchellhenke:
@reeganviljoen apologies, I didn't create an issue for this one, but I am happy to if it is helpful

joelhawksley (approved):

mitchellhenke:
Thank you for the review and merge!

PR #1912, opened 2023-11-16
Remove `with_variant` in API docs

tysongach:

The `with_variant` method was removed in v3.0.0 - this removes its mention from the docs.

Relevant changelog entry: https://github.com/ViewComponent/view_component/blob/main/docs/CHANGELOG.md#v300

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -374,16 +374,6 @@ with_request_url("/users/42", host: "app.example.com") do
 end
 ```
 
-### `#with_variant(variant)`
-
-Set the Action Pack request variant for the given block:
-
-```ruby
-with_variant(:phone) do
-  render_inline(MyComponent.new)
-end
-```
-
 ## Errors
 
 ### `AlreadyDefinedPolymorphicSlotSetterError`
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Remove `with_variant` in API docs.
+
+    *Tyson Gach*
+
 * Remove unsupported versions of Rails & Ruby from CI matrix.
 
     *Reegan Viljoen*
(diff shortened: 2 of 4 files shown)


tysongach:
There’s also a `with_variant` test helper mentioned in the testing docs: https://viewcomponent.org/guide/testing.html#action-pack-variants

I thought that might need removed, too, but it seems like that method is still available:

https://github.com/ViewComponent/view_component/blob/bfee458016a21bea3234bb76e74e0b481124663d/lib/view_component/test_helpers.rb#L126

Should that method be deprecated?

reeganviljoen:
@tysongach can you please add your entry the the changelog 🙏

docs/api.md
```diff
@@ -374,16 +374,6 @@ with_request_url("/users/42", host: "app.example.com") do
 end
 ```
 
-### `#with_variant(variant)`
```
Spone, line 377: This file is autogenerated with YARD, can you make sure the method and comments are removed from the source file as well?

tysongach: Ah, thanks for clarifying! I think I’m realizing this is more than just a docs change. 

So this is the `with_variant` test helper and still publicly available. It feels like it should’ve been deprecated and removed [alongside `Base#with_variant`](https://github.com/ViewComponent/view_component/pull/1041/files)?

I’ve added https://github.com/ViewComponent/view_component/pull/1912/commits/3e1ab7c3e63b501bbd2d230fa64128e07e6a0aec which removes the method, too. It looks like we have a bunch of tests still using that helper - shall I continue down this path of removal?

joelhawksley:
@tysongach I believe this change is incorrect: before v3.0.0, there was _another_ `with_variant` on `Base` that allowed the variant to be set at runtime. This lead to abuse of ActionPack variants beyond what we wanted support in the framework.

The remaining `with_variant` is a helper for setting variants in tests that should remain.

PR #1934, opened 2023-12-14
Allow translations to be inherited and overridden in subclasses

elia:

Inheriting from a component that uses translations in the template now requires all translations to be replicated in the subclass or it will break. 

I resorted to including all parent translation files in the subclass under the subclass `i18n_scope`, thus allowing subclasses to add or replace translations as they see fit.

### DIFF
--- a/docs/guide/translations.md
+++ b/docs/guide/translations.md
@@ -61,7 +61,7 @@ en:
   greeting: "Cheers!"
 ```
 
-The translation is available in subclasses of `ParentComponent`, allowing translations to be used as is or overridden by the child component:
+The translation is available in subclasses of `ParentComponent`, allowing translations to be used as-is or overridden by the subclass:
 
 ```yml
 # app/components/child_component.yml
--- a/docs/guide/translations.md
+++ b/docs/guide/translations.md
@@ -49,3 +49,32 @@ Access global translations via `helpers` or `I18n`:
 <%= helpers.t("hello") %>
 <%= I18n.t("hello") %>
 ```
+
+## Inheritance
+
+Translations are inherited from the component's parent class. Given a parent component with a translation file:
+
+```yml
+# app/components/parent_component.yml
+en:
+  hello: "Hello world!"
+  greeting: "Cheers!"
+```
+
+The translation is available in subclasses of `ParentComponent`, allowing translations to be used as is or overridden by the child component:
+
+```yml
+# app/components/child_component.yml
+en:
+  greeting: "Howdy!"
+```
+
+```rb
+# app/components/child_component.rb
+class ChildComponent < ParentComponent
+  def call
+    t(".hello") # => "Hello world!" (inherited)
+    t(".greeting") # => "Howdy!"    (overridden)
+  end
+end
+```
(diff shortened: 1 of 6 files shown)


docs/guide/translations.md
```diff
@@ -49,3 +49,34 @@ Access global translations via `helpers` or `I18n`:
 <%= helpers.t("hello") %>
 <%= I18n.t("hello") %>
 ```
+
+## Inheritance
+
+Translations are inherited from the component's parent class.
+For example, given a parent component with a translation file:
```
joelhawksley, line 56:
```suggestion
Translations are inherited from the component's parent class. Given a parent component with a translation file:
```

docs/guide/translations.md
```diff
@@ -49,3 +49,34 @@ Access global translations via `helpers` or `I18n`:
...
+
+```yml
+# app/components/parent_component.yml
+en:
+  hello: "Hello world!"
+  greeting: "Cheers!"
+```
+
+Then the translation will be available in subclasses of `ParentComponent`.
+This allows translations to be used as is or overridden by the child component:
```
joelhawksley, line 66:
```suggestion
The translation is available in subclasses of `ParentComponent`, allowing translations to be used as is or overridden by the child component:
```

joelhawksley (approved):

lib/view_component/translatable.rb
```diff
@@ -23,9 +23,13 @@ def i18n_scope
       def build_i18n_backend
         return if compiled?
 
-        self.i18n_backend = if (translation_files = sidecar_files(%w[yml yaml])).any?
-          # Returning nil cleans up if translations file has been removed since the last compilation
+        # We need to load the translations files from the ancestors so a component
+        # can inherit translations from its parent and is able to overwrite them.
+        vc_ancestors = ancestors.reverse.select { _1.is_a?(Class) && _1 < ViewComponent::Base }
+        translation_files = vc_ancestors.flat_map { _1.sidecar_files(%w[yml yaml]) }
```
BlakeWilliams, line 29:
```suggestion
        translations = ancestors.each_with_object([]) do |ancestor, sidecar_files|
          sidecar_files += side_car_files(%w[yml yaml]) if ancestor.is_a?(Class) && ancestor < ViewComponent::Base
        end
```

Thoughts on optimizing this to be a single loop? I imagine this will be called pretty often and I think it's still readable, so likely worth it?

BlakeWilliams (approved):
I had one thought, otherwise LGTM!

lib/view_component/translatable.rb
elia: Sure! I'll update 👍

docs/guide/translations.md
```diff
@@ -49,3 +49,32 @@ Access global translations via `helpers` or `I18n`:
...
+Translations are inherited from the component's parent class. Given a parent component with a translation file:
+
+```yml
+# app/components/parent_component.yml
+en:
+  hello: "Hello world!"
+  greeting: "Cheers!"
+```
+
+The translation is available in subclasses of `ParentComponent`, allowing translations to be used as is or overridden by the child component:
```
joelhawksley, line 64:
```suggestion
The translation is available in subclasses of `ParentComponent`, allowing translations to be used as-is or overridden by the child component:
```

PR #1938, opened 2023-12-17
Defer to built-in action caching

### LINKED ISSUE #1937: Use `setup-ruby`'s native bundle caching in our workflows
While working on #1308, I've noticed that we use `actions/cache` where this is potentially avoidable. We may be able to use the `setup-ruby` action's `bundler_cache` option to leave that to the action instead of loading it ourselves.

boardfish:
Closes #1937

Instead of using `actions/cache` directly, defer to the `setup-ruby` and `setup-node` actions' native support, which uses this under the hood.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,7 @@ jobs:
       with:
         ruby-version: 3.2
         bundler-cache: true
+        working-directory: 'view_component'
     - uses: actions/setup-node@v4
       with:
         node-version: 16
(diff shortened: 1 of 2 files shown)


Spone:
> is there a reason we run on 2.7 in these?

I'd say it's just legacy.

joelhawksley:
@boardfish any interest in bringing this PR up to date?

boardfish:
Yep, I'll take a look at this soon!

boardfish:
This is good to go – the failures on Rails `main` are unrelated.

joelhawksley:
@boardfish I think we might have one more real failure here ❤️

boardfish:
That's now fixed 👍

joelhawksley (approved):

PR #1950, opened 2024-01-03
Ensure HTML output safety

camertron:

This PR addresses a security concern affecting components that define a `#call` method. In most cases, HTML escaping is handled either by the template handler (ERB, Slim, HAML, etc) or by ActionView when the rendered string is appended to the output buffer. However, if a controller renders a component directly, the rendered string is not escaped and can be a vector for [cross-site scripting](https://owasp.org/www-community/attacks/xss/) (XSS) and similar attacks.

Consider the following component:

```ruby
class UnsafeComponent < ApplicationComponent
  def initialize(user_defined_input)
    @user_defined_input = user_defined_input
  end

  def call
    "<div>#{@user_defined_input}</div>"
  end
end
```

If the component is rendered directly by the controller...

```ruby
class MyController < ApplicationController
  def create
    render UnsafeComponent.new(params[:user_defined_input])
  end
end
```

...and `params[:user_defined_input]` is the string `<script>alert("foo")</script>`, the alert will be injected onto the page without being escaped, resulting in a browser alert. Since `params[:user_defined_input]` can contain arbitrary JavaScript code, cookies and other sensitive information can be exfiltrated to a 3rd-party.

This PR ensures that rendered output is HTML escaped, and if it is not, prints a warning and escapes it automatically. It does this for output from `#call` methods, output generated by template handlers, and for postambles.

Finally, it examines the MIME type of the request and only performs automatic escaping when generating HTML. In other words, JSON generated by eg. jbuilder will not be automatically escaped.

### Performance Impact

```
              render   366.000 i/100ms
              render      3.703k (± 6.1%) i/s -     36.966k in  10.021240s

         safe render   302.000 i/100ms
         safe render      3.522k (±11.5%) i/s -     34.730k in  10.005739s

Comparison:
              render:     3703.0 i/s
         safe render:     3522.1 i/s - same-ish: difference falls within error
```

Performance is impacted, but not to a significant degree 👍


joelhawksley (approved):
👏🏻

reeganviljoen:
@camertron can we please add this as a security advisory so dependabot can pick this up

camertron:
@reeganviljoen yes, that's the plan! I'll submit one as soon as this is merged and released 👍

Spone:
What are your thoughts on backporting this fix to the v2 as well?

lib/view_component/compiler.rb
```diff
@@ -16,6 +16,7 @@ class Compiler
     def initialize(component_class)
       @component_class = component_class
       @redefinition_lock = Mutex.new
+      @variants_rendering_templates = []
```
BlakeWilliams, line 19:
```suggestion
      @variants_rendering_templates = Set.new()
```

Thoughts on using a Set for faster lookups (although I imagine it's minor improvements but may add up?)?

BlakeWilliams (approved):
LGTM!

lib/view_component/compiler.rb
camertron: Sure, sounds good to me!

camertron:
@Spone we discussed that a bit internally and decided not to, but I could be persuaded otherwise.

PR #1953, opened 2024-01-05
Exclude I18n reserved keys from html escaping

nickcoyne:

With html translation keys, reserved keys are currently being made html safe, thus converting values like `true` to `"true"`. This is disabling some of these keys.

The existing code tries to match reserved keys using the `::I18n.reserved_keys_pattern`, however they will never match, as the pattern is expecting interpolated strings with the translation value. It will match `%{raise}`, but not `raise`, which is what we want. So these reserved key values are all subsequently html escaped.

This PR uses the `I18n::RESERVED_KEYS` constant directly, and excludes those keys from options that will be considered for escaping.

The key that I chose for testing was `:raise`. This meant that I needed a missing translation, so I've added another locale to the `I18n.available_locales`. This also matched the original scenario in my app that caused me to discover the issue.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Exclude html escaping of I18n reserved keys with `I18n::RESERVED_KEYS` rather than `I18n.reserved_keys_pattern`.
+
+    *Nick Coyne*
+
 * Update CI configuration to use `Appraisal`.
 
     *Hans Lemuet, Simon Fish*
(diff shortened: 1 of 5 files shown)


boardfish:
👋 Just a quick nudge to say #1308 means you'll need to rebase this branch for checks to pass.

nickcoyne:
> 👋 Just a quick nudge to say #1308 means you'll need to rebase this branch for checks to pass.

Thanks, done!

joelhawksley (approved):
👏🏻

PR #1958, opened 2024-01-08
Fix html escaping on non strings

### LINKED ISSUE #1955: #maybe_escape_html chokes when input is an Integer
My CI pipeline suddenly fails when upgrading from 3.8.0 to 3.9.0.

I believe this is related to #1950. 

### Steps to reproduce

- It seems like my code hands over an integer to [`#maybe_escape_html`](https://github.com/ViewComponent/view_component/blob/c8462c2113ca1771b681fc89eb025d8e8b685727/lib/view_component/base.rb#L312C1-L313C1). This is not expected by the code.

### Expected behavior

No Exception

### Actual behavior

```
     NoMethodError:
       undefined method `empty?' for 2016:Integer
             return text if text.nil? || text.empty?
                                             ^^^^^^^
     # /usr/local/bundle/ruby/3.1.0/gems/view_component-3.9.0/lib/view_component/base.rb:312:in `maybe_escape_html'
     # /usr/local/bundle/ruby/3.1.0/gems/view_component-3.9.0/lib/view_component/base.rb:326:in `safe_render_template_for'
     # /usr/local/bundle/ruby/3.1.0/gems/view_component-3.9.0/lib/view_component/base.rb:109:in `render_in'
```

**first thoughts**:

- One could convert the `text` to string first before calling `#empty?`, similar to how `Slot#html_safe?` works. See <https://github.com/ViewComponent/view_component/blob/main/lib/view_component/slot.rb#L99C6-L99C6>
  - Note that `to_s.html_safe?` should actually be not necessary, since Rails implements `html_safe?` on the `Numberic` and `Object` classes

### System configuration
**Rails version**: 7.0.8

**Ruby version**: 3.1.4

**Gem version**: 3.9.0

reeganviljoen:
closes #1955
 When an integer or any non-string is used inside call the html escaping throw an error(described in the linked issue)

I replaced `text.nil? || text.empty?` for with type safe `.blank?` so any type can be escaped as is allowed in templates

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -308,7 +308,7 @@ module ViewComponent
     end
 
     def maybe_escape_html(text)
-      text = text.to_s unless text.is_a?(String)# ensure we're working with a string because some things like active storage dont return strings
+      text = text.to_s unless text.is_a?(String) # ensure we're working with a string because some things like active storage dont return strings
 
       return text if request && !request.format.html?
       return text if text.nil? || text.empty?
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -308,6 +308,8 @@ module ViewComponent
     end
 
     def maybe_escape_html(text)
+      text = text.to_s # ensure we're working with a string because some things like active storage dont return strings
+
       return text if request && !request.format.html?
       return text if text.nil? || text.empty?
 
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -308,7 +308,7 @@ module ViewComponent
     end
 
     def maybe_escape_html(text)
-      text = text.to_s # ensure we're working with a string because some things like active storage dont return strings
+      text = text.to_s unless text.is_a?(String)# ensure we're working with a string because some things like active storage dont return strings
 
       return text if request && !request.format.html?
       return text if text.nil? || text.empty?
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -308,10 +308,8 @@ module ViewComponent
     end
 
     def maybe_escape_html(text)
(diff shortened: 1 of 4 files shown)


lib/view_component/base.rb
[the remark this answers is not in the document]
reeganviljoen: @tmaier thank you 🙏

test/sandbox/test/slotable_test.rb
```diff
@@ -752,4 +752,20 @@ def test_slot_names_can_start_with_call
       end
     end
   end
+
+  def test_inline_slot_html_escape_with_integer
```
camertron, line 756: These tests don't appear to assert anything?

test/sandbox/test/slotable_test.rb
```diff
@@ -752,4 +752,20 @@ def test_slot_names_can_start_with_call
...
+    render_inline InlineSlotHtmlEscapeComponent.new(heading: "lorem ipsum", counter: 12, url: "/some_url") do |component|
+      component.with_empty_state do
+        "empty state"
```
camertron, line 759: Would it make sense for this slot to return a non-string value? Otherwise we're testing the existing behavior and not the behavior mentioned in the issue.

camertron (changes requested):
Thanks for taking a look at this @reeganviljoen!

camertron:
Hey @reeganviljoen, in the interest of getting a release out today I forked your branch and created [another PR](https://github.com/ViewComponent/view_component/pull/1961) with a few small fixes. Thanks for tackling this!!

reeganviljoen:
@camertron sorry I didd't see this as I was in the gym 😄

camertron:
@reeganviljoen no worries! Sorry for closing, but your name is still in the changelog for this fix!

reeganviljoen:
@camertron no problem I fully understand the urgency, just letting you know I wasn't blowing off your response

PR #1960, opened 2024-01-08
Add output_preamble to match postamble

CrookedGrin:

Add output_preamble to match existing output_postamble, as suggested in [this issue](https://github.com/ViewComponent/view_component/issues/1931).

To keep this PR as simple as possible, we've just mirrored the existing output_postamble code.

This is a follow-up to [another PR](https://github.com/ViewComponent/view_component/pull/1932), incorporating the changes suggested there.


lib/view_component/base.rb
```diff
@@ -104,11 +104,16 @@ def render_in(view_context, &block)
...
+        rendered_template = safe_render_template_for(@__vc_variant).to_s
+
+        if output_preamble.blank? && output_postamble.blank?
+          rendered_template
         else
-          safe_render_template_for(@__vc_variant).to_s + safe_output_postamble
+          preamble = output_preamble.blank? ? "" : safe_output_preamble
+          postamble = output_postamble.blank? ? "" : safe_output_postamble
+
+          preamble + rendered_template + postamble
```
camertron, line 116: `#maybe_escape_html` already checks for blankness, so I think this code can be simplified a bit:

```suggestion
          safe_output_preamble + rendered_template + safe_output_postamble
```

camertron (changes requested):
Nice, thanks!

camertron:
Not sure what's going on with this PVC failure, will investigate.

lib/view_component/base.rb
CrookedGrin: Hey @camertron, I accepted your suggestion, but it looks like the build then started failing on the Primer tests. Any idea why this change could have done that?

camertron: Yeah, not sure, will look into it this afternoon 👍

camertron:
Ok, the failure is happening on PVC main too, so this PR shouldn't be blocked.

PR #1961, opened 2024-01-08
Fix html escaping non strings

camertron:
See: https://github.com/ViewComponent/view_component/pull/1958


camertron:
@reeganviljoen I'm not sure a test for JSON is necessary, since `html_escape({})` returns `"{}"` and calling `#to_json` on a hash or array returns a string. In all cases, escaping it should be fine if it's being emitted in an HTML context.

camertron:
PVC failures are unrelated to this change, i.e. are non-blocking 👍

PR #1965, opened 2024-01-08
Return in maybe_escape_html if in test environment

kathleenteamshares:

The changes [here](https://github.com/ViewComponent/view_component/compare/v3.8.0...v3.9.0#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9eR311) caused our specs to fail with `Failure/Error: subject { render_inline(component) } #<Double (anonymous)> received unexpected message :format with (no args)` due to being unable to call `format` on the double we define as a parameter.

Return `text` in `maybe_escape_html` when in a test environment.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Update `maybe_escape_html` to return when in `test` environment.
+
+    *Kathleen Cesar*
+
 * Add `output_preamble` to match `output_postamble`, using the same safety checks.
 
     *Kali Donovan*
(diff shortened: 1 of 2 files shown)


kathleenteamshares:
Not sure what is going on with the tests, but i think it is unrelated to this change.

camertron:
Hey @kathleenteamshares, thanks for bringing this up. Unfortunately I think we still want to escape (and possibly print a warning) in tests, since I think that's primarily the place devs will notice HTML safety issues and correct them before shipping (that was certainly the case for us).

> Not sure what is going on with the tests, but i think it is unrelated to this change.

Those failing tests are actually the ones I wrote to make sure HTML escaping happens automatically and that a warning is generated, so they are very much related to this change 😅 

I would suggest either 1) using a real request object in your tests, or 2) defining the `format` method on the double.

Issue #1976, opened 2024-01-15
Create a Strict Helpers mode


reeganviljoen:
## Feature request
I have been thinking about extending https://github.com/ViewComponent/view_component/issues/1848 by adding a config that disables the use of helpers through `helpers.<helper_name>` and forces you to use `use_helpers <helper_name>` when you need helpers

A possible example could be:
```
config.view_component.use_strict_helpers = true
```

### Motivation

This can help in developers that wish to make it easier to reason when using global state as is the intention with 
```
https://github.com/ViewComponent/view_component/issues/1848
```

joelhawksley:
Closing as not planned per https://github.com/ViewComponent/view_component/pull/1987#issuecomment-2936517879 ❤

PR #1987, opened 2024-02-12
Add strict helpers

reeganviljoen:
closses #1976 

Add a `config.view_component.strict_helpers_enabled` mode to view component which will throw an `ViewComponent::StrictHelperError` when `helpers.<some_helper>` is used.

This 

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -43,9 +43,8 @@ module ViewComponent
     RESERVED_PARAMETER = :content
     VC_INTERNAL_DEFAULT_FORMAT = :html
 
-    use_helpers :form_authenticity_token, :protect_against_forgery?, :config, :content_security_policy_nonce
-
     # For CSRF authenticity tokens in forms and Content Security Policy nonces
+    use_helpers :form_authenticity_token, :protect_against_forgery?, :config, :content_security_policy_nonce
 
     attr_accessor :__vc_original_view_context
 
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -228,7 +228,7 @@ module ViewComponent
     def helpers
       raise HelpersCalledBeforeRenderError if view_context.nil?
 
-      raise StrictHelperError if ViewComponent::Base.strict_helpers_enabled
+      raise StrictHelperError if !GlobalConfig.helpers_enabled || component_config.strict_helpers_enabled
       # Attempt to re-use the original view_context passed to the first
       # component rendered in the rendering pipeline. This prevents the
       # instantiation of a new view_context via `controller.view_context` which
@@ -519,6 +519,16 @@ module ViewComponent
         # Compile so child will inherit compiled `call_*` template methods that
         # `compile` defines
         compile
+
+        if child.superclass == ViewComponent::Base
+          child.define_singleton_method(:component_config) do
+            @@component_config ||= Rails.application.config.view_component.component_defaults.inheritable_copy
+          end
+        else
+          child.define_singleton_method(:component_config) do
+            @@component_config ||= superclass.component_config.inheritable_copy
+          end
+        end
 
         # Give the child its own personal #render_template_for to protect against the case when
         # eager loading is disabled and the parent component is rendered before the child. In
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -520,16 +520,6 @@ module ViewComponent
         # `compile` defines
         compile
 
-        if child.superclass == ViewComponent::Base
-          child.define_singleton_method(:component_config) do
-            @@component_config ||= Rails.application.config.view_component.inheritable_copy
-          end
-        else
-          child.define_singleton_method(:component_config) do
-            @@component_config ||= superclass.component_config.inheritable_copy
-          end
-        end
-
         # Give the child its own personal #render_template_for to protect against the case when
         # eager loading is disabled and the parent component is rendered before the child. In
         # such a scenario, the parent will override ViewComponent::Base#render_template_for,
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -522,6 +522,18 @@ module ViewComponent
         # `compile` defines
         compile
 
+        child.include ActiveSupport::Configurable
+
+        if child.superclass == ViewComponent::Base
+          child.define_singleton_method(:config) do
+            @@config ||= Rails.application.config.view_component.inheritable_copy
+          end
+        else
+          child.define_singleton_method(:config) do
+            @@config ||= superclass.config.inheritable_copy
+          end
+        end
+
         # Give the child its own personal #render_template_for to protect against the case when
         # eager loading is disabled and the parent component is rendered before the child. In
         # such a scenario, the parent will override ViewComponent::Base#render_template_for,
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require "action_view"
-require "active_support/configurable"
 require "view_component/collection"
 require "view_component/compile_cache"
 require "view_component/compiler"
@@ -231,7 +230,6 @@ module ViewComponent
     # @return [ActionView::Base]
     def helpers
       raise HelpersCalledBeforeRenderError if view_context.nil?
-
       # Attempt to re-use the original view_context passed to the first
       # component rendered in the rendering pipeline. This prevents the
       # instantiation of a new view_context via `controller.view_context` which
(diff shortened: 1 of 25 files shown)


lib/view_component/base.rb
```diff
@@ -224,7 +232,7 @@ def controller
     # @return [ActionView::Base]
     def helpers
       raise HelpersCalledBeforeRenderError if view_context.nil?
-
+      raise StrictHelperError if ViewComponent::Base.config.strict_helpers_enabled
```
joelhawksley, line 235: I wonder if this feature might be less confusing if we just had it be `config.helpers_enabled` and defaulting it to true?

joelhawksley (commented):
One thought on naming.

As a bigger question: what happens if one turns this on for their app and then wants to use a gem that provides ViewComponents that use helpers?

lib/view_component/base.rb
reeganviljoen: @joelhawksley I think that's a good idea, when I have some time I will apply this feedback

reeganviljoen: @joelhawksley I have changed the config to this, if you could review it again it would be a massive

joelhawksley:
@reeganviljoen thoughts on my comment from my review above?

> As a bigger question: what happens if one turns this on for their app and then wants to use a gem that provides ViewComponents that use helpers?

reeganviljoen:
> One thought on naming.
> 
> As a bigger question: what happens if one turns this on for their app and then wants to use a gem that provides ViewComponents that use helpers?

@joelhawksley I don't think this will break any existing gems, unless the user enables the config, however if the user does, it wont work unless the gem documents which helpers are used so the user can add them themselves

I do think this is a positive improvement however, as just adding helpers increases coupling to global state, wich I feel is bad and is against the whole point of view_component. If a gem wants to add helpers they should let the user decide what helpers and how much coupling to global state they want

boardfish:
If a gem providing ViewComponents uses helpers, I'm not sure we should mandate that it's subject to `strict_helpers` in the same way as the rest of the app.

Gems providing components that use helpers would need to be updated to use the new syntax, and waiting on providers to do that doesn't feel like an ideal experience for devs using ViewComponent to me. It's absolutely something gem maintainers should do post-haste, but I think configuration could make the transition period easier for consumers of these gems. 

I can see us going a few different ways with this:

1. Gems adhere to their own configuration. I don't think there's a layer of config in place to do this just yet – gems still inherit from app config. 
2. Enabling `strict_helpers` enforces this for gems too. Making this available as an option would be good, but I don't think it should be the only one.
3. We expose a lambda for `strict_helpers`. This could allow folks the flexibility to allowlist on a gem-by-gem basis, or report strict helper violations (e.g. to an error reporting platform/in logs). I'd like to see this become something we do more often potentially. If we introduce configuration that controls a breaking change, this would be good to allow folks to understand (through their test suite or at runtime) where changes need to be made.

joelhawksley:
@boardfish thanks for sharing your concerns- I generally agree with you. 

I think we might be better off having this be a macro folks can set on a component, likely on their ApplicationComponent.

reeganviljoen:
@joelhawksley noted, I wills see if I can work on something to macro based system working

danini-the-panini:
Some more potential solutions to the Gem problem:

1. Config is provided in a superclass, so if your app has a common parent `ApplicationComponent` which houses the strict helpers config, then it can be propagated to the children but not to external gem components. This would also allow devs working on legacy projects to migrate existing components one at a time to use strict helpers. Additionally this would allow gems to use their own strict helpers config by doing the same.
2. Make it a deprecation warning instead of an error, that way it won't break things but it will give devs the tool to migrate their existing components and potentially to make PRs into gems to migrate theirs as well

danini-the-panini:
Another useful feature might be allowing to pass in a block that gets called every time helpers is referenced, so that devs can hook into their error tracker e.g. Sentry

joelhawksley:
@reeganviljoen just checking in- are you planning to complete this work?

reeganviljoen:
@joelhawksley Absolutely, just been verry busy

reeganviljoen:
@joelhawksley what do you think about @danini-the-panini comment, I think its a good idea, I could work that into our current instrumentation setup in a seperate PR

reeganviljoen:
@joelhawksley I think that this api is complete as the main concern was gem compability but the recent addition to `use_helpers` provides a path to working. with gems in a strict mode

joelhawksley:
@reeganviljoen thanks for cleaning up this PR. For now, I'm going to mark it as blocked on @boardfish's work here: https://github.com/ViewComponent/view_component/issues/1945#issuecomment-2156025923 ❤️

boardfish:
This might actually be ideal as the first component-local config option due to the small scope of the PR. I'll see if I can grab some time this weekend to figure out if I can use this to make a start.

boardfish:
Had a very scrappy crack at this in https://github.com/ViewComponent/view_component/pull/2092/commits/35017a96df72d7bb5c6bfe8f5ac51ceec05e36e2 – it looks possible, and we could probably even ship that with a few more tests behind it and some cleanup (still need to document things and get the use of the private variable names vs. public method names straight). Going forward, though, we probably want to make a nice internal DSL that'll support migrating these config options across slowly but surely.

reeganviljoen:
@boardfish I aprreciate the help, if you want to pair on this issue I be happy to help.

boardfish:
There's perhaps an even cleaner solution in `ActiveSupport::Configurable` that [I've now found](https://github.com/ViewComponent/view_component/pull/2092/commits/56f28b94581e4dfce26da887640ecc22e93c0578) - we were previously including this but hadn't perhaps used it to its full extent as far as I'm aware. I've got a busy week ahead but I'd like to look at wrapping it up as soon as I can to unblock you - I'd appreciate any early feedback anyone has on [the PR](https://github.com/ViewComponent/view_component/pull/2092/files/313a9226016063b869b00d51ae3a37671231492f..56f28b94581e4dfce26da887640ecc22e93c0578).

reeganviljoen:
@boardfish your sugegsted rebase is done

boardfish:
Awesome ✨ In #2092, the [changes to `Base#inherited`](https://github.com/ViewComponent/view_component/pull/2092/files#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9eR518-R532) introduce component-local config – where you're currently checking the `helpers_` config options against `ViewComponent::Base`, you should be able to use `component_config` on the component instance instead. You'll also need to update some of the tests so that they don't stub the global config and instead use the value from the component's local config.

boardfish:
On reflection, I'm thinking that the implementation I've got [here](https://github.com/ViewComponent/view_component/pull/2092/files#diff-b69c75ffd85596e0b5fca1327059ecf3b2930f3f77004c78a268fa85e8ec8d9eR518-R532) probably might need to change a touch – that makes all global options available on component-local config, when in reality changing any settings other than the ones related to helpers won't do anything. I expect that might cause some confusion if left in.

Perhaps it would be good to have a `component_defaults` key in the global config which things that inherit from `ViewComponent::Base` will use an `inheritable_copy` of instead.

boardfish:
Made those changes myself, but there are a few icky commits left in – I figure that if the base PR is merged, those will be easier to clean up.

reeganviljoen:
@boardfish much thanks will take a look tonight

joelhawksley:
@reeganviljoen I'm going to move this to draft as it is pretty out of date and remains blocked by @boardfish's work ❤️

boardfish:
Hopefully it should be unblocked soon – I need to extract the component-local config work out of #2210 so that we can introduce this in V3.

reeganviljoen:
@joelhawksley I am sorry I haven't been too active lately, but I will be taking a look at this shortly

@boardfish I would really appreciate that but no rush, if push comes to shove I will rebase this branch to get something workable

boardfish:
@reeganviljoen https://github.com/ViewComponent/view_component/pull/2230 is in a position that you now may want to rebase off that branch instead. (Easier said than done by the looks of things – let me know if you need any help, and you can always force this branch back to [51e99fb](https://github.com/ViewComponent/view_component/commit/51e99fb61535163410e448964e7c1aff9b35f80e) (its current most recent commit) if things go sideways!)

joelhawksley:
@reeganviljoen I believe this PR is no longer blocked. Are you still hoping to land it?

reeganviljoen:
@joelhawksley Ill get working on it now, thanks

joelhawksley:
@reeganviljoen after giving this a fresh read, I've decided to close the PR and associated issue as not planned.

While I agree that `helpers` can be used improperly, I don't think it's up to us as framework authors to enforce its usage by consumers. The functionality mimics https://api.rubyonrails.org/classes/ActionController/Helpers.html, which does not have a configuration toggle.

If one wants to discourage access, I could see using a custom Rubocop linter or perhaps more aggressively, re-defining `helpers` on `ApplicationComponent` to `raise`.

Issue #1982, opened 2024-01-27
Support for rack-mini-profiler


htcarr3:
## Feature request
Currently only views and partials are broken out in the speed badge report that gets created by rack-mini-profiler. I assume this requires some tie-in on instrumentation. My apologies if this is already supported, I just couldn't find any examples or documentation for it, and maybe this is just a documentation issue (or maybe this is some configuration that makes sense to live on rack-mini-profiler)!

### Motivation
rack-mini-profiler seems to be the ubiquitous first stop for analyzing performance in rails/rack apps. Since ViewComponent (essentially) replaces partials in many instances, it seems natural to expect the rendering timings to show up there as well. Thank you for your consideration!

### Update/Example

After diving deeper into the instrumentation documentation for ViewComponent, and looking at the source code for Rack::MiniProfilerRails (which is essentially the official railtie), something like this seems to work:
``` ruby
# somewhere in the boot path for a rails app, like development.rb

Rack::MiniProfilerRails.subscribe('render.view_component') do |_name, start, finish, _id, payload|
  Rack::MiniProfilerRails.render_notification_handler(
    Rack::MiniProfilerRails.shorten_identifier(payload[:identifier]),
    finish,
    start
  )
end
```
If this seems correct, maybe a small addition to the instrumentation docs could be helpful? I'm happy to put together a PR.

htcarr3:
Related: #1490

joelhawksley:
@htcarr3 sure, please write a PR!

htcarr3:
> @htcarr3 sure, please write a PR!

I will put together a pr soon, thank you!

htcarr3:
@joelhawksley, there is certainly no rush on this, but I do have a pr up whenever you find some time to review

joelhawksley:
@htcarr3 thanks for the PR! Just merged it ❤

PR #1994, opened 2024-02-19
Docs: Use rack-mini-profiler for profiling ViewComponent rendering

htcarr3:

This PR aims to document how to profile ViewComponent rendering alongside regular rails views, partials, and controller actions with the rack-mini-profiler gem. This was discussed in issue #1982.


htcarr3:
I just realized that I included deleting the docs/Gemfile.lock file in this pr, I will fix that.

PR #2005, opened 2024-03-21
Use component path for generating RSpec files

neanias:

Generally, the path to an RSpec file should match the app path, but substituting `app` for `spec`. If a user changes the generation path for their components, then the generator should use an appropriate path that matches that. In the example where the user changes the generation path from `app/components` to `app/views/components`, the RSpec component spec should be created in `spec/views/components`.

The logic chosen here checks that the component is still being generated into the `app` dir, otherwise it will just fallback on the usual destination of `spec/components`. I've used simple string matching of `start_with?("app#{File::SEPARATOR}")`, but maybe using something in the `File` or `Pathname` classes would be more flexible.

Since this includes the `AbstractGenerator`, we can remove the `file_name` method as it's provided by the module.

(Hi, Simon!)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,16 +12,9 @@ nav_order: 5
 
 * Use component path for generating RSpec files.
 
-  When generating new RSpec files for components, the generator will use the
-  `view_component_path` value in the config to decide where to put the new spec
-  file. For instance, if the `view_component_path` option has been changed to
-  `app/views/components`, the generator will put the spec file in
-  `spec/views/components`. **If the `view_component_path` doesn't start with
-  `app/`, then the generator will fall back to `spec/components/`.**
+  When generating new RSpec files for components, the generator will use the `view_component_path` value in the config to decide where to put the new spec file. For instance, if the `view_component_path` option has been changed to `app/views/components`, the generator will put the spec file in `spec/views/components`. **If the `view_component_path` doesn't start with `app/`, then the generator will fall back to `spec/components/.**`
 
-  This feature is enabled via the
-  `config.view_component.generate.use_component_path_for_rspec_tests`
-  option. Set to `true` to enable this new functionality.
+  This feature is enabled via the `config.view_component.generate.use_component_path_for_rspec_tests` option, defaulting to `false`. The default will change to `true` in ViewComponent v4.
 
     *William Mathewson*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,21 @@ nav_order: 5
 
 ## main
 
+* Use component path for generating RSpec files.
+
+  When generating new RSpec files for components, the generator will use the
+  `view_component_path` value in the config to decide where to put the new spec
+  file. For instance, if the `view_component_path` option has been changed to
+  `app/views/components`, the generator will put the spec file in
+  `spec/views/components`. **If the `view_component_path` doesn't start with
+  `app/`, then the generator will fall back to `spec/components/`.**
+
+  This feature is enabled via the
+  `config.view_component.generate.use_component_path_for_rspec_tests`
+  option. Set to `true` to enable this new functionality.
+
+    *William Mathewson*
+
 ## 3.11.0
 
 * Fix running non-integration tests under Rails main.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,9 +12,9 @@ nav_order: 5
 
 * Use component path for generating RSpec files.
 
-  When generating new RSpec files for components, the generator will use the `view_component_path` value in the config to decide where to put the new spec file. For instance, if the `view_component_path` option has been changed to `app/views/components`, the generator will put the spec file in `spec/views/components`. **If the `view_component_path` doesn't start with `app/`, then the generator will fall back to `spec/components/.**`
+When generating new RSpec files for components, the generator will use the `view_component_path` value in the config to decide where to put the new spec file. For instance, if the `view_component_path` option has been changed to `app/views/components`, the generator will put the spec file in `spec/views/components`. **If the `view_component_path` doesn't start with `app/`, then the generator will fall back to `spec/components/.**`
 
-  This feature is enabled via the `config.view_component.generate.use_component_path_for_rspec_tests` option, defaulting to `false`. The default will change to `true` in ViewComponent v4.
+This feature is enabled via the `config.view_component.generate.use_component_path_for_rspec_tests` option, defaulting to `false`. The default will change to `true` in ViewComponent v4.
 
     *William Mathewson*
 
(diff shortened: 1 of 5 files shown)


joelhawksley (commented):
This generally makes sense to me! Two questions:

1) Do you think it's worth adding a test for this change?
2) Do you think this is a breaking change? If it isn't, I think we should at least add a more detailed changelog entry.

neanias:
Tests probably aren't a bad idea — I'll get one run up. Good question re breaking change. I suppose it is since it's actively changing generator behaviour... Rails allows overrides of builtin generators, so I can write some documentation about how to set it back to the old way? I can also expand on the changelog entry. Lmk what you think!

joelhawksley:
@neanias since it's a breaking change, we'll need to either hold it for the next major release or put it behind a config flag. I would vote we do the latter, adding a config flag that defaults to `false` for the new behavior. We can then remove the flag in the next major release!

neanias:
@joelhawksley Sounds like a plan. Should I put it under the `generate` collection of config options?

neanias:
The config option feels like a bit of a mouthful, but I think it's clear enough?

docs/CHANGELOG.md
```diff
@@ -10,6 +10,21 @@ nav_order: 5
...
+
+  When generating new RSpec files for components, the generator will use the
+  `view_component_path` value in the config to decide where to put the new spec
+  file. For instance, if you have your view components in
+  `app/views/components`, the generator will put the spec file in
+  `spec/views/components`. **If the `view_component_path` doesn't start with
+  `app/`, then the generator will fall back to `spec/components/`.**
+
+  This feature is enabled via the
+  `config.view_component.generate.check_component_path_for_rspec_components`
```
joelhawksley, line 23: How about?

```suggestion
  `config.view_component.generate.use_component_path_for_rspec_tests`
```

neanias: Rolls off the tongue much better!

test/sandbox/test/generators/rspec_generator_test.rb
boardfish, line 1: For completeness, it'd be good to include a test case here for if `generate.use_component_path_for_rspec_tests` is `true` but a custom component path isn't set.

boardfish:
👋 Nice to see you around here, @neanias! And thanks for a very complete PR 💯

test/sandbox/test/generators/rspec_generator_test.rb
neanias: Please hold...

boardfish:
Also, worth verifying if we need the same for Minitest.

neanias:
Just noticed I've used "you" in my documentation for the new config flag. Is that ok or should I change it?

joelhawksley:
> Just noticed I've used "you" in my documentation for the new config flag. Is that ok or should I change it?

I'd prefer to avoid "you". Mind making the change? I'll review the language in this PR for ya ❤️

neanias:
Ok, I think that's everything? Re the potential equivalent Minitest change, I think that should come under a separate PR

neanias:
Left "you" in the CHANGELOG 🤦

neanias:
God, really burning through the GHAction minutes with this one!

Spone (approved):
Good idea!

docs/CHANGELOG.md
```diff
@@ -10,6 +10,21 @@ nav_order: 5
...
+  When generating new RSpec files for components, the generator will use the
+  `view_component_path` value in the config to decide where to put the new spec
+  file. For instance, if the `view_component_path` option has been changed to
+  `app/views/components`, the generator will put the spec file in
+  `spec/views/components`. **If the `view_component_path` doesn't start with
+  `app/`, then the generator will fall back to `spec/components/`.**
+
+  This feature is enabled via the
+  `config.view_component.generate.use_component_path_for_rspec_tests`
+  option. Set to `true` to enable this new functionality.
```
joelhawksley, line 24:
```suggestion
  When generating new RSpec files for components, the generator will use the `view_component_path` value in the config to decide where to put the new spec file. For instance, if the `view_component_path` option has been changed to `app/views/components`, the generator will put the spec file in `spec/views/components`. **If the `view_component_path` doesn't start with `app/`, then the generator will fall back to `spec/components/.**1

  This feature is enabled via the
  `config.view_component.generate.use_component_path_for_rspec_tests`
  option. Set to `true` to enable this new functionality.
```

docs/CHANGELOG.md
```diff
@@ -10,6 +10,14 @@ nav_order: 5
 
 ## main
 
+* Use component path for generating RSpec files.
+
+When generating new RSpec files for components, the generator will use the `view_component_path` value in the config to decide where to put the new spec file. For instance, if the `view_component_path` option has been changed to `app/views/components`, the generator will put the spec file in `spec/views/components`. **If the `view_component_path` doesn't start with `app/`, then the generator will fall back to `spec/components/.`**
```
neanias, line 15:
```suggestion
When generating new RSpec files for components, the generator will use the `view_component_path` value in the config to decide where to put the new spec file. For instance, if the `view_component_path` option has been changed to `app/views/components`, the generator will put the spec file in `spec/views/components`. **If the `view_component_path` doesn't start with `app/`, then the generator will fall back to `spec/components/`.**
```

neanias: Also, should the body of the change not be indented? Rendering looks a bit weird when they're not indented with the "subject"?

lib/view_component/config.rb
```diff
@@ -79,6 +79,19 @@ def defaults
...
       # `"test/components/previews"` otherwise
+      #
+      # #### `#use_component_path_for_rspec_tests`
+      #
+      # Whether to use the `config.view_component_path` when generating new
+      # RSpec component tests:
+      #
+      #     config.view_component.generate.use_component_path_for_rspec_tests = true
+      #
+      # When this is `true`, the generator will use the `view_component_path` as
```
joelhawksley, line 90:
```suggestion
      # When set to `true`, the generator will use the `view_component_path` to
```

lib/view_component/config.rb
```diff
@@ -79,6 +79,19 @@ def defaults
...
+      # its guide for deciding where to generate the new RSpec component test.
```
joelhawksley, line 91:
```suggestion
      # decide where to generate the new RSpec component test.
```

lib/view_component/config.rb
```diff
@@ -79,6 +79,19 @@ def defaults
...
+      # For example, if the `view_component_path` option has been changed to
```
joelhawksley, line 92:
```suggestion
      # For example, if the `view_component_path` is
```

lib/view_component/config.rb
```diff
@@ -79,6 +79,19 @@ def defaults
...
+      # `app/views/components`, then the generator will create a new spec file
+      # in `spec/views/components/` rather than the usual `spec/components/`.
```
joelhawksley, line 94:
```suggestion
      # in `spec/views/components/` rather than the default `spec/components/`.
```

joelhawksley (approved):

joelhawksley:
The PVC failure is unrelated. Merging!

PR #2015, opened 2024-04-13
Fix inline component throws error with strip_trailing_whitespace

### LINKED ISSUE #2013: undefined method `rstrip!' for #<struct ViewComponent::InlineTemplate::Template
Hi, apologies for not opening a PR - the devcontainer configuration failed when I tried to load it (MacOS + Docker Desktop 4.22)

I'm running into issues with trailing whitespace and inline template handlers. Firstly, if you have a `strip_trailing_whitespace` and an `erb_template` in the same component, rendering will raise an error. You can see this happening with `SimpleInlineComponent` and `NestedComponent` in my example app.

This next part might be user error, but the reason I was trying to add these `strip_trailing_whitespace` calls was due to the fact stray newline characters are being added into the markup when my `NestedComponent` is rendered. I would appreciate any guidance on why this is happening and what I can do to fix it

Cheers

### Steps to reproduce
https://github.com/abr-storm/view-component-newlines

### Expected behavior
No error raised when visiting `/simple`

### Actual behavior
Error is raised

**Backtrace**:
```
undefined method `rstrip!' for #<struct ViewComponent::InlineTemplate::Template source="<div>Fooo</div>\n", language="erb", path="/Users/alex/src/view-component-newlines/app/components/simple_inline_component.rb", lineno=4>
```

### System configuration
**Rails version**:
7.1.3

**Ruby version**:
3.2.2

**Gem version**:
`main` (3.11.0)

reeganviljoen:
closes #2013 

Inline components throws
 ```
/Users/<redacted>/view_component/lib/view_component/compiler.rb:251:in `compiled_inline_template': undefined method `rstrip!' for an instance of ViewComponent::InlineTemplate::Template (NoMethodError)

 template.rstrip! if component_class.strip_trailing_whitespace?
``` 
 when  strip_trailing_whitespace is used, so this is an attempt to make `strip_trailing_whitespace` behave in inline components as they do for normal components
I removed `template.rstrip! if component_class.strip_trailing_whitespace?` as the template object is a `ViewComponent::InlineTemplate::Template` and not a string
I then added `template = template.source.dup` so tat the template string could be unfrozen for compile_template to remove the trailing white space

## Performance Imapct

before change
```
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-darwin21]
    inline_component   323.000 i/100ms
    inline_component      3.118k (± 4.8%) i/s -     31.331k in  10.075606s
```
After
```
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-darwin21]
strip_trailing_white_space
                       228.000 i/100ms
strip_trailing_white_space
                          3.057k (± 4.5%) i/s -     30.552k in  10.019824s
```

### CODE REVIEW

File: test/sandbox/app/components/nested_component.rb [hunk 1]
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+class NestedComponent < ViewComponent::Base
```
Spone, line 3: Can we make this component a bit simpler, while still having the test fail? I think it would help understanding what's going on.

reeganviljoen:
@Spone I have simplified the failing test with while keeping the same error

reeganviljoen:
it seems strip_trailing_whitespace doesn't work with inline templates, I have created a fix locally that seesm to work, I will push it up latter

Spone:
Awesome, that's really useful!

reeganviljoen:
@Spone I have updated the description, to reflect the actual bug, I then also fixed the test and have added the fix for the bug

reeganviljoen:
@Spone I write a benchmark by rewriting the old performance inline_componet on a new component and testing that, it seems this is actually faster by a small margin see the description for more details

camertron (approved):
This makes sense to me, thanks @reeganviljoen!

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Fixed a bug where inline templates where unable to remove trailing whitespace without throwing an error
```
```suggestion
* Fixed a bug where inline templates where unable to remove trailing whitespace without throwing an error.
```

Spone (approved):

PR #2021, opened 2024-05-10
use sidecar template on docs landing page

joelhawksley:
At RailsConf (and elsewhere) I got the feedback from folks that inline templates aren't for everyone. In the interest of not scaring newcomers away from the framework, I've opted to revert to using a sidecar template on the landing page of the docs site.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,27 +10,32 @@ A framework for creating reusable, testable & encapsulated view components, buil
 
 ## What's a ViewComponent?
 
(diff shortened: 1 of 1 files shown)


BlakeWilliams (approved):

Spone (approved):

reeganviljoen:
@joelhawksley I largely agree but I wonder why people don't like it. Maybe a community survey of some kind could give us some insight?

reeganviljoen:
Update I am planning on posting a poll on the LinkedIn rails group but I need a few options for why people don't like them, @joelhawksley since you have some feedback do you know what some options could be

PR #2012, opened 2024-04-08
add Ruby head and YJIT to CI

reeganviljoen:
### What are you trying to accomplish?

Add ruby head to CI as well as YJIT as discussed in #2010 by me and @joelhawksley 

### Anything you want to highlight for special attention from reviewers?
Ruby head seems to have to be failing on CI which needs further investigation into whether it is a ruby bug or if it really is a breaking change

PR #2010, opened 2024-04-04
Fix typo

cllns:
Small documentation fix

PR #2030, opened 2024-05-22
add Ruby head and YJIT to CI

reeganviljoen:
I created this pr as a continuation of #2012 as a separate branch so @joelhawksley and anyone else can add to this branch 

Add ruby head to CI as well as YJIT as discussed in #2010 by me and @joelhawksley 

Ruby head seems to have to be failing on CI which needs further investigation into whether it is a ruby bug or if it really is a breaking change


reeganviljoen:
@joelhawksley I have created this pr for the reasons listed in the description, can you please take another look at this when you can, I also know we are starting to fail on rails head so I will investigate that when I can

camertron:
Thanks for working on this @reeganviljoen! What would you think about always enabling YJIT? Is there a reason we would want to disable it, other than a particular Ruby version not supporting it?

We def need to fix those test failures for Ruby head and Rails head. I wonder if those could be separate PRs?

reeganviljoen:
@camertron I think it would be wise to enable both as some users still use MRI without YJIT, I am just a bit worried a bug in MRI is missed because we only test against YJIT

> We def need to fix those test failures for Ruby head and Rails head. I wonder if those could be separate PRs?
I have already started working ion the rails head issues, spoiler the backtrace is working very differently now, but  am having issues building ruby dev locally

camertron:
> I think it would be wise to enable both as some users still use MRI without YJIT, I am just a bit worried a bug in MRI is missed because we only test against YJIT

Ok that's fair, I just worry that our build matrix is getting awfully large 😵

> I have already started working ion the rails head issues, spoiler the backtrace is working very differently now, but am having issues building ruby dev locally

Awesome thank you! Let me know if I can help re: build issues.

reeganviljoen:
@camertron i do agree it's getting large, maybe we can remove all unsupported rails and ruby version from the matrix

camertron:
> maybe we can remove all unsupported rails and ruby version from the matrix

Yeah, I've considered that... I feel bad about it tho since so many people aren't using the latest versions, and we've always sort of said we'd support older Rails/Ruby versions as long as they were easy to support. So I guess we keep them in the matrix for now.

reeganviljoen:
@camertron maybe as a compromise we can have a sunsetting deadline for older version support ?

camertron:
@reeganviljoen yeah, I can bring it up again at our next sync meeting.

reeganviljoen:
> Awesome thank you! Let me know if I can help re: build issues.
@camertron Thanks much appreciated turns out I was missing a few build dependencies, it works lovely now, the ruby head issue is a mismatched regexp encoding(ASCII) with a matched string(UTF-8) I believe this is a bug, I will report it tomorrow to https://bugs.ruby-lang.org/

reeganviljoen:
@joelhawksley Rails Head is apssing locally for me but not in CI, I suspect a caching issue or something

reeganviljoen:
@joelhawksley @camertron @boardfish @Spone everything is finnaly passing on this PR can we get a look at it please 🙏

.github/workflows/ci.yml
```diff
@@ -26,34 +26,85 @@ jobs:
...
+          - rails_version: "6.1"
+            ruby_version: "head"
+            mode: "capture_patch_disabled"
+            yjit_mode: '1'
+          - rails_version: "6.1"
+            ruby_version: "head"
+            mode: "capture_patch_disabled"
+            yjit_mode: '0'
+
+
```
joelhawksley, line 107:
```suggestion
```

joelhawksley (approved):

joelhawksley:
@reeganviljoen thanks again, this is lovely ❤️

PR #2035, opened 2024-05-28
Fix Ruby head matcher

reeganviljoen:
Fix Ruby Main in Ci

when using `assert_match` and a string is represented in the regexp pattern with two different opening/closing quotes(`')

The error in Ruby main before the fix
```
Expected /undefined method `render_component'/ to match # encoding: US-ASCII
#    valid: true
"undefined method 'render_component' for an instance of #<Class:0x0000000111930960>".
```

we do not have Ruby main in Ci so it would be beneficial for reviewers to run the test suite against ruby head locally until it becomes a part of CI


reeganviljoen:
@boardfish I know you are working on more support for ruby 3.4 thats upcoming, can you take a look at this

docs/CHANGELOG.md
```diff
@@ -18,6 +18,10 @@ nav_order: 5
 
     *Reegan Viljoen*
 
+* Fixed ruby head matcher issue
```
joelhawksley, line 21:
```suggestion
* Fixed ruby head matcher issue.
```

reeganviljoen: thanks for the spot 👍

joelhawksley:
Thanks @reeganviljoen!

reeganviljoen:
> Thanks @reeganviljoen!

Its a pleasure 🎉

PR #2042, opened 2024-06-04
docs/index: add Workbrew.

MikeMcQuaid:
👋🏻 Thanks for the great project!

We're also using ViewComponents over here and ❤️ them.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -278,6 +278,7 @@ ViewComponent is built by over a hundred members of the community, including:
 * [Wecasa](https://www.wecasa.fr/)
 * [WIP](https://wip.co/)
 * [Within3](https://www.within3.com/)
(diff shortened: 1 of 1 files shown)


reeganviljoen:
@MikeMcQuaid great to hear 🎉 ,I see you have a CI failure that can be fixed by adding an entry to the changelog

joelhawksley:
@reeganviljoen I'm fine merging these without changelogs. Maybe we can relax the requirement to exclude changes to that markdown file?

MikeMcQuaid:
Thanks @reeganviljoen and @joelhawksley!

Hope you're well @joelhawksley, it's been a wee while!

reeganviljoen:
@joelhawksley i think thats fair, do we do that as a process thing or by making an exclusion in the actions check ?

PR #2061, opened 2024-07-19
Add ability to put previews into sidecar directory

sethherr:

Enable putting previews (`preview.rb` files) into the sidecar directory.

Organizing components into subdirectories (AKA sidecar) is a helpful way of organizing view components.

The organizing strategy of [view_component-contrib](https://github.com/palkan/view_component-contrib) puts the previews into the sidecar directory, and organizes an example component like this:

```
components/
  example/
    component.html
    component.rb
    preview.rb
    index.css
    index.js
```

Currently, `view_component-contrib` uses an initializer to enable this behavior - making this change would make it possible without an initializer.

Personally, I think it makes sense to organize the test folder in the same extended sidecar way and put the previews in there:

```
spec/
  components/
    example/
      component_spec.rb
      preview.rb
```

But either way, the previews are named `preview.rb` - so the matcher in `lib/view_component/preview.rb` needs to be updated to no longer have a `_` prefix.

This doesn't change the behavior of the generators, it only makes it so that if you decide to put the previews in the sidecar directory, it will be found. I wanted to make the smallest possible change that would enable the functionality.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+## 3.13.1
+
+* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory.
+
 ## 3.13.0
 
 * Add ruby head and YJIT to CI.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@ nav_order: 5
 
 ## main
 
-* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory alongside test files.
+* Look for preview files that end in `preview.rb` rather than `_preview.rb` to allow previews to exist in sidecar directory with test files.
 
     *Seth Herr*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,9 +10,9 @@ nav_order: 5
 
 ## main
 
-## 3.13.1
-
 * Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory alongside test files.
+
+    *Seth Herr*
 
 ## 3.13.0
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -12,7 +12,7 @@ nav_order: 5
 
 ## 3.13.1
 
-* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory.
+* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory alongside test files.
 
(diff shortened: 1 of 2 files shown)


sethherr:
@juanmanuelramallo is this change acceptable? 

If you think this change could cause false matches of non-previews, e.g. because someone has a component named `component_bpreview` - I could add a config to set the preview matcher string - but I didn't think matching incorrectly was likely and just changing the matcher seemed like the smallest, easiest fix.

juanmanuelramallo:
Hey @sethherr thanks for the PR! 

I'm not contributing that much lately to VC, so better ask @joelhawksley or @BlakeWilliams for a review

My preference is to keep non-production code away from the /app folder mostly to avoid loading those objects when running the app in production. But I'm curious to find what others think about it.

joelhawksley:
@sethherr - I agree with @juanmanuelramallo. As previews are not application code, I think it's improper to put them in the component folder.

I appreciate you taking the time to suggest this change, but I have decided not to accept it. ❤️ 

I would be open to allowing this to be set through configuration, though.

sethherr:
@joelhawksley, we're storing our previews in our test directory, as specified above in my comment:

```
spec/
  components/
    example/
      component_spec.rb
      preview.rb
```

So I don't understand your opposition to this - since we aren't doing anything "improper"?

I referenced `view_component-contrib` to make it clear that I wasn't the only one asking for the change.

I can put this in a configuration, but that adds another configuration setting - which I'm not sure is required if this change doesn't negatively impact anyone.

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+## 3.13.1
+
+* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory.
```
```suggestion
* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory alongside test files.
```

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+## 3.13.1
+
+* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory alongside test files.
+
```
joelhawksley, line 16:
```suggestion
* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory alongside test files.

    *Seth Herr*

```

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Look for preview files that end in `preview.rb` rather than `_preview.rb`, so they can be put in the sidecar directory alongside test files.
```
```suggestion
* Look for preview files that end in `preview.rb` rather than `_preview.rb` to allow previews to exist in sidecar directory with test files.
```

joelhawksley (approved):

PR #2063, opened 2024-07-30
Add experimental support for default slot values

joelhawksley:

This PR aims to close #1861 and #614, implementing an experimental approach for default slot values.

I added an experimental module that must be included to use the feature, adopting the API suggested by @BlakeWilliams in https://github.com/ViewComponent/view_component/pull/1861#issuecomment-1785454288.

Do we want to allow default value(s) for `renders_many` slots? I don't think we need to solve it now, but something to consider adding later.


reeganviljoen:
@joelhawksley would you mind adding docs about this, as I feel users would genuinely like this ❤️

test/sandbox/test/slotable_test.rb
```diff
@@ -764,4 +764,16 @@ def test_forwarded_slot_renders_correctly
 
     assert_text "Target content", count: 1
   end
+
+  def test_slotable_default
```
Spone, line 768: Shouldn't we test that specifying the slot correctly overrides the default?

Spone (changes requested):
Great to see this addition! Nice API :)

joelhawksley:
@reeganviljoen done, let me know what you think! https://github.com/ViewComponent/view_component/pull/2063/files#diff-867de23963ba79a34d995f1d16db1d42da92c63d86481bf59031a7d0ebc6f837

reeganviljoen (approved):
@joelhawksley LGTM🚀

Spone (approved):
👌

PR #2064, opened 2024-07-30
Add prefix to use_helper(s)

reeganviljoen:

I am adding a prefix option to `use_helpers` similar to some rails api's like `store_accesor`, `delegate`, etc

I added the api to both accept true in which case it will infer the modules name otherwise a name can be passed and that will be used, I chose this path as it seems it is consistent with rails api's like `store_accesor`, `delegate`, etc


reeganviljoen:
@joelhawksley would you mind having a look

docs/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ nav_order: 5
 
     *Seth Herr*
 
+* Add `prefix:` option to `use_helpers`
```
```suggestion
* Add `prefix:` option to `use_helpers`.
```

joelhawksley (approved):

PR #2069, opened 2024-08-06
remove individual code ownership

joelhawksley:
I'm wondering if it's time for us to consider removing individual code owners from the project, as different folks are now more involved than others were when we put this configuration in place. I think it would be healthier to move away from this ownership model.

What do you think @Spone @camertron @BlakeWilliams @boardfish?


camertron:
Ok with me. Has it been problematic lately?

camertron (approved):

boardfish:
I'm watching the repo so that I can triage and answer incoming requests, but being added as a reviewer thanks to CODEOWNERS is a nice signal that something's happening in an area I'm familiar with. So for that reason I'd probably prefer to keep it around (at least for me!), but I don't mind too much – folks can always @ me instead.

joelhawksley:
> but being added as a reviewer thanks to CODEOWNERS is a nice signal that something's happening in an area I'm familiar with

@boardfish my thinking here is that we should be defaulting to having anyone on the core team be able to review and approve any PR confidently as a default, instead of requiring certain folks to approve changes in certain areas.

@camertron no specific issues, I just noticed on some recent PRs that we were "requiring" reviews from code owners that didn't end up reviewing the PR for one reason or another.

PR #2072, opened 2024-08-09
Add helper instrumentation

reeganviljoen:

Add instrumentation to helpers, this will help people trying to migrate away from using helpers to log incidents to services like sentry

This was a suggestion by @danini-the-panini on #1987 and I feel it is a good addition

### What approach did you choose and why
Added a new method to the instrumentation that overrides helpers to instrument it

### DIFF
--- a/docs/guide/instrumentation.md
+++ b/docs/guide/instrumentation.md
@@ -30,6 +30,7 @@ end
 ```
 
 ## Instrument helpers
+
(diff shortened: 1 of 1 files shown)


camertron:
Hey @reeganviljoen, thanks for this! I don't think this is something we necessarily want in the framework. I would recommend people add it to their apps directly.

PR #2086, opened 2024-09-03
BREAKING: v4 minimum Ruby version should not be EOL

joelhawksley:
**This PR targets the v4 feature branch**

As time marches on, older versions of Ruby are EOL'd. We should align ourselves with this schedule to ensure our tooling can function properly. For example, we cannot use `dependabot` properly until we make this change.


camertron (approved):
Nice yeah, makes sense 👍

PR #2092, opened 2024-09-08
WIP: Extract strict_helpers_enabled? to component-local config

boardfish:
Based on #1987 

Aims to have `strict_helpers_enabled?` able to be set at an application level by `Rails.application.config.view_component` without affecting the base value for the setting stored on `ViewComponent::Base`, which may be consumed by gems.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -37,9 +37,8 @@ module ViewComponent
     RESERVED_PARAMETER = :content
     VC_INTERNAL_DEFAULT_FORMAT = :html
 
-    use_helpers :form_authenticity_token, :protect_against_forgery?, :config, :content_security_policy_nonce
-
     # For CSRF authenticity tokens in forms and Content Security Policy nonces
+    use_helpers :form_authenticity_token, :protect_against_forgery?, :config, :content_security_policy_nonce
 
     # Config option that strips trailing whitespace in templates before compiling them.
     class_attribute :__vc_strip_trailing_whitespace, instance_accessor: false, instance_predicate: false
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -37,6 +37,8 @@ module ViewComponent
     RESERVED_PARAMETER = :content
     VC_INTERNAL_DEFAULT_FORMAT = :html
 
+    use_helpers :form_authenticity_token, :protect_against_forgery?, :config, :content_security_policy_nonce
+
     # For CSRF authenticity tokens in forms and Content Security Policy nonces
 
     # Config option that strips trailing whitespace in templates before compiling them.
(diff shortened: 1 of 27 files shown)


lib/view_component/base.rb
```diff
@@ -518,6 +519,18 @@ def inherited(child)
         # `compile` defines
         compile
 
+        child.include ActiveSupport::Configurable
+
+        if child.superclass == ViewComponent::Base
+          child.define_singleton_method(:config) do
+            @@config ||= Rails.application.config.view_component.inheritable_copy
```
boardfish, line 526: I'll leave a comment to this effect, but ViewComponent::Base starts out with the defaults. These overrides allow subclasses to inherit from Rails.application.config to start with and cascade down from there.

To achieve our goal of making sure that changes to config don't broadly interfere with gems that want to integrate with ViewComponent, we should probably make it easier here for folks to override `config` with a base config.

test/sandbox/test/base_test.rb
```diff
@@ -154,4 +154,21 @@ def test_no_method_error_does_not_reference_missing_helper
     MESSAGE
     assert !exception_message_regex.match?(exception.message)
   end
+
+  def test_strict_helpers_enabled
+    with_config_option(:strict_helpers_enabled, false) do
+      top_level_component_class = Class.new(ViewComponent::Base)
+      refute ViewComponent::Base.config.strict_helpers_enabled?, ".strict_helpers_enabled? should be false by default"
+      # refute ViewComponent::Base.new.config.strict_helpers_enabled?, "#strict_helpers_enabled? should be false by default"
```
boardfish, line 162: Previously I was thinking `ViewComponent::Base` might have a `config` when instantiated, but that's probably not likely to be the case.

There's perhaps also a clash with `use_helper :config` that's in there by default.

boardfish:
Made some good progress today, but I'm struggling to diagnose the failure with the generator. Looks like it's not creating anything at all, but the generator does finish and run its `create_preview_file` method in full.

I should probably also remove `Config.current` in this PR.

test/sandbox/test/base_test.rb
boardfish: Moved component-level config to `component_config`. `ViewComponent::Base` itself still responds on `config`.

lib/view_component/base.rb
boardfish: I guess this'll be done by gems having their own `ApplicationComponent` equivalent now that directly inherits from `Base`.

reeganviljoen:
@boardfish sorry for not being supper active, I have just been extemly busy, I will try my best to make time to look at this this week

boardfish:
No problem at all! I think this is almost ready as far as getting back to parity with what we have already - just need to diagnose and fix that failing test. 

Because this touches config as a whole, I think there's probably wider discussion necessary about how to roll this out – it's a breaking change, and I think it might be difficult to fully preserve the old behaviour as an option.

joelhawksley:
> It's a breaking change, and I think it might be difficult to fully preserve the old behaviour as an option.

I think that approach makes sense, with the plan to remove the old behavior in v4 ❤️

boardfish:
That's good @joelhawksley – so just to be totally clear, the current state of this PR means that folks still interface with global config via `Rails.application.config.view_component`, but can no longer do so via `ViewComponent::Base.config`. Is that something we're happy to include in a v3 minor release?

I guess as far as v4, we can try and draw distinctions between global config (`Rails.application.config.view_component` and component-level config `ApplicationComponent.component_config`) in some follow-ups that make some options component-specific. Does that sounds good?

boardfish:
I've opened #2124 to introduce component-local config in isolation of Reegan's original PR, but I still need to fix the tests.

PR #2094, opened 2024-09-16
Update `with_controller_class` type documentation

TastyPi:

In Yard `[ActionController::Base]` means "an instance of `ActionController::Base`, but this method actually wants a subclass. Tools such as RubyMine complain that the expected type is incorrect when that's not true.

Fixed the doc to use `[Class<ActionController::Base>]` which is the correct type signatire for this meythod.

No

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -42,6 +42,10 @@ nav_order: 5
 
     *Petrik de Heus*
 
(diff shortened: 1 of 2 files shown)


joelhawksley:
@TastyPi thanks for the PR! Might you add a CHANGELOG entry?

TastyPi:
@joelhawksley Done

joelhawksley (approved):

PR #2100, opened 2024-09-25
Underscore internal methods that are public

joelhawksley:
We have plenty of internal methods that are not meant to be used externally, often tagged @private. In a recent office hours, @camertron, @blakewilliams and I noted that we should better indicate when an internal method should not be dependended on. I've done so here by adding a `__vc_` prefix as is convention in other parts of the codebase already.

As a separate exercise, we should probably re-evaluate our public API surface area and see if we can make any public methods private.

### DIFF
--- a/test/sandbox/test/rendering_test.rb
+++ b/test/sandbox/test/rendering_test.rb
@@ -79,12 +79,14 @@ class RenderingTest < ViewComponent::TestCase
   def test_render_without_template
     render_inline(InlineComponent.new)
 
+    assert_predicate InlineComponent, :__vc_compiled?
     assert_selector("input[type='text'][name='name']")
   end
 
   def test_render_child_without_template
     render_inline(InlineChildComponent.new)
 
+    assert_predicate InlineChildComponent, :__vc_compiled?
     assert_selector("input[type='text'][name='name']")
   end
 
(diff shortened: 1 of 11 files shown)


test/sandbox/test/rendering_test.rb
```diff
@@ -92,14 +92,12 @@ def test_raises_error_when_with_content_is_called_without_any_values
   def test_render_without_template
     render_inline(InlineComponent.new)
 
-    assert_predicate InlineComponent, :compiled?
```
camertron, line 95: Why take these assertions out?

camertron (commented):
Looks ok to me, but should we target this change for v4.0 since we're technically changing the public API?

test/sandbox/test/rendering_test.rb
joelhawksley: I thought it wasn't prudent to reference internal APIs in tests, but IMO thats not a strong enough reason to modify tests.

joelhawksley:
@camertron I'm torn. These APIs are not documented publicly, so I'm leaning towards this not being a breaking change.

camertron:
Yeah I hear you... I worry that we're going against semver by doing this though. Even though the methods in question aren't documented, they're still public, and semver is pretty clear about needing a major bump for API changes.

BlakeWilliams:
I could totally see someone using `compile` and a few of the other methods for _some_ reason so I tend to agree @camertron.

I think hiding the compiler like this makes a ton of sense though. I'd opt for v4, personally.

joelhawksley:
Closing for now, will re-open against v4 branch.

PR #2103, opened 2024-09-26
[v4] Remove code paths for unsupported versions of Rails

joelhawksley:
This PR removes the render monkey patch and other various accommodations for now-unsupported versions of Rails.


camertron (approved):
Yessssss

PR #2108, opened 2024-09-29
Add configuration to treat content as a slot

BlakeWilliams:
Like discussed many times before, this adds content as a slot to
ViewComponent so that we can begin separating the block provided when
rendering from content for performance and ergonomic reasons.

This adds two new methods:

* `content_is_a_slot!` - This makes the component _and its descendants_
  treat `content` as a slot, and the block provided to render calls _is
  always executed_.
* `do_not_use_content_as_a_slot!` - This keeps the existing behavior
  where `content` is lazily evaluated and is used to return content.

There's likely more implications from this change that need to be walked
through (like how slots are used, and easily setting content) before
this can be merged.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -135,7 +135,7 @@ module ViewComponent
       @__vc_content_evaluated = false
       @__vc_render_in_block = block
 
-      if self.class.send(:__vc_content_is_a_slot?)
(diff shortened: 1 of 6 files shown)


joelhawksley:
Closing as unplanned: https://github.com/ViewComponent/view_component/issues/1974#issuecomment-3025727862

PR #2109, opened 2024-09-30
Re-add `@private`, undocumented `.identifier` method that was only me…

joelhawksley:
…ant for internal framework use but was used by some downstream consumers. This method will be removed in a coming minor release.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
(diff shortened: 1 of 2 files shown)


joelhawksley:
@jcoyne @rickychilcott @navidemad mind reviewing this change?

jcoyne (approved):
This looks good to me.  If we expected more consumers, other than newrelic (which I think is doubtful), I'd consider using ActiveSupport::Deprecation as it's a bit more flexible, but this is sufficient.

PR #2115, opened 2024-10-01
Add template information to multiple template error messages.

joelhawksley:
In service of helping debug https://github.com/ViewComponent/view_component/issues/2114, this change fleshes out the template error messages to include the current list of templates.

### DIFF
--- a/lib/view_component/errors.rb
+++ b/lib/view_component/errors.rb
@@ -22,7 +22,7 @@ module ViewComponent
 
       if templates
         message << "\n"
-        message << "Templates:"
+        message << "Templates:\n"
         message << templates.map(&:inspect).join("\n")
       end
 
--- a/lib/view_component/errors.rb
+++ b/lib/view_component/errors.rb
@@ -17,8 +17,16 @@ module ViewComponent
   end
 
   class TemplateError < StandardError
-    def initialize(errors)
-      super(errors.join("\n"))
+    def initialize(errors, templates = nil)
+      message = errors.join("\n")
+
+      if templates
+        message << "\n"
+        message << "Templates:"
+        message << templates.map(&:inspect).join("\n")
+      end
+
+      super(message)
     end
   end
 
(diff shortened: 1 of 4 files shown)


lib/view_component/errors.rb
```diff
@@ -17,8 +17,16 @@ def initialize(klass_name)
...
   class TemplateError < StandardError
-    def initialize(errors)
-      super(errors.join("\n"))
+    def initialize(errors, templates = nil)
+      message = errors.join("\n")
+
+      if templates
+        message << "\n"
+        message << "Templates:"
+        message << templates.map(&:inspect).join("\n")
```
BlakeWilliams, line 26: do we want `inspect` output, or just the template name?

joelhawksley: All of the inspect output, as it shows what type of templates there are for the component and where they are defined:

```
Template file and inline render method found for variant 'phone' in VariantTemplateAndInlineVariantTemplateComponent. There can only be a template file or inline render method per variant.

Templates:

#<ViewComponent::Template:0x000000012a9980d8 @component=VariantTemplateAndInlineVariantTemplateComponent, @type=:file, @this_format=:html, @variant=:phone, @lineno=0, @path=\"/Users/joelhawksley/github/view_component/test/sandbox/app/components/variant_template_and_inline_variant_template_component.html+phone.erb\", @extension=\"erb\", @source=nil, @method_name=nil, @defined_on_self=true, @source_originally_nil=true, @call_method_name=\"call_phone\">

#<ViewComponent::Template:0x000000012a997ef8 @component=VariantTemplateAndInlineVariantTemplateComponent, @type=:inline_call, @this_format=:html, @variant=:phone, @lineno=nil, @path=nil, @extension=nil, @source=nil, @method_name=:call_phone, @defined_on_self=true, @source_originally_nil=true, @call_method_name=:call_phone>
```

camertron (approved):

PR #2118, opened 2024-10-03
Allow implicit locale in component file names

chloe-meister:

I'm proposing a fix for [Issue#2113](https://github.com/ViewComponent/view_component/issues/2113).

The refactor of the compiler in v.3.15 changed the way the format is used to define `@call_method_name`, used in the template to compile to component.
 
Once I had understood what the problem was, I tried to see why it was working in previous versions (<3.15). The method name was not defined in the same way back then, and [it didn't use the format](https://github.com/ViewComponent/view_component/blob/2d48c49563826814fd1c224b938ca76ec433b04f/lib/view_component/compiler.rb#L286-L292) at all, only the variant.

From this I assessed that it would be enough to exclude the locale from the format when building the `@call_method_name`. I created a new method that excludes anything before the `.` in formats received from the controller, and I use it to define if something needs to be appended to `call_`.

What might be a bit clumsy is this new `format_without_locale` method, as it literally takes the format symbol from `@this_format`, looks for a `.` character and _assumes_ the format is at the end.

I chose this solution however because it does not modify `@this_format`, which keeps the locale in if there is any.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,24 @@ nav_order: 5
 
     *Chloe Fons*
 
+* Wrap entire compile step in a mutex to make it more resilient to race conditions.
+
+    *Blake Williams*
+
+* Add [Niva]([niva.co](https://www.niva.co/)) to companies who use `ViewComponent`.
+
+    *Daniel Vu Dao*
+
+* Fix `preview_paths` in docs.
+
+    *Javier Aranda*
+
+## 3.17.0
+
+* Use struct instead openstruct in lib code.
+
+    *Oleksii Vasyliev*
+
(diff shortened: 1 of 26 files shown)


docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Fix bug where implicit locales in component filenames threw a `NameError`.
+
+  *Chloe Fons*
```
```suggestion
    *Chloe Fons*
```

joelhawksley:
Thanks for the PR! I should have time to look at this tomorrow and flesh out some edge cases ❤️

chloe-meister:
You're very welcome! Let me know how/if I can help!

PR #2121, opened 2024-10-05
Fix duplicate template error bug

BlakeWilliams:
This change fixes the bug reported where a component will report that it
has both a template and a `call` method defined despite only having an
HTML template.

I was able to reproduce this bug by pulling in Avo (like reported in the
initial bug report) and running `vegeta` (to trigger parallel requests
since I had a hunch this was a race condition):

```
echo "GET http://localhost:3000/avo/resources/cars/1" | vegeta attack -duration=2s
```

This reliably reproduced the error:

```
ActionView::Template::Error (Template file and inline render method found for Avo::CoverPhotoComponent. There can only be a template file or inline render method per component.
Template file and inline render method found for variant '' in Avo::CoverPhotoComponent. There can only be a template file or inline render method per variant.
Templates:
```

I added _a lot_ of debug logs to understand the race condition that I
thought was occurring, and realized that multiple threads are calling
`gather_templates` _and_ mutating the `@templates` array at the
same time. When looking at the old compiler code and realized that this likely
isn't new behavior. This led the investigation towards how we collect
and surface errors or otherwise might modify templates. It turns out
there's a difference in the new and old compiler code after the
refactor:

```ruby
# old
def template_errors
  @__vc_template_errors ||=

# new
def gather_template_errors(raise_errors)
  errors = []
```

_We're not memoizing the errors like we used to_. This is more correct
behavior, but explains how a race condition would make this error case
much more difficult to occur in older versions of the compiler.

This change brings us back to the old behavior by memoizing the errors
we collect in `gather_template_errors` but the `@templates` ivar is
still being mutated. I don't want to change _too much_ in this PR, but a
subsequent change might be wrapping the entire `compile` method in the
`redefinition_lock` (and renaming it to `compile_lock`) to avoid similar
issues in the future.

I did not include a test because it's dfficult to reproduce this race
condition reliably in the test environment.

I believe this _should_ close out https://github.com/ViewComponent/view_component/issues/2114

### DIFF
--- a/lib/view_component/compiler.rb
+++ b/lib/view_component/compiler.rb
@@ -88,7 +88,7 @@ module ViewComponent
           out = branches.each_with_object(+"") do |(conditional, branch_body), memo|
             memo << "#{(!memo.present?) ? "if" : "elsif"} #{conditional}\n  #{branch_body}\n"
           end
-          if default_template = templates.find { _1.variant.nil? && _1.default_format? }
+          if (default_template = templates.find { _1.variant.nil? && _1.default_format? })
             out << "else\n  #{default_template.safe_method_name}\nend"
           end
         end
--- a/lib/view_component/compiler.rb
+++ b/lib/view_component/compiler.rb
@@ -21,6 +21,7 @@ module ViewComponent
     end
 
     def compile(raise_errors: false, force: false)
+      raise TemplateError.new(template_errors) if compiled? && raise_errors && template_errors.any?
       return if compiled? && !force
       return if @component == ViewComponent::Base
 
@@ -30,7 +31,9 @@ module ViewComponent
         @component.superclass.compile(raise_errors: raise_errors)
       end
 
-      return if gather_template_errors(raise_errors).any?
+      if template_errors.present?
+        raise TemplateError.new(template_errors) if raise_errors
+      end
 
       if raise_errors
         @component.validate_initialization_parameters!
@@ -98,8 +101,8 @@ module ViewComponent
       end
     end
 
-    def gather_template_errors(raise_errors)
-      @_errors ||= begin
+    def template_errors
+      @_template_errors ||= begin
         errors = []
 
         errors << "Couldn't find a template file or inline render method for #{@component}." if @templates.empty?
@@ -159,8 +162,6 @@ module ViewComponent
 
           errors << "Colliding templates #{variant_names.sort.map { |v| "'#{v}'" }.to_sentence} found in #{@component}."
         end
-
-        raise TemplateError.new(errors, @templates) if errors.any? && raise_errors
 
         errors
       end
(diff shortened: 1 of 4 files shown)


BlakeWilliams:
The diff looks more drastic than the actual change. [consider using ignore whitespce when reviewing the change](https://github.com/ViewComponent/view_component/pull/2121/files?diff=split&w=1).

lib/view_component/compiler.rb
```diff
@@ -21,6 +21,7 @@ def compiled?
     end
 
     def compile(raise_errors: false, force: false)
+      raise TemplateError.new(template_errors) if compiled? && raise_errors && template_errors.any?
```
BlakeWilliams, line 24: I didn't have time to dive into this, but this was required to get the test suite passing in cases where there are `template_errors`.

I _think_ there might have been a change around how/when templates are compiled in the test environment, since the component looked to be pre-compiled, even though I was running a single test in isolation.

lib/view_component/compiler.rb
```diff
@@ -85,7 +88,9 @@ def define_render_template_for
           out = branches.each_with_object(+"") do |(conditional, branch_body), memo|
             memo << "#{(!memo.present?) ? "if" : "elsif"} #{conditional}\n  #{branch_body}\n"
           end
-          out << "else\n  #{templates.find { _1.variant.nil? && _1.default_format? }.safe_method_name}\nend"
+          if (default_template = templates.find { _1.variant.nil? && _1.default_format? })
+            out << "else\n  #{default_template.safe_method_name}\nend"
```
BlakeWilliams, line 92: I _think_ this leaves a template without a matching `end`? 🤔

lib/view_component/compiler.rb
BlakeWilliams: The templates are now `compiled? => true`, where they used to be `compiled? => false`, fwiw.

lib/view_component/errors.rb
```diff
@@ -20,12 +20,6 @@ class TemplateError < StandardError
     def initialize(errors, templates = nil)
       message = errors.join("\n")
 
-      if templates
```
joelhawksley, line 23: It's a nit, but we don't need to assign to `message` any more (we didn't before I added the debug output)

lib/view_component/compiler.rb
```diff
@@ -30,7 +30,12 @@ def compile(raise_errors: false, force: false)
         @component.superclass.compile(raise_errors: raise_errors)
       end
 
-      return if gather_template_errors(raise_errors).any?
+      if template_errors.present?
```
joelhawksley, line 33: Would you be up for adding a comment here briefly summarizing the steps to reproduce this issue? I just want to make sure we document it for those considering changing this class in the future.

joelhawksley (approved):
Needs changelog 😄

camertron (approved):
I'm not sure I fully understand the solution, but since we've validated this fixes the original issue, I'm ok approving it.

lib/view_component/compiler.rb
BlakeWilliams: I added a comment below, but I'd be happy to link to this PR here, although it's really several difference pieces that contribute since we'd need to comment here, and in a few other places like the template errors method.

I don't think we need it long-term though, since we should likely "just" use a mutex around the whole compilation step and avoid race conditions popping up in areas that aren't wrapped in a mutex like we're vulnerable to today.

joelhawksley (approved):

PR #2124, opened 2024-10-08
Make accommodations for component-local config

boardfish:
Explicitly declares access to current config as global where necessary, and makes it possible for us to introduce a component-local config that inherits from these global options in future.

Unsure of how this interacts with gems with engines that depend on ViewComponent, but I'm hoping in the long run it's a step in the right direction for them that'll allow them to have their own `ApplicationComponent` equivalents that inherit their config from ViewComponent's own defaults.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -95,7 +95,7 @@ namespace :docs do
     require "rails"
     require "action_controller"
     require "view_component"
-    ViewComponent::Base.config.view_component_path = "view_component"
+    ViewComponent::GlobalConfig.view_component_path = "view_component"
     require "view_component/docs_builder_component"
 
     error_keys = registry.keys.select { |key| key.to_s.include?("Error::MESSAGE") }.map(&:to_s)
--- a/app/controllers/concerns/view_component/preview_actions.rb
+++ b/app/controllers/concerns/view_component/preview_actions.rb
@@ -54,12 +54,12 @@ module ViewComponent
 
     # :doc:
     def default_preview_layout
-      ViewComponent::Base.config.default_preview_layout
+      GlobalConfig.default_preview_layout
     end
 
     # :doc:
     def show_previews?
-      ViewComponent::Base.config.show_previews
+      GlobalConfig.show_previews
     end
 
     # :doc:
@@ -102,7 +102,7 @@ module ViewComponent
     end
 
     def prepend_preview_examples_view_path
-      prepend_view_path(ViewComponent::Base.preview_paths)
+      prepend_view_path(GlobalConfig.preview_paths)
     end
   end
 end
--- a/app/helpers/preview_helper.rb
+++ b/app/helpers/preview_helper.rb
@@ -35,7 +35,7 @@ module PreviewHelper
       # Fetch template source via finding it through preview paths
       # to accomodate source view when exclusively using templates
       # for previews for Rails < 6.1.
-      all_template_paths = ViewComponent::Base.config.preview_paths.map do |preview_path|
+      all_template_paths = ViewComponent::GlobalConfig.preview_paths.map do |preview_path|
         Dir.glob("#{preview_path}/**/*")
       end.flatten
 
@@ -80,6 +80,6 @@ module PreviewHelper
   # :nocov:
 
   def serve_static_preview_assets?
-    ViewComponent::Base.config.show_previews && Rails.application.config.public_file_server.enabled
+    ViewComponent::GlobalConfig.show_previews && Rails.application.config.public_file_server.enabled
   end
 end
--- a/app/views/view_components/preview.html.erb
+++ b/app/views/view_components/preview.html.erb
@@ -1,5 +1,5 @@
 <% if @render_args[:component] %>
-  <% if ViewComponent::Base.config.render_monkey_patch_enabled || Rails.version.to_f >= 6.1 %>
+  <% if ViewComponent::GlobalConfig.render_monkey_patch_enabled || Rails.version.to_f >= 6.1 %>
     <%= render(@render_args[:component], @render_args[:args], &@render_args[:block]) %>
   <% else %>
     <%= render_component(@render_args[:component], &@render_args[:block]) %>
@@ -8,6 +8,6 @@
   <%= render template: @render_args[:template], locals: @render_args[:locals] || {} %>
 <% end %>
 
-<% if ViewComponent::Base.config.show_previews_source %>
+<% if ViewComponent::GlobalConfig.show_previews_source %>
   <%= preview_source %>
 <% end %>
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Make accommodations for component-local config to be introduced in future.
+
+    *Simon Fish*
+
 * Remove JS and CSS docs as they proved difficult to maintain and lacked consensus.
 
     *Joel Hawksley*
--- a/lib/rails/generators/abstract_generator.rb
+++ b/lib/rails/generators/abstract_generator.rb
@@ -29,7 +29,7 @@ module ViewComponent
     end
 
     def component_path
-      ViewComponent::Base.config.view_component_path
+      GlobalConfig.view_component_path
     end
 
     def stimulus_controller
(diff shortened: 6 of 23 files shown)


joelhawksley:
> Will be a breaking change if folks still use ViewComponent::Base.config to setup their config.

Unfortunately for us, it's the [first line of the API docs](https://viewcomponent.org/api.html#config--activesupportorderedoptions) 😬  so I think we'll need to make this change in 4.0 or implement it in a non-breaking fashion. How do you think we should proceed?

boardfish:
It might be possible to rename the `ViewComponent::Base.config` endpoint to `ViewComponent::Base.default_global_config` and preserve the existing interface. I suppose since the idea is that `ViewComponent::Base`'s idea of config shouldn't change and should instead be extended internally by gems, that should be fine.

This PR needed a rebase anyway, so I'll try to address that in the process.

boardfish:
I've made it so that `ViewComponent::Base.config`, `Rails.application.config.view_component`, and `ViewComponent::GlobalConfig` are all equivalent. My thoughts are that from here, we should set the pattern for component-local config going forward – that's currently done by #1987, but if that gets held up by other factors it could be done in isolation. 

Once we have the structure for component-local config in place, we can start looking at moving config options out of global and into local config one by one. There are a couple of ways we can go about this:

1. Have the default for an option be sourced from global config at the same endpoint it currently exists at. So for example, `Rails.application.config.component.strict_helpers_enabled = true` would mean that all descendants of `ViewComponent::Base` have strict helpers enabled unless they opt out. We would seek to deprecate this in the next major version.
2. Move options to component-local config in minor versions. This would be preferable in relation to the goal of making sure that engines that consume ViewComponent aren't subject to global config options in a way that would be detrimental for them, and would also mean that for particular minor versions we'd need to advise folks to set this config on, say, `ApplicationComponent` if they have it.

joelhawksley:
@boardfish sorry for the delay in reviewing your PR, I was on vacation last week and am just now getting to it!

Having seen this set of changes and #1987, I feel like I'm missing a clear sense of _exactly_ the end state we're hoping to migrate for. What would be helpful for me is something like this:

| Option | Old API | New API |
|--------|--------|--------|
| test_controller | TBD | TBD |
| render_monkey_patch_enabled | TBD | TBD |

For all of our configuration API. Do you think you could put together that list for us to discuss? I think we need to make sure we agree on the proposed API changes and that they are even feasible to make.

joelhawksley:
@boardfish I took a first crack at listing them all out. Here they are with my notes:

| Setting name                        | Description                                                                                                                                                                                                                                                    | Proposed changes for v4                                                                                                                                                         |
| ----------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| capture_compatibility_patch_enabled | Enables the experimental capture compatibility patch that makes ViewComponent compatible with forms, capture, and other built-ins. previews. Defaults to false.                                                                                                | None - This setting mokey-patches ActionView::Base, which we can only be done globally                                                                                          |
| component_parent_class              | The parent class from which generated components will inherit. Defaults to nil. If this is falsy, generators will use "ApplicationComponent" if defined, "ViewComponent::Base" otherwise.                                                                      | Move \- This shoule be moved to be under \`generate\`                                                                                                                           |
| config                              | Returns the value of attribute config.                                                                                                                                                                                                                         | Move \- Currently set on Base, I think this should be stored in the Rails application config, right?                                                                            |
| current                             | Returns the current ViewComponent::Config. This is persisted against this class so that config options remain accessible before the rest of ViewComponent has loaded. Defaults to an instance of ViewComponent::Config with all other documented defaults set. | Move \- Currently set on Base, I think this should be stored in the Rails application config, right?                                                                            |
| default_preview_layout              | A custom default layout used for the previews index page and individual previews. Defaults to nil. If this is falsy, "component_preview" is used.                                                                                                              | Move \- Currently set on Base, I think this should be stored in the Rails application config, right? I also think we could nest preview-related config under a \`preview\` key. |
| generate.\*                         | The subset of configuration options relating to generators.<br><br>All options under this namespace default to false unless otherwise stated.                                                                                                                  | Move \- Currently set on Base, I think this should be stored in the Rails/Engine application config, right?                                                                     |
| instrumentation_enabled             | Whether ActiveSupport notifications are enabled. Defaults to false.                                                                                                                                                                                            | Move \- Currently set on Base, I think this should be stored in the Rails/Engine application config, right?                                                                     |
| preview_controller                  | The controller used for previewing components. Defaults to ViewComponentsController.                                                                                                                                                                           | Move \- Currently set on Base, I think this should be stored in the Rails application config, right? I also think we could nest preview-related config under a \`preview\` key. |
| preview_paths                       | The locations in which component previews will be looked up. Defaults to ['test/components/previews'] relative to your Rails root.                                                                                                                             | Move \- Currently set on Base, I think this should be stored in the Rails application config, right? I also think we could nest preview-related config under a \`preview\` key. |
| preview_route                       | The entry route for component previews. Defaults to "/rails/view_components".                                                                                                                                                                                  | Move \- Currently set on Base, I think this should be stored in the Rails application config, right? I also think we could nest preview-related config under a \`preview\` key. |
| render_monkey_patch_enabled         | If this is disabled, use #render_component or #render_component_to_string instead. Defaults to true.                                                                                                                                                           | Remove (already done in v4 branch)                                                                                                                                              |
| show_previews                       | Whether component previews are enabled. Defaults to true in development and test environments.                                                                                                                                                                 | Move \- Currently set on Base, I think this should be stored in the Rails application config, right? I also think we could nest preview-related config under a \`preview\` key. |
| show_previews_source                | Whether to display source code previews in component previews. Defaults to false.                                                                                                                                                                              | Move \- Currently set on Base, I think this should be stored in the Rails application config, right? I also think we could nest preview-related config under a \`preview\` key. |
| test_controller                     | The controller used for testing components. Can also be configured on a per-test basis using #with_controller_class. Defaults to ApplicationController.                                                                                                        | Move \- Currently set on Base, I think this should be stored in the Rails application config, right? I also think we could nest preview-related config under a \`preview\` key. |
| use_deprecated_instrumentation_name | Whether ActiveSupport Notifications use the private name "!render.view_component" or are made more publicly available via "render.view_component". Will default to false in next major version. Defaults to true.                                              | Remove - This should be done in v4 IMO                                                                                                                                          |
| view_component_path                 | The path in which components, their templates, and their sidecars should be stored. Defaults to "app/components".                                                                                                                                              | Modify \- Wouldn't we want this to be view_component_paths?                                                                                                                     |

If it's easier, I shared a Google Doc with you we can collaborate on 🤷🏻

boardfish:
@joelhawksley Thanks for putting that together! A cursory look through seems to suggest that's all right, but `generate` jumps out at me as something that's both global within an app and could be configurable within an engine. I figure we'd handle this in a similar way to component-local config using inheritance. I'll have a more thorough look soon.

On reflection, [the implementation of GlobalConfig](https://github.com/ViewComponent/view_component/pull/2124/files#diff-381e48288d3e375d7089034fa91418a9f08c4e11a4e603c0c4c642aa4d2689dcR2) shouldn't change behaviour right this second, but I could do with understanding how that might need to change depending on whether we're in an engine or not. It might be possible for us to roll that out as opt-in before v4.

boardfish:
Okay, I've finally found some time to bring this one back up to scratch @joelhawksley – aware of #2085 and how we may want to release in April, so I'll try to make regular time to assist during the week.

What you've got there as far as config structure actually looks pretty perfect, but I think that what's referred to there as "Rails application/engine config" is really important to define properly. Essentially what I'm thinking is:

1. Truly global settings like the render patch should remain global
2. Apps and engines alike should define an `ApplicationComponent` that inherits from `ViewComponent::Base` and configure on that, e.g.

```
ViewComponent
|+ render_monkey_patch_enabled = true

Rails app
  ApplicationComponent
  |+ preview_paths = [path_1]
    SomeComponent (inherits parent config)

Engine
  ApplicationComponent
  |+ preview_paths = [path_2]
```

That alone probably won't make for the most comfortable user experience as configuring `ApplicationComponent` may seem unintuitive, which I think #2188 highlights really well as being something that matters a lot. So we should make the interface to config predictable, even though it may work this way under the hood.

That means we may want to decide whether to:
  - keep or retire the existing Rails configuration endpoint (`config.view_component`)
  - expose some kind of configure hook (`ViewComponent.configure do |config|`)
  - precreate an initializer with all the settings, as suggested

The config that is set up via those endpoints should probably be some kind of transient config that defines the defaults for an app's `ApplicationComponent`, so that we can do this:

```rb
Rails.application.configure do |config|
  config.view_component.preview_paths = [path_1]
```

and have this without needing to do anything inside `ApplicationComponent`:

```rb
ApplicationComponent.config.preview_paths = [path_1]
```

Gems that use ViewComponent should probably configure their _own_ `ApplicationComponent` individually – I think an accessible endpoint to do that matters less as long as we document how to create a gem/engine that uses ViewComponent.

I was having a go at what that might look like in #2092 under the understanding that this leap is probably impossible without a breaking change, which makes v4 good timing, I think. I found this tough to explain in a straightforward way, but I hope it's all clear – happy to grab some time for a chat if it would help.

boardfish:
Closing this in favor of #2210.

PR #2126, opened 2024-10-08
Add first class component cache

### LINKED ISSUE #234: Components should play nicely with Rails caching mechanisms

reeganviljoen:
closes #234

Add experimental fragment caching for ViewComponent that is true to the ViewComponent ethos, namely:

- caching remains component-local (no hidden global state required to work)
- caching that can easily be debugged with standard debugging tools like breakpoints

I chose to add an opt-in caching API:

- `ViewComponent::Cacheable` + a `cache_on` macro that declares which instance methods contribute to the cache key
- cache keys include component identity + template identity (variant/format) + declared dependencies (via `ActiveSupport::Cache.expand_cache_key`) + a component digest
- the component digest invalidates cached output when the component’s source changes (Ruby file, templates, i18n sidecars) and when rendered child ViewComponents change
- template dependency extraction compiles templates to Ruby and uses Prism to find `render` dependencies
  - ERB uses stdlib `ERB::Compiler` (no Temple)
  - Slim/Haml are loaded lazily and only when those handlers are used

- This is intentionally opt-in (global opt-in via `require "view_component/fragment_caching"` or per-component `include ViewComponent::Cacheable`)
- Added integration coverage proving parent cache invalidates when a child component template changes
- `cache_on` supports private methods and uses Rails cache key expansion for dependency values
- Known limitation: partial/layout string dependencies are not currently included in the digest (workaround: bump `RAILS_CACHE_ID`/`RAILS_APP_VERSION` on deploy)
- Follow-ups can extend the API (procs, conditionals/options, broader dependency coverage)


reeganviljoen:
@joelhawksley @Spone @camertron I would appreciate any thoughts on this.

joelhawksley (commented):
@reeganviljoen THANK YOU for taking a crack at building something functional to address the long-overdue need for caching in ViewComponent. 

To be honest with you, I have not used caching in Rails nearly enough to feel like I can properly assess this PR in terms of how it would integrate with the existing mental model of caching in Rails. I'm happy to learn that context, but I won't have the space to do so for at least a couple of weeks. 

@Spone @boardfish have y'all done much with Rails' caching? @reeganviljoen if you know of anyone else in the community who would be a good reviewer here, I'd be happy to have them.

reeganviljoen:
> @reeganviljoen THANK YOU for taking a crack at building something functional to address the long-overdue need for caching in ViewComponent. 
> 
> To be honest with you, I have not used caching in Rails nearly enough to feel like I can properly assess this PR in terms of how it would integrate with the existing mental model of caching in Rails. I'm happy to learn that context, but I won't have the space to do so for at least a couple of weeks. 
> 
> @Spone @boardfish have y'all done much with Rails' caching? @reeganviljoen if you know of anyone else in the community who would be a good reviewer here, I'd be happy to have them.

@joelhawksley I really aprreciate the honesty, I will see if I can find anyone with cache experience to take a crack at reviewing this ❤️

boardfish:
@joelhawksley Afraid it's not something I've really interacted with directly until now, sorry! I'd be coming from a similar place. Great to see @reeganviljoen working at it, though ✨

lib/view_component/base.rb
```diff
@@ -319,6 +331,15 @@ def __vc_render_in_block_provided?
       defined?(@view_context) && @view_context && @__vc_render_in_block
     end
 
+    def __vc_render_template(rendered_template)
+      # Avoid allocating new string when output_preamble and output_postamble are blank
+      if output_preamble.blank? && output_postamble.blank?
```
mikodagatan, line 336: Are there any chances only one of them would be blank?

reeganviljoen: yes they can be individually toggled

mikodagatan: Not sure, if I'm contributing to this. If it's safe to have one of them blank (since we don't have a statement to cover this having only one of them blank), then isn't it also safe to do the else statement `safe_output_preamble + rendered_template + safe_output_postamble` if both of them are blank?

reeganviljoen: don't worry to much about the established code, a look at the cache mechanism is where the bulk of the eyes are needed

lib/view_component/base.rb
```diff
@@ -495,6 +516,14 @@ def sidecar_files(extensions)
         (sidecar_files - [source_location] + sidecar_directory_files + nested_component_files).uniq
       end
 
+      def cache_on(*args)
+        class_eval <<~RUBY, __FILE__, __LINE__ + 1
+          def __vc_cache_args
+            #{args}
```
mikodagatan, line 522: I think it may be better to not just include the args. You could also include the class name of the component to the key of the cache. 

This way, you will not overwrite the cache for the first component with another component that has the same args

FirstComponent.new(foo:, bar:) 
SecondComponent.new(foo: bar:)

mikodagatan: Also, would you like to add the `updated_at` field if the arguments are model records? It would ensure that the cache is invalidated when you update the associated records.

lib/view_component/base.rb
```diff
@@ -50,6 +50,17 @@ def config
 
     attr_accessor :__vc_original_view_context
 
+    # Compoents can have a cache key that is used to cache the rendered output.
+    #
+    # @return [String]
+    def cache_key
```
BlakeWilliams, line 56: Can we separate this into a separate module, similar to how we implemented slots?

lib/view_component/base.rb
```diff
@@ -50,6 +50,17 @@ def config
...
+      @vc_cache_key = if defined?(__vc_cache_args)
+        Digest::MD5.hexdigest(
+          __vc_cache_args.map { |method| send(method) }.join("-")
```
BlakeWilliams, line 59: How does this differ from Rails caching? Should we leverage their caching infrastructure to keep parity behavior wise?

lib/view_component/base.rb
```diff
@@ -117,11 +128,12 @@ def render_in(view_context, &block)
             end
           end.to_s
 
-        # Avoid allocating new string when output_preamble and output_postamble are blank
-        if output_preamble.blank? && output_postamble.blank?
-          rendered_template
+        if cache_key.present?
+          Rails.cache.fetch(@vc_cache_key) do
```
BlakeWilliams, line 132:
```suggestion
          Rails.cache.fetch(cache_key) do
```

Lets avoid relying on this ivar since it requires `cache_key` to be called first. We should access one or the other consistently.

lib/view_component/base.rb
```diff
@@ -50,6 +50,17 @@ def config
 
     attr_accessor :__vc_original_view_context
 
+    # Compoents can have a cache key that is used to cache the rendered output.
+    #
+    # @return [String]
+    def cache_key
+      @vc_cache_key = if defined?(__vc_cache_args)
```
BlakeWilliams, line 57:
```suggestion
      return @__vc_cache_key if defined?(@__vc_cache_key)
      @__vc_cache_key ||= if defined?(__vc_cache_args)
```

I think we want to memoize this?

lib/view_component/base.rb
```diff
@@ -495,6 +516,14 @@ def sidecar_files(extensions)
         (sidecar_files - [source_location] + sidecar_directory_files + nested_component_files).uniq
       end
 
+      def cache_on(*args)
+        class_eval <<~RUBY, __FILE__, __LINE__ + 1
+          def __vc_cache_args
```
BlakeWilliams, line 521: I'm not sure I understand why this is defining a method if we're accepting symbols as arguments.

could this be:

```ruby
def cache_on(*args)
  @__vc_cache_args ||= args
end
```

lib/view_component/base.rb
reeganviljoen: @BlakeWilliams solid idea so I looked at https://api.rubyonrails.org/v6.0.6.1/classes/AbstractController/Caching.html#method-i-view_cache_dependencies
and implemented similar logic, only is the class_attribute is leaking across classes, do you have an idea why and how would you fix that

lib/view_component/base.rb
```diff
@@ -516,12 +510,16 @@ def sidecar_files(extensions)
...
-        class_eval <<~RUBY, __FILE__, __LINE__ + 1
-          def __vc_cache_args
-            #{args}
-          end
-        RUBY
+        self.__vc_cache_dependencies.push(*args)
+      end
+
+      def view_cache_dependencies
+        return unless __vc_cache_dependencies.any?
```
ekampp, line 520:
```suggestion
        return if __vc_cache_dependencies.none?
```

This has a lower cognitive load because it avoids having to flip the sentence in your read when you read any.

lib/view_component/base.rb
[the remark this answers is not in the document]
reeganviljoen: this has been replaced

lib/view_component/base.rb
reeganviljoen: @BlakeWilliams this has chnaged please take a look when you can 🙏

lib/view_component/base.rb
reeganviljoen: I have fixed the leaking

lib/view_component/base.rb
reeganviljoen: I generally disagree with that in when using a single condition but when using multiple conditions I prefer an if statement

lib/view_component/base.rb
reeganviljoen: on it

reeganviljoen: done

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,56 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    class_attribute :__vc_cache_dependencies, default: []
+
+    # For caching, such as #cache_if
+    #
+    # @private
+    def view_cache_dependencies
+      return unless __vc_cache_dependencies.present? && __vc_cache_dependencies.any?
```
ekampp, line 13: Consider this for reducing cognitive load:

```suggestion
      return if __vc_cache_dependencies.blank? || __vc_cache_dependencies.none?
```

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      __vc_cache_dependencies.map { |dep| send(dep) }.compact
```
ekampp, line 15:
```suggestion
      __vc_cache_dependencies.filter_map { |dep| send(dep) }
```

lib/view_component/cacheable.rb
reeganviljoen: I actually agree with this one, I generally disagree with using unless when using multiple conditions, thanks

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+  end
+
+  class_methods do
+    # For caching the component
+    def cache_on(*args)
+      __vc_cache_dependencies.push(*args)
+    end
+
+    def inherited(child)
```
JWShuff, line 50: We took this for a spin in a few cases of our cached partials and view components that we are familiar with and largely it works for the base cases. This is working to correctly bust a cached VC when it changes or things in the render path 'above' (partials or VCs) it change.

What we aren't seeing is when a VC renders another VC or partial as a child that downstream changes of the cached VC pick up changes and handle it. I'm going to fork your branch and offer some test cases based on an approach we use in the [vc fragment caching gem](https://github.com/patrickarnett/view_component-fragment_caching) (what we're currently solving this with)

Effectively we want some way for a child partial or VC change (in either the `.rb` or template) to bust the cached parent.

JWShuff (commented):
Reegan, really excited to see this. I'm going to take a shot at adapting our gem spec suite to your fork and open a PR at your fork with those specs by way of giving us a place to start that gives us the most cache functionality we can get.

JWShuff (commented):
Added a hypothetical approach to an integration test for the behavior.

lib/view_component/cacheable.rb
reeganviljoen: @JWShuff like touch: true in rails ?

reeganviljoen: also big thanks for testing it, I can look at adding the touch true thing later this week

lib/view_component/base.rb
reeganviljoen: @mikodagatan to get this across the line I will add this in the next pr with docs and other nice to haves

reeganviljoen:
@JWShuff I investigated a few of your suggestions, added a few, thanks, and others I am struggling with
luckily I think these aren't required for the initial feature and are more nice to haves that can be implemented in another pr(child cache busting)

reeganviljoen:
@joelhawksley with a live use case tested how do you feel about this now

reeganviljoen:
@joelhawksley, with this being tested @JWShuff, how do you feel about going forward with this?

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,56 @@
...
+    # For caching, such as #cache_if
+    #
+    # @private
+    def view_cache_dependencies
+      return if __vc_cache_dependencies.blank? || __vc_cache_dependencies.none?
+
+      __vc_cache_dependencies.filter_map { |dep| send(dep) }
+    end
+
+    # For handeling the output_preamble and output_postamble
```
joelhawksley, line 18:
```suggestion
    # For handling the output_preamble and output_postamble
```

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,56 @@
...
+    #
+    # @private
+    def __vc_render_template(rendered_template)
```
joelhawksley, line 21: IMO we should leave this inside Base as we'll be including this module manually to start.

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,56 @@
...
+      # Avoid allocating new string when output_preamble and output_postamble are blank
+      if output_preamble.blank? && output_postamble.blank?
+        rendered_template
+      else
+        safe_output_preamble + rendered_template + safe_output_postamble
+      end
+    end
+
+    # For determing if a template is rendered with cache or not
```
joelhawksley, line 30:
```suggestion
    # Render component from cache if possible
```

test/sandbox/test/rendering_test.rb
```diff
@@ -1240,4 +1240,21 @@ def test_request_param
...
+    assert_selector(".cache-component__cache-message", text: "foo bar")
+
+    render_inline(CacheComponent.new(foo: "foo", bar: "bar"))
+
+    assert_selector(".cache-component__cache-key", text: component.view_cache_dependencies)
+
+    render_inline(CacheComponent.new(foo: "foo", bar: "baz"))
+
+    refute_selector(".cache-component__cache-key", text: component.view_cache_dependencies)
+    refute_selector(".cache-component__cache-message", text: "foo bar")
```
joelhawksley, line 1258: Can we add an assertion for what _should_ be shown here?

lib/view_component/cacheable.rb
joelhawksley: @reeganviljoen I'd expect us to test this with changes to child partials.

joelhawksley (commented):
@reeganviljoen thanks for putting this together and testing it with @JWShuff. I'm interested in landing it as an optional, experimental module so we can make breaking changes outside of SemVer.

Besides the comments below, might you add some documentation? I'm happy to edit whatever you come up with ❤️

lib/view_component/base.rb
```diff
@@ -29,6 +30,7 @@ def config
       end
     end
 
+    include ViewComponent::Cacheable
```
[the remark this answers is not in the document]
reeganviljoen: done

test/sandbox/test/rendering_test.rb
reeganviljoen: done, thank you

lib/view_component/cacheable.rb
reeganviljoen: done, thank you

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,56 @@
...
+    #
+    # @private
+    def __vc_render_cacheable(rendered_template)
+      if view_cache_dependencies.present?
+        Rails.cache.fetch(view_cache_dependencies) do
```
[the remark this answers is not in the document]
reeganviljoen: @joelhawksley I don't really understand your question, the cache doesn't really care about how the component renders its content(i.e call, vs sidecar vs normal template vs inline template), it just cares about if the parameters marked for cache are changed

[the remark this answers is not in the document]
reeganviljoen: @joelhawksley I have added the variant and format to the cache digest to make sure that it is more unique and that a change to either would generate a unique cache key

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,44 @@
...
+  included do
+    class_attribute :__vc_cache_dependencies, default: [:format, :__vc_format]
+
+    # For caching, such as #cache_if
+    #
+    # @private
+    def view_cache_dependencies
+      return if __vc_cache_dependencies.blank? || __vc_cache_dependencies.none? || __vc_cache_dependencies.nil?
+
+      __vc_cache_dependencies.filter_map { |dep| send(dep) }.join("-")
```
tmaier, line 15: Could this use a logic similar to what rails does? This would work then also if the cache dependencies would have a more complex object than a string.

```ruby
        def retrieve_cache_key(key)
          case
          when key.respond_to?(:cache_key_with_version) then key.cache_key_with_version
          when key.respond_to?(:cache_key)              then key.cache_key
          when key.is_a?(Array)                         then key.map { |element| retrieve_cache_key(element) }.to_param
          when key.respond_to?(:to_a)                   then retrieve_cache_key(key.to_a)
          else                                               key.to_param
          end.to_s
        end
```

Source: https://github.com/rails/rails/blob/0367cbf265ff25546230d2f668aa887e09995509/activesupport/lib/active_support/cache.rb#L125-L133

lib/view_component/cacheable.rb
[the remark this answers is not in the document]
tmaier: When I read the code correctly, then a component with the same cache dependencies and maybe the same variant and format would lead to a collusion in the rails cache.

This is why I would recommend to also add the `identifier`.

lib/view_component/cacheable.rb
reeganviljoen: Firstly thank you :heart: , I really appreciate the help on this, secondly I think that might be a good idea, so I will look into it and hopefully put something together and let you know

lib/view_component/cacheable.rb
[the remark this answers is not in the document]
reeganviljoen: @tmaier thanks for that I have added `identifier` to :__vc_cache_dependencies

lib/view_component/cacheable.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
+
+module ViewComponent::Cacheable
+  extend ActiveSupport::Concern
+
+  included do
+    class_attribute :__vc_cache_dependencies, default: [:format, :__vc_format, :identifier]
```
tmaier, line 7: Should this be a Set so that we have automatic deduplication? This would allow that one calls `cache_on :current_locale` in the `ApplicationComponent`. Subsequent calls to `cache_on` would not lead to duplicates in `__vc_cache_dependencies`.

Side-note: `cache_on` with lambda support would also be interesting. E.g. `cache_on -> lambda { I18n.current_locale }`

lib/view_component/base.rb
```diff
@@ -293,6 +304,13 @@ def __vc_request
       @__vc_request ||= controller.request if controller.respond_to?(:request)
     end
 
+    # Fo use in caching
```
tmaier, line 307: Spelling error

tmaier:
I have two questions about cache invalidation. Please understand, that I am coming from the perspective of [Ruby on Rails Managing Dependencies](https://guides.rubyonrails.org/caching_with_rails.html#managing-dependencies)
If the following will not be supported intentionally, then it should be documented in the guide.

## 1.

Rails takes the digest of a template file into account when considering to invalidate a cached object.
When I understand it correctly, this is handled by `ActionView::Digestor` in Rails.

Translated to ViewComponents, one would need to take the `.rb` file, `.html.erb` file and the `.yml` file into account. Essentially, all sidecar files. Is this something already now supported? I guess not.
If this is by purpose, we could document a workaround to invalidate the cache with each deploy by changing the environment variable `RAILS_CACHE_ID`.

## 2.

If we would have (1), the next step would be, that dependencies would invalidate the cache of a component.

Assume you would have `OneComponent` with this template:

```erb
OneComponent
<%= render TwoComponent.new %>
```

Then a change to `TwoComponent` should lead to an invalidation of the cache of `OneComponent`.
Was this what @JWShuff referred to in his comments?
I think a work-around would be the pattern described in <https://guides.rubyonrails.org/caching_with_rails.html#external-dependencies>

reeganviljoen:
> I have two questions about cache invalidation. Please understand, that I am coming from the perspective of [Ruby on Rails Managing Dependencies](https://guides.rubyonrails.org/caching_with_rails.html#managing-dependencies) If the following will not be supported intentionally, then it should be documented in the guide.
> 
> ## 1.
> Rails takes the digest of a template file into account when considering to invalidate a cached object. When I understand it correctly, this is handled by `ActionView::Digestor` in Rails.
> 
> Translated to ViewComponents, one would need to take the `.rb` file, `.html.erb` file and the `.yml` file into account. Essentially, all sidecar files. Is this something already now supported? I guess not. If this is by purpose, we could document a workaround to invalidate the cache with each deploy by changing the environment variable `RAILS_CACHE_ID`.
> 
> ## 2.
> If we would have (1), the next step would be, that dependencies would invalidate the cache of a component.
> 
> Assume you would have `OneComponent` with this template:
> 
> ```
> OneComponent
> <%= render TwoComponent.new %>
> ```
> 
> Then a change to `TwoComponent` should lead to an invalidation of the cache of `OneComponent`. Was this what @JWShuff referred to in his comments? I think a work-around would be the pattern described in https://guides.rubyonrails.org/caching_with_rails.html#external-dependencies

after looking at this it seems we may need to build a `ViewComponent::Digestor` to do the same thing but for view component but I'll first need to dig a bit deeper on how rails does this

reeganviljoen:
@tmaier I have built a cache digetor that makes use of temple and prism to build an ast(as apposed to rails use of regexes) to build a digest, would you mind taking a look :pray:

reeganviljoen:
I will bring this pr up to date shortly to fix any conflicts

joelhawksley:
@tmaier might you be able to provide a follow-up review here? Thanks for all of your help with this PR so far ❤️

prostko:
This looks awesome thanks for the effort

reeganviljoen:
@joelhawksley @tmaier I have found some time to refactor this thing completely to fix your concerns and make any edge cases raised work, would you mind taking a look

rhulse:
Thanks @reeganviljoen for moving this forward. From a random person who is waiting for this to land so I can use on a large media website. :-)

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Add experimental support for caching.
```
```suggestion
* Add experimental support for caching by including `ViewComponent::ExperimentallyCacheable`.
```

lib/view_component/base.rb
```diff
@@ -49,12 +49,13 @@ def config
     include Rails.application.routes.url_helpers if defined?(Rails.application.routes)
     include ERB::Escape
     include ActiveSupport::CoreExt::ERBUtil
-
     include ViewComponent::InlineTemplate
     include ViewComponent::Slotable
     include ViewComponent::Translatable
     include ViewComponent::WithContentHelper
 
+    class_attribute :__vc_strip_trailing_whitespace, instance_accessor: false, instance_predicate: false, default: false
```
joelhawksley: Why was this moved? The associated comment was not moved.

lib/view_component/cache_digestor.rb
```diff
@@ -0,0 +1,91 @@
...
+        next unless template_source
+
+        sources << template_source
+
+        handler = path.split(".").last
+        dependencies = ViewComponent::TemplateDependencyExtractor.new(template_source, handler).extract
+        sources.concat(dependency_digests(dependencies))
+      end
+
+      component_class.sidecar_files(["yml"]).sort.each do |path|
```
joelhawksley: Is `yaml` also valid?

lib/view_component/prism_render_dependency_extractor.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+require "prism"
+
+module ViewComponent
+  class PrismRenderDependencyExtractor
```
joelhawksley: This is very similar to @jhawthorn's https://github.com/jhawthorn/actionview_precompiler/blob/main/lib/actionview_precompiler/render_parser.rb, perhaps we could reuse it instead of writing our own?

Also relevant: https://github.com/rails/rails/pull/50944 and https://github.com/rails/rails/pull/42458.

test/sandbox/test/rendering_test.rb
```diff
@@ -1339,6 +1339,90 @@ def test_around_render
     assert_text("Hi!")
   end
 
+  def test_inline_cache_component
+    return if Rails.version < "7.0"
```
joelhawksley: We no longer support < `7.0` so let's remove this and the other early returns in this file.

joelhawksley (commented):
Overall, this is in great shape. Lots of thoughts to consider before we land this but I think we're on the right track towards having something I'd be comfortable landing ❤️

docs/guide/caching.md
```diff
@@ -0,0 +1,50 @@
...
+  end
+end
+```
+
+Notes:
+
+- `cache_on` accepts method names; the returned values are expanded via `ActiveSupport::Cache.expand_cache_key` (so Active Record models, `GlobalID`, arrays, and plain strings work as expected).
+- Methods listed in `cache_on` may be private (the cache dependency reader calls `send`).
+- The cache key includes a digest of the component source (Ruby + templates + i18n sidecars) and rendered child ViewComponents.
+- Partial/layout string dependencies are not currently included in the digest; use `RAILS_CACHE_ID`/`RAILS_APP_VERSION` if you need to invalidate on deploy.
```
joelhawksley: Also, I think we should have a test that captures this shortcoming so we can fix it eventually.

docs/guide/caching.md
```diff
@@ -0,0 +1,50 @@
...
+require "view_component/fragment_caching"
+```
+
+Alternatively, you can opt in per-component by including `ViewComponent::Cacheable`.
+
+Components implement caching by marking the dependencies that should be included in the cache key using `cache_on`:
+
+```ruby
+class CacheComponent < ViewComponent::Base
+  include ViewComponent::Cacheable
```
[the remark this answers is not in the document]
reeganviljoen: done thanks

lib/view_component/cache_digestor.rb
```diff
@@ -0,0 +1,91 @@
...
+      return cached_digest if cached_digest
+
+      return "" if @visited_components.key?(component_class.name)
+
+      @visited_components[component_class.name] = true
+
+      sources = []
+
+      identifier = component_class.identifier
+      sources << file_contents(identifier) if identifier
```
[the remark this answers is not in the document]
reeganviljoen: sorry I this was something I forgot to remove while debugging

lib/view_component/cache_digestor.rb
reeganviljoen: it is

lib/view_component/fragment_caching.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
```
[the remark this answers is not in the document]
reeganviljoen: I see and that is a valid point so I removed it

reeganviljoen:
@joelhawksley I also added a benchmark for use to have an idea of the gain/overheads of this

```
ruby 4.0.0 (2025-12-25 revision 553f1675f3) +YJIT +PRISM [arm64-darwin24]
       non_cacheable    54.354k i/100ms
      cacheable_miss   168.000 i/100ms
       cacheable_hit     4.108k i/100ms
       non_cacheable    548.026k (± 1.5%) i/s    (1.82 μs/i) -     10.980M in  20.039446s
      cacheable_miss      1.609k (± 9.6%) i/s  (621.47 μs/i) -     31.920k in  20.053582s
       cacheable_hit     40.632k (± 5.5%) i/s   (24.61 μs/i) -    809.276k in  20.018553s

Comparison:
       non_cacheable:   548025.7 i/s
       cacheable_hit:    40631.9 i/s - 13.49x  slower
      cacheable_miss:     1609.1 i/s - 340.58x  slower
```

reeganviljoen:
After a few refactors to improve the benchmark numbers it now looks a lot better 
```
ruby 4.0.0 (2025-12-25 revision 553f1675f3) +YJIT +PRISM [arm64-darwin24]
       non_cacheable    51.748k i/100ms
      cacheable_miss    41.394k i/100ms
       cacheable_hit    39.364k i/100ms
       non_cacheable    558.656k (± 2.3%) i/s    (1.79 μs/i) -     11.178M in  20.019916s
      cacheable_miss    491.207k (± 2.2%) i/s    (2.04 μs/i) -      9.852M in  20.067487s
       cacheable_hit    396.384k (± 4.7%) i/s    (2.52 μs/i) -      7.912M in  20.036403s

Comparison:
       non_cacheable:   558656.4 i/s
      cacheable_miss:   491206.7 i/s - 1.14x  slower
       cacheable_hit:   396383.8 i/s - 1.41x  slower
```

reeganviljoen:
and finnaly I realised the reason the cache benchmarks were worse is because in casses where the coomponent is not expensive to render cache will most likely be too expensive therefore I added more complexitiy to the benchmark components and these are the results 
```
ruby 4.0.0 (2025-12-25 revision 553f1675f3) +YJIT +PRISM [arm64-darwin24]
       non_cacheable   181.000 i/100ms
      cacheable_miss   180.000 i/100ms
       cacheable_hit    30.238k i/100ms
       non_cacheable      1.868k (± 2.5%) i/s  (535.35 μs/i) -     37.467k in  20.071934s
      cacheable_miss      1.825k (± 2.4%) i/s  (547.84 μs/i) -     36.540k in  20.030874s
       cacheable_hit    297.514k (± 4.7%) i/s    (3.36 μs/i) -      5.957M in  20.081678s

Comparison:
       cacheable_hit:   297513.6 i/s
       non_cacheable:     1867.9 i/s - 159.27x  slower
      cacheable_miss:     1825.4 i/s - 162.99x  slower
  ```

File: lib/view_component/cache_digestor.rb [hunk 30]
```diff
@@ -0,0 +1,91 @@
...
+    def digest_for_component(component_class)
+      return "" unless component_class.respond_to?(:name)
+      return "" unless component_class <= ViewComponent::Base
+
+      cached_digest = @digests[component_class.name]
+      return cached_digest if cached_digest
+
+      return "" if @visited_components.key?(component_class.name)
+
+      @visited_components[component_class.name] = true
```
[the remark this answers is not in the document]
reeganviljoen: this has been reworked now thank you

reeganviljoen:
@joelhawksley im not sure if thd primer failure is related but everything I feel is done, I would appreciate a review when ever you can

docs/guide/caching.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+```erb
+<p><%= view_cache_dependencies.inspect %></p>
+
+<p><%= Time.zone.now %></p>
+<p><%= "#{foo} #{bar}" %></p>
+```
+
+`cache_on` accepts method names. Returned values are expanded via `ActiveSupport::Cache.expand_cache_key`, so Active Record models, `GlobalID`, arrays, and plain strings work as expected.
```
joelhawksley:
```suggestion
Returned values are expanded via `ActiveSupport::Cache.expand_cache_key`, so Active Record models, `GlobalID`, arrays, and plain strings work as expected.
```

docs/guide/caching.md
```diff
@@ -0,0 +1,45 @@
...
+# Caching
+
+Experimental
+{: .label }
+
+Caching is experimental.
+
+To enable caching, include `ViewComponent::ExperimentallyCacheable`.
+
+Components implement caching by marking dependencies using `cache_on`:
```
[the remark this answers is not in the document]
joelhawksley: Is it true that one _must_ call `cache_on`? What if you don't?

joelhawksley (commented):
Getting closer! Thank you for your hard work ❤️

reeganviljoen:
@joelhawksley and everyone who has reviewed or tested this so far, I have come back to this with a few fixes addressing the latest comments.

The main changes are:

- moved the API to `cache do ... end` and updated the docs around that
- kept caching opt-in per component while experimental
- switched the fragment reads/writes to go through Rails fragment caching APIs
- added `sidecar_templates` / `sidecar_translations`
- simplified the digestor and added `ViewComponent::CacheDigestor.digest`
- made cache hit/miss recording respect the instrumentation setting
- added/precomputed the component digest when template caching is enabled
- added coverage for ERB, Slim, Haml, and Jbuilder template handlers
- kept the partial/layout string dependency limitation documented and covered by tests

I think this now addresses the comments raised here, but I would really appreciate another pass before we try to land it.

@JWShuff @tmaier if either of you have time to test this against real cached component use cases again, that would be especially helpful.

Thanks again everyone for all the review and patience on this.

joelhawksley:
@reeganviljoen when you have some time, I'd be curious about your thoughts on @timburgan's PR ❤️

reeganviljoen:
@joelhawksley I have left a small comment on @timburgen's pr, the preliminary review I did leads me to believe it would be a good course of action to bring his commits in here

joelhawksley:
Closing in favor of https://github.com/ViewComponent/view_component/pull/2685, which I just merged! If you can, please test the latest release in your apps and provide feedback ❤️

PR #2135, opened 2024-10-16
Enable components to use `@request` and `request` methods/ivars

### LINKED ISSUE #2111: Can not use "request" as argument
### Steps to reproduce
Having `request` as a argument to a component fails, eg
```ruby
class ExampleComponent < ViewComponent::Base
  def initialize(request:)
    @request = request
  end
end
```
Running `render(ExampleComponent.new(request: "My Title"))` 
will result in:
```
private method `format' called for an instance of String
```

A repo with the error is at https://github.com/erikaxel/view_component-request-bug-replica

### System configuration
Using `request` worked before version 3.15, my assumption is that https://github.com/ViewComponent/view_component/pull/2079 is the culprit.

joelhawksley:
This replaces internal uses of `request` with `__vc_request` so any component that uses a request method or ivar will continue to work.

Closes #2111


lib/view_component/base.rb
```diff
@@ -286,7 +286,14 @@ def format
...
     def request
-      @request ||= controller.request if controller.respond_to?(:request)
+      __vc_request
+    end
+
+    # Enables consumers to override request/@request
+    #
+    # @private
+    def __vc_request
+      @__request ||= controller.request if controller.respond_to?(:request)
```
joelhawksley, line 296:
```suggestion
      @__vc_request ||= controller.request if controller.respond_to?(:request)
```

PR #2137, opened 2024-10-17
Allow assigning a spacer component between items in a rendered collection

nickcoyne:

This PR extends `with_collection` to allow an optional `spacer_component` to be specified. This component will be rendered between items in the rendered collection.

In use, this is a parallel for Rails' [partial rendering with a spacer template](https://guides.rubyonrails.org/layouts_and_rendering.html#spacer-templates) using the `spacer_template` option. 

The implementation is similar to that used in Rails' partial rendering, where the elements of the collection are simply joined with the content of the rendered spacer component.

I'm not sure that I'm completely following the existing patterns for the updates in `collection.rb`. Please suggest edits if necessary. Thanks!


lib/view_component/collection.rb
```diff
@@ -69,5 +70,15 @@ def component_options(item, iterator)
 
       @options.merge(item_options)
     end
+
+    def rendered_spacer(view_context)
+      if @spacer_component
+        spacer = @spacer_component.new
```
BlakeWilliams, line 76: Instead of `new`ing ourselves, would it make sense to accept a pre-initialized component? Thoughts being that we can support components that require arguments as a spacer?

nickcoyne: I considered that (but did not need it for my current use-case), so went with the simplest option, which also matches Rails spacer template implementation that does not accept any arguments.

nickcoyne: @BlakeWilliams do you want any other changes here? I feel like accepting components with arguments could be a future enhancement if the need arises.

joelhawksley: @nickcoyne up until now, ViewComponent has not handled component instantiation (and neither has Rails). I think I'd prefer we keep it that way for consistency's sake.

nickcoyne: @joelhawksley I've updated this to accept a component instance.

docs/guide/collections.md
```diff
@@ -110,3 +110,17 @@ class ProductComponent < ViewComponent::Base
   end
 end
 ```
+
+## Spacer Components
```
joelhawksley, line 114:
```suggestion
## Spacer components
```

docs/guide/collections.md
```diff
@@ -110,3 +110,17 @@ class ProductComponent < ViewComponent::Base
...
+
+Since 3.20.0
+{: .label }
+
+You can also specify a component instance to be rendered between instances of the main component by using the :spacer_component option:
```
joelhawksley, line 119:
```suggestion
Set `:spacer_component` as an instantiated component to render between instances:
```

docs/guide/collections.md
```diff
@@ -110,3 +110,17 @@ class ProductComponent < ViewComponent::Base
...
+
+```erb
+<%= render(ProductComponent.with_collection(@products, spacer_component: SpacerComponent.new)) %>
+
```
joelhawksley, line 123:
```suggestion
```

docs/guide/collections.md
```diff
@@ -110,3 +110,17 @@ class ProductComponent < ViewComponent::Base
...
+```
+
+ViewComponent will render the SpacerComponent component between each pair of ProductComponent components.
```
joelhawksley, line 126:
```suggestion
Which will render the SpacerComponent component between each pair of `ProductComponent`s.
```

lib/view_component/base.rb
```diff
@@ -509,9 +509,10 @@ def sidecar_files(extensions)
       # ```
       #
       # @param collection [Enumerable] A list of items to pass the ViewComponent one at a time.
+      # @param spacer_component [ViewComponent::Base] A spacer component instance to be rendered.
```
joelhawksley, line 512:
```suggestion
      # @param spacer_component [ViewComponent::Base] Component instance to be rendered between pairs of collection items.
```

joelhawksley (approved):

PR #2140, opened 2024-10-20
Relax ActiveSupport version now that Rails 8 beta is available

boardfish:


Spone (approved):

boardfish:
@camertron Thanks for releasing this! I have access to do it, but wasn't sure if I should.

camertron:
No problem! It was causing a CI failure in primer/view_components so I went ahead and released it. We could probably have let the release "bake" a little bit longer otherwise 👍

PR #2151, opened 2024-10-31
Call maybe_escape_html directly in compiled call method

joelhawksley:

This PR refactors the logic for calling `maybe_escape_html` to be embedded in the template call method instead when necessary instead of the existing roundabout approach of trying to determine whether it should be called at runtime.


camertron (approved):
Nice, makes sense to me 👍

PR #2153, opened 2024-10-31
Expose `.identifier`

joelhawksley:

As a follow-up to https://github.com/ViewComponent/view_component/issues/2106, I'm formally exposing `.source_location` as part of the public API, as it seems New Relic could make good use of this feature. (See their fallback in https://github.com/newrelic/newrelic-ruby-agent/blob/dev/CHANGELOG.md#v9150 💔 )

I made `.identifier` public and used the term internally, as `source_location` is a reserved word in Ruby.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Expose `.identifier` method as part of public API.
+
+    *Joel Hawksley*
+
 ## 3.20.0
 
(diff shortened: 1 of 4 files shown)


camertron (approved):
Cool 😎

daniel-rikowski:
This might break some applications using view_component-contrib. 

Their README recommends creating an `identifier` method as the Ruby/Rails counterpart to the JS/Stimulus controller identifiers. See https://github.com/palkan/view_component-contrib?tab=readme-ov-file#using-with-stimulusjs

PR #2156, opened 2024-11-04
Use ActionView::TemplateDetails for handling format and variant

sfnelson:

This change introduces `ActionView::TemplateDetails` as the primary abstraction for parsing and interpreting formats and variants. In a future PR I'll use this to implement template ordering based on the same logic that Rails uses for tie-breaks.

Discussion in https://github.com/ViewComponent/view_component/issues/2128

Rails already has complex and well-used logic for parsing template file names. ViewComponent's differences do not seem intentional, and using the Rails logic should make it easier for ViewComponents to avoid inconsistencies in the future.

I've split up the logic for the three different template types into separate classes. This helps reduce the conditional logic by using polymorphism instead. I've taken small steps and avoided any structural changes, but this lays the groundwork for more significant changes to the compiler logic.

I have broken compatibility with the previously supported `variant.name` feature. Rails explicitly does not allow period characters in variant names.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -24,6 +24,10 @@ nav_order: 5
 
     *Joel Hawksley*
 
+* BREAKING: Remove support for variant names containing `.` to be consistent with Rails.
+
+    *Stephen Nelson*
+
 * Ensure HTML output safety wrapper is used for all inline templates.
 
     *Joel Hawksley*
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 *.gem
 *.rbc
 .ruby-version
+.DS_Store
 /.config
 /coverage/assets
 /coverage/index.html
--- a/docs/index.md
+++ b/docs/index.md
@@ -193,6 +193,7 @@ ViewComponent is built by over a hundred members of the community, including:
 <img src="https://avatars.githubusercontent.com/sammyhenningsson?s=64" alt="sammyhenningsson" width="32" />
 <img src="https://avatars.githubusercontent.com/sampart?s=64" alt="sampart" width="32" />
 <img src="https://avatars.githubusercontent.com/seanpdoyle?s=64" alt="seanpdoyle" width="32" />
+<img src="https://avatars.githubusercontent.com/sfnelson?s=64" alt="sfnelson" width="32" />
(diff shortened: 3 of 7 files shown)


joelhawksley (commented):
Th

sfnelson:
@joelhawksley thanks, I've rebased and removed the unnecessary change to `format` in `Compiler`. I've left the change in `Template` because it was already exposed as `format` and it's consistent with the change I've made to caching on my WIP branch.

sfnelson:
@joelhawksley I think this code is ready for review. I'd think about this as a relatively minor internal refactor to ViewComponents that largely preserves compatibility at an API level – with the exception of dropping support for variants containing `.`.

I've run the benchmarks on my local dev machine and the results for this change are well within the margin of error vs V4. Attached if you're interested.

I have a second, larger branch ready to review but it's based on using Rails' TemplateDetails structure for describing format and variants, so I'd like to get this change merged first, if that's ok. I'll open a WIP PR that depends on this one.

[v4.benchmark.txt](https://github.com/user-attachments/files/17640765/v4.benchmark.txt)
[template-details.benchmark.txt](https://github.com/user-attachments/files/17640763/template-details.benchmark.txt)

joelhawksley (approved):
I think this is great! Thank you so much for your contribution ❤️

docs/CHANGELOG.md
```diff
@@ -24,6 +24,13 @@ nav_order: 5
 
     *Joel Hawksley*
 
+* BREAKING: Remove support for variant names containing `.`
+
+  This feature isn't supported by Rails and prevents the use of Rails logic
+  for file name parsing.
+
+  *Stephen Nelson*
```
joelhawksley, line 32:
```suggestion
    *Stephen Nelson*
```

docs/CHANGELOG.md
```diff
@@ -24,6 +24,13 @@ nav_order: 5
 
     *Joel Hawksley*
 
+* BREAKING: Remove support for variant names containing `.`
+
+  This feature isn't supported by Rails and prevents the use of Rails logic
+  for file name parsing.
```
joelhawksley, line 30:
```suggestion
* BREAKING: Remove support for variant names containing `.`, which isn't supported by Rails and prevents the use of Rails logic for file name parsing.
```

sfnelson:
Note, tests are failing due to Ruby 3.3.6 release

PR #2158, opened 2024-11-06
Refactor compiler to use ActionView's partial sorting algorithm

sfnelson:

Rails uses `ActionView::TemplateDetails::Requested` to sort available templates to find the best match for a given `LookupContext`, taking in to account the acceptable content types, variants, locale, and handlers available. ViewComponent implements a subset of this logic that leaves some gaps, i.e. https://github.com/ViewComponent/view_component/issues/2154 https://github.com/ViewComponent/view_component/issues/2128

This PR aims to bring the ViewComponent template selection logic in line with how Rails chooses a partial, and ensure that it won't drift in the future by using Rails' internal structures for the calculation.

1. I'm using `refine` to provide access to Rails' internal `@details` ivar from `LookupContext` that memoizes details about the rendering context for use in partial/template resolution. Using the Rails internal logic for template resolution ensures that ViewComponents will resolve templates in a consistent way and remain consistent in the future.
2. I'm using Procs instead of methods to compile `render_template_for`. This approach allows closing over the templates available to the component, which would not be possible with a compiled method body as a string.  The benefit of this approach is that the code is easy to read and maintain, and it's equivalent or slightly faster to render components. An alternative approach would be to generate constants that hold the data required for the template selection and refer to them in the method body, but I think the code would be harder to read and maintain.

I have run the included benchmarks and the CPU performance is marginally better vs v4. See attached.

[v4.benchmark.txt](https://github.com/user-attachments/files/17640916/v4.benchmark.txt)
[2128-variant.benchmark.txt](https://github.com/user-attachments/files/17673230/2128-variant.benchmark.txt)

The number of allocated objects per render is slightly better than v4 because it's using the LookupContext cached resolve structures instead of storing its own format and variants. There is some additional memory retained per component as the `Template` objects remain in scope after compilation, but the Template objects are small and the overhead is proportional to the number of component templates, it won't grow during application lifecycle.

My team has added this code to one of our larger projects and it will go to production in the next launch. I'll provide feedback, if any, once we have it.

I intend that this change has minimal impact on the observable API of ViewComponents. The benefits would be better alignment with Rails' behaviour, in line with the goal of seamless integration.

I've discussed this approach with @joelhawksley but I'm keen for feedback from other members of the community.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,6 +28,17 @@ nav_order: 5
 
     *Stephen Nelson*
 
+* BREAKING: Use ActionView's `lookup_context` for picking templates instead of the request format.
+
+  3.15 added support for using templates that match the request format, i.e. if `/resource.csv` is requested then
+  ViewComponents would pick `_component.csv.erb` over `_component.html.erb`.
+
+  With this release, the request format is no longer considered and instead ViewComponent will use the Rails logic
+  for picking the most appropriate template type, i.e. the csv template will be used if it matches the `Accept` header
+  or because the controller uses a `respond_to` block to pick the response format.
+
+    *Stephen Nelson*
+
 * Ensure HTML output safety wrapper is used for all inline templates.
 
     *Joel Hawksley*
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -9,6 +9,7 @@ require "view_component/config"
 require "view_component/errors"
 require "view_component/inline_template"
 require "view_component/preview"
+require "view_component/request_details"
 require "view_component/slotable"
 require "view_component/slotable_default"
 require "view_component/template"
@@ -63,6 +64,8 @@ module ViewComponent
       self.__vc_original_view_context = view_context
     end
 
+    using RequestDetails
+
     # Entrypoint for rendering components.
     #
     # - `view_context`: ActionView context from calling view
@@ -90,13 +93,12 @@ module ViewComponent
       # For i18n
       @virtual_path ||= virtual_path
 
-      # For template variants (+phone, +desktop, etc.)
-      @__vc_variant ||= @lookup_context.variants.first
+      # Describes the inferred request constraints (locales, formats, variants)
+      @__vc_requested_details ||= @lookup_context.vc_requested_details
 
       # For caching, such as #cache_if
       @current_template = nil unless defined?(@current_template)
       old_current_template = @current_template
-      @current_template = self
 
       if block && defined?(@__vc_content_set_by_with_content)
         raise DuplicateContentError.new(self.class.name)
@@ -108,7 +110,7 @@ module ViewComponent
       before_render
 
       if render?
-        rendered_template = render_template_for(@__vc_variant, __vc_request&.format&.to_sym).to_s
+        rendered_template = render_template_for(@__vc_requested_details).to_s
 
         # Avoid allocating new string when output_preamble and output_postamble are blank
         if output_preamble.blank? && output_postamble.blank?
@@ -156,7 +158,7 @@ module ViewComponent
         target_render = self.class.instance_variable_get(:@__vc_ancestor_calls)[@__vc_parent_render_level]
         @__vc_parent_render_level += 1
 
-        target_render.bind_call(self, @__vc_variant)
+        target_render.bind_call(self, @__vc_requested_details)
       ensure
         @__vc_parent_render_level -= 1
       end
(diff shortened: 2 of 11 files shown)


sfnelson:
@joelhawksley great to see you've merged my previous PR. I've updated this one, and I think it's ready for review. I discovered the reason why I was seeing more allocated objects and I've fixed the issue (I wasn't using the `ActionView::LookupContext` cache correctly).

lib/view_component/base.rb
```diff
@@ -267,11 +269,11 @@ def view_cache_dependencies
       []
     end
 
-    # For caching, such as #cache_if
+    # Expected by ActionView::Template::Renderable
     #
     # @private
     def format
-      @__vc_variant if defined?(@__vc_variant)
+      nil
```
sfnelson, line 276: I've spent some time investigating why this code returned the variant and I still don't really understand why. It was added during 1.x and it's been refactored numerous times. I don't think this has any meaningful contribution to caching anymore as Rails cache already uses the `LookupContext@details` ivar as an entry point to caching. Any cache generated from a ViewComponent render would already be keyed by the specific LookupContext details including locale, variants, formats, and handlers.

lib/view_component/base.rb
```diff
@@ -328,7 +330,7 @@ def content_evaluated?
     end
 
     def maybe_escape_html(text)
-      return text if __vc_request && !__vc_request.format.html?
+      return text if @current_template && !@current_template.html?
```
sfnelson, line 333: Note that current_template is set by `render_template_for` now, so this logic takes into account the format picked by the renderer (e.g. if the chosen template is `json` then the before/after strings won't be HTML sanitised).

reeganviljoen:
@joelhawksley hwo does this look for v4

sfnelson:
We deployed this branch to production yesterday. No perceivable difference in memory utilisation or CPU since deployment. I'll continue monitoring and report any issues we encounter (@joelhawksley as discussed).

reeganviljoen:
@joelhawksley any further reservationd or ideas to move thus forward

sfnelson:
@joelhawksley could you please give this PR a review?

joelhawksley:
@sfnelson @reeganviljoen sorry for not responding sooner, I was pulled away from OSS work to focus on other things at work, so I'm just now coming back to looking at ViewComponent. 

I'm going to start by running this PR through CI on the GitHub monolith ❤️

docs/CHANGELOG.md
```diff
@@ -28,6 +28,10 @@ nav_order: 5
 
     *Stephen Nelson*
 
+* BREAKING: Use ActionView's `lookup_context` for picking templates instead of the request format.
```
joelhawksley, line 31: Might we add an example of when this change would be breaking for a consumer?

lib/view_component/base.rb
joelhawksley: If that's the case, then let's just remove this method entirely. It's optional anyways: https://github.com/rails/rails/blob/main/actionview/lib/action_view/template/renderable.rb#L26

joelhawksley (approved):
Looks great, just two small items.

sfnelson:
@joelhawksley thanks for the feedback, I've addressed both points and rebased.

docs/CHANGELOG.md
```diff
@@ -28,6 +28,17 @@ nav_order: 5
...
+
+  3.15 added support for using templates that match the request format, i.e. if `/resource.csv` is requested then
+  ViewComponents would pick `_component.csv.erb` over `_component.html.erb`.
+
+  With this release, the request format is no longer considered and instead ViewComponent will use the Rails logic
+  for picking the most appropriate template type, i.e. the csv template will be used if it matches the `Accept` header
+  or because the controller uses a `respond_to` block to pick the response format.
+
+    *Stephen Nelson*
+
```
joelhawksley, line 41:
```suggestion
* BREAKING: Use ActionView's `lookup_context` for picking templates instead of the request format.

  `3.15` added support for using templates that match the request format: when`/resource.csv` is requested then ViewComponent would pick `_component.csv.erb` over `_component.html.erb`.

  With this release, the request format is no longer considered and instead ViewComponent will use the Rails logic for picking the most appropriate template type. For example, the `csv` template will be used if it matches the `Accept` header or because the controller uses a `respond_to` block to pick the response format.

    *Stephen Nelson*

```

kushniryb:
@joelhawksley 

Just curious, can we include this in v3.x version since it's unclear when v4 is coming?
We've encountered the same issue today and it's awesome that it's already addressed but we're wondering if it could potentially break anything if we try to port it?

joelhawksley:
@kushniryb my current goal is to ship `v4` by the end of this month. For what it's worth, we're running the latest `alpha5` release in production at GitHub ❤️

PR #2160, opened 2024-11-08
Document a propshaft approach to JS/CSS

### LINKED ISSUE #1064: How to configure stimulus using the new rails 7 asset pipeline
## Feature request

Hello, how could this be done? In the javascript documentation the configuration needed to use webpack is indicated, but I am interested in using stimulus using importmap and sprockets, as it will be in rails 7 by default, using the asset pipeline.

Could this be added to the documentation?

Thanks!

### Motivation

Motivated by those post by DHH:

- [Modern web apps without JavaScript bundling or transpiling](https://world.hey.com/dhh/modern-web-apps-without-javascript-bundling-or-transpiling-a20f2755)
- [Rails 7 will have three great answers to JavaScript in 2021+](https://world.hey.com/dhh/rails-7-will-have-three-great-answers-to-javascript-in-2021-8d68191b)

jkotchoff:
Resolves #1064

Given Rails 8 has been released without transpilation or bundled JavaScript by default, this documentation update demonstrates how to use Stimulus and CSS in a ViewComponent without Webpacker.

A more terse demonstration would be possible without the view component arguments and/or less complicated Stimulus behaviour however that might be less instructive.

One of the trickier aspects of binding Stimulus controllers is often determining the correct data-controller key to use, especially when namespaces and/or multi-word component names are involved which this example makes explicit.

Approach 1 outputs a stylesheet link tag into HTML potentially several times if a ViewComponent using this technique is re-used in a page several times. Aside from that affecting the aesthetic of the rendered source, it seems like a reasonable trade-off to ensure the ViewComponent is completely self-sufficient inclusive of styling in lieu of the `Encapsulating assets` approach. A sidecar loaded _.css_ (or .scss) file seems a like a logical place to look for styling.

Approach 2 demonstrates the Stimulus behaviour but not the css, that would likely involve dartsass-rails.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add documentation for JavaScript and CSS without webpacker
+
+    *Jason Kotchoff*
+
 * Ensure HTML output safety wrapper is used for all inline templates.
 
     *Joel Hawksley*
--- a/docs/guide/javascript_and_css.md
+++ b/docs/guide/javascript_and_css.md
@@ -8,6 +8,199 @@ parent: How-to guide
 
 While ViewComponent doesn't provide any built-in tooling to do so, it’s possible to include JavaScript and CSS alongside components.
 
+## Propshaft / Stimulus
+
+To use a [transpiler-less and bundler-less approach to JavaScript](https://world.hey.com/dhh/modern-web-apps-without-javascript-bundling-or-transpiling-a20f2755) (the default for Rails 8), Stimulus and CSS can be used inside ViewComponents one of two ways:
+
+### Upgrading a pre-Rails 8 app
+
+```ruby
+# Gemfile (then run `bundle install`)
+gem "importmap-rails" # JavaScript version/digests without transpiling/bundling
+gem "propshaft" # Load static assets like JavaScript/CSS/images without transpilation/webpacker
+gem "stimulus-rails" # Hotwire JavaScript approach
+```
+
+```js
+// app/javascript/controllers/application.js
+import { Application } from "@hotwired/stimulus"
+
+const application = Application.start()
+
+application.debug = false
+window.Stimulus   = application
+
+export { application }
+```
+
+```js
+// app/javascript/controllers/index.js
+import { application } from "controllers/application"
+```
+
+```ruby
+# config/importmap.rb
+pin "@hotwired/turbo-rails", to: "turbo.min.js"
+pin "application", preload: true
+pin "@hotwired/stimulus", to: "stimulus.min.js"
+pin "@hotwired/stimulus-loading", to: "stimulus-loading.js"
+```
+
+### Approach 1 - Default _app/components_ ViewComponent directory using named Stimulus controllers, no autoloading
+
+Locate CSS and Stimulus js with a ViewComponent. This example demonstrates a _HelloWorldComponent_ in an _examples_ namespace with a sidecar file naming approach:
+
+```console
+app/components
+├── ...
+├── examples
+|   ├── hello_world_component
+|   |   ├── hello_world_component_controller.js
+|   |   ├── hello_world_component.css
+|   |   └── hello_world_component.html.erb
+|   └── hello_world_component.rb
+├── ...
+```
+
+#### 1. Prepare _app/components_ as an asset path for css and ensure hot reloads of Stimulus JavaScript
+
+```ruby
+# config/application.rb
+config.assets.paths << "app/components"
+config.importmap.cache_sweepers << config.root.join("app/components")
+```
+
+#### 2. Pin ViewComponent Stimulus import map entries
+
+```ruby
+# config/importmap.rb
+pin_all_from "app/components"
+```
+
+#### 3. Expose the Stimulus controller with a named key:
+
+```ruby
+# app/javascript/controllers/index.js
+import HelloWorldComponentController from "examples/hello_world_component/hello_world_component_controller"
+application.register("examples--hello-world-component", HelloWorldComponentController)
+```
+
+#### 4. Implement the ViewComponent with custom CSS and Stimulus behaviour:
+
+```ruby
+# app/components/examples/hello_world_component.rb
+class Examples::HelloWorldComponent < ViewComponent::Base
+  def initialize(title:)
+    @title = title
+    super
+  end
+end
+```
+
+```erb
+<!-- app/components/examples/hello_world_component/hello_world_component.html.erb -->
+<%= stylesheet_link_tag "examples/hello_world_component/hello_world_component" %>
+
+<h1><%= @title %></h1>
+<p><%= content %></p>
+
+<div data-controller="examples--hello-world-component">
+  <p class="hello-world" id="<%= @id %>" data-examples--hello-world-component-target="output">
+    This div will be updated by the controller
+  </p>
+
+  <button data-action="click->examples--hello-world-component#greet">Toggle Greeting</button>
+</div>
+```
+
+```css
+/* app/components/examples/hello_world_component/hello_world_component.css */
+.hello-world {
+  color: blue;
+}
+```
+
+```js
+// app/components/examples/hello_world_component/hello_world_component_controller.js
+import { Controller } from "@hotwired/stimulus";
+
+export default class extends Controller {
+  static targets = ["output"]
+
+  initialize() {
+    console.log("Component initialized!");
+  }
+
+  connect() {
+    console.log("Component connected!");
+
+    this.outputTarget.textContent = "This div has been initialised by stimulus and will be updated when you click the button"
+  }
+
+  greet() {
+    const currentText = this.outputTarget.textContent;
+    this.outputTarget.textContent = currentText === "Hello from Stimulus!"
+      ? "Goodbye from Stimulus!"
+      : "Hello from Stimulus!";
+  }
+}
+```
+
+#### 5. Render the component in a rails view (or a [ViewComponent preview](previews.md)) to see the end result:
+
+```erb
+<!-- app/views/layouts/application.html.erb -->
+<body>
+  ...
+    <%= render(Examples::HelloWorldComponent.new(title: "Hello World!")) {
+      "<em>This</em> will demonstrate the use of <b>Stimulus</b> and <b>CSS</b> in a ViewComponent".html_safe
+      }
+    %>
+  ...
+```
+
+### Approach 2 - Autoloaded ViewComponents in a sub-directory
+
+Stimulus controllers [won't currently autoload](https://github.com/ViewComponent/view_component/issues/1064#issuecomment-1163314487) if ViewComponents are located at:
+
+```console
+app/components
+```
+
+a workaround is to put ViewComponents in a subdirectory:
+
+```console
+app/frontend/components
+```
+
+and then autoload them in import map:
+
+```ruby
+# config/importmap.rb
+pin_all_from "app/frontend/components", under: "controllers", to: "components"
+```
+
+which also requires adjustment of the ViewComponent defaults to account for the sub-directory path:
+
+```ruby
+# config/application.rb
+config.autoload_paths << Rails.root.join("app/frontend/components")
+config.importmap.cache_sweepers << Rails.root.join("app/frontend")
+config.assets.paths << Rails.root.join("app/frontend")
+config.view_component.view_component_path = "app/frontend/components"
+```
+
+allowing the autoloaded Stimulus controllers in views eg.
+
+```erb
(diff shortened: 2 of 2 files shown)


docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add documentation for JavaScript and CSS without webpacker
```
```suggestion
* Add documentation for JavaScript and CSS without webpacker.
```

joelhawksley (commented):
👋🏻 thanks for taking this on! As we don't use this approach at GitHub, might you be able to find someone you think would be a good reviewer of this change?

docs/CHANGELOG.md
jkotchoff: Thanks for the suggestion Joel. It looks like Github [uses](https://viewcomponent.org/viewcomponents-at-github.html#integrating-javascript-behaviors) a [Catalyst](https://catalyst.rocks/guide/introduction/) approach instead of Stimulus. I have to admit, I wasn't aware of Catalyst at all and it's really interesting. 

It helps explain the shadow root web component approach described in [Encapsulating assets](https://viewcomponent.org/guide/javascript_and_css.html#encapsulating-assets). Separately, that section appears to be [unchanged](https://github.com/ViewComponent/view_component/blob/2b09e11076770830a91b1e356b639b948db3a931/docs/building_components.md#encapsulating-sidecar-assets) since the original gem documentation was released. When I came across it, my immediate tendency was to overlook it (and the Webpacker section) given the [Rails buzz](https://world.hey.com/dhh/modern-web-apps-without-javascript-bundling-or-transpiling-a20f2755) around import maps / Stimulus / eliminating node atm. Perhaps there's an opportunity to mention Catalyst or contextualise Web Components a bit more in that section to help justify it's approach?

I believe it was @reeganviljoen who [suggested](https://github.com/ViewComponent/view_component/issues/1064#issuecomment-2135904924) documentation for this would be useful. I was blocked for awhile trying to figure this out in lieu of anointed documentation. Now that Rails defaults have been updated to remove Webpacker and [include](https://github.com/rails/rails/blob/1b4c73e2a17e592b007e00f083ef4f9b0703ebcf/Gemfile#L19) stimulus-rails, I suspect I'm not the only one per the activity in Issue #1064.

ViewComponent's [Governance](https://viewcomponent.org/CONTRIBUTING.html#governance) documentation suggests there are several folks with commit access to this gem. Does GitHub have a process for allocating reviewers to pull requests? I'm not very familiar with the project prior to this so wouldn't know who to approach.

docs/guide/javascript_and_css.md
```diff
@@ -8,6 +8,199 @@ parent: How-to guide
...
+|   |   └── hello_world_component.html.erb
+|   └── hello_world_component.rb
+├── ...
+```
+
+#### 1. Prepare _app/components_ as an asset path for css and ensure hot reloads of Stimulus JavaScript
+
+```ruby
+# config/application.rb
+config.assets.paths << "app/components"
```
reeganviljoen, line 69: one small issue I has with this config in the past is it digests everything in this path even your erb files, one way to solve this is to get all the files in components and remove any mot asset files

reeganviljoen:
@joelhawksley this config checks out with one small issue, otherwise I think its 100%

docs/guide/javascript_and_css.md
jkotchoff: I'd be interested in a better approach. I'm not sure how to avoid this?

docs/guide/javascript_and_css.md
```diff
@@ -8,6 +8,199 @@ parent: How-to guide
 
 While ViewComponent doesn't provide any built-in tooling to do so, it’s possible to include JavaScript and CSS alongside components.
 
+## Propshaft / Stimulus
+
+To use a [transpiler-less and bundler-less approach to JavaScript](https://world.hey.com/dhh/modern-web-apps-without-javascript-bundling-or-transpiling-a20f2755) (the default for Rails 8), Stimulus and CSS can be used inside ViewComponents one of two ways:
+
+### Upgrading a pre-Rails 8 app
```
camertron, line 15: Let's remove this section since upgrading your app is outside ViewComponent's scope.

docs/guide/javascript_and_css.md
```diff
@@ -8,6 +8,199 @@ parent: How-to guide
...
+
+```ruby
+# config/importmap.rb
+pin "@hotwired/turbo-rails", to: "turbo.min.js"
+pin "application", preload: true
+pin "@hotwired/stimulus", to: "stimulus.min.js"
+pin "@hotwired/stimulus-loading", to: "stimulus-loading.js"
+```
+
+### Approach 1 - Default _app/components_ ViewComponent directory using named Stimulus controllers, no autoloading
```
camertron, line 49: I'd really like to only have one section. Since app/components is the conventional place to put components, let's keep this section and remove the other one. It looks like propshaft now supports putting stimulus controllers in app/components anyway, so I don't know if there's much use in documenting another approach.

docs/guide/javascript_and_css.md
camertron: Could we do something like this?

```ruby
config.assets.paths += Dir.glob("app/components/**/*.js")
```

docs/guide/javascript_and_css.md
```diff
@@ -8,6 +8,199 @@ parent: How-to guide
...
+#### 5. Render the component in a rails view (or a [ViewComponent preview](previews.md)) to see the end result:
+
+```erb
+<!-- app/views/layouts/application.html.erb -->
+<body>
+  ...
+    <%= render(Examples::HelloWorldComponent.new(title: "Hello World!")) {
+      "<em>This</em> will demonstrate the use of <b>Stimulus</b> and <b>CSS</b> in a ViewComponent".html_safe
+      }
+    %>
```
camertron, line 158: No need to use `.html_safe` 😄 

```suggestion
    <%= render(Examples::HelloWorldComponent.new(title: "Hello World!")) do %>
      <em>This</em> will demonstrate the use of <b>Stimulus</b> and <b>CSS</b> in a ViewComponent
    <% end %>
```

docs/guide/javascript_and_css.md
```diff
@@ -8,6 +8,199 @@ parent: How-to guide
 
 While ViewComponent doesn't provide any built-in tooling to do so, it’s possible to include JavaScript and CSS alongside components.
 
+## Propshaft / Stimulus
+
+To use a [transpiler-less and bundler-less approach to JavaScript](https://world.hey.com/dhh/modern-web-apps-without-javascript-bundling-or-transpiling-a20f2755) (the default for Rails 8), Stimulus and CSS can be used inside ViewComponents one of two ways:
```
camertron, line 13: Can we link to the Rails guides or something a little more durable here instead of a blog post?

camertron (changes requested):
Thanks for working on this 😄

docs/guide/javascript_and_css.md
jkotchoff: Perhaps [this link](https://guides.rubyonrails.org/working_with_javascript_in_rails.html#choosing-between-import-maps-and-a-javascript-bundler) would be better?

docs/guide/javascript_and_css.md
jkotchoff: Without `.html_safe`, it renders like this:
Screenshot 2024-12-20 at 7 42 50 am

jkotchoff: ❤️ it!

docs/guide/javascript_and_css.md
jkotchoff: You're right. I wish there had been a more straightforward guide for this elsewhere but makes sense, doesn't belong here =]

docs/guide/javascript_and_css.md
jkotchoff: I'd also be super keen on showing a single approach. However, importing each stimulus controller with a named key:
```
import HelloWorldComponentController from "examples/hello_world_component/hello_world_component_controller"
application.register("examples--hello-world-component", HelloWorldComponentController)
```

is quite painful compared to autoloading them with:
```ruby
`pin_all_from "app/frontend/components", under: "controllers", to: "components"`
```

and as far as I'm aware, the issue [described here](https://github.com/ViewComponent/view_component/issues/1064#issuecomment-1163314487) hasn't been resolved in order to allow autoloading of stimulus controllers without the 'frontend' style subdirectory. I just tried all these steps again on a new Rails 8.0.1 project but the problem described in that issue remains.

Seems like there's some patching to be done in a library somewhere. For the time being, I thought it would be useful to show both approaches and then remove one of them when this is fixed.

jkotchoff:
@camertron, regarding https://github.com/ViewComponent/view_component/pull/2160#discussion_r1892967275, I'd love to avoid autoloading all assets in that `app/components` directory but when I have tried various approaches like the one you have suggested, I've been unable to make that work.

docs/guide/javascript_and_css.md
camertron: Hmm ok, any idea why https://github.com/rails/importmap-rails/pull/192 didn't fix the problem?

docs/guide/javascript_and_css.md
camertron: Yes, that looks good 👍

docs/guide/javascript_and_css.md
camertron: In any case, it sounds like this needs to be fixed upstream for the ideal experience. My concern is that the two approaches are confusing (not your fault) - we should just pick one and recommend it without expecting devs to make a choice. Such is the Rails Way™.

docs/guide/javascript_and_css.md
camertron: > I'd love to avoid autoloading all assets in that app/components directory but when I have tried various approaches like the one you have suggested, I've been unable to make that work.

What problems did you run into?

docs/guide/javascript_and_css.md
jkotchoff: I'm not sure why that didn't fix the problem. I've tried various approaches, maybe I'm missing something but it seems everyone is feeling the same pain per [the issue](https://github.com/ViewComponent/view_component/issues/1064) this is addressing.

I'm not sure which approach would be best to demonstrate at the moment, they're both suboptimal. I see a lot of examples of folks using the `frontend` subdirectory workaround however I really don't like that. I've gone with **Approach 1** for now for my own project however I really don't like that either and would prefer to be able to use **Approach 2** without the subdirectory 😂.

docs/guide/javascript_and_css.md
jkotchoff: `config.assets.paths` for the purpose of loading javascript with importmap-rails [expects this to contain an array of directories](https://github.com/rails/importmap-rails/blob/f588506eebc7b276f8b3539301633f93d2bda55d/lib/importmap/map.rb#L129-L141), not files.

Theoretically, you could probably create a directory structure for your view components like:
```console
app/components
├── ...
├── examples
|   ├── hello_world_component
|   |   ├── hello_world_component_javascript
|   |    |        ├── hello_world_component_controller.js
|   |   ├── hello_world_component.css
|   |   └── hello_world_component.html.erb
|   └── hello_world_component.rb
├── ...
```

and then do something like:
`config.assets.paths += Dir["app/components/**/"].select { |d| d.include?('javascript') }`

but that doesn't seem particularly nice.

docs/guide/javascript_and_css.md
jkotchoff: Hey @camertron, I note the thumbs up however I'm not sure how to interpret that. Are you suggesting you're happy to stick with both examples for the time being until the issue is resolved or are you suggesting you would prefer just one approach to be documented and if so, which one?

docs/guide/javascript_and_css.md
jkotchoff: Hey @camertron, I note the thumbs up here but I'm not sure how to interpret that in order to update the pull request. Are you suggesting we look to try and demonstrate a more complicated folder structure or are you suggesting you're now comfortable with the perhaps unavoidable usage of:
`config.assets.paths << "app/components"`

joelhawksley:
@jkotchoff in discussing this PR (and other asset-related topics) with the other ViewComponent maintainers, we've decided to remove this section from the docs entirely. It's proven difficult to keep up to date and the lack of consensus/convention in the community diminishes its value ❤️

jkotchoff:
Understood. A bit disappointing because I think the combo of view component and hotwire is really powerful for rails developers but in light of the challenges with autoloading I kind of understand. Perhaps you'll be open to a reintroduction of the docs if the asset loading becomes more straightforward in future.

PR #2164, opened 2024-11-15
Add Buildkite to list of organizations that use ViewComponent.

wooly:
Hello!

We have been using ViewComponent for a while now for a growing number of our components, so thought I should add us to the list of orgs.

Cheers
Steve


wooly:
Thanks @joelhawksley!

PR #2166, opened 2024-11-15
feat: Add --skip-suffix option to component generator

moekiorg:

In particular, when using namespaces, there are cases where the suffix component is not needed, and I want to create such a component with the generator

I added a `--skip-suffix` option that allows generating component files without the "_component" suffix, giving developers more flexibility in naming their components.


docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Add `--skip-suffix` option to component generator
```
```suggestion
* Add `--skip-suffix` option to component generator.
```

joelhawksley (approved):

PR #2184, opened 2024-12-18
Support rails 8 in v2

alhajrahmoun:

Relax ActiveSupport version constraint for v2 to support applications upgrading to Rails 8.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     view_component (2.83.0)
-      activesupport (>= 5.2.0, < 8.0)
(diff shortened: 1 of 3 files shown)


joelhawksley:
@alhajrahmoun are there any blockers for your team to upgrade to v3? We are about to release v4 and I'm hesitant to ship any non-security updates to the v2.x series ❤️

alhajrahmoun:
Thanks @reeganviljoen and @joelhawksley.

Understood! I was considering whether it's feasible to support Rails 8, as upgrading to version 3 will require some time and effort. However, if version 4 is coming soon, it seems like something we need to do sooner or later to avoid falling behind.

PR #2189, opened 2025-01-06
Add ruby 3.4 Support

reeganviljoen:

Add ruby 3.4 Support

Add ruby 3.4 to matrix

We should start looking into how chilled strings will affect view_component and possibly start to investigate any warnings from them


reeganviljoen:
@joelhawksley to reduce the size of the matrix we may want to also remove unsupported rubies(i.e 3.0)

joelhawksley:
@Spone @reeganviljoen agreed! We are going to formally drop support for unsupported rubies in v4: https://github.com/ViewComponent/view_component/pull/2102

PR #2202, opened 2025-02-04
Remove respond_to? check

tiagomenegaz:

This PR removes `respond_to?` from the following files:

- `app/controllers/concerns/view_component/preview_actions.rb`
- `lib/view_component/base.rb`
- `lib/view_component/collection.rb`
- `lib/view_component/slotable.rb`
- `lib/view_component/slotable_default.rb`
- `lib/view_component/translatable.rb`
- `test/sandbox/test/rendering_test.rb`
- `test/sandbox/test/slotable_test.rb`
- `view_component.gemspec`

Most of the changes were done by refactoring `respond_to?` to `class.method_defined` which is an alternative to that method. Please feel free to discuss these changes. 

Some checks aren't necessary because we don't support discontinued Ruby versions and the code can be written without `respond_to?`.

I noticed that some `respond_to?` are being used to support `ruby2_keywords`. I'm assuming it exists because we want to support old ruby versions. However, based on the gemfile - `ruby_version = (ENV["RUBY_VERSION"] || "~> 3.4").to_s` - we don't support Ruby 2 anymore. I'll address these keywords in a different PR.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -246,7 +246,7 @@ module ViewComponent
         super
       rescue => e # rubocop:disable Style/RescueStandardError
         e.set_backtrace e.backtrace.tap(&:shift)
-        raise e, <<~MESSAGE.chomp if view_context && e.is_a?(NameError) && helpers.respond_to?(method_name)
+        raise e, <<~MESSAGE.chomp if view_context && e.is_a?(NameError) && helpers.methods.include?(method_name.to_sym)
           #{e.message}
 
           You may be trying to call a method provided as a view helper. Did you mean `helpers.#{method_name}'?
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -287,7 +287,9 @@ module ViewComponent
     #
     # @private
     def __vc_request
-      @__vc_request ||= controller.request if controller.respond_to?(:request)
+      @__vc_request ||= controller.request
+    rescue NoMethodError
+      nil
     end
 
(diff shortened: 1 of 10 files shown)


lib/view_component/base.rb
```diff
@@ -287,7 +287,8 @@ def request
     #
     # @private
     def __vc_request
-      @__vc_request ||= controller.request if controller.respond_to?(:request)
+      @__vc_request ||= controller.request
+    rescue NoMethodError
```
tiagomenegaz, line 291: Maybe this change goes against this project's code style? 🤔

reeganviljoen:
@tiagomenegaz Firstly, thank you for this work 😄 
I wonder what the performance difference is. Are you able to run the benchmarks with and without this and post the results here?

tiagomenegaz:
> @tiagomenegaz Firstly, thank you for this work 😄 I wonder what the performance difference is. Are you able to run the benchmarks with and without this and post the results here?

No problem. Any recommendations for that? @reeganviljoen

joelhawksley:
@BlakeWilliams I looked at some of these with @tiagomenegaz. We were looking to see if we could get rid of these checks wholesale as many were implicitly for older versions of Rails. I'll review this!

joelhawksley:
Closing in favor of https://github.com/ViewComponent/view_component/pull/2219, added you as co-author!

tiagomenegaz:
> Closing in favor of #2219, added you as co-author!

Thanks for letting me know @joelhawksley

PR #2206, opened 2025-02-10
Support treat turbo_stream the same as html

bockets:

Rails sets the format for `turbo_stream` responses to `turbo_stream`, which then falls back to the `html` format. ViewComponents need to do the same so that variants are correctly selected.

This is a hack to unblock us from using ViewComponents with variants in production, so it's pretty hacky at the moment.

I hope this sparks further discussion from maintainers who are more familiar with best practices for what's going on here.

### DIFF
--- a/lib/view_component/compiler.rb
+++ b/lib/view_component/compiler.rb
@@ -78,7 +78,7 @@ module ViewComponent
                 "variant&.to_sym == #{template.variant.inspect}"
               else
                 [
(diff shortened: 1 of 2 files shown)


bockets:
Closing since this is already working on the `v4` branch.

bockets:
> Closing since this is already working on the `v4` branch.

I was mistaken, this actually isn't yet fixed on the `v4` branch.

PR #2208, opened 2025-02-11
Add Rubycritic Gem

tiagomenegaz:

This PR aims to add `rubycritic` with vanilla configurations since `Add rubycritic gem` is one of the items listed here:  https://github.com/ViewComponent/view_component/pull/2085

image

Run the following command
```shell
bundle install
rubycritic
```

Basic install using the existing gemspec without any additional configuration. 

It would be great to integrate this change with the existing pipelines. However, it could be done in a different PR.


Spone:
Hi @tiagomenegaz thanks for contributing! Can you elaborate on the rationale behind this change? I'm not sure why you suggest adding this dependency.

tiagomenegaz:
> Hi @tiagomenegaz thanks for contributing! Can you elaborate on the rationale behind this change? I'm not sure why you suggest adding this dependency.

Hi @Spone . Thanks for your comment. I'm adding this dependency because this is one of the items listed here: https://github.com/ViewComponent/view_component/pull/2085 by @joelhawksley . I'll add this to this PR's description.

joelhawksley:
@tiagomenegaz thanks for opening this PR! I've looked through the output from running `rubycritic` and I think we should consider adding each of its pieces separately, as it runs simplecov in a manner that appears to be incompatible with how we're using it already.

tiagomenegaz:
> I hope to be able to revisit this soon and figure out a path forward ❤️

Sure. I'll close this PR to keep everything clean. 🤞

PR #2209, opened 2025-02-13
Support opt-in Nokogiri HTML5 parsing

Noah-Silvera:

In view components rendered for test, we should allow opt-in Nokogiri::HTML5 parsing

Nokogiri::HTML is just an alias for Nokogiri::HTML4, so this is a non-breaking change.

I was motivated to make this change because in my own app, `Nokogiri::HTML4` was parsing my rendered HTML incorrectly (inserting spaces in the middle of data attributes),  while `Nokogiri::HTML5` parsed it correctly.

Example of the incorrect parsing

Screenshot 2025-02-13 at 10 43 37 AM
Screenshot 2025-02-12 at 5 27 44 PM

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -26,6 +26,10 @@ nav_order: 5
 
     *JP Balarini*
 
+* Support opt-in Nokogiri HTML5 parsing in view components rendered for test
+
+    *Noah Silvera*
+
 ## 3.21.0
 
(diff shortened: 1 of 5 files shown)


joelhawksley:
@Noah-Silvera thank you for your PR! I opted to just make this change the default in `v4`, which we should release in the next couple of months ❤️ https://github.com/ViewComponent/view_component/pull/2218

Issue #1945, opened 2023-12-22
Rails Engine support


tkowalewski:
I want to use view components in Rails Engine, but component generator produces component file without module (engine) name.
We can fix this manually always after component generation :|, but there should be support for `module_namespacing` like in templates in Rails.

### How to reproduce

Create new Rails Engine

```bash
rails plugin new dummy --mountable
```

Add `view_component` as dependency in `dummy.gemspec`

```ruby
spec.add_dependency "view_component", "~> 3.8"
```

and fix `dummy.gemspec` specs by adding `homepage`, `summary` ...

Install gems

```bash
bundle install
```

Require view component library in `lib/dummy/engine.rb` by adding at the top

```ruby
require "view_component"
```

Generate view component

```bash
% bin/rails generate component Example
      create  app/components/dummy/example_component.rb
      invoke  test_unit
      create    test/components/dummy/example_component_test.rb
      invoke  erb
      create    app/components/dummy/example_component.html.erb
```

And generated component (`app/components/dummy/example_component.rb`) looks like

```ruby
# frozen_string_literal: true

class ExampleComponent < ViewComponent::Base
  def initialize(text:)
    @text = text
  end

end
```

### Expected behaviour

ViewComponent should generate components within Rails Engine module like all other resources generated for Rails Engine.

### Expected component

```ruby
# frozen_string_literal: true

module Dummy
  class ExampleComponent < ViewComponent::Base
    def initialize(text:)
      @text = text
    end
  end
end
```

### The fix

To generate component within Rails Engine module we should use `module_namespacing` method in templates.
For example `lib/rails/generators/component/templates/component.rb.tt` should looks like:

```html
# frozen_string_literal: true

<% module_namespacing do -%>
class <%= class_name %>Component < <%= parent_class %>
<%- if initialize_signature -%>
  def initialize(<%= initialize_signature %>)
    <%= initialize_body %>
  end
<%- end -%>
<%- if initialize_call_method_for_inline? -%>
  def call
    content_tag :h1, "Hello world!"<%= ", data: { controller: \"#{stimulus_controller}\" }" if options["stimulus"] %>
  end
<%- end -%>
end
<% end -%>

```

reeganviljoen:
@joelhawksley @camertron @Spone @boardfish any thought's on this

boardfish:
Sounds to me like the solution is well understood here, so we can go ahead with the described change. @tkowalewski, would you like to open a pull request for this?

[...]

PR #2210, opened 2025-02-15
Reorganise configuration across Rails app and component classes

boardfish:
This PR aims to rework configuration such that individual component instances can be configured differently, with the end goal of ensuring that applications and engines can define their configuration separately. We're taking a scorched earth approach here, though it will be important to ensure that the migration path from v3 is straightforward.

NOTE: This branch is a WIP – I'll squish things down before it's ready to share.

#1945 sheds some good light on the kinds of problems we're looking to solve here. Big thanks to @Slotos for their really helpful comment on that issue.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -82,7 +82,7 @@ namespace :docs do
 
     instance_methods_to_document = meths.select { |method| method.scope != :class }
     class_methods_to_document = meths.select { |method| method.scope == :class }
-    configuration_methods_to_document = registry.get("ViewComponent::Config").meths.select(&:reader?)
+    configuration_methods_to_document = registry.get("ViewComponent::ApplicationConfig").meths.select(&:reader?)
     test_helper_methods_to_document = registry
       .get("ViewComponent::TestHelpers")
       .meths
--- a/app/controllers/concerns/view_component/preview_actions.rb
+++ b/app/controllers/concerns/view_component/preview_actions.rb
@@ -54,12 +54,12 @@ module ViewComponent
 
     # :doc:
     def default_preview_layout
-      Rails.application.config.view_component.default_preview_layout
+      Rails.application.config.view_component.previews.default_layout
     end
 
     # :doc:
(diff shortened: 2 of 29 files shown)


boardfish:
Gave this a day's work while my internet was down. I've gotten to a point where _all_ config lives on `Rails.application.config.view_component`, which I think is an ideal starting point. From here, I need to:

- review `view_component_paths` – it looks like we might need two variables, one for the default path for the generator to use and another for the lookup paths for components themselves that we remove from `Base#virtual_path`.
- move `test_controller` into component-local config as possibly the only local option so far.
- make sure local config is inheritable – it's behaving unusually. (Looks like I'm just duplicating [things that `ActiveSupport::Configurable` already does](https://github.com/rails/rails/blob/cf6ff17e9a3c6c1139040b519a341f55f0be16cf/activesupport/lib/active_support/configurable.rb#L30-L41).)

That hopefully gets us to the stage where everything _works_, but from there I'll need to:

- potentially look at trying this out in a gem with someone who's been affected to understand whether this really does unblock things for gems/engines that depend on ViewComponent
- update the documentation in this PR to lock in the changes, assuming everything's verified to be in the right place
- create a set of RuboCop cops with autocorrect (or a generator?) to ease the migration
- add deprecation warnings to v3

boardfish:
Chatted about this at maintainer office hours last night, and we came to the conclusion that:

- it should be possible for engines to define configuration local to them – component-level config shouldn't be the only option, because things like `test_controller` shouldn't sit so close to app code.
- we should start by making backwards-compatible changes that we can merge into v3 to make the migration path to v4 easier, rather than starting from v4 and working back
- once we have a pattern for engine-local and component-local options, we can share the work of moving all config options to their intended new locations.

boardfish:
Closing in favour of #2230 and #2231, but I'll be doing other things mentioned in here as followups (e.g. easing the migration with a codemod/cop).

PR #2220, opened 2025-03-10
Remove unnecessary usage of ruby2_keywords.

joelhawksley:

Remove unnecessary usage of `ruby2_keywords`.

I removed the usage wholesale as we do not support Ruby v2 in ViewComponent v4.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,4 +1,4 @@
-ruby_version: 2.5
+ruby_version: 3.2
 require:
   - standard
   - "rubocop-md"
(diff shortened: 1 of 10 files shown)


reeganviljoen:
@joelhawksley if we are removing ruby2_keywords, can't we remove it from our dependencies altogether

joelhawksley:
@reeganviljoen I'm not sure I follow what you're saying? I don't see any more references to `ruby2_keywords` on this branch.

reeganviljoen:
@joelhawksley Sorry I double checked and I see what you mean, I thought it would be in the gemspec but it seems I am wrong , sorry for the confusion

PR #2226, opened 2025-03-13
Loosen dependencies in v4

joelhawksley:

Loosen our dependencies in v4 to reduce the possibility of dependency deadlocks ("hell").

I updated our `gemspec` to depend only on major versions.


test/test_helper.rb
```diff
@@ -43,7 +43,7 @@ def self.warn(message)
 
 # Rails registers its own driver named "cuprite" which will overwrite the one we
 # register here. Avoid the problem by registering the driver with a distinct name.
-Capybara.register_driver(:vc_cuprite) do |app|
+Capybara.register_driver(:cuprite) do |app|
```
rmacklin, line 46: This change makes the preceding comment stale/confusing

joelhawksley: Oh! I had wondered why it was that way, but CI passed without the prefix. I'll revert this change 👍🏻

PR #2230, opened 2025-03-15
Migrate `strip_trailing_whitespace` to use component-local config under the hood

boardfish:
- Introduce component-local config
- Migrate strip_trailing_whitespace to component-local-config
- Update changelog

Following an initial exploration of this in #2210, I agreed with @joelhawksley and @camertron that we should take an incremental approach and roll out component- and engine-local config in V3, making deprecations that would be effective come v4. Our end goal here is to make it more practical to include view_component as a dependency in engines and gems, which are currently subject to the same config as the parent app they're included into.

trim_trailing_whitespace seemed to be an ideal candidate for being part of a component-local config entity, since (being an instance variable already) it effectively works in the same way. 

The ComponentLocalConfig module essentially replicates ActiveSupport::Configurable, but puts the config on `configuration` as opposed to `config`, which appears to point to `Rails.application.config` as-is. This could be changed down the line, but since I wasn't sure of what that `config` alias was used for right now, I essentially aliased the functionality. That having been said, we could always change this later if it turns out that it's not something that will cause much trouble - these two method names being similar could potentially be an issue when component-local config is more public.

Because the component local config module essentially aliases functionality we'd otherwise get from `ActiveSupport::Configurable`, it might be safe for us not to test it directly. It's minimally tested for inheritability in this PR. I had to mark a couple of methods as `nocov` since I'm almost certain that they're exercised during init, but Simplecov didn't appear to pick them up.


test/sandbox/app/components/configurable_component.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class ConfigurableComponent < ViewComponent::Base
+  configure do |config|
```
joelhawksley, line 4: 1) `configure` a pretty general name, which might lead to conflicts with other modules/dependencies/etc.
2) I wonder if there might be a less verbose option.

Did you consider other API options? I'd be curious to hear your thoughts and will think about other ways we could do this ❤️

boardfish: I reached for `ActiveSupport::Configurable` because it's part of Rails, which we're already dependent on. The interface is something I'd expect would feel comfortable and familiar by proxy, since it's already also used in Rails by app config, content security policy config, and other things. 

The PR for module config expects the same interface, but does more in setup to verify whether setting up the `config` accessor is correct and how it should go about it. Perhaps we could do something similar here... but then again, it feels far more likely that a) a module doesn't do anything unusual or is a `Rails::Engine` already, and b) a component would use some sort of gem to change its config directly, so I think I agree with you that we could do with making this config much more tightly scoped to ViewComponent as a gem. So I'm considering:

- moving config onto a semi-private instance variable (e.g. `@__vc_config` as we do with others)
- naming the accessors differently to avoid clashes (maybe `vc_config` and `configure_component`)

I wasn't quite as keen on class-level setters for config that live directly on the component class (e.g. as we had before with `trim_trailing_whitespace`). While it would only increase the number of methods on the class and not instances of it, I think we should try to keep the method surface of `ViewComponent::Base` low if we want to make sure it's extensible. I think it'd introduce the worry that if we introduced a component-local config option, it might clash with a class-level method that already exists on someone's component. In practice I'm not sure how often that comes up, but I think we've already seen issues and incompatibilities that come from naming clashes.

joelhawksley: @boardfish thanks for the context. I like both of your considerations. How about `view_component_config` and `configure_view_component`?

docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 5
 
 ## main
 
+* Introduce component-local config and migrate trim_trailing_whitespace to use it under the hood.
```
```suggestion
* Introduce component-local config and migrate `trim_trailing_whitespace` to use it under the hood.
```

joelhawksley (commented):
This looks good. What docs updates do you think we should make?

boardfish:
I was thinking we should have docs for the new config ready against v4 instead, but now that I think about it, it'd make sense to at least reflect the deprecation in the current docs. I'll do that next and accept that tweak to the changelog message.

docs/api.md
```diff
@@ -14,6 +14,10 @@ nav_order: 3
 
 Returns the current config.
 
+### `.identifier` → [String]
```
joelhawksley, line 17: I believe this whole file is auto-generated from code comments (and really shouldn't even be in source control on this branch). Mind making the changes in the comments instead?

docs/guide/templates.md
```diff
@@ -181,14 +181,18 @@ end
...
 Code editors commonly add a trailing newline character to source files in keeping with the Unix standard. Including trailing whitespace in component templates can result in unwanted whitespace in the HTML, eg. if the component is rendered before the period at the end of a sentence.
 
-To strip trailing whitespace from component templates, use the `strip_trailing_whitespace` class method.
+To strip trailing whitespace from component templates, use the `strip_trailing_whitespace` component-local config option.
 
 ```ruby
   # do strip whitespace
-  strip_trailing_whitespace
+  configure_component do |config|
```
joelhawksley, line 189: Let's do:

```suggestion
  configure_view_component do |config|
```

As it matches our generator config change coming in `v4` (`generate component` -> `generate view_component`)

joelhawksley (commented):
Looks good to me! Some minor notes.

docs/api.md
boardfish: Sure, I'll drop the changes to this file from this branch. Looks as though we do still need to make the changes to the docs generator for HTML safety, so I'll extract those out to another PR.

I think this might've been a case of old habits – while I'm in here I'll check that our documentation for testing changes to viewcomponent.org are up to date and in a good place.

joelhawksley (approved):

PR #2231, opened 2025-03-16
Delegate config to parent modules (no-op)

boardfish:
Adds a mechanism for inheriting config from a parent module. Since `Base` is generally used when referring to "global" config options, this doesn't do anything right now beyond exposing the ability to configure parent modules.

Modules can be configured by requiring and including `ViewComponent::Configurable`, which will expose a `config` accessor with a `view_component` configuration on the module. These will not interfere with existing configuration that might be granted to, e.g., a `Rails::Engine` by default.

This would give us a starting point to begin migrating options into engine-local config.

---

The separation between:
- module and/or Rails app-local config (this PR)
- exclusively Rails app-local config (`lib/view_component/engine.rb`)
- component-local config (#2230)

isn't perfect right now, but this PR and #2230 put us on the road to this.

It should only be possible to set options in each area as appropriate. Particularly with this PR, we should be concerned with making sure that module-local and exclusively Rails app-local config don't merge.

Because `ViewComponent::Base.config` is named as it is right now, the difference between component-local config (#2230) and module-local config isn't super clear, but we can rename parts of this interface in v4.

### DIFF
--- a/test/sandbox/test/base_test.rb
+++ b/test/sandbox/test/base_test.rb
@@ -146,4 +146,30 @@ class ViewComponent::Base::UnitTest < Minitest::Test
     MESSAGE
     assert !exception_message_regex.match?(exception.message)
   end
+
+  module TestModuleWithoutConfig
+    class SomeComponent < ViewComponent::Base
+
+    end
+  end
+
+  # Config defined on top-level module as opposed to engine.
+  module TestModuleWithConfig
+    include ActiveSupport::Configurable
+
+    configure do |config|
+      # To get this green.
+      config.view_component = ActiveSupport::InheritableOptions.new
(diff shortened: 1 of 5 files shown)


test/sandbox/test/base_test.rb
```diff
@@ -146,4 +147,54 @@ def test_no_method_error_does_not_reference_missing_helper
...
+
+    configure do |config|
+      config.view_component = ActiveSupport::InheritableOptions[test_controller: "AnotherController"]
+    end
+
+    include ViewComponent::Configurable
+
+    class SomeComponent < ViewComponent::Base
+    end
+  end
```
boardfish, line 191: Happy to move this setup out into other files if you'd rather.

joelhawksley (approved):

PR #2233, opened 2025-03-19
[v4] Remove unnecessary #format definitions, which returned nil

joelhawksley:

This PR looks to close https://github.com/ViewComponent/view_component/issues/1973 by removing our definition of `#format` methods, as:

1) @seanpdoyle [noted](https://github.com/ViewComponent/view_component/issues/1973#issuecomment-1889928719) they are optional
2) We were returning `nil` anyways. 

We can always revisit adding them back if needed, which I don't think would be a breaking change.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -52,6 +52,10 @@ nav_order: 5
 
     *Stephen Nelson*
 
+* Remove unnecessary `#format` methods that returned `nil`.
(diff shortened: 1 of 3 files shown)


tiagomenegaz (approved):
Approved based on the context provided and the non-breaking changes. Feel free to wait for @BlakeWilliams review if needed.

PR #2238, opened 2025-03-25
Remove unnecessary call to `view_renderer`

joelhawksley:

As noted in the code comment, this is no longer needed.


reeganviljoen (approved):
nice ❤️

PR #2255, opened 2025-04-01
ViewComponentsSystemTestController only exists in test env

joelhawksley:
Out of an abundance of caution, this PR changes `ViewComponentsSystemTestController` to only be loaded in the test environment.

### DIFF
--- a/app/controllers/view_components_system_test_controller.rb
+++ b/app/controllers/view_components_system_test_controller.rb
@@ -1,30 +1,27 @@
 # frozen_string_literal: true
 
-class ViewComponentsSystemTestController < ActionController::Base # :nodoc:
-  before_action :validate_test_env
-  before_action :validate_file_path
+if Rails.env.test?
+  class ViewComponentsSystemTestController < ActionController::Base # :nodoc:
+    before_action :validate_file_path
 
-  def self.temp_dir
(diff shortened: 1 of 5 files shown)


app/controllers/view_components_system_test_controller.rb
```diff
@@ -1,30 +1,27 @@
 # frozen_string_literal: true
 
-class ViewComponentsSystemTestController < ActionController::Base # :nodoc:
-  before_action :validate_test_env
-  before_action :validate_file_path
+if Rails.env.test?
```
BlakeWilliams, line 3: I can't think of a reason we'd need this outside of the test env, but I wonder if we should gate it here, or gate the `require` statement?

Then folks could opt-in to requiring it if they did need it outside the test env for whatever reason. but like I said, I don't know that there's a valid use-case for that.

BlakeWilliams (approved):

app/controllers/view_components_system_test_controller.rb
joelhawksley: Given the existing logic for raising if outside of the test env, I think we're safe here. I also don't see us loading this with `require`?

sfnelson: This change seems reasonable but it breaks in production – where Rails uses `eager_load`. See #2269.

PR #2265, opened 2025-04-06
Fix generation of docs that was broken due to HTML safety issues

boardfish:
Running `rake docs:build` after changes in #2230 resulted in some issues with HTML-escaped tags being generated, overriding existing working documentation. These changes mark output as HTML-safe so that docs are generated properly.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -142,4 +142,4 @@ namespace :docs do
   end
 end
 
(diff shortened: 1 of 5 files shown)


joelhawksley (approved):

joelhawksley:
Thanks @boardfish! I took the liberty of pushing a commit adding test coverage for the docs task ❤️

PR #2282, opened 2025-04-24
Pre-allocate instance variables for better compatibility with Object Shapes

joelhawksley:

@HParker and @jhawthorn both raised concerns that ViewComponent was a source of issues with Ruby's object shapes due to our general pattern of initializing internal instance variables when a component is rendered, which is after a component sets its own instance variables in its initializer. 

This PR aims to define as many internal instance variables as possible before a component is initialized.

I re-defined `.new`, setting the necessary instance variables before calling the component's initializer.


camertron:
I'm late to the party here, but do we know if this has a measurable performance impact? #JustCurious

joelhawksley:
@camertron I'm looking at a couple ways of measuring the performance impact in our real-world use case. Do you have any specific measurements you'd want to see?

camertron:
@joelhawksley not in particular, I'm mostly just curious to see the end result 😎

test/sandbox/test/rendering_test.rb
```diff
@@ -3,6 +3,20 @@
 require "test_helper"
 
 class RenderingTest < ViewComponent::TestCase
+  def self.new(...)
+    instance = allocate
```
jhawthorn, line 7: FYI allocate is slow in 3.4, it's just been fixed in trunk

PR #2291, opened 2025-05-02
Graduate SlotableDefault to be included by default

joelhawksley:

This PR graduates the SlotableDefault functionality to be included by default.

I removed the module and rewrote `get_slot` to incorporate the module's logic.


topherfangio:
@joelhawksley This looks cool! Will this fix https://github.com/ViewComponent/view_component/issues/2169 ? I'm running into this right now and was about to put together a workaround, but would love the functionality to "just work" 😄

joelhawksley:
Good call @topherfangio! I'll move this to draft and see about addressing that bug at the same time.

joelhawksley:
@topherfangio @billwatts I fixed https://github.com/ViewComponent/view_component/pull/2291 in this PR, will ship a release here in a minute if you wouldn't mind giving it a test?

lib/view_component/slotable.rb
boardfish, line 1: With the nesting in the conditional now, this feels a little harder to wrap my head around without as much context to this module. Could you add a few comments to explain what's going on at each step?

joelhawksley: @boardfish good call! I ended up streamlining the implementation while adding some comments. What do you think?

BillWatts:
@joelhawksley I loaded `4.0.0.alpha5` into my project and removed my local patch. Defaults look to be behaving as expected with this change.

PR #2293, opened 2025-05-09
[v4] remove unnecessary initializer in Base

joelhawksley:
BREAKING: Remove default initializer from `ViewComponent::Base`. Previously, `ViewComponent::Base` defined a catch-all initializer that allowed components without an initializer defined to be passed arbitrary arguments.


boardfish (commented):
How will this interact when `super` is called in a component's initializer? For a while that was a habit of mine because RuboCop would complain if you didn't call it. 

Assuming it breaks that, this might be worth making more noise about in the changelog.

EDIT: It's flagged as breaking, so that's probably sufficient.

boardfish (approved):

joelhawksley:
@boardfish `super` should work as before, as we're no longer inheriting from `ActionView::Base`. I added a test case just to be sure, though ❤️

PR #2295, opened 2025-05-13
[v4] BREAKING: Move config.view_component_path to config.generate.path

joelhawksley:
In auditing `config.view_component_path` for the coming configuration changes we're making, I realized that our usage of the option was riddled with bugs.

The issues all had to do with the underlying assumption that there can only be _one_ `view_component_path`, when in fact components can exist in any number of locations.

For example, we used it to strip part of the component file path for generating translation keys. This of course only worked if the component was in the directory set in `view_component_path`.

We also used it to populate the Rails statistics for ViewComponents. I opted to remove the stats functionality for now as we'd need a reliable way of counting components outside of a provided directory path.

The remaining use case (for determining where to put generated components) remains valid, so I opted to rename the option to `config.generate.path`.


joelhawksley:
Note: the Rails main failure is unrelated. I'm running into segfaults with the latest ruby HEAD even on `v4`.

boardfish (approved):
All looks good to me ✨ Nice to see some complexity go away, especially where this was used for the `identifier`.

I suppose for components that lived outside of the previously configured path, their identifiers will now change - will folks notice any impact in relation to that?

joelhawksley:
@boardfish that's a great callout! The only case that would change would be for defining component-specific translations in a global locale file, something we didn't document before. I added docs and better test coverage for that feature, but I'm guessing it's seen little to no use due to the lack of docs.

cc @palkan as you've done some work in `view_component_contrib` to support global translations.

palkan:
Thanks for pinging @joelhawksley!

Contrib's tests are green; we use custom identifier (actually, I had no idea about this one).

PR #2299, opened 2025-05-19
attempt to remove last nocovs in the project

joelhawksley:


joelhawksley:
@boardfish might you be up for trying to get test coverage working for these lines? I'm wary of leaving them uncovered since that will make them harder to refactor/remove later. No rush ❤️

PR #2308, opened 2025-05-20
Move component_parent_class to generator configuration namespace

joelhawksley:

`config.component_parent_class` is only used for generators. It should be in the `generate` namespace!

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,7 +10,7 @@ nav_order: 6
 
 ## main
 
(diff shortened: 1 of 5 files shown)


docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* BREAKING: `config.component_parent_class` is now `config.generate.component_parent_class`, moving the generator-specific option to generator configruation namespace.
```
```suggestion
* BREAKING: `config.component_parent_class` is now `config.generate.component_parent_class`, moving the generator-specific option to the generator configuration namespace.
```

PR #2310, opened 2025-05-21
v4 regression when using Turbo::StreamsHelper

sfnelson:

We have code that includes `Turbo::StreamsHelper` in our components and calls `<%= turbo_stream.<verb> %>` in the frontend. This may not be a supported use-case in v4 but I think it's worth a call-out as a breaking change.

Alternatively, consider delegating `format` to helpers.

This PR demonstrates the breaking change by changing the existing turbo stream component.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,10 @@ nav_order: 6
 
     *Joel Hawksley*
 
+* Add support for including Turbo::StreamsHelper
+
+    *Stephen Nelson*
+
 ## 4.0.0.alpha5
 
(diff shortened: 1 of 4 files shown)


joelhawksley:
Can you update this PR with a changelog and to add:

```ruby
    # For Turbo::StreamsHelper
    delegate :formats, :formats=, to: :helpers
```

To `base.rb`? Thanks!

docs/CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ nav_order: 6
 
     *Joel Hawksley*
 
+* Add support for including Turbo::StreamsHelper
```
```suggestion
* Add support for including Turbo::StreamsHelper.
```

PR #2317, opened 2025-05-21
[v4] BREAKING: Move previews-related configuration to separate namespace

joelhawksley:
This PR isolates the previews-related configuration into its own namespace for better organization and comprehension by consumers.


boardfish (approved):
Looks good! Should we backport this to v3 at some stage?

joelhawksley:
> Looks good! Should we backport this to v3 at some stage?

I'm less concerned about backward cross-compatibility for configuration changes, as they should only be a couple lines to change when bumping to the v4 release. What do you think?

PR #2324, opened 2025-05-23
Include .tt files in gem.

boof:

Make generators work again.

Since the generators work differently now (using .tt files now) they weren't packed in the gem anymore. I added a glob in the .gemspec to include files matching `lib/rails/**/*.tt`.
I could've just loosened the glob that includes all ruby files in `lib/rails` but I voted against the "include all" option since I assume that there must be a reason for being restrictive here.

This ain't covered by tests and it seems either nobody is using generators on the latest version (3.23.1) except me. Otherwise I would've spotted a Github issue I guess.

### DIFF
--- a/view_component.gemspec
+++ b/view_component.gemspec
@@ -30,6 +30,7 @@ Gem::Specification.new do |spec|
     "app/**/*",
     "docs/CHANGELOG.md",
     "lib/rails/**/*.rb",
+    "lib/rails/**/*.tt",
     "lib/view_component.rb",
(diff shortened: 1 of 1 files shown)


coiti:
According to #2274, the change was meant to prevent docs-build-related files from being added, but ended up affecting other necessary files, some of which were re-added in #2319.

Similarly in this case, I think all files under `lib/rails/**/*` are probably meant to be added to the release too (e.g. `USAGE` files). Maybe a maintainer can clear that up.

PR #2328, opened 2025-05-29
Reduce allocations in Slotable#register_default_slots

jdelStrother:
While profiling something unrelated I noticed a lot of unnecessary string allocations in Slotable#register_default_slots. How about this change to move it out of the `instance_methods` loop?

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Reduce string allocations during compilation.
+
+    *Jonathan del Strother*
+
 ## 3.23.2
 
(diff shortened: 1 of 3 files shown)


jdelStrother:
I've struggled to reproduce those CI failures locally, are they something I've introduced?

I wasn't sure how exhaustive the ruby-versions mentioned in render_inline_allocations were supposed to be, let me know if you rather I drop [that commit](https://github.com/ViewComponent/view_component/pull/2328/commits/632fe87d9ad54666899c99590f2c71c94a293a9d)

joelhawksley:
@jdelStrother those PVC failures are from the PVC test suite being tied to the `v4` branch ❤️

PR #2330, opened 2025-06-04
Remove deprecated #use_helper(s) in favor of include or helpers.

joelhawksley:
In reviewing the framework as a whole for the v4 release, we decided to remove support for the #use_helper(s) macro, as it does not align with broader Ruby patterns.


reeganviljoen (approved):
I will get working on moving this to a separate gem similar to how rails moved controller testing to a separate gem for users impacted by this

PR #2333, opened 2025-06-06
Reduce allocations in #with_collection

mvz:

Speed up rendering when there are lots of calls to `#with_collection` for the same componen, for example when rendering a collection of items, where each item contains a collection of sub-items.

I focused on reducing redundant allocation of string objects. This is because rack-mini-profiler's flamegraph showed me lots of repeats of the same string interpolation.

The code setting the expected number of allocations will have to be updated to include values for all the relevant Ruby and Rails versions.

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -665,7 +665,7 @@ module ViewComponent
 
       # @private
       def collection_counter_parameter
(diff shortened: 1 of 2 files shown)


joelhawksley:
Thank you for your contribution!

PR #2336, opened 2025-06-10
20250610 merge main

joelhawksley:


rmacklin:
The reason 41992931e81ea490855dee8af0925a118d9129d4 and bac7a24f12706dd7c88355786ad4f866a90ebdce are appearing here even though they were already in https://github.com/ViewComponent/view_component/pull/2334 is that that PR was squash-merged into `v4`, so those commits are technically (still) not part of the `v4` history. I think you'll have to avoid squash-merging in order to not have those commits keep showing up in subsequent PRs that merge `main` into `v4`.

PR #2348, opened 2025-06-12
Fix bug where response format was not set causing issues with Turbo Frames

joelhawksley:
Closes https://github.com/ViewComponent/view_component/issues/2212

Thanks to @gap777 and @d4be4st, I was able to isolate a gap in our compatibility with Turbo due to us not setting `#format`, which Rails then uses to set the `content_type` of the response. 

As noted in a code comment, we are not able to determine the format for multi-format components due to the order of operations Rails expects for renderables.

### DIFF
--- a/test/sandbox/app/components/turbo_content_type_component.rb
+++ b/test/sandbox/app/components/turbo_content_type_component.rb
@@ -1,0 +1,31 @@
+class TurboContentTypeComponent < ViewComponent::Base
+  def initialize(message: "Enter a message:", show_form: true)
+    @message = message
+    @show_form = show_form
+  end
+
+  def call
+    content_tag('turbo-frame', id: "test-frame") do
+      if show_form
+        form_content
+      else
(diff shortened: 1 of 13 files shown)


test/sandbox/app/components/turbo_content_type_component.rb
```diff
@@ -0,0 +1,21 @@
...
+  def form_content
+    form_tag("/submit", method: :post, data: {turbo_frame: "test-frame"}) do
+      content_tag(:div) do
+        content_tag(:h2, @message) +
+          text_field_tag("user_message", "", value: "Test message") +
+          content_tag(:br) +
+          submit_tag("Submit", id: "submit")
+      end
+    end
+  end
```
camertron, line 20: Can we put this into the template instead or pull it out as a separate component?

camertron (approved):
Cool, just one non-blocking comment.

camertron (approved):

PR #2353, opened 2025-06-17
add configuration page to docs

joelhawksley:
Closes https://github.com/ViewComponent/view_component/issues/2188

This PR adds a `Configuration` page to the docs and renames `component_parent_class` to `parent_class`, which _does_ default to `ApplicationComponent` if it exists.

### DIFF
--- a/app/controllers/concerns/view_component/preview_actions.rb
+++ b/app/controllers/concerns/view_component/preview_actions.rb
@@ -8,7 +8,7 @@ module ViewComponent
       prepend_view_path File.expand_path("../../../views", __dir__)
 
       around_action :set_locale, only: :previews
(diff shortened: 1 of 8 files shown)


docs/guide/configuration.md
```diff
@@ -0,0 +1,19 @@
...
+
+```ruby
+MyApplication.configure do
+  config.view_component.instrumentation_enabled = true
+  config.view_component.generate.path = "app/custom_components"
+  config.view_component.previews.controller = "MyPreviewController"
+end
+```
+
+For a list of available options, see [/api](/api.html#configuration).
```
joelhawksley, line 19:
```suggestion
For a list of available options, see [/api](/api#configuration).
```

PR #2374, opened 2025-06-26
BREAKING: Allow slots to be redefined, removing RedefinedSlotError

joelhawksley:
The following is now possible:

```ruby
class ButtonComponent < ViewComponent::Base
  renders_many :leading_items, types: {
    icon: IconComponent,
    svg: SvgComponent
  }
end

class SpecialButtonComponent < ButtonComponent
  renders_many :leading_items, types: {
    svg: SvgComponent,
    checkbox: CheckboxComponent
  }
end
```

Closes https://github.com/ViewComponent/view_component/issues/2311.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -22,11 +22,12 @@ nav_order: 6
 
     class SpecialButtonComponent < ButtonComponent
       renders_many :leading_items, types: {
-        svg: SvgComponent,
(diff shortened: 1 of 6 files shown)


joelhawksley:
In chatting this through with @BlakeWilliams, we decided to pause work on this change for now, as we don't feel the increase in complexity to the Slots mental model is worth it.

PR #2389, opened 2025-07-15
Use original virtual path when capturing blocks

### LINKED ISSUE #2386: [v4] Cannot use rails `translate` within a component slot
Follow-up on #2359, I was trying to upgrade VC to 4.0.0-rc2 in our main project and I think I encountered a `virtual_path` issue as well.
We use slots to feed small components like the `MediaObjectComponent` in this example, and we rely on Rails' lazy translation shortcut to fetch translations related to the partial where it's called.

I'm not sure if this is the expected behavior and we've been using VC incorrectly from the start, or if it's an edge case bug as we're the only beta testers using translations to feed component slots. So I'm seeking insights here today.
### Steps to reproduce
static_pages/home (Slim language)
```slim
= render MediaObjectComponent.new do |c|
  = c.with_media { 🔎 }
  = c.with_body(class: "lead") do
    = t(".visit_product_page")
```
### Expected behavior
In VC 3.x, the virtual_path was not altered, so the translation key was related to the calling partial, here `static_pages.home.visit_product_page`

### Actual behavior
Since 4.0, the translation key, is transformed to `media_object_component.body_component.visit_product_page`
**Backtrace**:

```
DRb::DRbRemoteError: Translation missing: fr.media_object_component.body_component.visit_product_page (ActionView::Template::Error)
            app/views/static_pages/home.html.slim:33
            app/components/media_object_component.rb:72:in 'MediaObjectComponent::BodyComponent#call'
```

The line 72 in the component is where the `body` slot is called.

### System configuration
**Rails version**: 8.0.2
**Ruby version**: 3.4.4
**Gem version**: 4.0.0rc2

joelhawksley:
Fixes #2386

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Fix bug where relative paths in `translate` didn't work in blocks passed to ViewComponents.
+
+    *Joel Hawksley*
+
 * Add SerpApi to "Who uses ViewComponent" list
 
     *Andy from SerpApi*
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -108,7 +108,7 @@ module ViewComponent
       self.class.__vc_compile(raise_errors: true)
 
       @view_context = view_context
-      old_virtual_path = view_context.instance_variable_get(:@virtual_path)
+      @old_virtual_path = view_context.instance_variable_get(:@virtual_path)
       self.__vc_original_view_context ||= view_context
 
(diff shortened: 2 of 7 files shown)


camertron (approved):
I was initially concerned that we might be preventing using component-scoped translations if we're always using the virtual path from the parent, but then I realized this change only affects component/slot _content_, which makes sense and should be fine. Thanks @joelhawksley 😄

camertron:
It looks like PVC tests are failing because somehow this change results in one fewer allocation 😅 Shouldn't be a merge blocker.

PR #2395, opened 2025-07-21
Fix: missing generators in gemfiles

Jell:
there's no `lib/rails` folder but there is a `lib/generators` folder, I believe this is a configuration bug that was introduced during a v4-related refactoring?

Attempting to address: https://github.com/ViewComponent/view_component/issues/2394

I looked into the expected convention for rails generators, noticed that the `lib/generators` folder was missing in the install folder of the gem on my local machine.

I'm hoping this is a simple typo / oversight, which I believe a maintainer should be able to quickly figure out from this PR! 😅 

I am not entirely sure how to write a spec for this, but perhaps that's not necessary for a gemspec typo?

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,9 +10,7 @@ nav_order: 6
 
 ## main
 
-## pending
-
-* Fix missing generators
+* Fix issue where generators were not included in published gem.
 
     *Jean-Louis Giordano*
 
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,12 @@ nav_order: 6
 
 ## main
 
+## pending
+
+* Fix missing generators
+
+    *Jean-Louis Giordano*
+
 ## 4.0.0.rc3
 
 * Reformat the avatars section to arrange them in a grid.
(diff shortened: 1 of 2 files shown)


Jell:
I'm not sure why the tests are failing after this change, I'm failing to understand if they are related or just flakey? I'm afraid I will have to let someone more knowledgeable about the codebase figure that out 😅

Jell:
Digging further into it, since ruby 3.5 is installed from "head" I'm guessing a new optimization landed on the latest version of ruby 3.5 which leads to fewer allocations, so I don't believe this is related to my change. Probably the expectation on the number of allocations should be updated to reflect that new optimization? 🤔

docs/CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ nav_order: 6
 
 ## main
 
+## pending
+
+* Fix missing generators
```
```suggestion
* Fix issue where generators were not included in published gem.
```

docs/CHANGELOG.md
```diff
@@ -10,6 +10,12 @@ nav_order: 6
 
 ## main
 
+## pending
+
```
joelhawksley, line 14:
```suggestion
```

PR #2407, opened 2025-07-30
Correct documentation regarding previews.paths option

dmrwebdev:
Hi all, first time contributing so hope I'm doing this right. Noticed in the documentation under `docs/guides/preview.md` that the old `previews_path` option was still being used in a few examples while migrating to 4.0. After changing my application to use the new `previews.paths` syntax previews worked again so assuming this is an oversight. If not sorry and feel free to close!


joelhawksley:
❤️ thanks!

PR #2409, opened 2025-08-01
[v4] Troubleshooting issue with Slim

edolix:

Trying to upgrade ViewComponent to [v4](https://github.com/ViewComponent/view_component/releases/tag/v4.0.0) in one of our apps, @xkraty and i noticed a lot of failures in CI like: `Couldn't find a template file or inline render method for SomeSlimComponent...`. I was like 🤔 
We use Slim everywhere and everything worked as expected with v3.

I _tried_ to dig a bit and i think it could be a race condition between Slim initialization and the cache logic for each component.
I may be wrong so apologies in case all of this doesn't make sense! 

In order:

In `ViewComponent::Base inherited` each component will be automatically compiled (and then cached) as soon as the file is loaded: https://github.com/ViewComponent/view_component/blob/6a3336297367b669c79af106ec0d7a99f158c194/lib/view_component/base.rb#L544-L547

The problem comes with the Slim gem that will register the `slim` template handler only at the [`action_view` hook](https://github.com/slim-template/slim/blob/main/lib/slim/railtie.rb#L6) causing a (possible) race condition within the `ViewComponent::Compiler.gather_templates` method:

https://github.com/ViewComponent/view_component/blob/6a3336297367b669c79af106ec0d7a99f158c194/lib/view_component/compiler.rb#L177-L179

`ActionView::Template.template_handler_extensions` might not include "slim" at that point.

All templates are gonna be memoized (with the related `template_errors` as well).

If few components got loaded (and automatically cached due to `ViewComponent::Base.inherited`) **before** the `action_view` hook (that will register slim as a valid template handler)... those components will always raise the `Couldn't find a template file or inline render method for ...`.

We created a Rails app trying to reproduce the issue with no luck. I'll try again tomorrow on it.

This PR is a basic/raw/dumb attempt to - at least - highlight the issue and give more context. 

Nothing - just thanks for the great VC gem! 🙌

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -141,7 +141,7 @@ GEM
       websocket-driver (~> 0.7)
     globalid (1.3.0)
       activesupport (>= 6.1)
-    haml (7.1.0)
+    haml (7.2.0)
       temple (>= 0.8.2)
       thor
       tilt
@@ -211,13 +211,13 @@ GEM
     nokogiri (1.19.0-x86_64-linux-musl)
       racc (~> 1.4)
     parallel (1.27.0)
-    parser (3.3.10.0)
+    parser (3.3.10.1)
       ast (~> 2.4.1)
       racc
     pp (0.6.3)
       prettyprint
     prettyprint (0.2.0)
-    prism (1.7.0)
+    prism (1.9.0)
     propshaft (1.3.1)
       actionpack (>= 7.0.0)
       activesupport (>= 7.0.0)
@@ -226,7 +226,7 @@ GEM
       date
       stringio
     public_suffix (7.0.2)
-    puma (7.1.0)
+    puma (7.2.0)
       nio4r (~> 2.0)
     racc (1.8.1)
     rack (3.2.4)
@@ -269,7 +269,7 @@ GEM
       zeitwerk (~> 2.6)
     rainbow (3.1.1)
     rake (13.3.1)
-    rdoc (7.0.3)
+    rdoc (7.1.0)
       erb
       psych (>= 4.0.0)
       tsort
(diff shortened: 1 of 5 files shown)


edolix:
note: today i found other two apps where v4 + Slim caused the CI to fail with the same errors

joelhawksley:
@edolix thank you for the report and for following up about seeing this issue elsewhere. I too was unable to reproduce it in https://github.com/ViewComponent/demo. (I've now added Slim to the test suite so you can try to reproduce it there if you'd like)

Please keep us posted as you work to reproduce it in isolation- we don't use Slim at GitHub so I can't dedicate much time to this issue at the moment ❤️

edolix:
@joelhawksley i found a way to reproduce it in this repo: https://github.com/edolix/view-component-bug-replica

I added piece by piece until i got the same behavior of my apps - i'm adding more info in the README but let me know if you see the same errors 🙏 
Thank you!

edolix:
@joelhawksley sorry to bug you directly but just to confirm: can you try to run the repo to validate the issue with Slim please? Thank you so much! 🙏

joelhawksley:
@edolix sorry, I haven't had the time to get to this. As GitHub does not use Slim, I'm not able to prioritize this PR over a couple of work priorities I have on my plate. If you'd like to work on a fix, I'd happily review what you come up with ❤️

edolix:
@joelhawksley no worries at all! I 💯 understand! 

As soon as I have some free time I'll try again 🤞

joelhawksley:
@edolix thanks again for providing the replica repo. I used it to narrow down the failure to https://github.com/ViewComponent/view_component/compare/4.0.0.alpha2...4.0.0.alpha3, basically when we removed the dependency on ActionView::Base.

cc @camertron do you have any ideas?

camertron:
This should be fixed by https://github.com/ViewComponent/view_component/pull/2547 😄

joelhawksley:
Closed by https://github.com/ViewComponent/view_component/pull/2547

PR #2411, opened 2025-08-02
Add ActiveSupport::Notifications instrumentation to compiler

jsolas:

Create a minimal extension point for ViewComponent::Compiler to enable sub-templates functionality through a dedicated gem. This addresses the need discussed in [this comment](https://github.com/ViewComponent/view_component/pull/451#issuecomment-2624314681) about creating an extensible compilation process.

The sub-templates gem will work by subscribing to this notification, receiving the compiled component, then searching for additional template files (e.g., header.html.erb, row.html.erb) in the component's directory and generating corresponding call_* methods with explicit locals validation.

I chose ActiveSupport::Notifications to emit a compile.view_component event after successful compilation. This approach is:

- Idiomatic Rails: Standard pattern for creating extension points
- Lightweight: Zero performance overhead when no subscribers are listening
- Non-invasive: Doesn't modify ViewComponent's internal compilation logic
- Decoupled: Extensions work independently without monkey-patching

- Performance: Zero overhead when no subscribers are listening. The notification only fires after successful compilation.
- Minimal scope: This is a focused change that adds only the essential hook needed for the sub-templates use case.
- Test coverage: Includes tests for notification emission and payload structure.


lib/view_component/compiler.rb
```diff
@@ -24,34 +24,40 @@ def compile(raise_errors: false, force: false)
...
 
       @lock.synchronize do
-        # this check is duplicated so that concurrent compile calls can still
-        # early exit
+        # This check is duplicated so that concurrent compile calls can still
+        # early exit before we instrument.
         return if compiled? && !force
 
-        gather_templates
+        ActiveSupport::Notifications.instrument(
```
joelhawksley, line 31: After thinking about how we might write the documentation for this feature, my gut tells me that we might be better served with something simpler.

What if we instead added a no-op public `after_compile` method to `Base` and call it inside `https://github.com/ViewComponent/view_component/blob/main/lib/view_component/base.rb#L539`? That way, consumers could define their own `after_compile` that takes additional actions.

jsolas: The `after_compile` hook seems much cleaner and avoids the Rails 7.1 issues. I’ll update the PR to use that approach.

fsateler: Perhaps using ActiveSupport::Callbacks would make it easier to inject multiple extensions into the base gem

jsolas:
Friendly ping @joelhawksley! Any chance to get feedback on this? Happy to make changes if needed

joelhawksley (changes requested):

lib/view_component/base.rb
```diff
@@ -611,6 +611,18 @@ def __vc_compiled?
...
 
+      # Called after a component class has been compiled.
+      #
+      # Extensions can override this instance method to run logic after
+      # compilation (e.g., generate helpers, register metadata, etc.).
+      #
+      # By default, this is a no-op. The compiler will invoke this method on an
+      # uninitialized instance using `allocate` to avoid requiring initializer
+      # arguments.
+      def after_compile
```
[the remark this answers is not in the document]
jsolas: Fixed! Now it's a class method, consistent with the other compilation steps.

PR #2424, opened 2025-08-06
Fix double rendering issue for partials that yield

camertron:

Fixes https://github.com/ViewComponent/view_component/issues/2423

`ViewComponent::Base` retains a reference to the view context's output buffer. When component templates are evaluated, they do so inside the context of `ViewComponent::Base`, and thus append to the retained `@output_buffer` instance. However, Rails creates a new output buffer when rendering partials, and temporarily reassigns its own `@output_buffer` instance variable. When rendering partials inside a view component, `ViewComponent::Base` forwards the block on to `view_context.render`, but since the block was created inside `ViewComponent::Base`, that's where it will be evaluated when called. This means the block will append to the original `@output_buffer` and not to the new one Rails creates for the partial. To mitigate the issue, this PR evaluates the block inside the view context with `instance_exec`, which causes the block to append to the correct buffer.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -22,6 +22,10 @@ nav_order: 6
 
     *Cameron Dutro*
 
+* Fix double rendering issue for partials that yield.
+
+    *Cameron Dutro*
+
 ## 4.0.0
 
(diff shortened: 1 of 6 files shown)


docs/CHANGELOG.md
```diff
@@ -22,6 +22,10 @@ nav_order: 6
 
     *Cameron Dutro*
 
+* Fix double rendering issue for partials that yield.
+
+    *Cameron Dutro*
```
rmacklin, line 27: 4.0.1 is already released, so I don't think this should be added as part of the 4.0.1 section

camertron: Woops, thank you 👍

doits:
thank you, LGTM!

PR #2432, opened 2025-08-18
Relative translation keys inside blocks resolve to partials scope instead of callers scope

AckermannTM:
Relative translations in a block use the path of the block, instead of the caller. See https://github.com/ViewComponent/view_component/issues/2413

Created an issue to describe the problem and was asked to create a failing test PR.
It took me some time to put the PR together 😅

I just hope I did everything correctly

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,7 @@ group :development, :test do
   gem "cuprite"
   gem "dry-initializer", require: true
   gem "erb_lint"
-  gem "haml", "~> 6"
+  gem "haml", "~> 7"
   gem "jbuilder", "~> 2"
   gem "m", "~> 1"
   gem "method_source", "~> 1"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -141,7 +141,7 @@ GEM
       websocket-driver (~> 0.7)
     globalid (1.3.0)
       activesupport (>= 6.1)
-    haml (6.4.0)
+    haml (7.0.2)
       temple (>= 0.8.2)
       thor
       tilt
@@ -428,7 +428,7 @@ DEPENDENCIES
   cuprite
   dry-initializer
   erb_lint
-  haml (~> 6)
+  haml (~> 7)
   jbuilder (~> 2)
   m (~> 1)
   method_source (~> 1)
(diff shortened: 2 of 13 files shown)


myabc:
Hi @AckermannTM thanks for creating the bug reproduction!  If I try running your test and inserting a `binding.irb` within the `render` block, I see that calling `t`/`translate` calls the `ActionView` implementation. Outside of the `render` block, it calls the `ViewComponent` implementation. This is actually what I would I would expect to see.

`bundle exec appraisal rails-8.0 rake test TEST=test/sandbox/test/components/relative_translation_partial_component_test.rb`

```rb
<%= render "shared/partial" do %>
  <%= t(".title") %>
  <% binding.irb %>
<% end %>
```

```
> method(:translate).source_location
=> ["/Users/alexbcoles/.gem/ruby/3.4.5/gems/actionview-8.0.2/lib/action_view/helpers/translation_helper.rb", 73]
```

vs

```rb
<%= render "shared/partial" do %>
  <%= t(".title") %>
<% end %>
<% binding.irb %>
```

```
> method(:translate).source_location
=> ["/Users/alexbcoles/git-repos/view_component/lib/view_component/translatable.rb", 93]
```

AckermannTM:
Hey @myabc, thanks for looking into this! That explanation makes sense. Inside a render block it is ActionView handling translations, and outside it falls back to the ViewComponent implementation. So this is not a ViewComponent issue but the expected ActionView behavior when rendering a partial inside a component.

I thought it might be a ViewComponent issue since it worked in version `3.23.2`, but now it is clear that this is expected behavior.

Looks like it’s time to replace all partials with components in my project 😄

Spone:
Thanks @AckermannTM for taking the time to create this PR and @myabc for chiming in!

Since it's not a ViewComponent issue, do you think it would be useful to modify the failing tests in order to document the expected behavior?

AckermannTM:
Hey @Spone , thanks for the suggestion! I'm sorry for replying this late.

I'm going to adjust the test to document this behavior.

AckermannTM:
Hey @Spone, I've applied the changes. Let me know if anything needs to be adjusted, and I'll fix it right away.

PR #2461, opened 2025-10-17
Declare actionview as explicit gem dependency

mikz:
lib/view_component.rb has `require "action_view"`, yet this gem did not have actionview declared as a dependency.

Gems depending on `view_component` would need to declare own dependency on actionview and copy the required version.
This PR tries to reduce the work and duplication by gem maintainers by delegating the responsibility here.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Make `ViewComponent::VERSION` accessible to other gems by default.
+
+    *Hans Lemuet*
+
 * Added Reinvented Hospitality to the list of companies using ViewComponent.
 
     *Torgil Zechel*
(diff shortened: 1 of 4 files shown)


docs/CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ nav_order: 6
 * Add Carwow to list of companies using ViewComponent.
 
     *Tom Lord*
+* Declare `actionview` as a `view_component` gem dependency.
```
Spone:
```suggestion

* Declare `actionview` as a `view_component` gem dependency.
```

mikz:
I’ll resolve the conflict later today and use a variable for the shared version constraint with activesupport.

mikz:
I see some issues with the lock files. Will regenerate then later today.

PR #2462, opened 2025-10-17
Properly guard against incomplete Rails module definitions

mikz:
Some parts of the code guard against undefined Rails module, but some parts do that in unsafe manner.

Good example: `defined?(Rails.env) && Rails.env`
Bad example: `defined?(Rails) && Rails.env`

There are gems (eg. rails-html-sanitizer https://github.com/rails/rails-html-sanitizer/blob/2a8fe8971b02b1b14abb6634e04af4e32b0057cd/lib/rails-html-sanitizer.rb#L10-L12), that define `module Rails` to extend it, but then it is incomplete and causes failures.

This change unifies accessing Rails module with proper `defines?` guarding against incomplete definition of the Rails module.

This pattern is already used on other places in the codebase:
* https://github.com/ViewComponent/view_component/blob/e63ef0eab17e73a7f52ae471b16b9e576c2f3cac/app/controllers/concerns/view_component/preview_actions.rb#L81
* https://github.com/ViewComponent/view_component/blob/e63ef0eab17e73a7f52ae471b16b9e576c2f3cac/app/controllers/concerns/view_component/preview_actions.rb#L98
* https://github.com/ViewComponent/view_component/blob/e63ef0eab17e73a7f52ae471b16b9e576c2f3cac/lib/view_component/engine.rb#L22
* https://github.com/ViewComponent/view_component/blob/e63ef0eab17e73a7f52ae471b16b9e576c2f3cac/lib/view_component/preview.rb#L7
* https://github.com/ViewComponent/view_component/blob/e63ef0eab17e73a7f52ae471b16b9e576c2f3cac/lib/view_component/template.rb#L150

Use view_component gem without Rails.

Guard clause `defined?` checking the nested access into the Rails module. Since it is already used within the codebase on other places.

I don't know what would be a good way to test this. A part of test suite that runs without Rails? Is that something you'd like to guarantee as a project?

### DIFF
--- a/lib/view_component/base.rb
+++ b/lib/view_component/base.rb
@@ -333,7 +333,7 @@ module ViewComponent
       []
     end
 
-    if defined?(Rails::VERSION) && (Rails::VERSION::MAJOR == 7 && Rails::VERSION::MINOR == 1)
+    if defined?(Rails::VERSION) && Rails::VERSION::MAJOR == 7 && Rails::VERSION::MINOR == 1
       # Rails expects us to define `format` on all renderables,
       # but we do not know the `format` of a ViewComponent until runtime.
       def format
--- a/lib/view_component/collection.rb
+++ b/lib/view_component/collection.rb
@@ -20,7 +20,7 @@ module ViewComponent
       components.each(&block)
     end
 
-    if defined?(Rails::VERSION) && (Rails::VERSION::MAJOR == 7 && Rails::VERSION::MINOR == 1)
+    if defined?(Rails::VERSION) && Rails::VERSION::MAJOR == 7 && Rails::VERSION::MINOR == 1
       # Rails expects us to define `format` on all renderables,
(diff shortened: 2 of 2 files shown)


Spone:
Thanks @mikz for your contribution!

I think it's a good idea to fix the guards as you suggest, but ViewComponent still [depends](https://github.com/ViewComponent/view_component/blob/main/view_component.gemspec#L36-L37) on ActiveSupport and ActionView.

> Is that something you'd like to guarantee as a project?

I'm not sure we want to guarantee this, because most of us are using ViewComponent with Rails, but if you'd like to work on a test suite that pass without Rails, we can certainly include it to the CI.

joelhawksley:
@mikz thank you for your contribution! I would welcome a CI suite to cover this change as our compatibility with Bridgetown is documented: https://viewcomponent.org/compatibility.html#bridgetown-static-site-generator

PR #2467, opened 2025-10-20
Support Rails 8.1

### LINKED ISSUE #2471: not compatible with rails 8.1
**Backtrace**:
Because view_component >= 4.0.2 depends on activesupport >= 7.1.0, < 8.1
  and view_component < 4.0.2 depends on activesupport >= 5.2.0, < 8.1,
  activesupport >= 5.2.0, < 8.1 is required.
So, because rails >= 8.1.0 depends on activesupport = 8.1.0
  and Gemfile depends on rails >= 8.1.0,
  version solving has failed.

### System configuration
**Rails version**: 8.1
**Ruby version**: 3.4.5
**Gem version**: 4.0.2

Spone:

Close #2471

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,8 +24,8 @@ jobs:
         bundler-cache: true
     - name: Run benchmarks
       run: |
-        bundle exec appraisal rails-8.1 rake partial_benchmark
-        bundle exec appraisal rails-8.1 rake translatable_benchmark
+        bundle exec appraisal rails-8.0 rake partial_benchmark
+        bundle exec appraisal rails-8.0 rake translatable_benchmark
   test:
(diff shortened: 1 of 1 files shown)


composerinteralia:
I just ran into this as well. It looks like this project uses https://github.com/thoughtbot/appraisal, so you may need to run the relevant appraisal command (`bundle exec appraisal install`?) to regenerate the Gemfile.lock files

Edit: Oh, I just noticed you are a maintainer, so you probably know that already 😅. Thank you!

morgoth:
Rails 8.1 was released, so would be nice to have it in

rainerborene:
The require statement on [view_component/base.rb](https://github.com/ViewComponent/view_component/blob/main/lib/view_component/base.rb#L4) should be moved to `ViewComponent::Configurable` concern to avoid the deprecation message.
```ruby
require "active_support/configurable"
```

PR #2475, opened 2025-10-24
Require view_component/version

Spone:

Make `ViewComponent::VERSION` accessible to other gems by default.

In [`view_component-form`](https://github.com/pantographe/view_component-form) we need to add conditionals to some specs, depending on ViewComponent version. Currently, calling `::ViewComponent::VERSION` from our code raises with `uninitialized constant ViewComponent::VERSION` if we don't require it manually.

I added a `require` to the main lib file. That's the default when you generate a new gem with `rails plugin new`.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Make `ViewComponent::VERSION` accessible to other gems by default.
+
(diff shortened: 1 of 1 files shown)


boardfish (approved):
Guessing from recent history that these CI runs aren't a blocker to merging? Certainly shouldn't be a problem for something like this.

PR #2474, opened 2025-10-24
Resolve ActiveSupport::Configurable deprecation warnings

boardfish:
Similar to rails/rails#53970, replicates its functionality inside `ViewComponent::Configurable`.

Closes #2469

PR #2477, opened 2025-10-24
Isolate ActiveSupport deprecation warning to Configurable module

hasghari:

Fix the deprecation warning from the `ActiveSupport::Configurable` module.

As I understand it, the `ViewComponent::Configurable` module isn't used by the view_component gem and I don't know the original reason for its inclusion. Instead of completely removing the dependency on the `ActiveSupport::Configurable` module, I've isolated its inclusion to only the `ViewComponent::Configurable`.

### DIFF
--- a/gemfiles/rails_8.0.gemfile.lock
+++ b/gemfiles/rails_8.0.gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: ..
   specs:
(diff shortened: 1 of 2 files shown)


Spone:
Thanks for your PR. I think we'll get rid of the dependency altogether, see #2474

Issue #2443, opened 2025-09-11
Possible regression in 4.0 regarding partials


markokajzer:
I'm not sure if this was ever intended to work this way, but in 3.x you could use the _component's methods and variables_ inside of a block passed to a partial. This does not seem to work in 4.0.

Example and reproduction repo below.
I'm sorry I could not grasp how to write up a PR with a failing test :(((

Is there a way to re-enable this behaviour? And if not, how would you handle it?

### Steps to reproduce

[Reproduction repo](https://github.com/markokajzer/view-component-4.x-repro)

```
bin/rails test test/components/test_component_test.rb
```

Inline example:

```
# app/components/test_component.rb
class TestComponent < ViewComponent::Base
  def initialize(title:)
    @title = title
  end

  attr_reader :title
end

# app/components/test_component.html.erb
<%= render "shared/card" do %>
  <%= title %>
<% end %>

# app/views/shared/_card.html.erb
<div>
  <%= yield %>
</div>

# app/components/test_component_test.rb
def test_component_renders_something_useful
  assert_includes(
    render_inline(TestComponent.new(title: "Hello, world!")).to_html,
    "Hello, world!"
  )
end
```

Test passes in 3.x (see [previous commit](https://github.com/markokajzer/view-component-4.x-repro/commit/2cdae0bf6fb66812ba27c79b7adb31997a34d696?w=1) in reproduction repo), but fails in 4.0

### Expected behavior
`title` inside of block passed to `card` partial resolves to `title` passed to `TestComponent`

### Actual behavior
Partial context does not have access to `TestComponent` context.
`@title` is `nil`
`title` method is not defined

**Backtrace**:

```
ActionView::Template::Error: undefined method `title' for an instance of #<Class:0x0000000125f70280>
    app/components/test_component.html.erb:2:in `block in call'
    app/views/shared/_card.html.erb:2
    app/components/test_component.html.erb:1:in `call'
    test/components/test_component_test.rb:8:in `test_component_renders_something_useful'
```

### System configuration
**Rails version**: 8.0.2.1

**Ruby version**: 3.3.6

**Gem version**: 4.0.2 (works in 3.x)

markokajzer:
Looks like this might be somewhat related to https://github.com/ViewComponent/view_component/issues/2413

23tux:
I have exactly the same issue, but I think it has to do with this PR https://github.com/ViewComponent/view_component/pull/2424

[...]

PR #2484, opened 2025-10-29
Capture partial block in the components context

23tux:

- When a component renders a partial, it's block is evaluated in the context of the `__vc_original_view_context`
- Because of this, inside of the block you don't have access to the component's instance

I used a separate `#capture` for the partials block, so that it get's appended into a new OutputBuffer. This output is then caputured by the `__vc_original_view_context`

I'm not sure about performance here, as creating a separate OutputBuffer by using `#capture` is an overhead.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,9 @@ nav_order: 6
 
 ## main
 
+* Capture partial block in the components context
+    *23tux*
+
 * Resolve deprecation warning for `ActiveSupport::Configurable`.
 
     *Simon Fish*


markokajzer:
i can confirm this closes #2443 🎉

doits:
👍 This probably also makes methods and variables defined `attr_accessor` et al available inside the block. I was just about to create an issue about this then found #2443 and this PR, glad there's a solution waiting to be merged :-)

camertron:
Awesome, thanks for this @23tux ❤️ I believe main's tests are passing now, could you rebase and see if we can get a clean CI run on this PR?

joelhawksley:
Closed by https://github.com/ViewComponent/view_component/pull/2510

PR #2502, opened 2025-11-19
Fix bug where error line numbers were incorrect in Rails 8.1

joelhawksley:
Rails 8.1 added a newline to the compiled ERB output in https://github.com/rails/rails/pull/53731, so we need to change our class_eval'd line number depending on the Rails version at compile time, but only for ERB.

I also made some appraisal updates here that get us closer to passing on Rails and Ruby main, but we are blocked by Nokogiri not supporting Ruby 4. I'm guessing that will change in the next couple of weeks ❤️

### DIFF
--- a/lib/view_component/template.rb
+++ b/lib/view_component/template.rb
@@ -57,7 +57,7 @@ module ViewComponent
         # Rails 8.1 added a newline to the compiled ERB output in
         # https://github.com/rails/rails/pull/53731
         lineno =
-          if Rails::VERSION::MAJOR == 8 && Rails::VERSION::MINOR > 0 && details.handler == :erb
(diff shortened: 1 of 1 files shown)


lib/view_component/template.rb
```diff
@@ -45,11 +54,20 @@ class Inline < Template
       def initialize(component:, inline_template:)
         details = ActionView::TemplateDetails.new(nil, inline_template.language.to_sym, nil, nil)
 
+        # Rails 8.1 added a newline to the compiled ERB output in
+        # https://github.com/rails/rails/pull/53731
+        lineno =
+          if Rails::VERSION::MAJOR == 8 && Rails::VERSION::MINOR > 0 && details.handler == :erb
```
Spone:
```suggestion
          if Rails::VERSION::MAJOR >= 8 && Rails::VERSION::MINOR > 0 && details.handler == :erb
```

PR #2506, opened 2025-12-02
remove Who Uses ViewComponent section

joelhawksley:
We originally added this section to the docs to provide social proof of who was adopting ViewComponent. With the framework now well adopted by the community, I think it's time to retire this part of the docs. It has lost its utility and contributions to the list muddy up the changelog.

I also have a hunch that it's not a great idea to be linking off to company sites like this for SEO purposes, as we don't want to imply endorsement of any given corporate entity.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,10 +28,6 @@ nav_order: 6
 
 ## 4.1.1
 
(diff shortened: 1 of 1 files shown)


Spone:
Definitely agreed. What about cleaning up the Changelog as well?

joelhawksley:
@Spone great idea. Done!

PR #2507, opened 2025-12-02
ignore failed head

hakanensari:
Just cleaning up CI. Ruby head failures are informative and, in a way, expected.


joelhawksley:
Agreed. Let's revisit once 4.0 lands and/or Nokogiri is compatible.

PR #2540, opened 2026-01-31
Add support for slot arguments

CyberAP:

This PR adds support for slot arguments.

```ruby
# greeting_component.rb
class GreetingComponent < ViewComponent::Base
  renders_one :message
end
```

```erb
<%# greeting_component.html.erb %>
<div class="greeting">
  <%= message("Hello", "World") %>
</div>
```

```erb
<%# index.html.erb %>
<%= render GreetingComponent.new do |component| %>
  <% component.with_message do |greeting, name| %>
    <%= greeting %>, <%= name %>!
  <% end %>
<% end %>
```

Returning:

```erb
<div class="greeting">
  Hello, World!
</div>
```

Slot arguments provide 2 major benefits to a component approach:

1. Isolation – parent components no longer need to access private component instance to grab necessary data for rendering
2. Configuration – slots could be used in more sophisticated ways previously not possible

Right now ViewComponent only supports basic slots without arguments. This means you have to follow patterns like this in order to get the data you need for rendering:

```erb
<%# index.html.erb %>
<%= render GreetingComponent.new do |component| %>
  <% component.with_message %>
    <%= component.greeting %>, <%= component.name %>!
  <% end %>
<% end %>
```

This breaks one of the key component principles: Encapsulation. If the parent component knows about internal state of the child component they become coupled. That leads to unreliable refactorings and degraded component's reusability.

A common pattern of customizing a slot would be to use lambda slots:

```ruby
  renders_one :header, ->(classes:, &block) do
    content_tag :h1, class: classes, &block
  end
end
```

Unfortunately this approach does not allow us to pass data to the block directly, meaning it's actually the inverse of what we want: it passes data to a child component, not the child component passing data to the parent.

Another problem with this approach is that we can't customize what exactly we render inside the lambda.

Slot arguments allow us to solve these problems with elegance. Instead of component instance we access slot arguments passed to the block:

```erb
<%# index.html.erb %>
<%= render GreetingComponent.new do |component| %>
  <% component.with_message do |greeting, name| %>
    <%= greeting %>, <%= name %>!
  <% end %>
<% end %>
```

This means we are in full control of what's rendered inside the slot and also never touch component internals, allowing for safer refactorings.

With slot arguments we can now support more advanced cases of using slots:

```erb
<%# greeting_component.html.erb %>
<div class="greetings">
  <% greetings.each_with_index do |greeting, index| %>
    <%= message(greeting, index) %>
  <% end %>
</div>
```

```erb
<%# index.html.erb %>
<%= render GreetingComponent.new do |component| %>
  <% component.with_message do |greeting, index| %>
    <%= index % 2 == 0 ? greeting : '' %>!
  <% end %>
<% end %>
```

This feature has been [previously discussed](https://github.com/ViewComponent/view_component/issues/1810) but there still is no working alternative to it.

This approach is not new and is inspired by [Scoped Slots in Vue](https://vuejs.org/guide/components/slots.html#scoped-slots) and [Child Component as Function pattern in React](https://reactpatterns.js.org/docs/function-as-child-component/).

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -18,6 +18,10 @@ nav_order: 6
 
     *Joel Hawksley*
 
+* Support slot arguments.
+
+    *Stanislav Lashmanov*
+
 ## 4.2.0
 
 * Fix translation scope resolution in deeply nested component blocks (3+ levels). Translations called inside deeply nested slot blocks using `renders_many`/`renders_one` were incorrectly resolving to an intermediate component's scope instead of the partial's scope where the block was defined. The fix captures the virtual path at block definition time and restores it during block execution, ensuring translations always resolve relative to where the block was created regardless of nesting depth.
--- a/docs/guide/slots.md
+++ b/docs/guide/slots.md
@@ -219,6 +219,92 @@ end
 
 _Note: While a lambda is called when the `with_*` method is called, a returned component isn't rendered until first use._
 
+## Passing arguments to slots
+
+Slots can receive arguments during rendering. This is useful when the component template has data that the slot content needs access to.
+
+```ruby
+# greeting_component.rb
+class GreetingComponent < ViewComponent::Base
+  renders_one :message
+end
+```
+
+```erb
+<%# greeting_component.html.erb %>
+<div class="greeting">
+  <%= message("Hello", "World") %>
+</div>
+```
+
+```erb
+<%# index.html.erb %>
+<%= render GreetingComponent.new do |component| %>
+  <% component.with_message do |greeting, name| %>
+    <%= greeting %>, <%= name %>!
+  <% end %>
+<% end %>
+```
+
+Returning:
+
+```erb
+<div class="greeting">
+  Hello, World!
+</div>
+```
+
+For `renders_many` slots, call `.call()` on each slot item:
+
+```ruby
+# list_component.rb
+class ListComponent < ViewComponent::Base
+  renders_many :items
+end
+```
+
+```erb
+<%# list_component.html.erb %>
+<ul>
+  <% items.each_with_index do |item, index| %>
+    <li><%= item.call(index) %></li>
+  <% end %>
+</ul>
+```
+
+```erb
+<%# index.html.erb %>
+<%= render ListComponent.new do |component| %>
+  <% component.with_item do |index| %>
+    Item <%= index + 1 %>
+  <% end %>
+  <% component.with_item do |index| %>
+    Item <%= index + 1 %>
+  <% end %>
+<% end %>
+```
+
+Returning:
+
+```erb
+<ul>
+  <li>Item 1</li>
+  <li>Item 2</li>
+</ul>
+```
+
+Keyword arguments are also supported:
+
+```erb
+<%# In the component template %>
+<%= my_slot(name: "Alice", role: "Admin") %>
+
+<%# When setting the slot %>
+<% component.with_my_slot do |name:, role:| %>
+  <%= name %> (<%= role %>)
+<% end %>
+```
+
 ## Rendering collections
 
 Since 2.23.0
--- a/lib/view_component/slot.rb
+++ b/lib/view_component/slot.rb
@@ -43,7 +43,27 @@ module ViewComponent
     # If there is no slot renderable, we evaluate the block passed to
     # the slot and return it.
     def to_s
-      return @content if defined?(@content)
+      call
+    end
+
+    # Renders the slot content, optionally passing arguments to the content block.
+    #
+    # This allows slot content blocks to receive arguments from the component
+    # that defines the slot, enabling patterns like:
+    #
+    #   # In the component template:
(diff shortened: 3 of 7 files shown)


CyberAP:
Thanks for taking your time to look into this @joelhawksley!

> ### Real-world use-case
> Can you provide an example of how you're looking to use this feature? Specifically, I'm looking for enough context to know what is being blocked by this PR so that I can understand if this proposed solution is the best option.

For sure! I am going to use more abstract components than in the real world to remove the project context from this. I am also going to use HAML since that's how I write my components and I am more familiar with it than with erb. Hope that still can help you evaluate this.

Let's say I have a TableComponent, which renders multiple TableRowComponents through configuration:

```haml
-# BaseAppComponent
= render TableComponent.new(rows: rows)
```

```haml
-# TableComponent
%table
  %thead
    -# There could be other configurable content
  %tbody
    - rows.each do |row|
      = render TableRowComponent.new(row: row)
```

```haml
-# TableRowComponent
%tr
  - row.columns.each do |column|
    %td= column.text
```

This is a simple and good enough abstraction to render the tables.

But now I need to customize some of the rows and insert content in between them: add edit buttons or show a warning message. I am left with only 2 options: 

1. Add more configuration to the TableRowComponent
2. Use `html` instead of `text` to render the contents of each table cell

Option 1 makes the component aware of all the possible content that could go inside a table cell. If I needed to support other types of content I would have to modify the TableRowComponent implementation. This makes component less reusable and harder to work with:

```haml
-# TableRowComponent
%tr
  - row.columns.each do |column|
    - if column.warning
      %td
        = render WarningComponent.new(warning: column.warning)
    - elsif column.button
      %td
        = render ButtonComponent.new(**column.button)
    - else
      %td= column.text
```

Option 2 is less secure because now we're inserting HTML, rather than working with Rails rendering. For my particular case I'd like to avoid relying on `.html_safe` completely since this content will still be rendered with ViewComponents.

Now let's imagine we had slot arguments. The implementation would change to this:

```haml
-# BaseAppComponent
= render TableComponent.new(rows: rows) do |instance|
  - instance.with_after_row do |row, row_index|
    - if row_index == row_with_button_index
      %tr
        %td{ colspan: 2 }
          = render ButtonComponent.new(**button_options)
    - if row_index == row_with_warning_index
      %tr
        %td{ colspan: 2 }
          = render WarningComponent.new(warning: warning) 
```

```haml
-# TableComponent
%table
  %thead
    -# There could be other configurable content
  %tbody
    - rows.each_with_index do |row, row_index|
      = render TableRowComponent.new(row: row)
      = after_row(row, row_index)
```

Now our TableComponent can be reused anywhere and we can insert any type of content in between rows, without TableComponent and TableRowComponent being aware which types of content they can support.

----

> ### Explicit argument definition
> The proposed implementation does not define what arguments can be passed to a slot, instead allowing callers to pass whatever they want! One of ViewComponent's key advantages is clearly communicating what context is needed to render views. What might it look like to define parameters when registering a slot?

That's a very good point. WDYT of adding named arguments to `renders_one` and `renders_many` methods to define the slot arguments?

```ruby
renders_one :after_row, with_arguments: [:row, :row_index] # for ordered arguments
renders_one :header, with_arguments: { :text => true } # for named arguments, true means required
```

I am not against even limiting this for now to just named arguments only. This matches how Vue handles slots for example and would still work for my case as well.

Happy to hear your thoughts about this.

CyberAP:
@joelhawksley I believe this was closed by accident.

joelhawksley:
@CyberAP it was, my apologies!

CyberAP:
> That being said, I forgot that we _already_ accept arguments for slots that render other components: https://viewcomponent.org/guide/slots.html#component-slots. Have you considered leveraging that API? What I like about that approach is that 1) it already exists! and 2) it allows for defining arguments the standard Ruby way instead of with a custom format like you suggested.

@joelhawksley indeed I did!

That was actually the first thing I tried but it's actually the opposite of what I want. It allows you to pass data **to** a slot. But I want to pass data **from** a slot to another component or logic.

This is a crucial difference: passing data from a slot allows us to decouple components. With component slots our host component becomes aware of the component that renders a slot. If I wanted a slot to render different components I would have to describe all the possible variants that a slot can render using lambda slots. For me personally that's no different from passing another argument to the component itself and configuring the display this way. I've tried to pass a block to a lambda slot but that didn't work out for me unfortunately.

Slot arguments invert this paradigm: now the host component does not control how and what is rendered in the slot, it's the receiving component that uses that slot and controls what to render inside of it.

joelhawksley:
> Slot arguments invert this paradigm: now the host component does not control how and what is rendered in the slot, it's the receiving component that uses that slot and controls what to render inside of it.

I can see why you need this functionality now, but I don't think Slots is the right place to put it. The API is already quite complex and a source of confusion to new and seasoned users of ViewComponent. 

For now, I'd recommend building your own DSL that gives you the functionality you want on top of the existing slots API. 

I'm sorry we weren't able to find a better solution for you here ❤️

PR #2547, opened 2026-02-06
Fix load order issues for 3rd-party template handlers

camertron:

This PR is a fix for the bug described by https://github.com/ViewComponent/view_component/issues/2448 and https://github.com/ViewComponent/view_component/pull/2409. It simply loads `ActionView::Base` before `ViewComponent::Base`, which ensures all template handlers are registered correctly before ViewComponent starts compiling templates.

As @edolix discovered in https://github.com/ViewComponent/view_component/pull/2409 and reproduced in https://github.com/edolix/view-component-bug-replica (thank you!), `ActionView::Template.template_handler_extensions` does not contain `"slim"` before some components are loaded. As a result, VC can find no template for the affected components. Even when the slim template handler does eventually get registered with ActionView, VC doesn't notice because it memoizes the result of the template lookup.

As @joelhawksley correctly surmised, the bug was introduced when `ViewComponent::Base` stopped inheriting from `ActionView::Base`. The inheritance originally caused `ActionView::Base` to load before `ViewComponent::Base`, and, crucially, _triggered ActionView load hooks_. The load hooks feature is the mechanism Slim (and I would imagine other template engines) use to register themselves. Removing the inheritance causes `ActionView::Base` to load at a different point during the boot process, and thus delays the template engines from registering their handlers as well.

The fix is really simple: just require `action_view/base`. There should be no negative side-effects of doing so, since inheriting from `ActionView::Base` effectively requires `action_view/base` and thereby reproduces the previous behavior.


edolix:
Thank you @camertron!

As I was looking at the VC source code, I imagined it was something "simple but not that obvious" - at least for me without any knowledge of the code base :) 

Thank you! 🙏

joelhawksley (approved):

PR #2595, opened 2026-04-07
Add support for Turbo-streaming ViewComponents.

### LINKED ISSUE #1106: Turbo-Streaming ViewComponents

joelhawksley:
Closes #1106.

### DIFF
--- a/docs/guide/turbo_streams.md
+++ b/docs/guide/turbo_streams.md
@@ -1,0 +1,87 @@
+---
+layout: default
+title: Turbo Streams
+parent: How-to guide
+---
+
+# Turbo Streams
+
+ViewComponents can be used with [Turbo Streams](https://turbo.hotwired.dev/handbook/streams) to broadcast updates over WebSockets.
+
+## Rendering in controllers
+
+In a controller action, render a component inside a Turbo Stream response using `render_to_string` or `view_context.render`:
+
+```ruby
+class MessagesController < ApplicationController
+  def create
+    @message = Message.create!(message_params)
+
+    Turbo::StreamsChannel.broadcast_append_to(
+      "messages",
+      target: "messages",
+      html: render_to_string(MessageComponent.new(message: @message))
+    )
+  end
+end
+```
+
+## Broadcasting later with ActiveJob
+
+To broadcast asynchronously via ActiveJob (using `broadcast_action_later_to`), components must be serializable so they can be passed to the background job.
+
+### Setup
+
+Include `ViewComponent::Serializable` in the component and use `.serializable` instead of `.new` when broadcasting:
+
+```ruby
+class MessageComponent < ViewComponent::Base
+  include ViewComponent::Serializable
+
+  def initialize(message:)
+    @message = message
+  end
+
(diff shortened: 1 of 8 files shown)


docs/guide/turbo_streams.md
```diff
@@ -0,0 +1,87 @@
...
+  after_create_commit :broadcast_append
+
+  private
+
+  def broadcast_append
+    Turbo::StreamsChannel.broadcast_action_later_to(
+      "messages",
+      action: :append,
+      target: "messages",
+      renderable: MessageComponent.serializable(message: self),
```
gap777: This is surprising. I thought `serializable` was producing an artifact for passing into ActiveJob, but that `renderable` was describing something that could render HTML for the front end. Isn't this mixing unrelated things?

bensheldon:
This definitely does what it says it does, and is very simple by simply intercepting `new` to assign a property. So I'm 👍🏻 on this without any project-lifecycle context.

Specifically as to project lifecycle, I think this API is fairly limiting and I wonder if it becomes difficult to extend. e.g. do the limitations become inherent to this implementation.

I'd offer that one alternative would be to create a Proxy object that looked something like this:

```ruby
class ViewProxy
  def initialize(target, **kwargs)
    @target = target
    @initialize_kwargs = kwargs
    @with_component_kwargs ||= []
  end
  
  def with_component(**kwargs)
    @with_component_kwargs << kwargs
  end
  # etc. whatever
  
  def render
    @target.new(@initialize_kwargs).then do |instance|
      @with_component_kwargs.each { |k| instance.with_component(**kwargs) }
      # etc.
    end.render
  end
end

module Serializable
  def serializable(**kwargs)
    ViewProxy.new(self.class, **kwargs)
  end
end

MyComponent.serializable(foo: "bar").with_component(foo: "bar")
```

bensheldon:
This can be closed. Implemented in #2682 🎉

PR #2617, opened 2026-04-24
Yielded content gets insertet at wrong location when used with Rails' form helpers

doits:
When passing the from helper to a shared partial which yields some content, the content gets inserted at the wrong location.

In this test it should be inside the label tag – `<label>world</label>` – but it is rendered outside: `world <label></label>`.

For now this MR just adds a failing test for you to reproduce which is the way bug reports are expected according to the issue template.

Is there somebody that can take a look at this? Thank you very much.


joelhawksley:
Thank you for the failing test PR! I'll continue this work in https://github.com/ViewComponent/view_component/pull/2619.

PR #2619, opened 2026-04-24
Fix yielded content rendered at wrong location with Rails form helpers

joelhawksley:
## Problem

When a ViewComponent renders a partial that uses a Rails form helper (like `form.label`) with a yielding block, the yielded content is rendered **outside** the helper's tag instead of inside it.

**Expected:** `<label>world</label>`
**Actual:** `world <label></label>`

Reported in #2617 — this PR provides the fix.

## Root Cause

`form_with` creates form builders with `@template_object` pointing to the **component**. When the form builder is later used inside a partial (e.g. `form.label :something do ... end`), it calls `@template_object.capture(builder, &block)` — capturing on the **component's** `@output_buffer`.

However, Rails' `Template#render` → `_run` has already replaced `view_context.@output_buffer` with a **fresh `OutputBuffer`** for the partial. The compiled partial template code writes to this new buffer, while the component's `@output_buffer` still references the original one set during `render_in`.

This mismatch means `form.label`'s capture reads from an empty (stale) buffer, and the block's output leaks into the partial's main output stream instead of being wrapped by the label tag.

## Fix

Override `capture` in `ViewComponent::Base` to synchronize `@output_buffer` with `view_context.output_buffer` before delegating to `super`. This ensures form helpers that capture through the component use the same buffer that the block's template code writes to.

```ruby
def capture(*, **, &block)
  old_output_buffer = @output_buffer
  @output_buffer = view_context.output_buffer if view_context
  super
ensure
  @output_buffer = old_output_buffer
end
```

## Test

Adds a test case with a component that renders a partial using `form.label` with `yield`, verifying the yielded content appears inside the label tag.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Fix yielded content rendered at wrong location when using form helpers.
+
+    *Joel Hawksley*, *Markus*
+
 ## 4.8.0
 
 * Add `compile.view_component` ActiveSupport::Notifications event for eager compilation at boot time.
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -28,7 +28,6 @@ nav_order: 6
 
     *Joel Hawksley*
 
-
 ## 4.8.0
 
 * Add `compile.view_component` ActiveSupport::Notifications event for eager compilation at boot time.
--- a/test/sandbox/test/rendering_test.rb
+++ b/test/sandbox/test/rendering_test.rb
@@ -1379,6 +1379,10 @@ class RenderingTest < ViewComponent::TestCase
     assert_text "hello world", exact: true, normalize_ws: true
   end
 
+  def test_render_partial_with_yield_form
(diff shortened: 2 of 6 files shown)


joelhawksley:
@camertron might you be able to review this PR? We can also look at it during office hours tomorrow ❤️

test/sandbox/test/rendering_test.rb
```diff
@@ -1379,6 +1379,10 @@ def test_render_partial_with_yield
     assert_text "hello world", exact: true, normalize_ws: true
   end
 
+  def test_render_partial_with_yield_form
+    assert_includes render_inline(PartialWithYieldFormComponent.new).css('label').to_html, 'world'
```
joelhawksley:
```suggestion
    assert_includes render_inline(PartialWithYieldFormComponent.new).css("label").to_html, "world"
```

camertron (approved):
Awesome, this makes sense to me 👍

docs/CHANGELOG.md
```diff
@@ -24,6 +28,7 @@ nav_order: 6
 
     *Joel Hawksley*
 
+
```
```suggestion
```

Issue #2513, opened 2025-12-04
Wrong I18n Scope used for slots in nested components in Ruby on Rails


henrikbjorn:
In our application after upgrading to the new major version 4 we are seeing issues with how i18n scopes in our rails application is resolved when used with components.

We have a table component with the following api that uses 3 levels of nesting in the component.

``` erb
<# app/views/test/index.html.erb %>
  <%= render(TableComponent.new) do |table| %>
    <% table.with_row do |row| %>
      <% row.with_cell do %>
        <%= t('.hello') %>
      <% end %>
    <% end %>
  <% end %>
```

Previously this would be resolved to `test.index.hello` i18n key. After our upgrade it will try `table_component.hello` instead.

Recently there was merged something that looked like a fix for this on the main branch. But it dosen't resolve the issue.

Curiously it works if the nested is only 2 levels and not 3 as in my example (at least on the main branch)

### Steps to reproduce
Here is a single file test case that shows the bug https://gist.github.com/henrikbjorn/71bfad72e8810d4ac3eae2223c67d493

Running it will fail with

```
Fetching https://github.com/viewcomponent/view_component.git
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
Run options: --seed 17537

# Running:

F

Finished in 0.012763s, 78.3515 runs/s, 156.7030 assertions/s.

  1) Failure:
SlotI18nTest#test_t('.key')_in_nested_slot_should_use_view's_i18n_scope [test.rb:77]:
Expected 'test.index.hello' but got: <table><tr><td>      Hello from table
</td></tr></table>.
Expected # encoding: ASCII-8BIT
#    valid: true
"<table><tr><td>      Hello from table\n</td></tr></table>" to include "Hello from view".

1 runs, 2 assertions, 1 failures, 0 errors, 0 skips
```

I have created a fix locally for our application, which added to the test case https://gist.github.com/henrikbjorn/689b8295ed5dd29797a876d5ac06db7f will make the test pass 

```
Fetching https://github.com/viewcomponent/view_component.git
Fetching gem metadata from https://rubygems.org/...........
Resolving dependencies...
Run options: --seed 46458

# Running:

.

Finished in 0.012941s, 77.2738 runs/s, 154.5476 assertions/s.

1 runs, 2 assertions, 0 failures, 0 errors, 0 skips
```

The fix as a module, not sure if it is good enough for a PR, since I am not really familiar with view component internals

``` ruby
module VirtualPathFix
[...]

PR #2633, opened 2026-05-12
Fix i18n scope for nested lambda-backed slots

artinboghosian:

Fix relative t(".key") translations inside nested lambda-backed slots resolving against the wrong scope. Instead of resolving to the partial where the block was defined, they resolved to an intermediate component's virtual path.
Fixes the bug reported in #2513 which persisted after #2520.

PR #2520 introduced `__vc_content_block_virtual_path` — capturing the virtual path at block-definition time and restoring it at block-evaluation time. This correctly fixed non-lambda slots via Slot#to_s.

However, lambda-backed slots are evaluated immediately inside `__vc_set_slot` rather than deferred to `Slot#to_s`. The lambda branch was updated in #2520 to wrap the block evaluation with `with_captured_virtual_path`, but used `@old_virtual_path` (the intermediate parent component's path) instead of the already-captured `slot.__vc_content_block_virtual_path` (the original partial's path).

The existing `slot.__vc_content_block_virtual_path` is defined via `attr_writer`, making it write-only. Rather than exposing it through an `attr_reader`, the captured path is stored in a local variable within __vc_set_slot and passed directly to the lambda branch. Happy to use a different approach if the project prefers exposing the reader instead.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Fix translation scope resolution in nested lambda-backed slots. Relative `t(".key")` calls inside lambda-backed slots were resolving against an intermediate component's scope instead of the original partial's scope where the block was defined.
+
+    *Artin Boghosian*
+
 ## 4.10.0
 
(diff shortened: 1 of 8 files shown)


artinboghosian:
@joelhawksley The failing check is also failing on main branch.

artinboghosian:
@joelhawksley thanks for merging, appreciate all your hard work

PR #2645, opened 2026-06-01
PR->PR: Replace `serializable` interceptor with Proxy object

bensheldon:

This is a PR demonstrating an alternative to serializing a View Component by using a Proxy object. I did it on top of #1106.


bensheldon:
@joelhawksley fyi, just wanted to show you an alternative that I think I like better than the other method I first championed.

bensheldon:
Replaced with #2682.

PR #2662, opened 2026-06-30
Fix compiler crash on Rails main after template_handler_extensions removal

luizkowalski:
Rails removed `ActionView::Template.template_handler_extensions` in [rails/rails#57874](https://github.com/rails/rails/pull/57874), so the compiler crashes with `NoMethodError` when gathering sidecar templates on Rails main.

This reads the registered extensions through `ActionView::Template::Handlers.extensions` instead. That method is the supported read-path API and exists on every supported Rails version (7.1+), so no version check is needed.

Tested: full sandbox suite green on Rails 8.1; confirmed the old call raises and the new one works against the actual PR commit.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,10 @@ nav_order: 6
 
     *Peter Yates*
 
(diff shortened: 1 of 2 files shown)


luizkowalski:
the error seems unrelated to the changes.

joelhawksley (approved):

PR #2679, opened 2026-08-17
Use minitest-memory for allocation tests

joelhawksley:
Replace the custom `allocation_stats` integration with `minitest-memory` so allocation failures provide richer diagnostics and the test suite can use its broader memory assertions in the future.

The version-aware wrapper continues selecting allocation ranges by Ruby minor version, then delegates the global count assertion to `minitest-memory`. The existing render and collection allocation thresholds remain unchanged.

## Testing

- `bundle exec ruby -Itest test/sandbox/test/rendering_test.rb --name '/allocations/'`
- `bundle exec standardrb test/test_helper.rb test/sandbox/test/rendering_test.rb`


joelhawksley:
Thanks @sferik!

PR #2682, opened 2026-08-18
Add support for Turbo-streaming ViewComponents via proxy object

### LINKED ISSUE #1106: Turbo-Streaming ViewComponents

bensheldon:
Replaces #2645 as a freestanding PR. Alternative to #2595. Closes #1106.

### DIFF
--- a/docs/guide/turbo_streams.md
+++ b/docs/guide/turbo_streams.md
@@ -1,0 +1,102 @@
+---
+layout: default
+title: Turbo Streams
+parent: How-to guide
+---
+
+# Turbo Streams
+
+ViewComponents can be used with [Turbo Streams](https://turbo.hotwired.dev/handbook/streams) to broadcast updates over WebSockets.
+
+## Rendering in controllers
+
+In a controller action, render a component inside a Turbo Stream response using `render_to_string` or `view_context.render`:
+
+```ruby
+class MessagesController < ApplicationController
+  def create
+    @message = Message.create!(message_params)
+
+    Turbo::StreamsChannel.broadcast_append_to(
+      "messages",
+      target: "messages",
+      html: render_to_string(MessageComponent.new(message: @message))
+    )
+  end
+end
+```
+
+## Broadcasting later with ActiveJob
+
+To broadcast asynchronously via ActiveJob (using `broadcast_action_later_to`), components must be serializable so they can be passed to the background job.
+
+Include `ViewComponent::Serializable` in the component and use `.render_later` instead of `.new` when broadcasting:
+
+```ruby
+class MessageComponent < ViewComponent::Base
+  include ViewComponent::Serializable
+
+  def initialize(message:)
+    @message = message
+  end
+end
+
+class Message < ApplicationRecord
+  after_create_commit :broadcast_append
+
+  private
+
+  def broadcast_append
+    Turbo::StreamsChannel.broadcast_action_later_to(
+      "messages",
+      action: :append,
+      target: "messages",
+      renderable: MessageComponent.render_later(message: self),
+      layout: false
+    )
+  end
+end
+```
+
+The component is serialized when the job is enqueued and deserialized when the job runs. The job renders the component via `ApplicationController.render` and broadcasts the resulting HTML over ActionCable.
+
+### Slots
+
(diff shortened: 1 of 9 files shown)


docs/guide/turbo_streams.md
```diff
@@ -0,0 +1,102 @@
...
+
+### How it works
+
+- `.render_later(*args, **kwargs)` returns a `ViewComponent::Serializable::Proxy` that stores the component class and initialization arguments without instantiating the component.
+- `ViewComponent::ActiveJobSerializer` handles converting the proxy to and from a JSON-safe format. It's automatically registered when ActiveJob is loaded.
+- ActiveRecord objects passed as arguments are serialized via GlobalID, just like any other ActiveJob argument.
+
+### Limitations
+
+- Blocks passed to `render_later`, `render_in`, or slot calls raise `ViewComponent::Serializable::UnserializableError`. Use component-based slots instead.
```
joelhawksley, line 101: I'm feeling torn on this constraint.

On one hand, ViewComponents are generally only passed blocks when rendered inside templates, so this probably isn't a limitation most folks will hit.

But I wonder if we even need to have a proxy at all if we could instead render the component _before_ passing it to `broadcast_action_later_to`. Yes, you'd be paying the rendering cost inside the callback in your example instead of in a background job, but _most_ ViewComponents should be very cheap to render.

What do you think about taking that approach? FWIW, I am totally fine with your PR as it stands, just wondering if there's something I'm missing about this other potential approach.

bensheldon: > I wonder if we even need to have a proxy at all if we could instead render the component before passing it to broadcast_action_later_to. 

That's what `broadcast_to` basically does; my intent to support `_later` is specifically to allow deferring the rendering.

I agree that deferred rendering is not completely transparent: you still have to design the View Componant to be deferrable by ensuring that all of its inputs are serializable (no blocks, procs, unmaterialized active record queries, etc.). But I want to make it _possible_.

I don't feel very strongly about doing it as a Proxy, or just storing the serialized data inside of the base View Component. I liked the Proxy because I found it easier to extend and have a clear contract of like "this is serializable, everything else isn't" and I felt like made it easier to add functionality, like adding serializable `with_` slot methods.

Also, sorta an aside, but I remember DMing about this with @camertron and it doesn't look like it made it into the issue: the tricky bit here is that on a View Component, it's tricky to intercept `initialize` because the method is defined on the View Component, not the Base, and Collections introspect the arguments (specifically for the `_counter` argument, but maybe others). The argument introspection makes it tricky to do the other thing I tried, which was simply prepending a module on top with `def initialize(...)` on `inherited`. But using `...` argument forwarding broke Collection's argument introspection and it felt a lot more heavyhanded/performance-sapping to change the introspection.

joelhawksley (approved):

docs/guide/turbo_streams.md
joelhawksley: I had figured you wanted to defer rendering. I think this is fine. Merging!

bensheldon: AWESOME! 🚀

PR #2711, opened 2026-09-03
Resolve declared component dependencies that never opted in

erikaxel:
## Problem

`# Template Dependency: SomeComponent` is the escape hatch for component renders static analysis can't see — a class held in a local variable, or reached through a helper module. The [caching guide](https://viewcomponent.org/guide/caching.html) documents naming components this way.

`CacheDigest.explicit_component_dependencies` translates the declared class name into the synthetic path the component is digested under, but it resolves the name through `constantize_component`, which only accepts classes that already include `ViewComponent::ExperimentallyCacheable`. For anything else the raw constant name stays in the dependency list and is handed to the Digestor as a template path. Nothing resolves there, so the declaration silently does nothing and the only signal is a confusing log line naming a class where a path is expected:

```
Couldn't find template for digesting: ErbComponent
```

That defeats the point of the hatch: the dependencies you need it for are exactly the ones whose target you may not control or want to modify.

## Fix

Resolve any `ViewComponent::Base` descendant in `explicit_component_dependencies`, and register it so `Resolver` can synthesize a template for it — being named by a declaration is itself the opt-in. `component_for` resolves through the same path, since the registry can now hold components that never included the module.

The declared component's template, Ruby class, sidecar files, and superclasses are digested, so editing any of them busts caches that depend on it. Rendering the declared component is still uncached; only components that include the module and declare `cache_on` cache their own output.

Constants that aren't components (and names that aren't constants at all) are left in the dependency list untouched, as before.

## Tests

- `test_declared_components_resolve_without_opting_into_caching` — a declaration naming a component with no `__vc_cacheable?` now resolves to its synthetic path.
- `test_cache_digest_changes_when_a_component_declared_without_opting_in_changes` — end to end through the Digestor: editing `ErbComponent`'s template changes the digest of a component that declares it. This fails on `main`.
- `test_declared_names_that_are_not_components_are_left_alone` — replaces the old assertion that a non-opted-in component resolved to nothing, and covers a plain class and an unknown constant instead.

Full suite passes on the default Gemfile and on `gemfiles/rails_7.1.gemfile` (the two failures there are the pre-existing `RenderingAllocationsTest` entries with no baseline for the Ruby version I ran locally).

## Not covered here

The guide also suggests `# Template Dependency:` for modules included into a component. A module still resolves to nothing, since it has no `virtual_path` and no template or sidecar files to hash. That needs a separate mechanism and seemed out of scope for this fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,12 @@ nav_order: 6
 
 ## main
 
+* Resolve `# Template Dependency: SomeComponent` declarations naming a component that hasn't included `ViewComponent::ExperimentallyCacheable`.
+
+    The declaration was previously dropped: the class name was emitted to the Digestor verbatim, which looked for a template at that path, found none, and logged `Couldn't find template for digesting: SomeComponent`. That made the escape hatch unusable for exactly the dependencies it exists for, since a component reached through a local variable or a helper module is often one you don't control and can't add the module to. Naming a component in a declaration now digests it, opted in or not.
+
+    *Erik Axel Nielsen*
+
 ## 4.15.0
 
 * Add experimental caching support, opt-in per component via `include ViewComponent::ExperimentallyCacheable`.
--- a/docs/guide/caching.md
+++ b/docs/guide/caching.md
@@ -187,7 +187,7 @@ The same works in a template, where the branch is often the more natural place f
 <%= render component.new(post: @post) %>
 ```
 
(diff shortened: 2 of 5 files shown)


erikaxel:
I have marked this as draft until we have considered #2714. Also, this overlaps quite a bit with #2713

PR #2712, opened 2026-09-03
Raise errors while computing cache digests

erikaxel:
## Problem

The component digest machinery rescued every unexpected error and returned a neutral value:

* `CacheDigest::Resolver#find_templates` returned `[]`
* `CacheDigest.constantize_component` returned `nil`
* `CacheDigest.partial_paths_in` returned `[]`
* `DependencyTracking#find_dependencies` fell back to Rails' dependencies

A misconfiguration, failed autoload, raising component hook, parser failure, or filesystem error could therefore make a component untracked. Its fragment digest would stop changing and stale HTML could be served with no indication that digest computation had failed.

Rails does not rescue exceptions from dependency tracking or digest-tree construction. It only creates a missing digest node when template lookup normally returns no template.

## Fix

Remove the broad rescues and let unexpected digest errors propagate, matching Rails' behavior. Expected missing constants continue to return `nil` through `safe_constantize`, and an ordinary unresolved template continues to use Rails' existing missing-node behavior.

## Tests

Public `.cache_digest` integration tests use real cacheable components to verify that failures propagate through the complete digest stack:

* A dependency referenced from a component's Ruby source raises while autoloading, exercising resolver synthesis and component constantization.
* A dependency referenced from a component template raises while autoloading, exercising Rails dependency tracking and component constantization.
* A real digest source cannot be read, exercising filesystem failure propagation through the resolver.

The tests do not stub ViewComponent internals or call private APIs.

Validated locally with:

* `bundle exec rake all_tests`
* the focused caching test suite
* StandardRB on the touched Ruby files

The full suite completed with 637 Minitests, 5 engine compatibility tests, and 8 RSpec examples, with no failures.

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Report the errors swallowed while computing a component's cache digest, instead of degrading to an untracked component with no indication that anything went wrong. Digest errors are now raised in local environments and logged at `warn` elsewhere, configurable with `config.view_component.raise_on_cache_digest_errors`.
+
+    *Erik Axel Nielsen*
+
 ## 4.15.0
 
 * Add experimental caching support, opt-in per component via `include ViewComponent::ExperimentallyCacheable`.
--- a/docs/api.md
+++ b/docs/api.md
@@ -294,6 +294,19 @@ A custom default layout used for the previews index page and individual previews
 
     config.view_component.previews.default_layout = "preview_layout"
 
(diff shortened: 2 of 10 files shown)


docs/CHANGELOG.md
```diff
@@ -10,6 +10,10 @@ nav_order: 6
 
 ## main
 
+* Raise errors encountered while computing a component's cache digest, instead of degrading to an untracked component that can serve stale fragments.
```
```suggestion
* Raise errors encountered while computing a component's cache digest instead of degrading to an untracked component that can serve stale fragments.
```

PR #2715, opened 2026-09-09
Remove redundant quotation

chrislo:
I'm not sure this quotation adds much beyond what is already said in the "Extract general-purpose ViewComponents" section and I found the association with this particular individual a bit jarring as I was reading this section of the docs today, particularly in light of this projects clear and welcoming code of conduct.


chrislo:
@joelhawksley thank you 👍

